PR #280, opened 2014-11-25
Helper methods for resource loading

Turbo87:
This PR adds URL-based `linesOf()` and `contentOf()` methods that can be used to load URLs returned from e.g. `ClassLoader.getResource(...)`.


Turbo87:
deprecated by #402

PR #285, opened 2014-12-04
Added AssertProvider for linking objects to their Assert

liefke:
Currently if I have an additional Assert class, I have to create (or generate) a new Assertions class to allow the fluent form MyAssertions.assertThat(MyObject).condition().

This is ok if the tested object (MyObject) is a real application object, because it can't know anything about the Assert class (which is usually only available during test time).

But if MyObject is an object just for testing as well, e.g. because it is a page object ([see here for an introduction to page objects](https://code.google.com/p/selenium/wiki/PageObjects)), this extra Assertions class is not really necessary and the disadvantage dominates:
You have either more than one custom Assertions class (where you can't use a static import for assertThat in the same test class), or you have to combine all different Assert from different packages into one big Assertions class - what may lead to dependency cycles and other architectural problems.

I know that there is the concept of a AssertDelegateTarget, but if you use that you have to create (and know) the Asserts by yourself. And Assertions.assertThat(new MyDelegateAssert(testObject)).containsText("...") is not as fluent as Assertions.assertThat(testObject).containsText("...").

So I have created a small patch of Assertions, to support the definition of the associated Assert in the tested object.

Here a short example of the usage:
```java
public class Button implements AssertProvider<ButtonAssert> {
  public String getText() {
    return ...;
  }
  public ButtonAssert assertThat() { 
    return new ButtonAssert(this); 
  } 
}

public class ButtonAssert extends Assert<ButtonAssert, Button> {
  ...
  public ButtonAssert containsText(String text) {
    ...
  }
}

void testMethod() {
  Button button = ...;
  Assertions.assertThat(button).containsText("Test");
}
```

I hope you see the advantages of this usecase and add it to AssertJ as well.

### DIFF
--- a/src/main/java/org/assertj/core/api/AssertProvider.java
+++ b/src/main/java/org/assertj/core/api/AssertProvider.java
@@ -1,0 +1,44 @@
+package org.assertj.core.api;
+
+/**
+ * Provides a {@link Assert} for the current object.
+ * 
+ * <p>Used to map an object to its Assert without having to create a new "Assertions" class.</p>
+ * 
+ * Usage: <pre><code class='java'>
+ * public class Button implements AssertProvider<ButtonAssert> {
+ *   public ButtonAssert assertThat() { 
+ *     return new ButtonAssert(this); 
+ *   } 
+ * }
+ * 
+ * public class ButtonAssert extends Assert<ButtonAssert, Button> {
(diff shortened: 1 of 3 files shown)


joel-costigliola:
Integrated, thx.

PR #286, opened 2014-12-07
Add Maps.doesNotContainKeys

dorzey:
Maps had doesNotContainsKey, but was missing doesNotContainsKeys. This was inconsistent when considering that containsKey and containsKeys were present.

Please let me know if anything needs to be addressed.


dorzey:
:+1:

PR #303, opened 2014-12-22
Skeleton PathAssert, plus tests -- NOT FOR INCLUSION. Yet.

fge:
Skeleton based on existing code. I modeled it on FileAssert. In fact I saw nothing related to anything java.nio.file in the code, so I created it from scratch...

Preview only. I know the coding style does not fit; knowing that I use IDEA, do you happen to have a style file available?

Also, about error messages: there seem to be a lot of factories; FileAssert seems to basically use one per message! Is that on purpose? Can I create a single factory for Path-specific operations (including exists, non exists and absolute)?

Please comment! I'll add more test when the base is good enough for you.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractPathAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractPathAssert.java
@@ -354,4 +354,52 @@ public abstract class AbstractPathAssert<S extends AbstractPathAssert<S>>
         paths.assertIsSymbolicLink(info, actual);
         return myself;
     }
+
+    /**
+     * Assert that a given path is absolute
+     *
+     * <p>Note that the fact that a path is absolute does not mean that it is
+     * {@link Path#normalize() normalized}: {@code /foo/..} is absolute, for
+     * instance, but it is not normalized.</p>
+     *
+     * <p>Sample use:</p>
+     *
+     * <pre><code class="java">
+     * // unixFs is a Unix FileSystem
+     * final Path path1 = unixFs.getPath("/foo/bar");
+     * final Path path2 = unixFs.getPath("foo/bar");
+     *
+     * // The following assertion succeeds:
+     * assertThat(path1).isAbsolute();
+     *
+     * // The following assertion fails:
+     * assertThat(path2).isAbsolute();
+     *
+     * // windowsFs is a Windows FileSystem
+     * final Path path3 = windowsFs.getPath("c:\\foo");
+     * final Path path4 = windowsFs.getPath("foo\\bar");
+     * final Path path5 = windowsFs.getPath("c:foo");
+     * final Path path6 = windowsFs.getPath("\\foo\\bar");
+     *
+     * // The following assertion succeeds:
+     * assertThat(path3).isAbsolute();
+     *
+     * // The following assertion fails...
+     * assertThat(path4).isAbsolute();
+     *
+     * // ... And these one also fail!
+     * assertThat(path5).isAbsolute();
+     * assertThat(path6).isAbsolute();
+     * </code></pre>
+     *
+     * @return self
+     *
+     * @see Path#isAbsolute()
+     * @see Path#toRealPath(LinkOption...)
+     */
+    public S isAbsolute()
+    {
+        paths.assertIsAbsolute(info, actual);
+        return myself;
+    }
 }
--- a/src/main/java/org/assertj/core/api/AbstractPathAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractPathAssert.java
@@ -440,4 +440,45 @@ public abstract class AbstractPathAssert<S extends AbstractPathAssert<S>>
         paths.assertIsNormalized(info, actual);
         return myself;
     }
+
+    /**
+     * Assert whether the tested path is canonical
+     *
+     * <p>For Windows users, this assertion is no different than {@link
+     * #isAbsolute()}. For Unix users, this assertion ensures that the tested
+     * path is the actual file system resource, that is, it is not a {@link
+     * Files#isSymbolicLink(Path) symbolic link} to the actual resource, even if
+     * the path is absolute.</p>
+     *
+     * <p>Example:</p>
+     *
+     * <pre><code class="java">
+     * // fs is a Unix filesystem
+     * // Create a directory
+     * final Path basedir = fs.getPath("/tmp/foo");
+     * Files.createDirectories(basedir);
+     * // Create a file in this directory
+     * final Path file = basedir.resolve("file");
+     * Files.createFile(file);
+     * // Create a symbolic link to that file
+     * final Path symlink = basedir.resolve("symlink");
+     * Files.createSymbolicLink(symlink, file);
+     *
+     * // The following assertion succeeds
+     * assertThat(file).isCanonical();
+     *
+     * // The following assertion fails
+     * assertThat(symlink).isCanonical();
+     * </code></pre>
+     *
+     * @throws PathsException an I/O error occurred while evaluating the path
+     *
+     * @see Path#toRealPath(LinkOption...)
+     * @see Files#isSameFile(Path, Path)
+     */
+    public S isCanonical()
+    {
+        paths.assertIsCanonical(info, actual);
+        return myself;
+    }
 }
--- a/src/main/java/org/assertj/core/api/AbstractPathAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractPathAssert.java
@@ -250,4 +250,58 @@ public abstract class AbstractPathAssert<S extends AbstractPathAssert<S>>
         paths.assertIsRegularFile(info, actual);
         return myself;
     }
+
+    /**
+     * Assert that a given path is a directory
+     *
+     * <p><strong>Note that this method will follow symbolic links.</strong> If
+     * you are a Unix user and wish to assert that a path is a symbolic link
+     * instead, use {@link #isSymbolicLink()}.</p>
+     *
+     * <p>This assertion first asserts the existence of the path (using {@link
+     * #exists()}) then checks whether the path is a directory.</p>
+     *
+     * <p>Examples:</p>
+     *
+     * <pre><code class="java">
+     * // fs is a Unix filesystem
+     *
+     * // Create a regular file, and a symbolic link to that regular file
+     * final Path regularFile = fs.getPath("regularFile");
+     * final Path symlink = fs.getPath("symlink");
+     * Files.createFile(regularFile);
+     * Files.createSymbolicLink(symlink, regularFile);
+     *
+     * // Create a directory, and a symbolic link to that directory
+     * final Path dir = fs.getPath("dir");
+     * final Path dirSymlink = fs.getPath("dirSymlink");
+     * Files.createDirectories(dir);
+     * Files.createSymbolicLink(dirSymlink, dir);
+     *
+     * // Create a nonexistent entry, and a symbolic link to that entry
+     * final Path nonExistent = fs.getPath("nonexistent");
+     * final Path dangling = fs.getPath("dangling");
+     * Files.createSymbolicLink(dangling, nonExistent);
+     *
+     * // the following assertions fail because paths do not exist:
+     * assertThat(nonExistent).isRegularFile();
+     * assertThat(dangling).isRegularFile();
+     *
+     * // the following assertions fail because paths exist but are not
+     * // directories:
+     * assertThat(regularFile).isRegularFile();
+     * assertThat(symlink).isRegularFile();
+     *
+     * // the following assertions succeed:
+     * assertThat(dir).isRegularFile();
+     * assertThat(dirSymlink).isRegularFile();
+     * </code></pre>
+     *
+     * @return self
+     */
+    public S isDirectory()
+    {
+        paths.assertIsDirectory(info, actual);
+        return myself;
+    }
 }
--- a/src/main/java/org/assertj/core/api/AbstractPathAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractPathAssert.java
@@ -304,4 +304,54 @@ public abstract class AbstractPathAssert<S extends AbstractPathAssert<S>>
         paths.assertIsDirectory(info, actual);
         return myself;
     }
+
+    /**
+     * Assert that a given path is a symbolic link
+     *
+     * <p>This assertion first asserts the existence of the path (using {@link
+     * #existsNoFollow()}) then checks whether the path is a symbolic link.</p>
+     *
+     * <p>Examples:</p>
+     *
+     * <pre><code class="java">
+     * // fs is a Unix filesystem
+     *
+     * // Create a regular file, and a symbolic link to that regular file
+     * final Path regularFile = fs.getPath("regularFile");
+     * final Path symlink = fs.getPath("symlink");
+     * Files.createFile(regularFile);
+     * Files.createSymbolicLink(symlink, regularFile);
+     *
+     * // Create a directory, and a symbolic link to that directory
+     * final Path dir = fs.getPath("dir");
+     * final Path dirSymlink = fs.getPath("dirSymlink");
+     * Files.createDirectories(dir);
+     * Files.createSymbolicLink(dirSymlink, dir);
+     *
+     * // Create a nonexistent entry, and a symbolic link to that entry
+     * final Path nonExistent = fs.getPath("nonexistent");
+     * final Path dangling = fs.getPath("dangling");
+     * Files.createSymbolicLink(dangling, nonExistent);
+     *
+     * // the following assertion fails because the path does not exist:
+     * assertThat(nonExistent).isSymbolicLink();
+     *
+     * // the following assertions fail because paths exist but are not symbolic
+     * // links
+     * assertThat(regularFile).isSymbolicLink();
+     * assertThat(dirSymlink).isSymbolicLink();
+     *
+     * // the following assertions succeed:
+     * assertThat(symlink).isSymbolicLink();
+     * assertThat(dangling).isSymbolicLink();
+     * assertThat(dir).isSymbolicLink();
+     * </code></pre>
+     *
+     * @return self
+     */
+    public S isSymbolicLink()
+    {
+        paths.assertIsSymbolicLink(info, actual);
+        return myself;
+    }
 }
--- a/src/main/java/org/assertj/core/api/AbstractPathAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractPathAssert.java
@@ -1,0 +1,88 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2014 the original author or authors.
+ */
+package org.assertj.core.api;
+
+import org.assertj.core.internal.Paths;
+import org.assertj.core.util.PathsException;
+import org.assertj.core.util.VisibleForTesting;
+
+import java.nio.file.ClosedFileSystemException;
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.ProviderMismatchException;
+import java.nio.file.spi.FileSystemProvider;
+
+/**
+ * Assertions for {@link Path} objects
+ *
+ * <p>Note that some assertions have two versions: a normal one and a "raw" one
+ * (for instance, {@code hasParent()} and {@code hasParentRaw()}. The difference
+ * is that normal assertions will {@link Path#toRealPath(LinkOption...)
+ * canonicalize} or {@link Path#normalize() normalize} the tested path and,
+ * where applicable, the path argument, before performing the actual test.
+ * Canonicalization includes normalization.</p>
+ *
+ * <p>Canonicalization may lead to an I/O error if a path does not exist, in
+ * which case the given assertions will fail with a {@link PathsException}. Also
+ * note that {@link Files#isSymbolicLink(Path) symbolic links} will be followed
+ * if the filesystem supports them. Finally, if a path is not {@link
+ * Path#isAbsolute()} absolute}, canonicalization will resolve the path against
+ * the process' current working directory.</p>
+ *
+ * <p>These assertions are filesystem independent. You may use them on {@code
+ * Path} instances issued from the default filesystem (ie, instances you get
+ * when using {@link java.nio.file.Paths#get(String, String...)}) or from other
+ * filesystems. For more information, see the {@link FileSystem javadoc for
+ * {@code FileSystem}}.</p>
+ *
+ * <p>Furthermore:</p>
+ *
+ * <ul>
+ *     <li>Unless otherwise noted, assertions which accept arguments will not
+ *     accept {@code null} arguments; if a null argument is passed, these
+ *     assertions will throw a {@link NullPointerException}.</li>
+ *     <li>It is the caller's responsibility to ensure that paths used in
+ *     assertions are issued from valid filesystems which are not {@link
+ *     FileSystem#close() closed}. If a filesystems is closed, assertions will
+ *     throw a {@link ClosedFileSystemException}.</li>
+ *     <li>Some assertions take another {@link Path} as an argument. If this
+ *     path is not issued from the same {@link FileSystemProvider provider} as
+ *     the tested path, assertions will throw a {@link
+ *     ProviderMismatchException}.</li>
+ *     <li>Some assertions may need to perform I/O on the path's underlying
+ *     filesystem; if an I/O error occurs when accessing the filesystem, these
+ *     assertions will throw a {@link PathsException}.</li>
+ * </ul>
+ *
+ * @param <S> self type
+ *
+ * @see Path
+ * @see java.nio.file.Paths#get(String, String...)
+ * @see FileSystem
+ * @see FileSystem#getPath(String, String...)
+ * @see FileSystems#getDefault()
+ * @see Files
+ */
+public abstract class AbstractPathAssert<S extends AbstractPathAssert<S>>
+    extends AbstractAssert<S, Path>
+{
+    @VisibleForTesting
+    protected Paths paths = Paths.instance();
+
+    protected AbstractPathAssert(final Path actual, final Class<?> selfType)
+    {
+        super(actual, selfType);
+    }
+}
--- a/src/main/java/org/assertj/core/api/AbstractPathAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractPathAssert.java
@@ -739,4 +739,83 @@ public abstract class AbstractPathAssert<S extends AbstractPathAssert<S>>
         paths.assertStartsWithRaw(info, actual, other);
         return myself;
     }
+
+    /**
+     * Check whether this path ends with another path
+     *
+     * <p>This assertion will attempt to canonicalize the tested path and
+     * normalize the path given as an argument before performing the actual test.
+     * </p>
+     *
+     * <p>Note that the criterion to determine success is determined by the
+     * path's name elements; therefore, {@code /home/foobar/baz} does
+     * <em>not</em> end with {@code bar/baz}.</p>
+     *
+     * <p>Examples:</p>
+     *
+     * <pre><code class="java">
+     * // fs is a Unix filesystem; the current directory is supposed to be /home
+     * // tested will be canonicalized to /home/joe/myfile
+     * final Path tested = fs.getPath("/home/jane/../joe/myfile");
+     * final Path good = fs.getPath("joe/myfile");
+     * // this path will be normalized to joe/otherfile
+     * final Path bad = fs.getPath("joe/myfile/../otherfile");
+     *
+     * // the following assertion succeeds:
+     * assertThat(tested).endsWith(good);
+     *
+     * // the following assertion fails:
+     * assertThat(tested).endsWith(bad);
+     * </code></pre>
+     *
+     * @param other the other path
+     * @return self
+     *
+     * @throws PathsException failed to canonicalize the tested path (see class
+     * description)
+     */
+    public S endsWith(final Path other)
+    {
+        paths.assertEndsWith(info, actual, other);
+        return myself;
+    }
+
+    /**
+     * Check whether this path ends with another path
+     *
+     * <p><em>This assertion will not perform any canonicalization (on the
+     * tested path) or normalization (on the path given as an argument); see the
+     * class description for more details. If this is not what you want, use
+     * {@link #endsWith(Path)} instead.</em></p>
+     *
+     * <p>This may lead to some surprising results; for instance, path {@code
+     * /home/foo} does <em>not</em> end with {@code foo/.} since the last name
+     * element of the former ({@code foo}) is different from the last name
+     * element of the latter ({@code .}).</p>
+     *
+     * <p>Examples:</p>
+     *
+     * <pre><code class="java">
+     * // fs is a Unix filesystem
+     * final Path tested = fs.getPath("/home/joe/myfile");
+     * final Path good = fs.getPath("joe/myfile");
+     * // Even though the path below is the same as "good" when normalized,
+     * // this assertion will not normalize.
+     * final Path bad = fs.getPath("harry/../joe/myfile");
+     *
+     * // The following assertion succeeds:
+     * assertThat(tested).endsWithRaw(good);
+     *
+     * // But the following assertion fails:
+     * assertThat(tested).endsWithRaw(bad);
+     * </code></pre>
+     *
+     * @param other the other path
+     * @return self
+     */
+    public S endsWithRaw(final Path other)
+    {
+        paths.assertEndsWithRaw(info, actual, other);
+        return myself;
+    }
 }
--- a/src/main/java/org/assertj/core/api/AbstractPathAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractPathAssert.java
@@ -481,4 +481,93 @@ public abstract class AbstractPathAssert<S extends AbstractPathAssert<S>>
         paths.assertIsCanonical(info, actual);
         return myself;
     }
+
+    /**
+     * Assert whether the tested path has the expected parent path
+     *
+     * <p><em>This assertion will perform canonicalization of the tested path
+     * and of the given argument before performing the test; see the class
+     * description for more details. If this is not what you want, use {@link
+     * #hasParentRaw(Path)} instead.</em></p>
+     *
+     * <p>Checks that the tested path has the given parent. This assertion will
+     * fail both if the tested path has no parent, or has a different parent
+     * than what is expected.</p>
+     *
+     * <p>Examples:</p>
+     *
+     * <pre><code class="java">
+     * // fs is a Unix filesystem
+     * final Path actual = fs.getPath("/dir1/dir2/file");
+     * // this path will be normalized to "/dir1/dir2"
+     * final Path good = fs.getPath("/dir1/dir3/../dir2/.");
+     * final Path bad = fs.getPath("/dir1");
+     *
+     * // the following assertion succeeds
+     * assertThat(actual).hasParent(good);
+     *
+     * // the following assertion fails
+     * assertThat(actual).hasParent(bad);
+     * </code></pre>
+     *
+     * @param expected the expected parent path
+     * @return self
+     *
+     * @see Path#getParent()
+     */
+    public S hasParent(final Path expected)
+    {
+        paths.assertHasParent(info, actual, expected);
+        return myself;
+    }
+
+    /**
+     * Assert whether the tested path has the expected parent path
+     *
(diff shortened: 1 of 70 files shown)


joel-costigliola:
@fge do you plan to continue working on this issue in a near future ?

fge:
Sorry for the delay, but yes I do. I am working on some stuff for myself at the moment, if you wish that this be done rapidly then I can dedicate my time to it!

Note about tests (meh, always strange to be talking about tests for a test framework): in order to verify the accuracy of assertions I write, I'd like to rely on [memoryfs](https://github.com/marschall/memoryfilesystem) since it works relatively well, would that be an option?

joel-costigliola:
I'm planning the 2.0 release in few weeks and Path assertions need definitely to be in 2.0, it would be nice if you could finish this within 2/3 weeks but if you don't that's ok, I'll finish the work, no problem.

I'm not against using memoryfs if it allows to to test things that would be really difficult with a real file system but I'm a little bit worried that some tests would pass with memoryfs and fail with a real file system. My opinion is then using the real file system as first option and memoryfs in second.
I'm all hear if you have additional arguments.

fge:
Well, memoryfs allows to test things like, for instance, [`Files.isSameFile()`](http://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#isSameFile-java.nio.file.Path-java.nio.file.Path-); with a Windows filesystem you just cannot do that, whereas with an emulated Unix filesystem (which memoryfs does just fine) you can.

Among other things this would also allow to test the results of whether a given `FileAttributeView` is supported etc etc.

OK, resuming work on it. First things first, expect an amended commit, rebased on `master`, with your concerns addressed. I'll then add more tests.

Note: given the variety of filesystem implementations, some tests are going to be difficult to write at all; for instance, check that two `Path`s have the same content. If both support the creation of `FileChannel`s then it is as easy as creating `MappedByteBuffer`s out of them and check whether those are equals -- easy; if, however, the backing filesystem of a `Path` does _not_ support the creation of a `FileChannel`, the "good old" method of reading byte arrays and comparing will be the way.

Well, I am probably getting ahead of myself at the moment. For now --> code.

fge:
OK, updated with _at least_ javadoc. However:

> Another missing thing that I really care about : javadoc with code example(s) for every assertion API.

Where? In the `AbstractAssert*` implementation?

In the meantime, I'll continue writing assertions...

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,107 @@
...
+     */
+    public S doesNotExist()
+    {
+        paths.assertNotExists(info, actual);
+        return myself;
+    }
+
+    /**
+     * Assert that a given path is absolute
+     *
```
joel-costigliola: add code example

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,107 @@
...
+     * Assert that a given path exists
+     *
+     * <p><strong>IMPORTANT NOTE:</strong> this method will NOT follow symbolic
+     * links (provided that the underlying {@link FileSystem} of this path
+     * supports symbolic links at all).</p>
+     *
+     * <p>This means that even if the path in question is a symbolic link
+     * but its target is nonexistent, this assertion will
+     * <strong>succeed</strong>.</p>
+     *
```
joel-costigliola: add code example including  standard file and a symbolic link to illustrate the comment

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,107 @@
...
+     * Assert that a given path does not exist
+     *
+     * <p><strong>IMPORTANT NOTE:</strong> this method will NOT follow symbolic
+     * links (provided that the underlying {@link FileSystem} of this path
+     * supports symbolic links at all).</p>
+     *
+     * <p>This means that this assertion will return true even if the path is
+     * a symbolic link but the target of the symbolic link does <em>not</em>
+     * exist.</p>
+     *
```
joel-costigliola: add code example including standard file and a symbolic link to illustrate the comment

joel-costigliola:
I have added notes where javadoc code examples were missing.

If you have time (don't worry if you don't) I would like to see these assertions:
* `hasParent`
* `hasNoParent`
* `startsWith`
* `endsWith`
* `hasRoot`

fge:
I will add code examples as requested.

As to assertions, the ones you mention are planned already; but there are a lot more I plan to add.

Before that however: do you agree with the added dependency? Also, do you agree about the base test code? Note that I am a TestNG user, so my JUnit code may turn out to be lacking.

And yeah, code style... I'd rather fix it in one go once everything is written, if you don't mind?

src/test/java/org/assertj/core/internal/paths/Paths_assertIsAbsolute_Test.java
```diff
@@ -0,0 +1,65 @@
...
+import static org.assertj.core.test.TestData.someInfo;
+import static org.assertj.core.test.TestFailures.failBecauseExpectedAssertionErrorWasNotThrown;
+import static org.assertj.core.util.FailureMessages.actualIsNull;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@SuppressWarnings("AutoBoxing")
+public class Paths_assertIsAbsolute_Test
+    extends PathsBaseTest_NoFileSystem
```
joel-costigliola: I would just get rid of `PathsBaseTest_NoFileSystem`, override `setup` calling `PathsBaseTest#setup` and initialising mock. Having a class `PathsBaseTest_NoFileSystem` with a `initPath` feels wrong to me.
IMO, if you need a base class for test that uses a mock path, its name should reflect that, maybe something like `WithMockPathsBaseTest` ...

joel-costigliola:
I agree adding a test dependency.

for the code style, don't worry too much, I'll rebase your commits and take of format discrepancies.

more assertions, like it ! :)

src/test/java/org/assertj/core/internal/paths/Paths_assertIsAbsolute_Test.java
fge: The problem is that for some assertions I'll need more than one path (example: test that two paths share the same filesystem; test that the target of a symlink is a given path; etc etc).

For tests that only use paths, I really want to avoid having to initialize a FileSystem. What is more, and that is a real problem to me, JUnit's `@{Before,After}Class` methods are static (unlike TestNG's), which means that I have to initialize/destroy a FileSystem after _each test_; that is expensive.

I'm not sure about how to tackle this problem, really. I _do_ want to avoid having to create a `FileSystem` for each and every test, especially for classes which do not need to have one in the first place; and ideally I'd like to have the filesystem created and closed per test _class_ not per test, and have the test class initialize it with the contents needed, not the base class.

This is a piece of cake with TestNG, not so with JUnit :(

fge:
I completely rebased the commits so that they look cleaner.

Note about reallyExists() vs exists(): I have chosen this naming even though another solution would have been to have a boolean argument telling whether to follow symlinks or not. Is that OK?

joel-costigliola:
I find `reallyExists()` confusing, it makes me think that `exists()` is not **really** doing its job properly.
I suggest to use `exists(boolean followSymlink)` and `exists()` as a shortcut for `exists(true)` since by default links are followed in the JDK, cf `Files#exists(Path path, LinkOption... options)`.
WDYT ?

Small change needed in javadoc comments (which are nice btw), you need to enclose code with `<pre><code class='java'>` and `</code></pre>` instead of `<pre>` and `</pre>`, if code is not java replace `class='java'` by the target language (ex: class='sh') - it uses [highlight.js](https://highlightjs.org/static/demo/) the hood.

fge:
> I find reallyExists() confusing, it makes me think that exists() is not really doing its job properly.
I suggest to use exists(boolean followSymlink) and exists() as a shortcut for exists(true) since by default links are followed in the JDK, cf Files#exists(Path path, LinkOption... options).
WDYT ?

OK, will do.

> Small change needed in javadoc comments (which are nice btw), you need to enclose code with <pre><code class='java'> and </code></pre> instead of <pre> and </pre>, if code is not java replace class='java' by the target language (ex: class='sh') - it uses highlight.js the hood.

Nice addition! I should look at that for my own javadoc... Will do.

fge:
Uh, pom is currently failing because of a missing mockito-core version dependency. Parent pom needs fixing?

joel-costigliola:
raah, I have forgotten to push maven parent pom modifications ! :( 
It's done now but you need to maven install parent pom locally (it's a SNAPSHOT - 1.3.2-SNAPSHOT). I have just released 1.3.2 release but it takes couple hours to be published in maven central. 
Once available, I will use it for assertj-core pom.

joel-costigliola:
It should be fixed now as parent pom 1.3.2 is [available](http://search.maven.org/#artifactdetails|org.assertj|assertj-parent-pom|1.3.2|pom).

fge:
Confirmed, it fixes it! By the way, mockito 1.10.17 is out.

About javadoc: should I heavily comment Paths as well as AbstractPathAssert?

fge:
Ah, yeah, also, what is the preferred format for error messages? Sometimes I see that there are messages preceded by a newline before "expected", sometimes not...

fge:
OK, can you please carefully review the last assertion?

It is the first assertion which can throw an IOException; as such I modeled the tests etc on an equivalent `FileAssert` which can also fail due to an I/O error.

I have also noted that there are files which use tabs for indentation; most of them use spaces...

Meanwhile, writing other assertions!

(yeah, also, you may have seen that I use a `@ClassRule` when I do need a `FileSystem` to run assertions; my JUnit-fu is poor since I'm a TestNG user, and in TestNG, per-class test resources don't need to be static...)

fge:
OK, another one to review carefully is `.hasParent()`. I have mimicked the equivalent `File` assertion, but this is far from being the equivalent of `Path`'s `.getParent()`.

joel-costigliola:
Don't waste your time on `Paths`, I just want awesome javadoc on the API.
Add javadoc on `Paths` if you feel it's worth it.

Error messages are not consistent, I have started migrated them to start with a new line, use 2 spaces before displaying a value on a new line, a good example is [ShouldBeEqualWithinOffset](https://github.com/joel-costigliola/assertj-core/blob/c15e3aca0bee8cfd1f251c78c38edfe9b15f4d60/src/main/java/org/assertj/core/error/ShouldBeEqualWithinOffset.java):
```
Expecting:
  <8.1>
to be close to:
  <8.0>
by less than <0.01> but difference was <0.1>.
(a difference of exactly <0.01> being considered valid)
```

I will update mockito in a future commit, I don't have specific reasons to use the latest version but it's always good to be up to date.

What is exactly the "last" assertion you want me to review carefully ?

fge:
> Don't waste your time on Paths, I just want awesome javadoc on the API.

OK, will do that.

> Error messages are not consistent, I have started migrated them to start with a new line, use 2 spaces before displaying a value on a new line

OK, will do that.

> What is exactly the "last" assertion you want me to review carefully ?

Uh yeah, I've rebased a lot so the track has been lost.

I'm talking about `.has{,No}Parent()` assertions. The thing is that assertions as they are currently written behave completely differently than `Path`'s `.getParent()`; the latter will not attempt any normalization and/or resolution, the assertion does (`.toRealPath()` performs normalization, by the way, but `.toAbsolutePath()` does not!).

In a similar vein, I'm writing the assertion for `.startsWith()` at the moment and if I were to do it in a similar vein to `.has{No,}Parent()` then I'd `.toRealPath()` on both `actual` and the argument before checking the result.

Is that what you want?

joel-costigliola:
My first thoughts would be that we should stick to the JDK behaviour to avoid confusing users, is it possible not to do any kind of normalization and/or resolution ?

fge:
Well, it is possible, but then this is not what the equivalently named assertion in `File` does. The choice is yours.

fge:
OK, I have transformed `.has{,No}Parent()` to behave like their relevant `Path` methods.

It may have surprising results though since with this modification, `/home` is not a parent of `/home/foo/..` and `/home/..` has a parent ;) And this is unlike what the equivalent assertions in `Files` do.

fge:
OK, since I rebase etc a little too heavily, I have set up my gh-pages branch so that it contain the javadoc.

The result for `AbstractPathAssert` can be viewed [here](http://fge.github.io/assertj-core/org/assertj/core/api/AbstractPathAssert.html).

Note that building the javadoc with a 1.7 JDK fails:

> [ERROR] Exit code: 1 - javadoc: error - invalid flag: -Xdoclint:none

Not very surprising since this option is 1.8 only.

joel-costigliola:
my bad, some java 8 option leaked into the parent pom, parent pom has been fixed (1.3.3), I'm just waiting it to be available in laven central before pushing the commit that will make assertj-core use it.

in the meantime, use java 8 :o)

Some improvements to be done on the javadoc:
* start at first column like [containsOnlyOnce](http://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/ObjectEnumerableAssert.html#containsOnlyOnce%28T...%29)
* I don't really like comment qt the end of the line because you only have the OK/KO inofrmation at the end, I prefer a comment line before successful/failed assertions, for example
```java
     assertThat(symlink).exists();       // success
     assertThat(symlink).exists(true);   // success
     assertThat(symlink).exists(false);  // success

     assertThat(dangling).exists();      // FAILURE
     assertThat(dangling).exists(true);  // FAILURE
     assertThat(dangling).exists(false); // success
```
should preferably be:
```java
// these assertions succeed:
assertThat(symlink).exists(); 
assertThat(symlink).exists(true);
assertThat(symlink).exists(false);

// these assertions fail:
assertThat(dangling).exists(); 
assertThat(dangling).exists(true);
// these assertion succeeds:
assertThat(dangling).exists(false);
```
or
```java
// these assertions succeed:
assertThat(symlink).exists(); 
assertThat(symlink).exists(true);
assertThat(symlink).exists(false);
assertThat(dangling).exists(false);

// these assertions fail:
assertThat(dangling).exists(); 
assertThat(dangling).exists(true);
```

joel-costigliola:
I just realize that `assertThat(symlink).exists(false);` is confusing, I suggest to get rid of `exists(Boolean)` and add `existsWithoutFollowingSymlink()` to cover the case where links are not followed. I know you will find the name long (yes, I know you now !) if you find a shorter name as clear as this one, I'm ok to use it.

joel-costigliola:
I have second thoughts on `hasParent`, I'm not sure what the best option is, on one hand it is definitely confusing for users when paths are not normalized but on the other hand I don't like when assertj does not follow the JDK semantics ... 
WDYT ? Shall we normalize or not ?

joel-costigliola:
javadoc can be built now, just update assertj-core.

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,446 @@
...
+     * <p>Sample use:</p>
+     *
+     * <pre><code class="java">
+     *     assertThat(somePath).isAbsolute();
+     * </code></pre>
+     *
+     * <p>Sample paths on Unix and Windows which are absolute, or not:</p>
+     *
+     * <ul>
```
joel-costigliola: I would rather see the various examples in the code section with a comment line before each case (not at the end).

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,446 @@
...
+     * target of this path exists or not.</p>
+     *
+     * <p>Also note that the fact that a path is absolute does not mean that it
+     * is {@link Path#normalize() normalized}: {@code /foo/..} is absolute, for
+     * instance, but it is not normalized.</p>
+     *
+     * <p>Sample use:</p>
+     *
+     * <pre><code class="java">
+     *     assertThat(somePath).isAbsolute();
```
joel-costigliola: let's start code example at  first column

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,446 @@
...
+     * ({@code .}) and that its only parent components ({@code ..}), if any, are
+     * at the beginning of the path.</p>
+     *
+     * Sample usage:
+     *
+     * <pre><code class="java">
+     *     assertThat(path).isNormalized();
+     * </code></pre>
+     *
+     * <p>For instance, {@code a/b} on Unix systems is normalized but {@code
```
joel-costigliola: same comment as above, 
Move the various examples in the code section with a comment line before each case (not at the end).

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,446 @@
...
+     * following shell code is entered:</p>
+     *
+     * <pre><code class="sh">
+     *     # Create an empty file
+     *     $ touch file
+     *     # Create a symbolic link to that file
+     *     $ ln -s file symlink
+     * </code></pre>
+     *
+     * <p>then provided that {@code file} and {@code symlink} are absolute
```
joel-costigliola: Move the examples in the code section with a comment line before each case (not at the end).

fge:
So...

> I just realize that assertThat(symlink).exists(false); is confusing, I suggest to get rid of exists(Boolean) and add existsWithoutFollowingSymlink() to cover the case where links are not followed.

Well, there will also be a `.isSymbolicLink()` assertion; so maybe just get rid of the boolean version and mention this assertion in `.exists()` javadoc?

(note that all of `Files.is*()` return false if the path does not exist; as I intend to use the JDK for this, I will also document that and recommend that people use `.exists().isRegularFile()` for instance, instead of just `.isRegularFile()`; except for `.isSymbolicLink()` of course!)

> I have second thoughts on hasParent, I'm not sure what the best option is, on one hand it is definitely confusing for users when paths are not normalized but on the other hand I don't like when assertj does not follow the JDK semantics ... 
WDYT ? Shall we normalize or not ?

Well, the current behavior certainly differs from what the equivalent assertion in `Files` does; maybe a solution would be to be able to write:

```java
assertThat(path).asCanonicalPath().assertions().go().here();
```

This means there would be two Path assertion classes though... But that would leave the liberty for "power JSR 203" users (like... Me...) to play with petty Paths :p

joel-costigliola:
I agree, let's just use `exists()` that checks symlinks.

`assertThat(path).asCanonicalPath().assertions().go().here();` is too long for me.
I would rather use normalized Path assertions by default or give use something like: 
`assertThat(fs.getPath("/dir1/dir2/file")).normalized().hasParent(fs.getPath("/dir1/dir2/."))`

No strong opinion yet on that, using normalized Path assertions has my preference as it is intuitive, we would just have to mention in the javadoc that it differs slightly from the JDK.

We might also ask the maliling-list but it's not really active.

fge:
(Javadoc updated, please review)

> No strong opinion yet on that, using normalized Path assertions has my preference as it is intuitive, we would just have to mention in the javadoc that it differs slightly from the JDK.

Normalized is one thing, but should they also be absolute?

Personally I'd prefer "raw" assertions by default; this is mainly because I develop over JSR 203 ([here](https://github.com/fge/java7-fs-dropbox) and [here](https://github.com/fge/java7-fs-box) are two examples; but I also have [this](https://java7-fs-more)), therefore I have to test pretty tricky situations which a "normal", default-filesystem-only user will never encounter.

But I also am of the opinion that the most common case should be the easiest to write, and apart from deriving from the standard naming and provide an `assertThatRawPath()` along with an `assertThat()`, I don't see a clean solution at the moment...

> We might also ask the maliling-list but it's not really active.

Well, there is nothing to lose by asking, and this is a pretty important topic imho.

I'll post a mail...

fge:
Bonus question: what do you expect `.hasRoot()` to do? `Path` has a `.getRoot()` operation

fge:
> I suggest to get rid of exists(Boolean) and add existsWithoutFollowingSymlink() to cover the case where links are not followed.

`.existsNoFollow()`?

fge:
> No strong opinion yet on that, using normalized Path assertions has my preference as it is intuitive, we would just have to mention in the javadoc that it differs slightly from the JDK.

What about we add `.hasRaw{No,}Parent()`?

The big advantage is that `.has{No,}Parent` could just reuse them; the only difference is that the "non raw" assertions would attempt to fully resolve the `Path`s first (using `.toRealPath()`)

fge:
Ohwell, while we are at it, why not three versions:

* the regular version,
* the raw version,
* the canonical version.

The difference:

* the regular version would normalize the paths and turn them into absolute paths;
* the raw version would do nothing;
* the canonical version would call .toRealPath(), which means it would follow symlinks BUT would also fail with an IOException if the path does not exist.

Comments?

fge:
OK, amend to the above comment; new proposals: for each affected assertion there will be a `*Raw()` version. For instance, `.hasParentRaw()` vs `.hasParent()`, etc. Three versions are a little too much.

The non raw versions will perform two steps on the actual path and, when relevant, the path given as an argument:

* normalization,
* turning them into absolute paths (NOT canonical, see below).

Now, the question is: should the non raw versions also check the paths' existence? This may not be desirable imho. One may simply want to check that she is given a path, and that this path, for instance, begins with another one, whether the target path is anything valid.

Again, that will differ from `Files`' equivalent assertions in the sense that when you check whether a `File` has a parent, the assertion will check for the existence of both the file and the parent. But then there are also `.exists*()` tests which are here for just that.

In the meanwhile I'll continue coding some more assertions, until there is an agreement on this. Note that I'll also post this plan on the mailing list.

fge:
Followup... Yes, I know, I ask a lot of questions ;)

Specifically, how should error messages be handled if the user chooses the "non raw" operations? That is, which path should be displayed? Should that be the canonical one or the one on which the assertion is done?

For instance, let's say that the user submits "foo" has a path, and its canonical form is "/tmp/x/foo"; now, it wants to know whether it has parent "/tmp/y".

Right now, the message displays that "foo" doesn't have the expected parent. I don't believe this is the correct message to be displayed but I'd rather ask before changing that...

(also, I'm currently rewriting the javadoc in addition to modifying the tests; javadoc on my github.io page _is not_ up to date yet, I prefer to fix it all before)

fge:
OK, so:

* complete rebase AGAIN; this is based on the discussion on the mailing list;
* the javadoc can be reviewed [here](http://fge.github.io/assertj-core/org/assertj/core/api/AbstractPathAssert.html);
* at this stage, I just propose it for inclusion.

As to the last point, I am fully aware that there is work to be done yet. But I believe that it is in a pretty good shape to be included, even though the code could be better (it can always be!). Still, you have to start somewhere and while the code currently does not have all that `AbstractFileAssert` has (I'm thinking about the content assertions in particular), it also has more than that.

BUT... It depends on how the Javadoc reads to you. This is probably the most important part of this code, in fact!

joel-costigliola:
on error messages, I would display the canonical form because it's the one used in assertions and it should not be difficult for users to match the absolute and relative paths used in assertions.

I will have a look at javadoc and make my comments

joel-costigliola:
My feedback, I would rename `existsNoFollow` to `existsNoFollowLinks` to use the same wording as the [LinkOption#NOFOLLOW_LINKS](http://docs.oracle.com/javase/7/docs/api/java/nio/file/LinkOption.html#NOFOLLOW_LINKS).

The javadoc is nice but the code examples should cover the obvious use cases, for example in [exists](http://fge.github.io/assertj-core/org/assertj/core/api/AbstractPathAssert.html#exists%28%29) we should add this successful assertion:
```java  
assertThat(someFile).exists();
```  
my other comments are made directly in the code.

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     * Files.createSymbolicLink(dangling, nonExistent);
+     *
+     * // Create a regular file, and a symbolic link pointing to it
+     * final Path someFile = fs.getPath("somefile");
+     * Files.createFile(someFile);
+     * final Path symlink = fs.getPath("symlink");
+     * Files.createSymbolicLink(symlink, someFile);
+     *
+     * // The following assertion succeeds
+     * assertThat(symlink).exists();
```
joel-costigliola: add `assertThat(someFile).exists();`

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     * // fs is a Unix filesystem
+     * // Create a symbolic link whose target does not exist
+     * final Path nonExistent = fs.getPath("nonexistent");
+     * final Path dangling = fs.getPath("dangling");
+     * Files.createSymbolicLink(dangling, nonExistent);
+     *
+     * // Create a regular file, and a symbolic link pointing to it
+     * final Path someFile = fs.getPath("somefile");
+     * Files.createFile(someFile);
+     * final Path symlink = fs.getPath("symlink");
```
joel-costigliola: rename `symlink` to `someFileSymLink` to make it obvious what the link target is.

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     * // The following assertions succeed
+     * assertThat(symlink).existsNoFollow();
+     * assertThat(dangling).existsNoFollow();
+     * </code></pre>
+     *
+     * @return self
+     *
+     * @see Files#exists(Path, LinkOption...)
+     */
+    public S existsNoFollow()
```
joel-costigliola: rename to `existNoFollowLinks`

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     *
+     * <pre><code class="java">
+     * // fs is a Unix filesystem
+     * // Create a symbolic link whose target does not exist
+     * final Path nonExistent = fs.getPath("nonexistent");
+     * final Path dangling = fs.getPath("dangling");
+     * Files.createSymbolicLink(dangling, nonExistent);
+     *
+     * // The following assertions succeed
+     * assertThat(symlink).existsNoFollow();
```
joel-costigliola: nonExistent instead of symlink

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     * <p>Examples:</p>
+     *
+     * <pre><code class="java">
+     * // fs is a Unix filesystem
+     * // Create a symbolic link whose target does not exist
+     * final Path nonExistent = fs.getPath("nonexistent");
+     * final Path dangling = fs.getPath("dangling");
+     * Files.createSymbolicLink(dangling, nonExistent);
+     *
+     * // The following assertions succeed
```
joel-costigliola: add example of a standard existing path

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     * that it can be used to assert the existence of a symbolic link even if
+     * its target is invalid.</p>
+     *
+     * <p>Examples:</p>
+     *
+     * <pre><code class="java">
+     * // fs is a Unix filesystem
+     * // Create a symbolic link whose target does not exist
+     * final Path nonExistent = fs.getPath("nonexistent");
+     * final Path dangling = fs.getPath("dangling");
```
joel-costigliola: rename `dangling` to `linkToNonExistent` to make things more obvious (you almost don't have to read the code initializing `linkToNonExistent` to understand the assertions)

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     * Files.createSymbolicLink(dangling, nonExistent);
+     *
+     * // The following assertions succeed
+     * assertThat(symlink).existsNoFollow();
+     * assertThat(dangling).existsNoFollow();
+     * </code></pre>
```
joel-costigliola: add an example of a failing assertion, e.g: 
`assertThat(nonExistent).existsNoFollow();`

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     * @see Files#exists(Path, LinkOption...)
+     */
+    public S existsNoFollow()
+    {
+        paths.assertExistsNoFollow(info, actual);
+        return myself;
+    }
+
+    /**
+     * Assert that a given path does not exist
```
joel-costigliola: add `.`

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     *
+     * <pre><code class="java">
+     * // fs is a FileSystem
+     * // Create a symbolic link "foo" whose nonexistent target is "bar"
+     * final Path foo = fs.getPath("foo");
+     * final Path bar = fs.getPath("bar");
+     * Files.createSymbolicLink(foo, bar);
+     *
+     * // The following assertion fails:
+     * assertThat(foo).doesNotExist();
```
joel-costigliola: add `assertThat(bar).doesNotExist();` as successful assertion (before this one)

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     * // the following assertions fail because paths do not exist:
+     * assertThat(nonExistent).isRegularFile();
+     * assertThat(dangling).isRegularFile();
+     *
+     * // the following assertions fail because paths exist but are not regular
+     * // files:
+     * assertThat(dir).isRegularFile();
+     * assertThat(dirSymlink).isRegularFile();
+     *
+     * // the following assertions succeed:
```
joel-costigliola: move up, it's better to show succeeding asssertions first.

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     *
+     * // Create a directory, and a symbolic link to that directory
+     * final Path dir = fs.getPath("dir");
+     * final Path dirSymlink = fs.getPath("dirSymlink");
+     * Files.createDirectories(dir);
+     * Files.createSymbolicLink(dirSymlink, dir);
+     *
+     * // Create a nonexistent entry, and a symbolic link to that entry
+     * final Path nonExistent = fs.getPath("nonexistent");
+     * final Path dangling = fs.getPath("dangling");
```
joel-costigliola: rename dangling to linkToNonExistent

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     * // the following assertions fail because paths do not exist:
+     * assertThat(nonExistent).isRegularFile();
+     * assertThat(dangling).isRegularFile();
+     *
+     * // the following assertions fail because paths exist but are not
+     * // directories:
+     * assertThat(regularFile).isRegularFile();
+     * assertThat(symlink).isRegularFile();
+     *
+     * // the following assertions succeed:
```
joel-costigliola: move up

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     *
+     * // Create a directory, and a symbolic link to that directory
+     * final Path dir = fs.getPath("dir");
+     * final Path dirSymlink = fs.getPath("dirSymlink");
+     * Files.createDirectories(dir);
+     * Files.createSymbolicLink(dirSymlink, dir);
+     *
+     * // Create a nonexistent entry, and a symbolic link to that entry
+     * final Path nonExistent = fs.getPath("nonexistent");
+     * final Path dangling = fs.getPath("dangling");
```
joel-costigliola: rename dangling to linkToNonExistent

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+    {
+        paths.assertIsCanonical(info, actual);
+        return myself;
+    }
+
+    /**
+     * Assert whether the tested path has the expected parent path
+     *
+     * <p><em>This assertion will perform canonicalization of the tested path
+     * and of the given argument before performing the test; see the class
```
joel-costigliola: add a link back to the javadoc class

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     * fail both if the tested path has no parent, or has a different parent
+     * than what is expected.</p>
+     *
+     * <p>Examples:</p>
+     *
+     * <pre><code class="java">
+     * // fs is a Unix filesystem
+     * final Path actual = fs.getPath("/dir1/dir2/file");
+     * // this path will be normalized to "/dir1/dir2"
+     * final Path good = fs.getPath("/dir1/dir3/../dir2/.");
```
joel-costigliola: add final Path normalized = fs.getPath("/dir1/dir2"); to show that it will pass the assertion

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     * final Path bad = fs.getPath("/dir1");
+     *
+     * // the following assertion succeeds
```
joel-costigliola: assertThat(actual).hasParent(normalized);

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     * </p>
+     *
+     * <p>Examples:</p>
+     *
+     * <pre><code class="java">
+     * // unixFs is a Unix filesystem
+     * final Path path1 = fs.getPath("/usr");
+     * final Path path2 = fs.getPath("/");
+     * final Path path3 = fs.getPath("foo");
+     * final Path path4 = fs.getPath("/usr/lib");
```
joel-costigliola: add: final Path pathWithoutParentButNotNormalized = fs.getPath("/foo/..");

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     *
+     * // the following assertions succeed
+     * assertThat(path1).hasNoParentRaw();
+     * assertThat(path2).hasNoParentRaw();
+     * assertThat(path3).hasNoParentRaw();
+     *
+     * // the following assertion fails
+     * assertThat(path4).hasNoParentRaw();
```
joel-costigliola: add: assertThat(pathWithoutParentButNotNormalized).hasNoParentRaw();

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     * for example, {@code /home/foobar/baz} <em>does not</em> start with {@code
+     * /home/foo}.</p>
+     *
+     * <p>Examples:</p>
+     *
+     * <pre><code class="java">
+     * // fs is a Unix filesystem
+     * final Path tested = fs.getPath("/home/joe/myfile");
+     * // this path will be canonicalized to "/home/joe"
+     * final Path good = fs.getPath("/home/jane/../joe/.");
```
joel-costigliola: add: final Path home = fs.getPath("/home");

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     * final Path bad = fs.getPath("/home/harry");
+     *
+     * // the following assertion succeeds:
+     * assertThat(tested).startsWith(good);
```
joel-costigliola: add: assertThat(tested).startsWith(home);

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     * /../home/foo} does <em>not</em> start with {@code /home} since the first
+     * name element of the former ({@code ..}) is different from the first name
+     * element of the latter ({@code home}).</p>
+     *
+     * <p>Examples:</p>
+     *
+     * <pre><code class="java">
+     * // fs is a Unix filesystem
+     * final Path tested = fs.getPath("/home/joe/myfile");
+     * final Path good = fs.getPath("/home/joe");
```
joel-costigliola: add: final Path home = fs.getPath("/home");

src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -0,0 +1,821 @@
...
+     * final Path bad = fs.getPath("/home/../joe");
+     *
+     * // the following assertion succeeds:
+     * assertThat(tested).startsWithRaw(good);
```
joel-costigliola: add: assertThat(tested).startsWithRaw(home);

joel-costigliola:
@fge any update on this issue ? Do you want me to finish it ?

fge:
Yes, please do.

Unfortunately, I've been very busy with other stuff lately... Sorry :(

Note however that the first commit is a proposed fix for issue #318 and that the code currently depends on this fix.

joel-costigliola:
No problem, thanks for your work !

joel-costigliola:
finally done, thanks @fge for the work !

fge:
First of all, I am sorry for not having had the time to complete my branch to your expectations...

And, uh, next question, in which version will those tests show up? ;)

joel-costigliola:
2.0 planned in a few weeks.

Issue #257, opened 2014-10-22
Request: BDD style for soft asserts


protobufel:
To mirror the BDD style for regular asserts, I am requesting the same for soft asserts.

lpandzic:
@protobufel  take a look at #321. Comments are welcome.

joel-costigliola:
Just to be clear, we need `BDDSoftAssertions` and `JUnitBDDSoftAssertions`.

PR #321, opened 2015-01-08
implemented issue #257

lpandzic:
BDDSoftAssertions for issue 257.

Example usage (using lambdas):

```java
BDDSoftAssertions.then(softly -> {
    softly.assertThat(mansion.guests()).as("Living Guests").isEqualTo(7)
    softly.assertThat(mansion.kitchen()).as("Kitchen").isEqualTo("clean");
});
```

Note that assertAll is called for users in the then method and hence there is some duplication involved regarding assertAll mainly to mark it private and prevent users from using it (since it is redundant). Javadoc is marked with comment why it is private.

### DIFF
--- a/src/main/java/org/assertj/core/api/BDDSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/BDDSoftAssertions.java
@@ -1,0 +1,116 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2014 the original author or authors.
+ */
+package org.assertj.core.api;
+
+import static org.assertj.core.groups.Properties.extractProperty;
+
+import java.util.List;
+
+/**
+ * <p>
+ * BDD style variation of {@link SoftAssertions}.
+ * </p>
+ *
+ * <p>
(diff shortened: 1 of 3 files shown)


joel-costigliola:
This is not what was expected, the idea was to mimic SoftAssertions and renaming `assertThat` by `then`.

lpandzic:
Ok, but I it is more readable like this than to have multiple thens followed by a assertAll.

joel-costigliola:
I disagree, I don't like to have to write a lambda and it does read as nicely as the "standard" example, moreover it would only work for java 8.

Having to call `softly.assertAll();` is indeed bothersome, that's why we have created [JUnitSoftAssertions](https://github.com/joel-costigliola/assertj-examples/blob/master/assertions-examples/src/test/java/org/assertj/examples/JUnitSoftAssertionsExamples.java) rule which will take care of calling it.

In addition to write `BDDSoftAssertions`, it is also needed to write `JUnitBDDSoftAssertions`.

lpandzic:
> Having to call softly.assertAll(); is indeed bothersome, that's why we have created JUnitSoftAssertions rule which will take care of calling it.

Which is, while less, still bothersome. The second disadvantage of that approach is that it forces JUnit dependency.

> I disagree, I don't like to have to write a lambda and it does read as nicely as the "standard" example, moreover it would only work for java 8.

Would this solution be better suitable for java-8 branch under a different name?

joel-costigliola:
I'm afraid not, I still find that having `then` and `assertThat` with the lambda at the same time does not read well.

lpandzic:
Has your stance about this one changed since AssertJ is now on Java 8?

joel-costigliola:
I haven't changed my mind as the example were already showing java 8 code.

lpandzic:
And regarding https://github.com/joel-costigliola/assertj-core/issues/332?
Sorry to bug you, but I really think this would help java 8 users.

joel-costigliola:
I would change my mind if you can find other users willing to have that feature.

joel-costigliola:
can we comment on #332 in #332 ? Anyway I haven't changed my mind on this on either

lpandzic:
Ok, thanks for response.

PR #325, opened 2015-01-19
Removed unused ThrowingRunnable interface

Turbo87:
We use the Callable interface now instead of this custom Runnable interface


Turbo87:
now that I've tried using this with `Callable` I think we should probably keep this interface as a class that implements the `Callable<Void>` interface and returns `null` or something like that. Having to return `null` all the time is a huge pain...

joel-costigliola:
Well spotted. Done in fe00a8c7dbb4dfc2756b068c9b17fcbb475c379b.

PR #328, opened 2015-02-04
introduce CheckStyle/ImportControl for restricting layer dependencies

robert-szucs:
The first commit covers the existing package dependency graph; it compiles without warning.

The second commit is a possible solution to break the existing cycles; I just analyzed the dependency graph, not the actual code so maybe you would do the breaking on an other way; anyhow this shows how to remove an edge so you can remove the choice of yours.

Feel free to contact me if you needs more information on this topic.


joel-costigliola:
Thanks, I'll have a look at that.

joel-costigliola:
I'm closing this issue not because I'm probably not use CheckStyle for that, if I were to use a tool, that would be [macker](https://innig.net/macker/).

PR #330, opened 2015-02-13
Java 8 friendly throwable capture (and assert) API

bric3:
This PR proposes an API that allows a better integration with Java 8 on the 1.7.x branch.

The proposed code covers more need than the one introduced in AssertJ 2.0.0-SNAPSHOT on master in eb0cddd02de8415efa4276de6533f2d403e68576 for issue #255.

1. Users won't have to wait for AssertJ 2.0.0 ...if AssertJ 1.7.2 is released ;)
2. The API don't use the `java.util.concurrent.Callable`
  a. This is cleaner package wise as such test are not about concurrent stuff
  b. `Callable` is limited to raise `Exception`, this PR custom callable allows `Throwable`
3. The API is split in two composable methods.

  The change introduced in eb0cddd02de8415efa4276de6533f2d403e68576 forces the user to _assert_ at the same place he want to _act_. 

  ```java
  assertThatExceptionThrownBy(() -> service.sendMail()).isInstanceOf(...);
  ```

  This is uncomfortable when one wants to follow BDD or AAA styles, hence this PR proposes a `capturing` API

  ```java
  // given
  /* setup fixture */

  // when
  Throwable thrown = capture(() -> service.sendMail());

  // then
  assertThat(thrown).isInstanceOf(...);
  ```

4. The API was also designed for brevity.

The idea comes from my own [StackOverflow answer](http://stackoverflow.com/questions/17426313/how-to-verify-that-an-exception-was-not-thrown/17428439#17428439) and we are users of this code since we are running Java 8 already.

As a heavy github user, I'm open for comment, review, change

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
@@ -1,19 +1,20 @@
 /*
  * Created on Jan 28, 2011
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this Throwable except in compliance
  * with the License. You may obtain a copy of the License at
- * 
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
  * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
  * specific language governing permissions and limitations under the License.
- * 
+ *
  * Copyright @2011 the original author or authors.
  */
 package org.assertj.core.api;
 
+import org.assertj.core.internal.Failures;
 import org.assertj.core.internal.Throwables;
 import org.assertj.core.util.VisibleForTesting;
 
@@ -37,6 +38,13 @@ public abstract class AbstractThrowableAssert<S extends AbstractThrowableAssert<
 
 	protected AbstractThrowableAssert(A actual, Class<?> selfType) {
 		super(actual, selfType);
+	}
+
+	protected S hasBeenThrown() {
+		if(actual == null) {
+			throw Failures.instance().failure("Expected a throwable to have been raised");
+		}
+		return myself;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


joel-costigliola:
Hi @bric3,

I had only a quick look, here's some comments though.
 
I like the idea of separating _when_ and _then_ steps.

I would rename:
* `capture` to `captureThrowable` to make the intent clearer or even better (?) `catchThrowable`
* `assertThatCaptured` to `assertThatThrowableThrownBy` or `assertThatThrownBy` to keep naming consistency with `assertThatExceptionThrownBy`
* `AssertedBehaviorCallable` to `AssertedCallable`, I don't think `Behavior` add much information

I'm not too fond of introducing `AssertedBehaviorCallable` but I get your points, need to think about it a little bit more.

I'm planning to release AssertJ 2.0 in the next weeks so users won't have to wait for 1.7.2 ;)
And I will also release AssertJ for java 8 soon after (as  AssertJ 3.0).

Merci Brice !

bric3:
Hi @joel-costigliola

> I would rename:
> * `capture` to `captureThrowable` to make the intent clearer or even better (?) `catchThrowable`

I first choose `capture` for API brevity, though I agree this word may be too broad and conflict with other APIs if statics are imported. I will rename to `catchThrowable` as it's shorter and express the same intent.

> * `assertThatCaptured` to `assertThatThrowableThrownBy` or `assertThatThrownBy` to keep naming consistency with `assertThatExceptionThrownBy`

IMHO `assertThatExceptionThrownBy` should be removed as it is limited in use as described in my first comment, there's still time as 2.0.0 is not yet out in the wild. In the sake for API brevity and API alignment with the previous point I prefer `assertThatThrownBy`.

> * `AssertedBehaviorCallable` to `AssertedCallable`, I don't think `Behavior` add much information

Good point, although this interface name should matter much in Java 8, let's have a better name. I've given some more thought I guess that `ThrowingCallable` maybe even better, thoughts ?

Thanks for the feedback.

How is NZ ? :wink:

joel-costigliola:
Integrated, thanks.

bric3:
Cool, thanks to you :)

PR #333, opened 2015-02-15
Issue 238 - Use extracting with SoftAssertions

jcgay:
I'm not a `cglib` ninja, if someone has a better way to do it, I take it :)

Have changed the parent-pom version to get license header v2015.


joel-costigliola:
Awesome ! Will review it to make it part of assertj 2.0

jcgay:
Cool, Thx !

PR #334, opened 2015-02-15
Fix usage of "as" with "extracting" in tests

jcgay:
Just found misuse while hacking around.


joel-costigliola:
Well spotted, I will add that misuse in the doc.

PR #336, opened 2015-02-17
added @SafeVarargs to methods that perform save operations on varargs parameters

lpandzic:


joel-costigliola:
thanks

joel-costigliola:
Actually, the tests don't pass because soft assertions requires method to proxy not to be final.
This had a positive outcome to remove all the @SafeVarargs ffrom the API class to make soft assertions work.

lpandzic:
Shame. :/

PR #342, opened 2015-02-27
Add CharSequence containsOnlyDigits assertion

kamilszymanski:


joel-costigliola:
I will review that at my return from holidays.
Thanks.
Le 28 févr. 2015 07:58, "Kamil Szymański" <notifications@github.com> a
écrit :

joel-costigliola:
Thanks for the contribution.
One thing to change is that the assertion should fail in case of an empty String, after all an empty does not contain anything and thus no digits.

kamilszymanski:
I disagree.
I would expect containsOnlyDigits assertion to fail only if char sequence contains any non-digit characters.
Since empty char sequence contains no non-digit characters in my opinion it should pass.

joel-costigliola:
I'm not convinced, I still think that my suggestion is the most intuitive behaviour.

kamilszymanski:
How does empty char sequence break contract of containing only digits?
What exception's message should say? `Expecting: <CharSequence> to contain only digits but it was empty`?

Questions mentioned above convince me that containsOnlyDigits should not fail for empty char sequence - it doesn't contain non-digit characters.
Or to put it the other way, all (even if all means zero characters) of its characters are digits.

joel-costigliola:
I understand your point but still disagree.
Following your rational, an empty string would also contain only vowels and
we would have found a string that contains only digit but also contains
only vowels.
And what about null string, with your rational one can argue that it
contains only digits.

kamilszymanski:
I'm nowhere near to being convinced that containsOnlyDigits should fail for empty string and I see you're not convinced it should pass - maybe we should ask other developers (top committers to assertj-core?) which option makes sense in their opinion.

joel-costigliola:
We agree to disagree ;-)
If a majority of users share your opinion with you, I will agree to it.
I let you post a message in the mailing list : in https://groups.google.com/forum/?fromgroups=#!forum/assertj

kamilszymanski:
Posted request for feedback: https://groups.google.com/forum/?fromgroups=#!topic/assertj/TGDzfJusqrw

tedyoung:
I don't think there's a "right" answer, but I'm going to side with Joel on this. My expectation without reading the docs would be that if the String was empty or null, that it would fail. My main concern would be a false positive, i.e., passes for an unexpectedly empty String.

elefevre:
I think that it should *not* fail if the string is empty.
My rationale is that, if you do want to see it fail when empty, then you can always add a call .isNotEmpty().

tedyoung:
Let's look at this from a consistency point of view: would you expect the .containsOnly() assertions (e.g., `stringList.containsOnly("John", "Sally")`), they all fail if the actual List (or Map, etc.) are null or empty. This is no different: if the String (or CharSequence) is empty, an IllegalArgumentException should be thrown, just like the rest of the API.

joel-costigliola:
A (short) majority of users prefer `containsOnlyDigits()` to fail for empty char sequences.  
Let's implement it like that.

joel-costigliola:
One more argument in favor of returning false on empty String : https://commons.apache.org/proper/commons-lang/javadocs/api-3.3.2/org/apache/commons/lang3/StringUtils.html#isNumeric%28java.lang.CharSequence%29

kamilszymanski:
I don't see much correlation between isNumeric and contains onlyDigits when it comes to handling empty CharSequence, but I we want to find some examples of how well known and used libraries handle such situations I guess [StringUtils.conatinsOnly](https://commons.apache.org/proper/commons-lang/javadocs/api-3.3.2/org/apache/commons/lang3/StringUtils.html#containsOnly%28java.lang.CharSequence,%20java.lang.String%29) is a better example.

kamilszymanski:
Adding to discussion last year's most active contributors to get more feedback.

@mhaligowski, @jcgay, @abalhier, @AlexBischof how in your opinion containsOnlyDigits assertion should behave for empty char sequences?

tedyoung:
@kamilszymanski Since the purpose of isNumeric is _exactly the same_ as containsOnlyDigits (per the docs for `isNumeric`: "Checks if the CharSequence contains only Unicode digits."), it certainly seems like a perfect example to me. Especially since in 3.0 they explicitly changed it to be false on empty, per the docs again: "3.0 Changed "" to return *false* and not true".

Personally I feel like this discussion has gone on too long considering how important this issue is. Let's just return false and get it done.

joel-costigliola:
I concur, let's just return false and get it done.

joel-costigliola:
Just wondering if we should rename the assertion to `isNumeric` ?

tedyoung:
While that would align it with the commons-lang3 name, I find it somewhat misleading, e.g.:

is `"+300"` Numeric? How about `"234.44"`? 

Both of those would return `false` from commons-lang3's `isNumeric()` method. Personally I find that confusing, because "numeric" has a specific meaning to me, i.e., it has the attribute of being a number. Using `containsOnlyDigits` is more precise.

joel-costigliola:
I have to agree, let's keep `containsOnlyDigits`, it's clearer.

kamilszymanski:
I don't think we will get much more feedback on this issue.
Since there were more votes for failing the assertion for empty CharSequence I'll implement it like that.

joel-costigliola:
Integrated after minor changes, thanks.

PR #344, opened 2015-03-07
changing \n to %n

FaustXVI:


src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
```diff
@@ -156,13 +156,13 @@ public S hasSize(int expected) {
   /**
    * Verifies that the actual {@code CharSequence} has the expected line count.
    * <p/>
-   * A line is considered to be <a name="lt">terminated</a> by any one of a line feed ('\n'), a carriage return ('\r'),
```
joel-costigliola: this comment should not be changed, `\n` is the correct value.

src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -1092,7 +1092,7 @@ public static String contentOf(File file) {
   /**
    * Loads the text content of a file into a list of strings with the default charset, each string corresponding to a
    * line.
-   * The line endings are either \n, \r or \r\n.
```
joel-costigliola: this comment should not be changed, \n is the correct value.

src/main/java/org/assertj/core/api/SoftAssertionError.java
```diff
@@ -36,7 +39,7 @@ public SoftAssertionError(List<String> errors) {
   }
 
   private static String createMessage(List<String> errors) {
-    StringBuilder msg = new StringBuilder("\nThe following ");
+    StringBuilder msg = new StringBuilder("%nThe following ");
```
joel-costigliola: this needs a String.format call or using System.lineSeparator().

src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -1092,7 +1092,7 @@ public static String contentOf(File file) {
...
+   * The line endings are either %n, \r or \r%n.
```
joel-costigliola: this comment should not be changed, \n is the correct value.

src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
```diff
@@ -156,13 +156,13 @@ public S hasSize(int expected) {
...
+   * A line is considered to be <a name="lt">terminated</a> by any one of a line feed ('%n'), a carriage return ('\r'),
```
joel-costigliola: this comment should not be changed, \n is the correct value.

src/main/java/org/assertj/core/util/Strings.java
```diff
@@ -83,7 +83,8 @@ public static String concat(Object... objects) {
    * @return the formatted string if any args were given
    */
   public static String formatIfArgs(String message, Object... args) {
-    return Arrays.isNullOrEmpty(args) ? message : format(message, args);
+    return Arrays.isNullOrEmpty(args) ? format(message.replaceAll("%","%%").replaceAll("%%n","%n")).intern() : format
```
joel-costigliola: I don't think we need to intern the String, if the reason to do that was AbstractAssert_describedAs_with_text_description_Test.should_set_description failure, we should simply change the test and use `assertThat(assertions.descriptionText()).isEqualTo(description);` instead checking for String to be the same.
A comment would be welcome to explain the trick with % replacements.

src/main/java/org/assertj/core/util/Files.java
```diff
@@ -303,7 +303,7 @@ private static String loadContents(File file, Charset charset) throws IOExceptio
 
   /**
    * Loads the text content of a file into a list of strings, each string corresponding to a line. The line endings are
-   * either \n, \r or \r\n.
+   * either %n, \r or \r%n.
```
joel-costigliola: let's keep `\n` here

src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
FaustXVI: Yep, my bad… and I think this is the case for all comments right ? Including in src/main/java/org/assertj/core/error/ShouldBeInstance.java param description.

src/main/java/org/assertj/core/util/Strings.java
FaustXVI: That's exactly the reason but I didn't want to change the tests while changing production code and without talking to you about it first. Now I'll do :D
I agree it do need a comment but I rather find a more elegent maner…

FaustXVI:
I think the production code is now OK. If you do think so, I'll rebase my history to make it more consistant and continue with the tests.

joel-costigliola:
Integrated, Thanks Xavier !

Issue #353, opened 2015-03-30
Easier flatExtracting usage


joel-costigliola:
For example, instead having to implements `Extractor<BasketBallPlayer, List<BasketBallPlayer>>`  to be able to call:
```java
// teamMatesExtractor() returns a new Extractor<BasketBallPlayer, List<BasketBallPlayer>>
assertThat(list(noah, james)).flatExtracting(teamMatesExtractor()).contains(irving, rose);
```
I just want to write (assuming `BasketBallPlayer` has a `teamMate` property):
```java
assertThat(list(noah, james)).flatExtracting("teamMates").contains(irving, rose);
```
or the type safe version:
```java
assertThat(list(noah, james)).flatExtracting("teamMates", BasketBallPlayer.class)
                             .contains(irving, rose);
```

AlexBischof:
Implemented the non-typesafe variant but i have the feeling that it was too easy. Basically i was overloading AbstractIterableAssert.flatExtracting/AbstractObjectArrayAssert.flatExtracting and used internally Extractors.byName.

After a short feedback i would create also the typesyfe variant.

joel-costigliola:
Thanks that was quick  ! :)

Can you create a PR ? It's just easier to comment code.
Can you check that you have the correct formatting preferences, there were so much diff due to reformatting that it is hard to see what has really changed.

From what I saw, we should (if possible) probably reuse the existing `flatExtracting` your code seems to duplicate it, I also don't understand the NPE.

AlexBischof:
I have refactored and reformatted the code and created a PR. I also removed the exception wrapping which was a result of a manually created extractor of the test classes.

joel-costigliola:
Thanks, integrated with few improvements.

PR #354, opened 2015-03-31
#353 Easier flatExtracting usage

AlexBischof:
Added flatExtracting(String) to AbstractIterableAssert and AbstractObjectArrayAssert and reused overloaded flatExtracting. Interally an extractor is created for the property name.

I've converted the IllegalArgumentException into a NPE so that the behavior of the existing flatExtracting is same to the new one, as you can see from the old tests.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
@@ -347,15 +347,15 @@ public abstract class AbstractIterableAssert<S extends AbstractIterableAssert<S,
    * subclasses of the given type).
    * <p/>
    * Example:
-   *
+   * 
    * <pre><code class='java'>
    * List&lt;Number&gt; numbers = new ArrayList&lt;Number&gt;();
    * numbers.add(1);
    * numbers.add(2L);
-   *
+   * 
    * // successful assertion:
    * assertThat(numbers).hasAtLeastOneElementOfType(Long.class);
-   *
+   * 
    * // assertion failure:
    * assertThat(numbers).hasAtLeastOneElementOfType(Float.class);
    * </code></pre>
@@ -378,17 +378,17 @@ public abstract class AbstractIterableAssert<S extends AbstractIterableAssert<S,
    * subclasses of the given type).
    * <p/>
    * Example:
-   *
+   * 
    * <pre><code class='java'>
    * List&lt;Number&gt; numbers = new ArrayList&lt;Number&gt;();
    * numbers.add(1);
    * numbers.add(2);
    * numbers.add(3);
-   *
+   * 
    * // successful assertion:
    * assertThat(numbers).hasOnlyElementsOfType(Number.class);
    * assertThat(numbers).hasOnlyElementsOfType(Integer.class);
-   *
+   * 
    * // assertion failure:
    * assertThat(numbers).hasOnlyElementsOfType(Long.class);
    * </code></pre>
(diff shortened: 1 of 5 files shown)


src/main/java/org/assertj/core/api/AbstractIterableAssert.java
```diff
@@ -924,6 +925,45 @@ public S usingDefaultElementComparator() {
...
+     * homer.getChildren().add(bart);
+     * homer.getChildren().add(lisa);
+     * homer.getChildren().add(maggie);
+     *
+     * CartoonCharacter pebbles = new CartoonCharacter("Pebbles Flintstone");
+     * CartoonCharacter fred = new CartoonCharacter("Fred Flintstone");
+     * fred.getChildren().add(pebbles);
+     *
+     *
+     * assertThat(Lists.newArrayList(homer, fred)).flatExtracting("children").containsOnly(bart, lisa, maggie, pebbles);
```
joel-costigliola: get rid of `Lists.` as we use `newArrayList` directly when creating `cartoonCharacters`, it will make the example lighter.
It would also be good to put `.containsOnly` in the next line, aligned on `.flatExtracting`

src/main/java/org/assertj/core/api/AbstractIterableAssert.java
```diff
@@ -924,6 +925,45 @@ public S usingDefaultElementComparator() {
 	return new ListAssert<>(result);
   }
 
+    /**
+     * Extract the Iterable values from Iterable's elements under test by applying an Iterable extracting function based
```
joel-costigliola: I feel we could simplify this explanation by avoiding to mention the extracting function as it is an implementation detail. Maybe something like:
```
Extract from Iterable's elements the Iterable values corresponding to the given property/field name and concatenate them into a single list becomes a new object under test.
```

src/test/java/org/assertj/core/api/objectarray/ObjectArrayAssert_flatExtracting_Test.java
```diff
@@ -68,10 +68,4 @@ public void should_allow_assertions_on_empty_result_lists() throws Exception {
     assertThat(new CartoonCharacter[] { bart, lisa, maggie }).flatExtracting(children).isEmpty();
   }
 
-  @Test
```
joel-costigliola: why is this test removed ?

src/test/java/org/assertj/core/api/objectarray/ObjectArrayAssert_flatExtractingEasy_Test.java
```diff
@@ -0,0 +1,63 @@
...
+  @Test
+  public void should_allow_assertions_on_joined_lists_when_extracting_children() {
+    assertThat(new CartoonCharacter[] { homer, fred }).flatExtracting("children").containsOnly(bart, lisa, maggie,
+        pebbles);
+  }
+
+  @Test
+  public void should_allow_assertions_on_empty_result_lists() throws Exception {
+    assertThat(new CartoonCharacter[] { bart, lisa, maggie }).flatExtracting("children").isEmpty();
+  }
```
joel-costigliola: add test to assess NPE when one of the elements is null

AlexBischof: The NPE of the older version was related to the extractor implementation in the test class. With the new version an IllegalArgumentException is thrown. My first idea was to rethrow the IllegalArgumentException as a NPE but that doesn't look good. So, just a test for the IllegalArgumentException?

     private final Extractor<CartoonCharacter, List<CartoonCharacter>> children = new Extractor<CartoonCharacter, List<CartoonCharacter>>() {
        @Override
        public List<CartoonCharacter> extract(CartoonCharacter input) {
          return input.getChildren();
        }
      };

joel-costigliola: Yes ! `IllegalArgumentException` express the behavior ina better way than NPE

joel-costigliola:
`IterableAssert_flatExtractingEasy_test` as now the code throws an `IllegalArgumentException`.
I'm fixing that but please run the unit tests before pushing code.

AlexBischof:
Sorry for that but i always run mvn clean test or install before i push. The problem is/was that the test class has a wrong name for the maven convention (begin or end with Test (casesensitive)). So the test classes should end with 'Test'. Does it makes sense that i rename the classes in this PR or would you like to do it?

joel-costigliola:
You are indeed right, I will rename the class.

Issue #355, opened 2015-03-31
Feature request: isEqualToIgnoringWhitespace assertion on String


mkordas:
It could be something simillar to Hamcrest's `org.hamcrest.Matchers#equalToIgnoringWhiteSpace`.

From it's Javadoc:
>Creates a matcher of String that matches when the examined string is equal to the specified expectedString, when whitespace differences are (mostly) ignored. To be exact, the following whitespace rules are applied:
all leading and trailing whitespace of both the expectedString and the examined string are ignored
any remaining whitespace, appearing within either string, is collapsed to a single space before comparison

>For example:
`assertThat("   my\tfoo  bar ", equalToIgnoringWhiteSpace(" my  foo bar"))`

joel-costigliola:
Agreed, with the same semantics as hamcrest to avoid confusing users.

AlexBischof:
My commit should solve this issue but i am not able to get the test running. He fails always with "Joda" which i am not using.

No offense but the design of the CharSequenceAssertBaseTest is totally not understandable. Time to implement the issue <5min, time to get the test running >1h until now and still not running

joel-costigliola:
PR please ! ;-)

mkordas:
Thanks guys for fast response. I agree that semantics should be the same, but:
* we should add more documentation and define what is whitespace
* we should specify how it behaves with strings with Windows/Unix newline characters
* we should change `WhiteSpace` to `Whitespace` as it is used on http://en.wikipedia.org/wiki/Whitespace_character

joel-costigliola:
The behaviour should be described in javadoc WITH code examples, I always find it easier to understand behaviour through examples.

PR #358, opened 2015-04-01
#355 isEqualToIgnoringWhitespace assertion on String

AlexBischof:
added assertEqualIgnoringWhitespaces to AbstractCharSequenceAssert which is using the logic of the corresponding hamcrest matcher.

The related test CharSequenceAssert_isEqualToIgnoringWhitespaces_Test still fails because something goes wrong with the CharSequenceAssertBaseTest design which i couldn't figure out yet.

Examples still missing...

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
@@ -708,6 +708,25 @@ public abstract class AbstractCharSequenceAssert<S extends AbstractCharSequenceA
     return myself;
   }
 
+   /**
+   * Verifies that the actual {@code CharSequence} is equal to the given one, ignoring whitespace differences (mostly).
+   * <p>
+   * Example :
+   *
+   * <pre><code class='java'>
+   * // assertion will pass
+   * assertThat(&quot; my\tfoo bar &quot;).isEqualToIgnoringWhitespace(&quot; my foo bar&quot;);
+   *
+   * // assertion will fail
+   * assertThat(&quot; my\tfoo bar &quot;).isEqualToIgnoringWhitespace(&quot; my foobar&quot;);
+   * </code></pre>
+   *
+   * </p>
+   *
+   * @param expected the given {@code CharSequence} to compare the actual {@code CharSequence} to.
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual {@code CharSequence} is not equal ignoring whitespace differences to the given one.
+   */
   public S isEqualToIgnoringWhitespace(CharSequence expected) {
     strings.assertEqualsIgnoringWhitespace(info, actual, expected);
     return myself;
--- a/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
@@ -708,4 +708,8 @@ public abstract class AbstractCharSequenceAssert<S extends AbstractCharSequenceA
     return myself;
   }
 
+  public S isEqualToIgnoringWhitespaces(CharSequence expected) {
+    strings.assertEqualsIgnoringWhitespaces(info, actual, expected);
+    return myself;
+  }
 }
--- a/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
@@ -708,8 +708,8 @@ public abstract class AbstractCharSequenceAssert<S extends AbstractCharSequenceA
     return myself;
   }
 
-  public S isEqualToIgnoringWhitespaces(CharSequence expected) {
-    strings.assertEqualsIgnoringWhitespaces(info, actual, expected);
+  public S isEqualToIgnoringWhitespace(CharSequence expected) {
+    strings.assertEqualsIgnoringWhitespace(info, actual, expected);
     return myself;
   }
 }
--- a/src/main/java/org/assertj/core/internal/Strings.java
+++ b/src/main/java/org/assertj/core/internal/Strings.java
@@ -17,6 +17,7 @@ import static java.lang.String.format;
 import static org.assertj.core.error.ShouldBeEmpty.shouldBeEmpty;
 import static org.assertj.core.error.ShouldBeEqual.shouldBeEqual;
 import static org.assertj.core.error.ShouldBeEqualIgnoringCase.shouldBeEqual;
+import static org.assertj.core.error.ShouldBeEqualIgnoringWhitespace.shouldBeEqualIgnoringWhitespace;
 import static org.assertj.core.error.ShouldBeNullOrEmpty.shouldBeNullOrEmpty;
 import static org.assertj.core.error.ShouldContainCharSequence.shouldContain;
 import static org.assertj.core.error.ShouldContainCharSequence.shouldContainIgnoringCase;
@@ -340,7 +341,7 @@ public class Strings {
    */
   public void assertEqualsIgnoringWhitespace(AssertionInfo info, CharSequence actual, CharSequence expected) {
       if (!areEqualIgnoringWhitespace(actual, expected)) {
-        throw failures.failure(info, shouldBeEqual(actual, expected));
+        throw failures.failure(info, shouldBeEqualIgnoringWhitespace(actual, expected));
       }
   }
 
--- a/src/main/java/org/assertj/core/internal/Strings.java
+++ b/src/main/java/org/assertj/core/internal/Strings.java
@@ -337,14 +337,14 @@ public class Strings {
    * @param expected the expected {@code CharSequence}.
    * @throws AssertionError if the given {@code CharSequence}s are not equal.
    */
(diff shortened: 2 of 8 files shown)


src/main/java/org/assertj/core/internal/Strings.java
```diff
@@ -330,6 +330,44 @@ private boolean areEqualIgnoringCase(CharSequence actual, CharSequence expected)
   }
 
   /**
+   * Verifies that two {@code CharSequence}s are equal, ignoring case considerations.
```
mkordas: It's not about `case considerations`

src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
```diff
@@ -708,4 +708,8 @@ public S inUnicode() {
     return myself;
   }
 
+  public S isEqualToIgnoringWhitespaces(CharSequence expected) {
```
joel-costigliola: should be `isEqualToIgnoringWhitespace` (not Whitespace**s**)

src/main/java/org/assertj/core/internal/Strings.java
```diff
@@ -330,6 +330,44 @@ private boolean areEqualIgnoringCase(CharSequence actual, CharSequence expected)
...
+   *
+   * @param info contains information about the assertion.
+   * @param actual the actual {@code CharSequence}.
+   * @param expected the expected {@code CharSequence}.
+   * @throws AssertionError if the given {@code CharSequence}s are not equal.
+   */
+  public void assertEqualsIgnoringWhitespaces(AssertionInfo info, CharSequence actual, CharSequence expected) {
```
joel-costigliola: `assertEqualsIgnoringWhitespace`

src/main/java/org/assertj/core/internal/Strings.java
```diff
@@ -330,6 +330,44 @@ private boolean areEqualIgnoringCase(CharSequence actual, CharSequence expected)
...
+      if (areEqualIgnoringWhitespaces(actual, expected)) {
```
joel-costigliola: can be simplified (no `return;` needed) with:
```java
if (!areEqualIgnoringWhitespaces(actual, expected)) {
   throw failures.failure(info, shouldBeEqual(actual, expected));
}

Comment (AlexBischof) [inline src/main/java/org/assertj/core/internal/Strings.java]: I have oriented myself on the method assertEqualsIgnoringCase. Should this method also be simplified? Otherwise it feels i little bit inconsistent.

**Comment by joel-costigliola** [comment]
@AlexBischof  `CharSequenceAssert_isEqualToIgnoringWhitespace_Test` is supposed to verify that calling `isEqualToIgnoringWhitespace` will call call `assertEqualsIgnoringWhitespace` with the same expected String but you use `" my\tfoo bar "` and `" my foo bar "` which are different.
Use the same String it will work.

File: src/main/java/org/assertj/core/internal/Strings.java (hunk 4 above)
Comment (joel-costigliola) [inline src/main/java/org/assertj/core/internal/Strings.java]: It should but not as part of this PR.

**Comment by AlexBischof** [comment]
Thank you, now i understand how the design works.

File: src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java [hunk 5]
```diff
@@ -708,4 +708,8 @@ public S inUnicode() {
     return myself;
   }
 
+  public S isEqualToIgnoringWhitespace(CharSequence expected) {
```
joel-costigliola: Javadoc (with code examples) missing

src/main/java/org/assertj/core/internal/Strings.java
```diff
@@ -330,6 +331,47 @@ private boolean areEqualIgnoringCase(CharSequence actual, CharSequence expected)
...
   /**
+   * Verifies that two {@code CharSequence}s are equal, ignoring any changes in whitespace.
+   *
+   * @param info contains information about the assertion.
+   * @param actual the actual {@code CharSequence}.
+   * @param expected the expected {@code CharSequence}.
+   * @throws AssertionError if the given {@code CharSequence}s are not equal.
+   */
+  public void assertEqualsIgnoringWhitespace(AssertionInfo info, CharSequence actual, CharSequence expected) {
+      if (!areEqualIgnoringWhitespace(actual, expected)) {
```
joel-costigliola: should check that `actual` is not null with `assertNotNull(info, actual);` (I know it's missing in `assertEqualsIgnoringCase`).

src/main/java/org/assertj/core/internal/Strings.java
```diff
@@ -330,6 +331,47 @@ private boolean areEqualIgnoringCase(CharSequence actual, CharSequence expected)
...
+        throw failures.failure(info, shouldBeEqual(actual, expected));
+      }
+  }
+
+  private boolean areEqualIgnoringWhitespace(CharSequence actual, CharSequence expected) {
+      if (actual == null) {
```
joel-costigliola: with the previous comment, `actual` is not null but if `expected` is, let's return `false`

src/main/java/org/assertj/core/internal/Strings.java
```diff
@@ -330,6 +331,47 @@ private boolean areEqualIgnoringCase(CharSequence actual, CharSequence expected)
...
+          return expected == null;
+      }
+
```
joel-costigliola: remove blank lines

src/main/java/org/assertj/core/internal/Strings.java
```diff
@@ -330,6 +331,47 @@ private boolean areEqualIgnoringCase(CharSequence actual, CharSequence expected)
...
+
+      return stripSpace(actual).equals(stripSpace(expected));
+  }
+
+  private String stripSpace(CharSequence toBeStripped) {
```
joel-costigliola: why not just remove all whitespaces ? this will simplify this method.
if I'm right, we should rename it to `removeAllWhitespaces`

src/test/java/org/assertj/core/internal/strings/Strings_assertEqualsIgnoringWhitespace_Test.java
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * Copyright 2012-2015 the original author or authors.
+ */
+package org.assertj.core.internal.strings;
+
+import org.assertj.core.api.AssertionInfo;
+import org.assertj.core.internal.StringsBaseTest;
+import org.junit.Test;
+
+import static org.assertj.core.error.ShouldBeEqualIgnoringCase.shouldBeEqual;
```
joel-costigliola: this is not the expected error, it should be `ShouldBeEqualIgnoringWhitespace`, let's create this class along with its unit test.

src/test/java/org/assertj/core/internal/strings/Strings_assertEqualsIgnoringWhitespace_Test.java
```diff
@@ -0,0 +1,82 @@
...
+import static org.mockito.Mockito.verify;
+
+/**
+ * Tests for <code>{@link org.assertj.core.internal.Strings#assertEqualsIgnoringWhitespace(org.assertj.core.api.AssertionInfo, CharSequence, CharSequence)} </code>.
+ *
+ * @author Alex Ruiz
+ * @author Joel Costigliola
+ * @author Alexander Bischof
+ */
+public class Strings_assertEqualsIgnoringWhitespace_Test extends StringsBaseTest {
```
joel-costigliola: missing test: should_fail_if_actual_is_not_null_and_expected_is

src/test/java/org/assertj/core/internal/strings/Strings_assertEqualsIgnoringWhitespace_Test.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  @Test
+  public void should_pass_if_both_Strings_are_the_same() {
+    String s = "Yoda";
+    strings.assertEqualsIgnoringWhitespace(someInfo(), s, s);
+  }
+
+  @Test
+  public void should_pass_if_both_Strings_are_equal_but_not_same() {
```
joel-costigliola: It seems that test verifyies the same thing as `should_pass_if_both_Strings_are_equal_ignoring_whitespace`

src/test/java/org/assertj/core/internal/strings/Strings_assertEqualsIgnoringWhitespace_Test.java
```diff
@@ -0,0 +1,82 @@
...
+    strings.assertEqualsIgnoringWhitespace(someInfo(), " my\tfoo bar ", new String(arrayOf(' ', 'm', 'y', ' ', 'f', 'o', 'o', ' ', 'b', 'a', 'r')));
+  }
+
+  @Test
+  public void should_pass_if_both_Strings_are_equal_ignoring_whitespace() {
```
joel-costigliola: there should be more successful cases, not just one where strings differ by one character, ex:
* `" "` vs `"      \t  \t"`
* `"  abc"` vs `"abc "`
* ...
a parametrized test is the best option to write this kind of test

joel-costigliola:
Integrated, thanks

Issue #356, opened 2015-04-01
assert matchers for java.net.URI/java.net.URL


tokuhirom:
I want an assert matcher for java.net.UR[IL].

joel-costigliola:
Ok, what assertions ?
Can you contribute ?

tokuhirom:
e.g. `assertThat(url).hasSchemeEqualsTo("http").hasPathEquals("/path/to/foo")`.

joel-costigliola:
Let's keep it simple and use: `assertThat(url).hasScheme("http").hasPath("/path/to/foo")`.

AlexBischof:
The current design of my branch uses only UriAssert. Any URL's can be asserted via a convenience method in Assertions so that there is no code duplication. As a drawback the following test code would fail:

      AbstractUriAssert<?> assertions = Assertions.assertThat(url);
      assertThat(assertions.actual).isSameAs(url);

Thoughts?

PascalSchumacher:
+1 for adding these assertions

PR #360, opened 2015-04-03
#356 Assertions for URI and URL

AlexBischof:
Created structures and packages for the issue containing also the assertions hasScheme and hasPathEquals (test classes included). But one problems remains with the core api tests because java.net.URL cannot be mocked because of final class. I guess that powermock or JMockit is not an option so how should i proceed?


joel-costigliola:
Thanks Alex.
We should stick to the JDK API and use `hasPath` instead of `hasPathEquals`

AlexBischof:
In the issue the name was hasPathEquals ;)

joel-costigliola:
Yep, this is something that I commented in #356

joel-costigliola:
Fixed after much additional work to support URL.

Issue #240, opened 2014-09-19
rename file/stream assertion hasContentEqualTo to hasSameContentAs. 


joel-costigliola:
* deprecate ̀hasContentEqualTo ̀ 
* rename  ̀ShouldHaveEqualContent ̀ to  ̀ShouldHaveSameContent ̀

PR #362, opened 2015-04-04
#240 rename file/stream assertion hasContentEqualTo to hasSameContentAs

AlexBischof:
Simple refactoring as described in the issue #240. hasContentEqualTo is marked as deprecated and uses internally hasSameContentAs


joel-costigliola:
Good PR, thanks. Integrated

Issue #365, opened 2015-04-07
Asserting for interfaces resulting in Ambiguous method error under Java 8


mariuszs:
```java
import org.assertj.core.api.Assertions;

public class Ambiguous {

    public static void main(String... args) {
        Assertions.assertThat(caughtException());
    }

    public static <E extends Exception> E caughtException() {
        return null;
    }
}
```

Error message (for Java 8 only!):

    javac -source 1.8 Ambiguous.java  -classpath assertj-core-3.0.0.jar                                                                     
    Ambiguous.java:6: error: reference to assertThat is ambiguous
            Assertions.assertThat(caughtException());
                      ^
      both method assertThat(CharSequence) in Assertions and method assertThat(Throwable) in Assertions match
    1 error

Explanation:

http://stackoverflow.com/questions/29499847/ambiguous-method-in-java-8-why/29500086#29500086

Maybe we can change this API, deprecate conflicting methods and remove it in future releases?

mariuszs:
title misspelled

mariuszs:
title misspelled

joel-costigliola:
Fixed I hope.
I have let other method `assertThat` that took interface or generic type extending interface.

joel-costigliola:
Oh ... and thanks !

joel-costigliola:
@mariuszs Can you give it a try ? I'm about to release 3.1.0 this week.

mariuszs:
Nice :) Sorry for delay from my site.

This code works for me:

```java
import org.assertj.core.api.StrictAssertions;

public class Ambiguous {

    public static void main(String... args) {
        StrictAssertions.assertThat(caughtException());
    }

    public static <E extends Exception> E caughtException() {
        return null;
    }
}
```

mariuszs:
This code works too:

```java
    @Test
    public void bdd() throws Exception {

        // given: an empty list
        List myList = new ArrayList();

        // when: we try to get the first element of the list
        when(() -> myList.get(1));

        // then: we expect an IndexOutOfBoundsException
        StrictAssertions.assertThat(caughtException())
                .isInstanceOf(IndexOutOfBoundsException.class)
                .hasMessage("Index: 1, Size: 0")
                .hasNoCause();
    }
```

Maybe we need also something like BDDStrictExceptions

joel-costigliola:
sweet, thanks for the feedback

PR #366, opened 2015-04-08
Fix/strict assertions

mariuszs:
Proposed fix for #365


joel-costigliola:
@mariuszs  it is almost impossible to see what has really changed :(
what did you change exactly ?

mariuszs:
OK :) The idea was to leave all assertions for not ambiguous types into StrictAssert with other code. Ambiguous assertions are leaved in Assertions. So, If someone is using AssertJ with Java8 and has Ambigouos error during compile, then can simply change Assertions with StrictAssertions. 

This for example can help with catch-exception and allows for writing custom assertion for business exception (`assertThat(ex).hasErrorCode(300)`) etc.

src/main/java/org/assertj/core/api/StrictAssertions.java
```diff
@@ -0,0 +1,1240 @@
...
+import org.assertj.core.data.MapEntry;
+import org.assertj.core.data.Offset;
+import org.assertj.core.groups.Properties;
+import org.assertj.core.groups.Tuple;
+import org.assertj.core.util.Files;
+import org.assertj.core.util.FilesException;
+import org.assertj.core.util.introspection.FieldSupport;
+
+/**
+ * Entry point for assertion methods for different data types. Each method in this class is a static factory for the
```
joel-costigliola: Explain why we need this class, otherwise it is not easy to understand, using the example from the PR comment would be a good idea.

PR #368, opened 2015-04-11
Fixes #352 : Add hasSameTimeAs(String dateAsString) assertion

### LINKED ISSUE #352: Add hasSameTimeAs(String dateAsString) assertion
```
assertThat(Date date).hasSameTimeAs(String dateAsString)
```

mkordas:


src/main/java/org/assertj/core/api/AbstractDateAssert.java
```diff
@@ -2188,6 +2188,51 @@ public S hasSameTimeAs(Date date) {
...
   /**
+   * Same assertion as {@link #hasSameTimeAs(Date)} but given date is represented as String either with one of the
+   * supported defaults date format or a user custom date format (set with method {@link #withDateFormat(DateFormat)}).
+   * <p/>
+   * Beware that the default formats are expressed in the current local timezone.
+   * <p/>
+   * Example:
+   *
+   * <pre><code class='java'>
+   * // assertion will pass
```
joel-costigliola: the code examples are not helpful if you don't define date, moreover I'm pretty sure that one the "succeeding" assertion fails.

src/test/java/org/assertj/core/api/date/DateAssert_hasSameTimeAsOtherDate_Test.java
```diff
@@ -0,0 +1,59 @@
...
+  public ExpectedException thrown = none();
+
+  @Test
+  public void should_verify_that_actual_has_time_equals_to_expected() {
+	Date date = new Date();
+	Timestamp timestamp = new java.sql.Timestamp(date.getTime());
+	assertThat(date).hasSameTimeAs(timestamp);
+	assertThat(timestamp).hasSameTimeAs(date);
+  }
+
```
joel-costigliola: I would add a test that takes String parameter

src/main/java/org/assertj/core/api/AbstractDateAssert.java
```diff
@@ -2188,6 +2188,51 @@ public S hasSameTimeAs(Date date) {
...
+   * @return this assertion object.
+   * @throws AssertionError if the actual {@code Date} is {@code null}.
+   * @throws NullPointerException if given date as String is {@code null}.
+   * @throws AssertionError if the actual {@code Date} time is not equal to the time from date represented as
+   *           String.
+   * @throws AssertionError if the given date as String could not be converted to a Date.
+   * @author Michal Kordas
+   */
+  public S hasSameTimeAs(String dateAsString) {
+	dates.hasSameTimeAs(info, actual, parse(dateAsString));
```
joel-costigliola: the implementation is good, I think we could improve the error message which looks like:
```
Expecting
 <2003-04-26T13:01:02>
to have time:
 <1051318861000L>
but was:
 <1051318862000L>
```
I would prefer a message that shows we are comparing the time's of two dates, for example

```
Expecting
 <2003-04-26T13:01:01>
and:
 <2003-04-26T13:01:02>
to represent the same time but was not.
Actual's time was:
 <1051318861000L>
and expected:
 <1051318862000L>
```

mkordas: Done

src/test/java/org/assertj/core/api/date/DateAssert_hasSameTimeAsOtherDate_Test.java
mkordas: Added new file with dedicated tests only for `String` version

src/main/java/org/assertj/core/api/AbstractDateAssert.java
mkordas: Indeed, fixed

src/main/java/org/assertj/core/error/ShouldHaveSameTime.java
```diff
@@ -0,0 +1,38 @@
...
+     * @param actual the actual value in the failed assertion.
+     * @param expected the expected timestamp.
+     * @return the created {@code ErrorMessageFactory}.
+     */
+    public static ErrorMessageFactory shouldHaveSameTime(Date actual, Date expected) {
+        return new ShouldHaveSameTime(actual, expected);
+    }
+
+    private ShouldHaveSameTime(Date actual, Date expected) {
+        super("%nExpecting%n <%s>%nto represent same time as:%n <%s>%nbut the time was different.%nActual time is%n "
```
joel-costigliola: Minor change: use two spaces indentation after a new line, I know not all messages use that but it's something that I would like to, so for new error messages let's use that indentation.

mkordas: Done

joel-costigliola:
Integrated, thanks.

PR #371, opened 2015-04-14
refactored extractor to java 8 function

lpandzic:
Alternative approach is to make Extractor extend Function and make extract a default function.


joel-costigliola:
Besides using standard JDK classes, is there any advantage to use Function over Extractor ?

The main point for me is to be able to use a lambda to write extractors.

I like Extractor because it expresses well the intention (`extract` method vs `apply`), my other concern is too keep `master`  and `java-8` branches close to each other to have easy merge from `master` to `java-8` (so that AssertJ for java 8 gets improvements done in `master` branch).

lpandzic:
Yes, the main point is to use JDK classes.
The other advantage is that you can use default methods provided in Function, but this can also be achieved by simply extending Function and making extract method default (extract points to apply in order to keep Extractor a functional interface).

joel-costigliola:
closed after deleting java8 branch

joel-costigliola:
closing this PR, not because it's wrong but not proven useful for the moment (might change later).

lpandzic:
Ok, np.

PR #375, opened 2015-04-17
Fixes #374 : Use strict instead of lenient date parsing by default

### LINKED ISSUE #374: Use strict instead of lenient date parsing mode by default
While parsing String representing the date, [setLenient(false)](http://docs.oracle.com/javase/7/docs/api/java/text/DateFormat.html#setLenient(boolean)) mode should be used.

It solves some date parsing anomalies, such as allowing January 32nd to be interpreted as February first or even January -1st as December 31st. More examples can be found [here](http://stackoverflow.com/questions/7606387/what-is-the-use-of-lenient).

The commit message should say that it's a braking change so that when I do the release notes I won't forget it.

We need a way to set the parsing leniency so that users can switch to the old behaviour. The (new) default behaviour being not to be lenient.

##### Design question
* Should this be static and global AssertJ parameter? Then we need to add a method `setStrictDateParsing(boolean)` in Assertions delegating the real work to the same method in `AbstractDateAssert`.
* Otherwise we may add e.g. `usingLenientDateParsing()` to `AbstractDateAssert` to parse date leniently just for one particular assertion.

mkordas:
This change breaks backward compatibility.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractDateAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractDateAssert.java
@@ -2345,6 +2345,45 @@ public abstract class AbstractDateAssert<S extends AbstractDateAssert<S>> extend
   }
 
   /**
+   * Instead of using default strict date/time parsing it is possible to enforce mode when for default date formats
+   * parser uses heuristics to interpret inputs that do not precisely match supported date formats (lenient parsing).
+   *
+   * With strict parsing, inputs must match exactly date/time format.
+   *
+   * Example:
+   *
+   * <pre><code class='java'>
+   * final Date date = Dates.parse("2001-02-03");
+   * final Date dateTime = parseDatetime("2001-02-03T04:05:06");
+   * final Date dateTimeWithMs = parseDatetimeWithMs("2001-02-03T04:05:06.700");
+   *
+   * AbstractDateAssert.setLenientDateParsing(true);
+   *
+   * // assertions will pass
+   * assertThat(date).isEqualTo("2001-01-34");
+   * assertThat(date).isEqualTo("2001-02-02T24:00:00");
+   * assertThat(date).isEqualTo("2001-02-04T-24:00:00.000");
+   * assertThat(dateTime).isEqualTo("2001-02-03T04:05:05.1000");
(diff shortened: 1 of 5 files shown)


src/main/java/org/assertj/core/api/AbstractDateAssert.java
```diff
@@ -2345,6 +2345,45 @@ public S withDateFormat(String userCustomDateFormatPattern) {
...
+   * assertThat(date).hasSameTimeAs("2001-02-04"); // different date
+   * assertThat(dateTime).hasSameTimeAs("2001-02-03 04:05:06"); // leniency does not help here
+   * </code></pre>
+   *
+   * To revert to default strict date parsing, call {@code useLenientDateParsing(false)}.
+   *
+   * @param value whether lenient parsing mode should be enabled or not
+   * @author Michal Kordas
+   */
+  public static void useLenientDateParsing(boolean value) {
```
joel-costigliola: I prefer to rename `useLenientDateParsing` to `setLenientDateParsing`.
When I read `useLenientDateParsing` the first thing that comes to my mind is that the parsing will be lenient.

src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -1193,6 +1193,43 @@ public static String contentOf(File file) {
...
+   * assertThat(date).hasSameTimeAs("2001-02-04"); // different date
+   * assertThat(dateTime).hasSameTimeAs("2001-02-03 04:05:06"); // leniency does not help here
+   * </code></pre>
+   *
+   * To revert to default strict date parsing, call {@code useLenientDateParsing(false)}.
+   *
+   * @param value whether lenient parsing mode should be enabled or not
+   * @author Michal Kordas
+   */
+  public static void useLenientDateParsing(boolean value) {
```
joel-costigliola: Rename to `setLenientDateParsing`.

src/main/java/org/assertj/core/api/AbstractDateAssert.java
mkordas: Agreed, done

joel-costigliola:
Integrated, thanks.

PR #382, opened 2015-04-23
Fix 381

### LINKED ISSUE #381: Extraction of nested field/property combination fails
Today I discovered that extraction of field/property combinations fails for `AbstractIterableAssert`. It either expects properties or fields. It seem to fail for altering combinations.

I depend on a class from a lib from Google, which has a public field `file` of type `File`. If I want to extract `file.name`, where name is a method (`File#getName()`), I get the following exception:
>org.assertj.core.util.introspection.IntrospectionError: 
>Can't find any field or property with name 'file.name'.
>Error when introspecting fields was :
>No getter for property 'file' in com.android.tools.lint.Warning 

Here is some example code to demonstrate the issue:
```java
List<Warning> warnings = lint.getWarnings();   
assertThat(warnings).extracting("file.name").containsExactly("AndroidManifest.xml");
```

By the way, it succeeds when I try `file.path`, cause `path`is a field of type `File`.

A look at assertj's `ByNameSingleExtractor` class seems to support my assumption (see line 48-52).

So my question is, if this is on purpose or if it is an issue?

Because when looking at other "path reflection" solutions, for instance at Apache's commons-lang, the resolution of nested properties always accepts anything, fields, getters, indexes, maps...

a11n:
Fixes #381.

I tried to keep changes to a minimum. I only needed to extend some of the test fixtures in order to test "altering nested property/field combinations".

Furthermore I decided to copy some stuff into ByNameSingleExtractor.java to avoid a variety of changes when for instance moving PropertySupport to the Introspection package.

I used the code style mentioned in CONTRIBUTING.md ... however I noticed that ByNameSingleExtractor.java seems to be not formatted conforming to that style. However I DID NOT reformat the code in order to avoid "unrelated changes".

### DIFF
--- a/src/main/java/org/assertj/core/extractor/ByNameSingleExtractor.java
+++ b/src/main/java/org/assertj/core/extractor/ByNameSingleExtractor.java
@@ -22,6 +22,8 @@ import org.assertj.core.util.introspection.IntrospectionError;
 import java.util.Map;
 
 class ByNameSingleExtractor<T> implements Extractor<T, Object> {
+		private static final String SEPARATOR = ".";
+		
   private final String propertyOrFieldName;
 
   ByNameSingleExtractor(String propertyOrFieldName) {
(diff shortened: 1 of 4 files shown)


a11n:
Please let me know if you'd like me to squash the commits into one. I let them separated in order to highlight the changes.

joel-costigliola:
Thanks for the contribution, I will review it shortly.

joel-costigliola:
Integrated, I have just moved the test to `ByNameSingleExtractorTest`.
Thanks !

PascalSchumacher:
I do not know if this intended, but a side-effect? of this pull request is that AbstractIterableAssert#usingElementComparatorOnFields() now works for properties without a backing field.

a11n:
Thank you for bringing this up. I will investigate this.

joel-costigliola:
`usingElementComparatorOnFields` supports properties not backed by a field or a field (which can be private, javadoc is wrong on this last point).
It does not support nested values mixing fields and properties (which is what #381 fixed on extraction).

For consistency sake's and intuitive behaviour, all comparison by value strategies should support mixed nested field/property.

PR #383, opened 2015-04-24
replace deprecated junit.framework.Assert class with org.junit.Assert

PascalSchumacher:


joel-costigliola:
I won't merge this PR as I would prefer use AssertJ assertions instead of JUnit.
I haven't taken the time to do it with the JUnit to AssertJ script but this is in my todolist (#386)

PR #388, opened 2015-04-25
Fixes #367 : Add isFinal and isNotFinal assertions for Class objects

### LINKED ISSUE #367: Feature request: More assertions on Class objects
Current code to assert that class is final:
```
import static java.lang.reflect.Modifier.isFinal;

final boolean isClassFinal = isFinal(utilClass.getModifiers());
assertThat(isClassFinal).isTrue();
```
After introducing additional assertions:
```
assertThat(utilClass).isFinal();
```

mkordas:


joel-costigliola:
Thanks, I have just added the missing test for `ShouldBeFinal` (and slightly improved the error message).

mkordas:
Thank you a lot. Indeed, I've missed this test and now error message is a bit better.

PR #397, opened 2015-05-02
Fixes #394 : Add isEqualByComparingTo and isNotEqualByComparingTo for Comparables

### LINKED ISSUE #394: Feature request: assertion to check that two comparable objects are equal
Sample class:
```java
  class A implements Comparable<A> {
    @Override public int compareTo(A o) {
      return 0;
    }
  }
```
This assertion passes:
```java
    assertThat(new A().compareTo(new A())).isZero();
```
Would be nice to have more fluent assertion for that:
```java
    assertThat(new A()).equalsTo(new A());
```
`equalsTo` is just a guess. I was thinking also about `comparesTo`, `comparesAsEqualTo`. I'm open to other ideas.

mkordas:


joel-costigliola:
Thanks !
As `isEqualByComparingTo` and `isNotEqualByComparingTo` are now available for all subclasses of `AbstractComparableAssert`, we should add the corresponding unit tests like `BigDecimals_assertNotEqualByComparison_Test` and `BigDecimals_assertEqualByComparison_Test`.

Lot of work, I know !

mkordas:
In `BigDecimals_assertEqualByComparison_Test` and `BigDecimals_assertNotEqualByComparison_Test` I see tests for static methods of `BigDecimals`. Same is for `Bytes`, `Characters` and so on. In my code I don't add any specific methods to each type, so what am I supposed to test?

joel-costigliola:
forget my comment ... I wanted to make sure that it worked for double, float, ... but I think we can avoid to write tests for that as it is already covered by `Comparables_assertEqualByComparison_Test`

mkordas:
Agreed :)

src/main/java/org/assertj/core/api/ComparableAssert.java
```diff
@@ -25,6 +25,24 @@
 public interface ComparableAssert<S extends ComparableAssert<S, A>, A extends Comparable<? super A>> {
 
   /**
+   * Verifies that the actual value is equal to the given one by invoking <code>{@link Comparable#compareTo(Object)}</code>.
+   * @param other the given value to compare the actual value to.
```
joel-costigliola: add a code example

src/main/java/org/assertj/core/api/ComparableAssert.java
```diff
@@ -25,6 +25,24 @@
...
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual value is {@code null}.
+   * @throws AssertionError if the actual value is not equal to the given one.
```
joel-costigliola: equal is not used in that case

src/main/java/org/assertj/core/api/ComparableAssert.java
```diff
@@ -25,6 +25,24 @@
...
+  S isEqualByComparingTo(A other);
+
+  /**
+   * Verifies that the actual value is not equal to the given one by invoking <code>{@link Comparable#compareTo(Object)}</code>.
+   * @param other the given value to compare the actual value to.
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual value is {@code null}.
+   * @throws AssertionError if the actual value is equal to the given one.
+   */
+  S isNotEqualByComparingTo(A other);
```
joel-costigliola: same javadoc comment as previous

joel-costigliola:
I would be nice to add an example to assertj-examples it you find the time.

mkordas:
I've added examples to the Javadoc and created PR for assertj-examples: joel-costigliola/assertj-examples#21

joel-costigliola:
Thanks, good work.

PR #401, opened 2015-05-05
#4 : Add missing javadoc code examples for DoubleAssert.

PascalSchumacher:


joel-costigliola:
Thanks.
I don't think it adds a lot of value to use `Double.valueOf()` in the examples.
Can you update the example using `double` directly ?

Cheers

PascalSchumacher:
I was not sure about the value of `Double.valueOf()` myself.

Pro: The examples actually test something (a bit more realistic).
Con: The examples are more complicated.

I guess it's good to make the examples as simple as possible. I will update the pull request.

Cheers, Pascal

joel-costigliola:
Yeah, I agree that the examples related to `isEqualTo` and `isNotEqualTo` may seem "dumb" but it does not hurt.

joel-costigliola:
Integrated, thx.

PR #402, opened 2015-05-06
Added Resources class for resource loading from URLs

dorzey:
Addressed the feedback on https://github.com/joel-costigliola/assertj-core/pull/280 and moved the work done by @Turbo87

### DIFF
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -1219,23 +1219,23 @@ public class Assertions {
 
   /**
    * Loads the text content of a URL, so that it can be passed to {@link #assertThat(String)}. <p> Note that this will
-   * load the entire file in memory. </p>
+   * load the entire contents in memory. </p>
    *
    * @param url the URL.
    * @param charset the character set to use.
-   * @return the content of the file.
+   * @return the content of the URL.
    * @throws NullPointerException if the given charset is {@code null}.
-   * @throws FilesException       if an I/O exception occurs.
+   * @throws FilesException   if an I/O exception occurs.
    */
   public static String contentOf(URL url, Charset charset) { return URLs.contentOf(url, charset); }
 
   /**
    * Loads the text content of a URL, so that it can be passed to {@link #assertThat(String)}. <p> Note that this will
-   * load the entire file in memory. </p>
+   * load the entire contents in memory. </p>
    *
    * @param url the URL.
    * @param charsetName the name of the character set to use.
-   * @return the content of the file.
+   * @return the content of the URL.
    * @throws IllegalArgumentException if the given character set is not supported on this platform.
    * @throws FilesException           if an I/O exception occurs.
    */
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -39,7 +39,7 @@ import org.assertj.core.groups.Properties;
 import org.assertj.core.groups.Tuple;
 import org.assertj.core.util.Files;
 import org.assertj.core.util.FilesException;
-import org.assertj.core.util.Resources;
+import org.assertj.core.util.URLs;
 import org.assertj.core.util.introspection.FieldSupport;
 
 /**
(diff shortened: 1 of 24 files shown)


Turbo87:
awesome, thanks! 

Just one little remark: the Resources class actually does no resource loading at all and only operates on URLs/Streams, so it might be better to call it `URLs` instead. I would expect the Resources class instead to accept a string and do the ClassLoader calls for me. Obviously that can't be exported in the Assertions class because it would conflict with the file loading methods, but I think it would still be useful.

joel-costigliola:
Good point indeed, let's go for `URLs`.

dorzey:
Renamed. Is there anything else that needs addressing?

Turbo87:
I think this is good to merge, but as I mentioned in the last comment it would be cool to have something like the following too:

```java
public class Resources {
    public static String contentOf(String resource) {
        return URLs.contentOf(ClassLoader.getSystemResource(resource));
    }

    // ...
}
```

```java
String resource = Resources.contentOf("some-resource.json");
```

joel-costigliola:
@Turbo87 well why not but let's remember that assertj is an assertion library not a IO utils lib.

Turbo87:
My plan is to use it roughly like this:

```java
assertThat(result).isEqualTo(Resources.contentOf("test1.json"))
```

joel-costigliola:
In that case, we could introduce a new String assertions:
```java
assertThat(result).isEqualToContentOfResource("test1.json");
assertThat(result).isEqualToContentOfURL("htpp://foo.com/tests/test1.json");
assertThat(result).isEqualToContentOfFile(test1.json);
```

dorzey:
Can this be merged? And we raise a separate issue for the extensions suggested.

Turbo87:
:+1: from my side

src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -1211,6 +1214,89 @@ public static String contentOf(File file) {
   }
 
   // --------------------------------------------------------------------------------------------------
+  // URL/Resource methods : not assertions but here to have a single entry point to all AssertJ features.
+  // --------------------------------------------------------------------------------------------------
+
+  /**
+   * Loads the text content of a URL, so that it can be passed to {@link #assertThat(String)}. <p> Note that this will
+   * load the entire file in memory. </p>
```
joel-costigliola: Javadoc should not mention file.

src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -1211,6 +1214,89 @@ public static String contentOf(File file) {
...
+   *
+   * @param url the URL.
+   * @param charset the character set to use.
+   * @return the content of the file.
+   * @throws NullPointerException if the given charset is {@code null}.
+   * @throws FilesException       if an I/O exception occurs.
```
joel-costigliola: let's rename `FilesException` to `RuntimeIOException`, URL are not only about file.

src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -1211,6 +1214,89 @@ public static String contentOf(File file) {
...
+  // URL/Resource methods : not assertions but here to have a single entry point to all AssertJ features.
+  // --------------------------------------------------------------------------------------------------
+
+  /**
+   * Loads the text content of a URL, so that it can be passed to {@link #assertThat(String)}. <p> Note that this will
+   * load the entire file in memory. </p>
+   *
+   * @param url the URL.
+   * @param charset the character set to use.
+   * @return the content of the file.
```
joel-costigliola: Idem.

src/test/java/org/assertj/core/util/URLs_contentOf_Test.java
```diff
@@ -0,0 +1,62 @@
...
+    assertFalse(missingFile.exists());
+
+    thrown.expect(FilesException.class);
+    URLs.contentOf(missingFile.toURI().toURL(), Charset.defaultCharset());
+  }
+
+  @Test
+  public void should_load_resource_from_url_using_charset() {
+    // NB: UTF-8 must be supported by every Java implementation
+    assertEquals(expectedContent, URLs.contentOf(sampleResourceURL, Charset.forName("UTF-8")));
```
joel-costigliola: use `assertThat`

joel-costigliola:
Just minor comments but otherwise PR is ok.

joel-costigliola:
Integrated, thx !

Issue #379, opened 2015-04-20
Add support for primitive Optional classes


jodastephen:
The Java 8 release contains support for `Optional` (yay!). But it does not appear to have support for `OptionalInt`, `OptionalLong` or `OptionalDouble`. It would be great to have support for these.

See [our version](https://github.com/OpenGamma/OG-Commons/blob/master/modules/collect/src/test/java/com/opengamma/collect/OptionalDoubleAssert.java) for `OptionalDouble`.

joel-costigliola:
Indeed they are missing.

When comparing double with a tolerance, we use `within` syntatic sugar to improve readability, following the same approach for `OptionalDouble` would give:
```java
assertThat(OptionalDouble.of(10.0)).isCloseTo(8.0, within(2.0));
```
Would you be happy with that ?

jodastephen:
In our version (linked above) we have 

* isPresent()
* isEmpty()
* hasValue(double expected)
* hasValue(double expected, Offset<Double> tolerance)
* hasValue(double expected, double tolerance)

The method name `isCloseTo` with a tolerance seems fine to us.

AlexBischof:
I would take it.

joel-costigliola:
@AlexBischof to be clear for `OptionalDouble` value check, let's define these assertions: 

```java
assertThat(OptionalDouble.of(8.0)).hasValue(8.0);

// within is a helper method to create an Offset
assertThat(OptionalDouble.of(10.0)).hasValueCloseTo(8.0, within(2.0));
```

AlexBischof:
Thanks for clarifying. 

Finally i have a PR-ready branch but wondering about the code duplication (e.g. OptionalXXXShouldBeEmpty, OptionalXXXShouldBePresent, AbstractOptionalXXXAssert etc.). 

It should be no problem to composite the error and AbstractOptionalXXXAssert classes. Is there anything i should keep in mind

joel-costigliola:
You can add new methods to `OptionalShouldBeEmpty` for `OptionalInt`, `OptionalLong` or `OptionalDouble`, no need to create another class (don't forget to add the corresponding unit tests in `OptionalShouldBeEmpty_create_Test`).

ex:
```public static OptionalShouldBeEmpty shouldBeEmpty(OptionalInt optional)```

joel-costigliola:
@AlexBischof what about your PR ? is it ready ?

AlexBischof:
[...]

PR #403, opened 2015-05-06
Add support for primitive Optional classes #379

AlexBischof:
This PR contains an extension of the API for OptionalInt, OptionalLong and OptionalDouble as described in the corresponding issue. It contains tests and documentation and I have oriented myself on existing code. I also extended BDDAssertions as well as Softassertions for those classes.

Here are some examples:
```java
assertThat(OptionalDouble.of(8)).hasValueCloseTo(8.0, within(0d));
assertThat(OptionalDouble.of(8)).hasValue(8.0);

assertThat(OptionalInt.of(7)).hasValue(7);
assertThat(OptionalLong.of(7)).hasValue(7);

assertThat(OptionalInt.empty()).isEmpty();
```

One thing seems not right, yet. The generic class Optional contains a value whereas an OptionalXXX has a value. 
```java
assertThat(Optional.of("Hello World")).contains("Hello World");
assertThat(OptionalDouble.of(8)).hasValue(8.0);
```
It is not konsistent, is it?

Do you need more help on that?


src/main/java/org/assertj/core/api/AbstractOptionalIntAssert.java
```diff
@@ -0,0 +1,128 @@
...
+     * @throws AssertionError if actual value is empty.
+     * @throws AssertionError if actual is null.
+     * @throws AssertionError if actual has not the value as expected.
+     */
+    public S hasValue(int expectedValue) {
+        isNotNull();
+        if (!actual.isPresent()) throw failure(shouldContain(expectedValue));
+        //Reuses integers functionality, catches possible assertionerror and throw correct one
+        try {
+            integers.assertEqual(info, actual.getAsInt(), expectedValue);
```
joel-costigliola: In that case, just comparing `int` is easier:
```java
if (expectedInt != actual.getAsInt()) throw Failures.instance().failure(info, shouldContain(actual, expectedValue));
```

src/main/java/org/assertj/core/api/AbstractOptionalLongAssert.java
```diff
@@ -0,0 +1,128 @@
...
+ * Assertions for {@link java.util.OptionalLong}.
+ *
+ * @author Jean-Christophe Gay
+ * @author Alexander Bischof
+ */
+public abstract class AbstractOptionalLongAssert<S extends AbstractOptionalLongAssert<S>> extends
+    AbstractAssert<S, OptionalLong> {
+
+    @VisibleForTesting
+    Integers integers = Integers.instance();
```
joel-costigliola: `Integers` in `AbstractOptionalLongAssert` ??? 
Anyway, we don't need that as per my previous comment.

src/main/java/org/assertj/core/error/OptionalShouldBeEmpty.java
```diff
@@ -23,18 +24,48 @@
  */
 public class OptionalShouldBeEmpty extends BasicErrorMessageFactory {
 
-  private OptionalShouldBeEmpty(Object optionalValue) {
-	super("%nExpecting an empty Optional but was containing value: <%s>.", optionalValue);
-  }
+    private OptionalShouldBeEmpty(Object optionalValue) {
```
joel-costigliola: add a `String` parameter to get the `Optional` type to have more precise error message:
```
"Expecting an empty %s but was containing value: <%s>."
```
That should give for `OptionalDouble`: 
```
"Expecting an empty OptionalDouble but was containing value: <1.0>."
```

src/main/java/org/assertj/core/error/OptionalShouldContain.java
```diff
@@ -12,44 +12,78 @@
...
-  /**
-   * Indicates that an {@link java.util.Optional} is empty so it doesn't contain the expected value.
-   *
-   * @param expectedValue the value we expect to be in an {@link java.util.Optional}.
-   * @return a error message factory.
-   */
-  public static OptionalShouldContain shouldContain(Object expectedValue) {
-	return new OptionalShouldContain(expectedValue);
-  }
+    private OptionalShouldContain(Object actual, Object expected) {
```
joel-costigliola: Similar comment as `OptionalShouldBeEmpty` to add optional type in the message

AlexBischof:
I fixed the comparison accordingly (int, long, double) and improved the OptionalShouldXXX classes so that the actual optional class appears in the error message.

joel-costigliola:
cool, thanks !

joel-costigliola:
I just read your comment on `hasValue` vs `contains`, it is indeed inconsistent but I think `hasValue`is a good name which works well wth `.hasValueCloseTo(8.0, within(0d));` whereas `.containsCloseTo(8.0, within(0d));` is a little bit weird.

We can't replace `contains` by `hasValue` in optional assertions, it would break the API.
If someone requests it, we might just add `hasValue` as an alias of `contains`.

joel-costigliola:
Merged, thanks.

Issue #393, opened 2015-04-26
Feature request: extend containsSequence assertions


mkordas:
Right now it's only possible to do positive `containsSequence` assertion using varargs:
https://github.com/joel-costigliola/assertj-core/blob/ca732b2d513984e086eface4aaafeb2af7b57555/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java#L365

Would be nice to have:
* assertion using `Iterable` as argument in addition to varargs
* negative assertion `doesNotContainSequence`

joel-costigliola:
I agree.

PR #407, opened 2015-05-07
Fixes #398

### LINKED ISSUE #398: Iterables.toArray does not return correctly typed array
In order to fix #393 I wanted to use `org.assertj.core.util.Iterables#toArray` for the proposed convenience methods.

However I noticed that the aforementioned `toArray` somehow does not return a correctly typed array. It returns an array of objects instead.

Here is some test code to reproduce the issue:
```java
  @Test
  public void should_return_array_of_given_iterable_type() throws Exception {
    ArrayList<CharSequence> values = new ArrayList<>();
    values.add("one"); values.add("two");
    
    CharSequence[] result = Iterables.toArray(values);
  }
```

Fails with:
`java.lang.ClassCastException: [Ljava.lang.Object; cannot be cast to [Ljava.lang.CharSequence;`

I debugged several `toArray` calls and it always returns `Object[]`.

Am I doing something wrong?

I furthermore noticed that the `Iterables` class is marked to be removed? Just curious because I am working on a contribution for #393 where this util class would be nice to have :-)

a11n:
Extended current Iterables utils in order to implement feature #393.

My idea is to just convert the iterable to an array to be able to reuse the current implementation.

**AbstractCharSequenceAssert.java**
```java
/* omitted javadoc for clarity */
  public S contains(Iterable<CharSequence> values) {
    strings.assertContains(info, actual, Iterables.toArray(values, CharSequence.class));
    return myself;
  }
```

and

```java
/* omitted javadoc for clarity */
  public S containsSequence(Iterable<CharSequence> values) {
    strings.assertContainsSequence(info, actual, Iterables.toArray(values, CharSequence.class));
    return myself;
  }
```

### DIFF
--- a/src/main/java/org/assertj/core/util/Iterables.java
+++ b/src/main/java/org/assertj/core/util/Iterables.java
@@ -15,6 +15,7 @@ package org.assertj.core.util;
 import static java.util.Collections.emptyList;
 import static org.assertj.core.util.Lists.newArrayList;
 
+import java.lang.reflect.Array;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


joel-costigliola:
Thanks for the PR, I like the idea

a11n:
Cool :-) I have the PR (code + tests) for #393 ready.

Since it depends on this PR I have not submitted it yet.

joel-costigliola:
done, thanks.

PR #409, opened 2015-05-08
Added convenience methods as requested in #393

a11n:


joel-costigliola:
Integrated thanks.
I have juts relaxed the bounded parameter to `Iterable<? extends CharSequence>` to be able to pass  `Iterable<String>`

Issue #363, opened 2015-04-05
Add OffsetTime and OffsetDateTime 


joel-costigliola:

checketts:
With this work happening in the Java 8 branch will a new Java 8 release of AssertJ happen any time soon, or should I just build my own RC to begin using? I want to start using it and give feedback/pull requests.

joel-costigliola:
It just happened :)
Feedback and PR always welcome

AlexBischof:
What is the status here? Can i take it?

joel-costigliola:
you can, thx.

AlexBischof:
Ok, created PR for OffsetTime and looking now at OffsetDateTime.

AlexBischof:
PR contains now assertion for OffsetDateTime, too.

joel-costigliola:
Sweet, thanks, will have a look at it next week or so

PR #410, opened 2015-05-10
Added OffsetTime assertion (#363)

AlexBischof:
This PR contains the OffsetTime assertion integration which is kind of similar to LocalTime assertion. Until now it differs only in the assertion isEqualsToIgnoringTimezone. 
I have tried to create tests accordingly and added also some missing tests for LocalTime (e.g. SoftAssertions).


joel-costigliola:
Integrated, thanks :+1:

PR #420, opened 2015-05-22
Fixes #417: Increased text font size of javadoc

### LINKED ISSUE #417: Use same font size in javadoc between code and normal text
The normal text font size is too small, it should have the size close to code sample.

AlexBischof:


joel-costigliola:
Integrated, that was really helpful as I suck at web stuff, thanks !

Issue #415, opened 2015-05-16
Feature Request: Add isCloseToPercentage to numbers


AlexBischof:
In my current project i need to check values with an offset which is a percentage value. So i need to calculate the offset like this:

```java
        double actualSum = 101d;
        double expectedSum = 100d;
        //should differ only within 0.5 percent
        double percentage = 0.5d;

        double calculatedOffset = percentage * expectedSum / 100;
        assertThat(actualSum).isCloseTo(expectedSum, offset(calculatedOffset));
```
Can we create a convenience method to encapsulate that calculation so that it would look like this
```java
assertThat(actualSum).isCloseToPercentage(expectedSum, offset(0.5d));
```

AlexBischof:
Having started with implementing this i think it is better to use it that way
```java
assertThat(9.0).isCloseTo(10.0, Percentage.percentage(0.5));
```

joel-costigliola:
Ok but I suggest to use `withinPercentage` to be consistent with the existing API.

AlexBischof:
I see that *isCloseTo* is only in *RealNumbers* which is not up to date anymore because every number has this method now. I would like to update this by pulling this method to *Numbers*. Therefore a refactoring of Numbers is needed.
```java
class Numbers<NUMBER extends Number & Comparable<NUMBER>>
```
@joel-costigliola Do you see any problems with that?

AlexBischof:
I am almost finished so do you mean that way?

```java
assertThat(9.0).isCloseTo(10.0, Percentage.withinPercentage(0.5));
assertThat(9.0).isCloseTo(10.0, Assertions.withinPercentage(0.5));
```

joel-costigliola:
Yes, `Assertions.withinPercentage` delegating to `Percentage.withinPercentage`

joel-costigliola:
Just to be sure, if you want to express that actual is close to expected by less than 5%, you would write:

```java
assertThat(9.0).isCloseTo(10.0, withinPercentage(5));
```

right ?

AlexBischof:
Yes. Basically it is just a convenience method so that the following is the same:
```java
assertThat(9.0).isCloseTo(10.0, offset(0.5));
assertThat(9.0).isCloseTo(10.0, offset(10.0*0.05));
```

joel-costigliola:
:+1:

PR #422, opened 2015-05-24
#415 Added isCloseToPercentage to numbers

AlexBischof:
This PR contains changes so that numbers (incl. BigDecimals) can be checked by closeTo using a percentage value. Internally an offset value is calculated from expected and percentage and is then use the same way as isCloseTo with offset.

In order to integrate this PR as best as possible i had to refactor some things:
* Deleted *BigDecimals.abs*, uses now Jdk7 Method
* Pulled *assertCloseTo* to Numbers (was only in *Doubles* and *Floats*)

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractByteAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractByteAssert.java
@@ -15,6 +15,7 @@ package org.assertj.core.api;
 import java.util.Comparator;
 
 import org.assertj.core.data.Offset;
+import org.assertj.core.data.Percentage;
 import org.assertj.core.internal.Bytes;
 import org.assertj.core.internal.ComparatorBasedComparisonStrategy;
 import org.assertj.core.util.VisibleForTesting;
@@ -375,7 +376,43 @@ public abstract class AbstractByteAssert<S extends AbstractByteAssert<S>> extend
     bytes.assertIsCloseTo(info, actual, expected, offset);
     return myself;
   }
-  
+
+  /** {@inheritDoc} */
+  @Override
+  public S isCloseTo(Byte expected, Percentage<Byte> percentage) {
+    bytes.assertIsCloseToPercentage(info, actual, expected, percentage);
+    return myself;
+  }
+
+    /**
+     * Verifies that the actual number is close to the given one within the given percentage.<br>
+     * If difference is equal to the percentage value, assertion is considered valid.
+     * <p>
+     * Example with double:
+     *
+     * <pre><code class='java'>
+     * // assertions will pass:
+     * assertThat(11.0).isCloseTo(new Double(10.0), withinPercentage(20d));
+     *
+     * // if difference is exactly equals to the computed offset (1.0), it's ok
+     * assertThat(11.0).isCloseTo(new Double(10.0), withinPercentage(10d));
+     *
+     * // assertion will fail
+     * assertThat(11.0).isCloseTo(new Double(10.0), withinPercentage(5d));
+     * </code></pre>
+     *
+     * @param expected the given number to compare the actual value to.
+     * @param percentage the given positive percentage between 0 and 100.
+     * @return {@code this} assertion object.
+     * @throws NullPointerException if the given offset is {@code null}.
+     * @throws NullPointerException if the expected number is {@code null}.
+     * @throws AssertionError if the actual value is not equal to the given one.
+     */
+  public S isCloseTo(byte expected, Percentage<Byte> percentage) {
+      bytes.assertIsCloseToPercentage(info, actual, expected, percentage);
+      return myself;
+  }
+
   @Override
   public S usingComparator(Comparator<? super Byte> customComparator) {
     super.usingComparator(customComparator);
--- a/src/main/java/org/assertj/core/api/AbstractByteAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractByteAssert.java
@@ -377,7 +377,30 @@ public abstract class AbstractByteAssert<S extends AbstractByteAssert<S>> extend
     return myself;
   }
 
-  /** {@inheritDoc} */
+    /**
+     * Verifies that the actual number is close to the given one within the given percentage.<br>
+     * If difference is equal to the percentage value, assertion is considered valid.
+     * <p>
+     * Example with byte:
+     *
+     * <pre><code class='java'>
+     * // assertions will pass:
+     * assertThat((byte)11).isCloseTo(Byte.valueOf(10), withinPercentage((byte)20));
+     *
+     * // if difference is exactly equals to the computed offset (1), it's ok
+     * assertThat((byte)11).isCloseTo(Byte.valueOf(10), withinPercentage((byte)10));
+     *
+     * // assertion will fail
+     * assertThat((byte)11).isCloseTo(Byte.valueOf(10), withinPercentage((byte)5));
+     * </code></pre>
+     *
+     * @param expected the given number to compare the actual value to.
+     * @param percentage the given positive percentage between 0 and 100.
+     * @return {@code this} assertion object.
+     * @throws NullPointerException if the given offset is {@code null}.
+     * @throws NullPointerException if the expected number is {@code null}.
+     * @throws AssertionError if the actual value is not equal to the given one.
+     */
   @Override
   public S isCloseTo(Byte expected, Percentage<Byte> percentage) {
     bytes.assertIsCloseToPercentage(info, actual, expected, percentage);
@@ -388,17 +411,17 @@ public abstract class AbstractByteAssert<S extends AbstractByteAssert<S>> extend
      * Verifies that the actual number is close to the given one within the given percentage.<br>
      * If difference is equal to the percentage value, assertion is considered valid.
      * <p>
-     * Example with double:
+     * Example with byte:
      *
      * <pre><code class='java'>
      * // assertions will pass:
-     * assertThat(11.0).isCloseTo(new Double(10.0), withinPercentage(20d));
+     * assertThat((byte)11).isCloseTo((byte)10, withinPercentage((byte)20));
      *
-     * // if difference is exactly equals to the computed offset (1.0), it's ok
-     * assertThat(11.0).isCloseTo(new Double(10.0), withinPercentage(10d));
+     * // if difference is exactly equals to the computed offset (1), it's ok
+     * assertThat((byte)11).isCloseTo((byte)10, withinPercentage((byte)10));
      *
      * // assertion will fail
-     * assertThat(11.0).isCloseTo(new Double(10.0), withinPercentage(5d));
+     * assertThat((byte)11).isCloseTo((byte)10, withinPercentage((byte)5));
      * </code></pre>
      *
      * @param expected the given number to compare the actual value to.
--- a/src/main/java/org/assertj/core/api/AbstractIntegerAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractIntegerAssert.java
@@ -211,14 +211,59 @@ public abstract class AbstractIntegerAssert<S extends AbstractIntegerAssert<S>> 
     return myself;
   }
 
-    /** {@inheritDoc} */
+    /**
+     * Verifies that the actual int is close to the given one within the given offset.<br>
+     * If difference is equal to offset value, assertion is considered valid.
+     * <p>
+     * Example:
+     *
+     * <pre><code class='java'>
+     * // assertions will pass:
+     * assertThat(5).isCloseTo(Integer.valueOf(7), within(3));
+     *
+     * // if difference is exactly equals to the offset, it's ok
+     * assertThat(5).isCloseTo(Integer.valueOf(7), within(2));
+     *
+     * // assertion will fail
+     * assertThat(5).isCloseTo(Integer.valueOf(7), within(1));
+     * </code></pre>
+     *
+     * @param expected the given int to compare the actual value to.
+     * @param offset the given positive offset.
+     * @return {@code this} assertion object.
+     * @throws NullPointerException if the given offset is {@code null}.
+     * @throws AssertionError if the actual value is not equal to the given one.
+     */
     @Override
   public S isCloseTo(Integer expected, Offset<Integer> offset) {
     integers.assertIsCloseTo(info, actual, expected, offset);
     return myself;
   }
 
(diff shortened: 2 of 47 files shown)


src/main/java/org/assertj/core/api/AbstractByteAssert.java
```diff
@@ -375,7 +376,43 @@ public S isCloseTo(Byte expected, Offset<Byte> offset) {
     bytes.assertIsCloseTo(info, actual, expected, offset);
     return myself;
   }
-  
+
+  /** {@inheritDoc} */
```
joel-costigliola: Unfortunately, you can't inherit javadoc as it shows a code example with `double`.
this comment is relevant in other Assert classes

src/main/java/org/assertj/core/api/AbstractByteAssert.java
```diff
@@ -375,7 +376,43 @@ public S isCloseTo(Byte expected, Offset<Byte> offset) {
...
+     * </code></pre>
+     *
+     * @param expected the given number to compare the actual value to.
+     * @param percentage the given positive percentage between 0 and 100.
+     * @return {@code this} assertion object.
+     * @throws NullPointerException if the given offset is {@code null}.
+     * @throws NullPointerException if the expected number is {@code null}.
+     * @throws AssertionError if the actual value is not equal to the given one.
+     */
+  public S isCloseTo(byte expected, Percentage<Byte> percentage) {
```
joel-costigliola: code example should show bytes, not double.

src/main/java/org/assertj/core/api/AbstractIntegerAssert.java
```diff
@@ -210,35 +211,48 @@ public S isCloseTo(int expected, Offset<Integer> offset) {
...
-   * </code></pre>
-   *
-   * @param expected the given Integer to compare the actual value to.
-   * @param offset the given positive offset.
-   * @return {@code this} assertion object.
-   * @throws NullPointerException if the given offset is {@code null}.
-   * @throws NullPointerException if the expected Integer is {@code null}.
-   * @throws AssertionError if the actual value is not equal to the given one.
-   */
+    /** {@inheritDoc} */
```
joel-costigliola: same comment : revert back to the initial javadoc that was showing int

src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -985,6 +988,104 @@ public static Index atIndex(int index) {
...
+     */
+    public static Percentage<BigDecimal> withinPercentage(BigDecimal value) {
+        return withPercentage(value);
+    }
+
+    /**
+     * Assertions entry point for Double {@link org.assertj.core.data.Percentage} to use with isCloseTo assertions for percentages.
+     * <p/>
+     * Typical usage :
+     * <p/>
```
joel-costigliola: remove `<p/>` the following `<pre>` tag takes care of appropriate spacing

src/main/java/org/assertj/core/data/Percentage.java
```diff
@@ -0,0 +1,80 @@
...
+    public static <T extends Number> Percentage<T> withPercentage(T value) {
+        checkNotNull(value);
+        checkBoundaries(value.doubleValue());
+        return new Percentage<>(value);
+    }
+
+    private static void checkBoundaries(double value) {
+        if (value < 0 || value > 100)
+            throw new IllegalArgumentException(
+                String.format("The percentage value <%s> should be between zero and hundred.", value));
```
joel-costigliola: use `... between 0 and 100.`, it's more user friendly

src/main/java/org/assertj/core/api/AbstractByteAssert.java
AlexBischof: I have oriented myself on AbstractDoubleAssert which is full inheritDoc but ok, i will change it.

src/main/java/org/assertj/core/internal/BigDecimals.java
```diff
@@ -50,20 +57,31 @@ public BigDecimals(ComparisonStrategy comparisonStrategy) {
...
-    throw failures.failure(info, shouldBeEqual(actual, other, offset, differenceAbsoluteValue));
+    if (areNotEqual(actual, other, offset))
+        throw failures.failure(info, shouldBeEqual(actual, other, offset, actual.subtract(other).abs()));
   }
 
-  // borrowed from java 7 API ... to remove when we will be using Java 7 instead of java 6.
-  private BigDecimal abs(final BigDecimal bigDecimal) {
-    return (bigDecimal.signum() < 0 ? bigDecimal.negate() : bigDecimal);
-  }
+    private boolean areNotEqual(BigDecimal actual, BigDecimal other,Offset<BigDecimal> offset) {
```
joel-costigliola: rename to `areNotCloseEnough`, areNotEqual is confusing

src/main/java/org/assertj/core/internal/Bytes.java
```diff
@@ -12,55 +12,70 @@
...
+        if (absDiff > offset.value) throw failures.failure(info, shouldBeEqual(actual, expected, offset, absDiff));
     }
 
+    public void assertIsCloseToPercentage(AssertionInfo info, Byte actual, Byte other,
+                                Percentage<Byte> percentage) {
+        assertNotNull(info, actual);
+        checkPercentageIsNotNull(percentage);
+        checkNumberIsNotNull(other);
 
+        if (org.assertj.core.util.Objects.areEqual(actual, other)) return;
```
joel-costigliola: this check is not needed, please remove it

src/main/java/org/assertj/core/internal/Doubles.java
```diff
@@ -91,8 +94,26 @@ protected boolean isEqualTo(Double actual, Double expected, Offset<?> offset) {
...
   }
+
+  public void assertIsCloseToPercentage(final AssertionInfo info, final Double actual, final Double expected,
+                              final Percentage<Double> percentage) {
+      checkPercentageIsNotNull(percentage);
+      checkNumberIsNotNull(expected);
+      assertNotNull(info, actual);
+
+      // doesn't use areEqual method relying on comparisonStrategy attribute
+      if (Objects.areEqual(actual, expected)) return;
```
joel-costigliola: same comment: to remove

src/main/java/org/assertj/core/internal/Longs.java
```diff
@@ -62,4 +66,20 @@ public void assertIsCloseTo(AssertionInfo info, Long actual, Long expected, Offs
...
+        checkPercentageIsNotNull(percentage);
+        checkNumberIsNotNull(other);
+
+        if (org.assertj.core.util.Objects.areEqual(actual, other)) return;
+
+        Offset<Double> calculatedOffset = offset(percentage.value * other / 100d);
+
+        Long absDiff = abs(other - actual);
+        if (absDiff > calculatedOffset.value) throw failures.failure(info, ShouldBeEqualWithinPercentage
+            .shouldBeEqualWithinPercentage(actual, other, percentage, absDiff));
```
joel-costigliola: static import `shouldBeEqualWithinPercentage`, it makes the code more readable
applies to the other classes

src/test/java/org/assertj/core/error/ShouldBeEqualWithinPercentage_create_Test.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Test
+  public void should_create_error_message() {
+    String message = factory.create(new TestDescription("Test"), new StandardRepresentation());
+    assertThat(message).isEqualTo(String.format("[Test] %n" +
+                                  "Expecting:%n" +
+                                  "  <12.0>%n" +
+                                  "to be close to:%n" +
+                                  "  <10.0>%n" +
+                                  "by less than <10.0> percent but difference was <20.0> percent.%n" +
```
joel-costigliola: will be better with:

`"by less than 10.0% but difference was 20.0% percent.`

src/main/java/org/assertj/core/error/ShouldBeEqualWithinPercentage.java
```diff
@@ -0,0 +1,48 @@
...
+                                                                                       Percentage<T> percentage,
+                                                                                       T difference) {
+        double expectedPercentage = difference.doubleValue() / expected.doubleValue() * 100d;
+        return new ShouldBeEqualWithinPercentage(actual, expected, percentage, expectedPercentage);
+    }
+
+    private <T extends Number> ShouldBeEqualWithinPercentage(Number actual, Number expected, Percentage<T> percentage,
+                                                             double expectedPercentage) {
+        super("%nExpecting:%n  <%s>%nto be close to:%n  <%s>%n" +
+              "by less than <%s> percent but difference was <%s> percent.%n" +
```
joel-costigliola: better if the message looked like : `"by less than 10.0% but difference was 20.0% percent.`

src/test/java/org/assertj/core/internal/bigdecimals/BigDecimals_assertIsCloseToPercentage_Test.java
```diff
@@ -0,0 +1,84 @@
...
+        bigDecimals.assertIsCloseToPercentage(someInfo(), ONE, ZERO, null);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void should_fail_if_percentage_is_negative() {
+        bigDecimals.assertIsCloseToPercentage(someInfo(), ONE, ZERO, withPercentage(valueOf(-1)));
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void should_fail_if_percentage_is_greater_hundred() {
```
joel-costigliola: `is_greater_than_one_hundred`

src/test/java/org/assertj/core/internal/bigdecimals/BigDecimals_assertIsCloseToPercentage_Test.java
```diff
@@ -0,0 +1,84 @@
...
+    @Test
+    public void should_pass_if_difference_is_less_than_given_percentage() {
+        bigDecimals.assertIsCloseToPercentage(someInfo(), ONE, ONE, withPercentage(valueOf(1)));
+        bigDecimals.assertIsCloseToPercentage(someInfo(), ONE, TWO, withPercentage(valueOf(50)));
+    }
+
+    @Test
+    public void should_pass_if_difference_is_equal_to_given_percentage() {
+        bigDecimals.assertIsCloseToPercentage(someInfo(), ONE, ONE, withPercentage(ZERO));
+        bigDecimals.assertIsCloseToPercentage(someInfo(), ONE, TWO, withPercentage(valueOf(100)));
```
joel-costigliola: In that case, the difference = 1 but the allowed margin is 2, so the difference != allowed margin.
you need to move this to `should_pass_if_difference_is_less_than_given_percentage`

src/test/java/org/assertj/core/internal/bigdecimals/BigDecimals_assertIsCloseToPercentage_Test.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Test(expected = IllegalArgumentException.class)
+    public void should_fail_if_percentage_is_greater_hundred() {
+        bigDecimals.assertIsCloseToPercentage(someInfo(), ONE, ZERO, withPercentage(valueOf(101)));
+    }
+
+    @Test
+    public void should_pass_if_difference_is_less_than_given_percentage() {
+        bigDecimals.assertIsCloseToPercentage(someInfo(), ONE, ONE, withPercentage(valueOf(1)));
+        bigDecimals.assertIsCloseToPercentage(someInfo(), ONE, TWO, withPercentage(valueOf(50)));
```
joel-costigliola: If I'm not wrong, the difference = 1 and the allowed margin is 2*50%=1, so the difference == allowed margin and not < margin.
you need to move this to `should_pass_if_difference_is_equal_to_given_percentage`

src/test/java/org/assertj/core/test/ErrorMessages.java
```diff
@@ -75,6 +75,10 @@ public static String offsetValueIsNotPositive() {
     return "The value of the offset should be greater than zero";
   }
 
+  public static String percentageValueIsInRange(Number number) {
+        return String.format("The percentage value <%s> should be between zero and hundred.",number.doubleValue());
```
joel-costigliola: `0 and 100`

AlexBischof:
Fixed your mentioned points. 
Thank you for your time/comments. It looks now more cleaner.

joel-costigliola:
Thanks for your hard work, will review that this we and integrate it for the 2.1.0 release that I'm planning to do this next week.

joel-costigliola:
I almost integrated the PR but I realized that expressing percentage with the same number type as numbers compared is weird, I should be able to write
```java
assertThat(5.0).isCloseTo(6.0, withinPercentage(20));
``` 
IMO, a percentage should only be constructed with an `int` or a `double`, it does not have to follow the compared number types, I should be able to write:
```java
assertThat(101).isCloseTo(100, withinPercentage(0.5));
assertThat(20f).isCloseTo(25f, withinPercentage(10));
``` 

I also changed `the Percentage.toString` to return directly `values%`.
 
I will add this changes this week end unless you strongly disagree - with good arguments ;-).

Last point, it would be nice if you could follow the code formatting convention (I know you are using Idea and it's not easy to use eclipse formatting preferences) but it would be easier for code review and integration as I would not have to reformat all the submitted code.
You might give a shot to : https://plugins.jetbrains.com/plugin/6546

AlexBischof:
I also stumbled upon this problem but did not followed it further because *within* has the same problem. Maybe *isCloseTo(A expected, Percentage<? extends Number> percentage)* could be a solution for both. 

I'll spend some time for that formatting issue.

AlexBischof:
Finally get the formatting problem solved. I had thought the plugin is already installed but **Lesson learned**: Eclipse Integration != Eclipse Code Formatter

joel-costigliola:
Awesome if you have fixed the formatting issue, it does not have to be perfect, most important thing  is getting the spacing right.

`isCloseTo` and `within` let you express a difference, in that case I think it's ok to express the difference with the same number type, it would not make sense to express a difference of `0.1` when comparing integers.

Percentage are best expressed with integer or double, I don't think we need anything else.

joel-costigliola:
Integrated ... with additional refactoring :)

PR #424, opened 2015-05-26
Fixed #423

### LINKED ISSUE #423: Add "expected" value to AbstractIterableAssert.containsExactly assertion message
The AbstractIterableAssert message does not contain the expected value, only the actual value. This makes it harder to use in a parameterized setting, like the one below:

    // Verifies that searching for values returns only those values
    for (String storageType : Arrays.asList("FILE", "TEXT", "URL", "XML")) {
        List<String> storage = searchFor(storageType);
        assertThat(storage).containsExactly(storageType);
    }

The error message does not tell me which what I was looking for, but only what I got -- which of the four iterations failed?

    java.lang.AssertionError: 
    Actual and expected should have same size but actual size is:
     <2>
    while expected is:
     <1>
    Actual was:
    <["XML", "TEXT"]>

This can be contrasted with a simple `assertThat("hi").isEqualTo("ho")`:

    org.junit.ComparisonFailure: 
    Expected :"ho"
    Actual   :"hi"

It would be nice if the error message also included expected value:

    java.lang.AssertionError: 
    Actual and expected should have same size but actual size is:
     <2>
    while expected is:
     <1>
    Expected was:           <-- New
    <["TEXT"]>              <-- New
    Actual was:
    <["XML", "TEXT"]>

AlexBischof:
Extended signature for ShouldHaveSameSizeAs

### DIFF
--- a/src/main/java/org/assertj/core/error/ShouldHaveSameSizeAs.java
+++ b/src/main/java/org/assertj/core/error/ShouldHaveSameSizeAs.java
@@ -28,14 +28,14 @@ public class ShouldHaveSameSizeAs extends BasicErrorMessageFactory {
    * @param expectedSize the expected size.
    * @return the created {@code ErrorMessageFactory}.
    */
-  public static ErrorMessageFactory shouldHaveSameSizeAs(Object actual, Object actualSize, Object expectedSize) {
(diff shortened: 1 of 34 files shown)


joel-costigliola:
Integrated

Issue #426, opened 2015-06-01
MapEntry class and map assertions using it as parameters are not usable from OSGI projects.


sarod:
asertj 2.0.0 MANIFEST.MF contains the following Private-Package attribute:
```
Private-Package: org.assertj.core.data,org.assertj.core.description,org.
 assertj.core.error,org.assertj.core.extractor,org.assertj.core.groups,o
 rg.assertj.core.internal,org.assertj.core.presentation,org.assertj.core
 .util,org.assertj.core.util.introspection,org.assertj.core.util.xml
```

This prevents usage of MapEntry class and map assertions using it as parameters from OSGI projects.

MapEntry should either be moved out of core.data package or core.data package should be marked public in OSGI metadata.

sarod:
It seems that more packages are missing like core.extractor, core.description, core.error.

In fact I guess only core.internal package (and maybe util) are really private packages

PR #427, opened 2015-06-01
Fix Issue#426 OSGI Metadata did not include "core.data" as Export-Package.

sarod:
Fix Issue #426 OSGI Metadata did not include "core.data" as Export-Package.

It is now assumed that only core.internal and core.util are "Private"

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,8 +76,8 @@
           <instructions>
             <Export-Package>
               !org.assertj.core.internal,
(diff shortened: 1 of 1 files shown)


joel-costigliola:
Sweet, thank you.
Will integrate that for the next release that is coming next week.

joel-costigliola:
Integrated, thanks.

PR #430, opened 2015-06-03
Implement 'containsSame' on 'OptionalAssert'

nicolaiparlog:
There are times when it is necessary to assert that an `Optional` not only contains an equal instance but a specific one, i.e. the _same_ which is given to the assertion. (E.g. when working with classes which do not properly implement `equals` or the implemented `equals` is - for some reason or another - not sufficient in a given situation.)

These assertions should then fail:

```java
	assertThat(Optional.of(new String("something"))).containsSame(new String("something"));
	assertThat(Optional.of(new Integer(10))).containsSame(new Integer(10));
```

This pull request implements `containsSame` which tests for reference equality.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractOptionalAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractOptionalAssert.java
@@ -15,6 +15,7 @@ package org.assertj.core.api;
 import static org.assertj.core.error.OptionalShouldBeEmpty.shouldBeEmpty;
 import static org.assertj.core.error.OptionalShouldBePresent.shouldBePresent;
 import static org.assertj.core.error.OptionalShouldContain.shouldContain;
+import static org.assertj.core.error.OptionalShouldContain.shouldContainSame;
 
 import java.util.Optional;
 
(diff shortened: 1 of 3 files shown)


joel-costigliola:
@nicolaiparlog your code is far behind the latest commit on java-8 branch, your code can't be directly merged/rebased since `OptionalShouldContain` has changed.

Please rebase your on top of the latest java-8 commit.

`OptionalShouldContain` message has been changed to:
```java
"%nExpecting:%n  <%s>%nto contain:%n  <%s>%nbut did not.", actual, expected
```
which is fine, no need to update it, just take care of `containsSame`, speaking of it I suggest to change your message to:
```java
"%nExpecting:%n  <%s>%nand:%n  <%s>%nto contain the same value %s (i.e. compared with ==) but did not.";
``` 
Something we might do is that if actual and expected option description are equals, then differentiate them using object hascode, this si already done in `ShouldBeEqual` and it gives something like:
```java
// both actual and expected toString are identical : Person[name=Jake]
// we differentiate them with hash code (in hexa)  
Expecting:
 <"Person[name=Jake] (Person@76724357)">
to be equal to:
 <"Person[name=Jake] (Person@47a34181)">
but was not."
```

Oh ... please add a test in `OptionalShouldContain_create_Test` for `containsSame` message ;-)

nicolaiparlog:
See #431 for a continuation of this discussion.

PR #436, opened 2015-06-11
CheckReturnValue annotation on *Assert objects

sul3n3t:
A common mistake programmers make when using assertj is to create an assertion object but forget to call any methods on it (e.g. `assertThat(obj);` vs. `assertThat(obj).isTrue();`).

By adding the`@CheckReturnValue` annotation to `public static *Assert assertThat(*)` methods, static analysis tools will warn when programmers forget to call any methods on the returned assertion object.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractStandardSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractStandardSoftAssertions.java
@@ -36,6 +36,7 @@ public abstract class AbstractStandardSoftAssertions extends AbstractSoftAsserti
    * @param actual the actual value.
    * @return the created assertion object.
    */
+  @CheckReturnValue
   public BigDecimalAssert assertThat(BigDecimal actual) {
 	return proxy(BigDecimalAssert.class, BigDecimal.class, actual);
   }
@@ -46,6 +47,7 @@ public abstract class AbstractStandardSoftAssertions extends AbstractSoftAsserti
    * @param actual the actual value.
    * @return the created assertion object.
    */
+  @CheckReturnValue
   public BooleanAssert assertThat(boolean actual) {
 	return proxy(BooleanAssert.class, Boolean.class, actual);
   }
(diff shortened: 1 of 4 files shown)


src/main/java/org/assertj/core/api/AbstractStandardSoftAssertions.java
```diff
@@ -36,6 +36,7 @@
    * @param actual the actual value.
    * @return the created assertion object.
    */
+  @CheckReturnValue
```
cushon: Note that you can also apply `@CheckReturnValue` to classes (or packages) which has the same affect as annotation all of the elements. If you go that route, Error Prone defines a [`@CanIgnoreReturnValue`](http://errorprone.info/api/latest/com/google/errorprone/annotations/CanIgnoreReturnValue.html) annotation you can use to opt-out specific methods.

joel-costigliola:
Good idea, addresses https://github.com/joel-costigliola/assertj-core/issues/157.
Will have a look at it soon.

joel-costigliola:
One blocking problem : the dependency on `com.google.code.findbugs` annotations, by design assertj does not depend on anything to avoid dependency version constraints for users.

Is there another way of solving this issue ?

joel-costigliola:
I'm closing this PR as assertj-core does not ship with dependencies.

JnRouvignac:
I think you can use the maven shade plugin to avoid depending on an external library?
From what I saw, error-prone checks the annotation's class simple name and ignores the package. So this should work.

Am I right @cushon?
Would that work for you @joel-costigliola?

joel-costigliola:
@JnRouvignac I believe this has been addressed by https://github.com/joel-costigliola/assertj-core/pull/695 as findbug detects any `CheckReturnValue` annotations.

JnRouvignac:
Great! I had not seen this.
@sul3n3t I guess you may give it a try!

PR #441, opened 2015-06-16
Fix signature of Extractors.toStringMethod() to avoid compile error.

shimikano:
The return type of `Extractors.toStringMethod()` should be changed to `Extractor<Object, String>` in order to avoid compile errors due to the wildcard.

This seems to have been untested previously.


joel-costigliola:
Good catch, we missed this one.
Thanks for the PR, will integrate it soon.

joel-costigliola:
Integrated, thanks.

PR #443, opened 2015-06-18
Support of Optional inner-type checking.

adericbourg:
Eg: 

    assertThat(Optional.of("something")).containsInstanceOf(String.class);


joel-costigliola:
just in time for 3.1.0 :)

joel-costigliola:
Integrated, thanks

PR #445, opened 2015-06-26
Added hasField assertion to ObjectAssert and fixed one NPE

librucha:


joel-costigliola:
thanks, will review it shortly.

src/main/java/org/assertj/core/api/AbstractObjectAssert.java
```diff
@@ -190,4 +191,85 @@ public S isEqualToComparingFieldByField(A other) {
...
+   *
+   * TolkienCharacter frodo = new TolkienCharacter("Frodo", 33);
+   *
+   * assertThat(frodo).hasField("name", "Frodo");       // pass
+   * assertThat(frodo).hasField("name", "not_equals");  // fail
+   * assertThat(frodo).hasField("age", 33);             // fail because age is private withou public getter
+   * assertThat(frodo).hasField(null, 33);              // fail
+   * assertThat(frodo).hasField("age", null);           // fail
+   *
+   * TolkienCharacter frodo = new TolkienCharacter(null, 33);
```
joel-costigliola: rename `frodo` to `noname` otherwise the example is confusing

src/main/java/org/assertj/core/util/introspection/PropertySupport.java
```diff
@@ -173,8 +173,8 @@ private boolean isNestedProperty(String propertyName) {
    * @throws IntrospectionError if target object does not have a property with a matching name.
    */
   public <T> T propertyValueOf(String propertyName, Class<T> clazz, Object target) {
-    // returns null if target is null as we can't extract a property from a null object
-    if (target == null) return null;
+    // returns null if target or property name is null as we can't extract a property from a this object
+    if (target == null || propertyName == null) return null;
```
joel-costigliola: good catch

joel-costigliola:
What you have described in the javadoc is not `hasField` but `hasProperty`, the implementation does not correspond to `hasField` or `hasProperty` but `hasFieldOrProperty`.   

What was your intention in the first place, checking field, property or both ?

librucha:
In the first place it was probably hasProperty and hasPropertyWithValue, but I choose wrong name. 
I'll rename it.

joel-costigliola:
Thanks, I have integrated your work with minor improvements:
* `hasPropertyWithValue` error message display actual and expected value
* private fields are read by default unless `Assertions.setAllowExtractingPrivateFields(false)` is called
* few javadoc fixes

PR #453, opened 2015-07-04
Add assertThatThrownBy & catchThrowable to WithAssertions

cykl:
This patch adds `assertThatThrownBy` and `catchThrowable` to the `WithAssertions` interface. 

Having to import `Assertions` because those two very useful methods are absent from the interface is annoying. Lets add them !


joel-costigliola:
Agreed, it was a miss more than a deliberate decisions.

joel-costigliola:
Fixed, thanks !

Issue #438, opened 2015-06-12
Javadoc: Reduce/remove gap between "Example:" heading and example code


PascalSchumacher:
It would be nice to reduce or remove the gap between the "Example:" heading and example code in the javadoc, so that the content assist in IDEs show more actual content.

E.g: https://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/AbstractIterableAssert.html#containsOnly%28T...%29

joel-costigliola:
Hmmm you don't syntax highlighting, I suspect `https` as this URL looks fine : http://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/AbstractIterableAssert.html#contains%28T...%29

Maybe HTTPS prevents hightlight.js to be loaded ...

PascalSchumacher:
Yes the highlighting works with HTTP.

Does highlighting work for you in Eclipse? For me it does not.

My complaint was, that without code highlighting the gap between the "Example:" heading and example code is very big: https://github.com/PascalSchumacher/stuff/blob/master/assertj-javadoc-example-margin.png and looks like wasted space.

With highlighting it looks o.k., so feel free to close this issue.

joel-costigliola:
No, eclipse can't preserve syntax highlighting as it is done through javascript.

I might consider reducing the space so that it looks better in eclipse but I still want to keep the nice looking generated javadoc.

joel-costigliola:
Best solution so far : 
* remove starting and ending in examples (lot of work but worth it!)
* increment padding in `hljs` class to 0.6em

PascalSchumacher:
If there is anything I can help with I can create a pull request.

joel-costigliola:
yep, have a look at what I did in assertj-guava (see this commit https://github.com/joel-costigliola/assertj-guava/commit/841ee1b60df12a59b5e7528901a07e9d18c0f618), you can generate the javadoc with maven and if you are happy with the result, we have to do the same for assertj-core, that is reformatting the code examples like in the mentionned commit.

PascalSchumacher:
The javadoc examples of assertj-guava look much improved in Eclipse!

Thanks I lot. :)

I will create a pull request with the changes for assertj.

joel-costigliola:
:+1:

PR #464, opened 2015-07-16
#438: Better spacing for javadoc code examples

PascalSchumacher:
I tried to follow https://github.com/joel-costigliola/assertj-guava/commit/841ee1b60df12a59b5e7528901a07e9d18c0f618 

Once we are satisfied with this change I will create another pull request for the 3.0 classes.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -83,9 +83,7 @@ public abstract class AbstractAssert<S extends AbstractAssert<S, A>, A> implemen
    * defined by the user with {@link #overridingErrorMessage(String, Object...)}.
    * <p>
    * Example :
-   * 
-   * <pre><code class='java'>
-   * public TolkienCharacterAssert hasName(String name) {
+   * <pre><code class='java'> public TolkienCharacterAssert hasName(String name) {
    *   // check that actual TolkienCharacter we want to make assertions on is not null.
    *   isNotNull();
    * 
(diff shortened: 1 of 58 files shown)


joel-costigliola:
Thanks a lot ! 
Don't create a PR for 3.x, 3.x is really just a branch containing java 8 specific stuff, I usually report issues in 2.x to 3.x by merging master (= latest 2.x) in java-8 branch (latest 3.x).

PascalSchumacher:
Thanks for the info. :)

I will let you merge to the java-8 branch, but I guess there could be some java 8 specific classes e.g. AbstractLocalDateAssert which might need similar changes?

joel-costigliola:
You are correct indeed, e.g. `AbstractOptionalAssert`, if you want to address it, just create another PR specific to those classes.

src/main/java/org/assertj/core/api/AbstractAssert.java
```diff
@@ -83,9 +83,7 @@ protected WritableAssertionInfo getWritableAssertionInfo() {
    * defined by the user with {@link #overridingErrorMessage(String, Object...)}.
    * <p>
    * Example :
-   * 
-   * <pre><code class='java'>
-   * public TolkienCharacterAssert hasName(String name) {
+   * <pre><code class='java'> public TolkienCharacterAssert hasName(String name) {
```
PascalSchumacher: I'm wondering if the space between `<pre><code class='java'>` and the start of the code should be removed?

joel-costigliola: No because the next lines start also with a space, they should have the same indentation.

joel-costigliola:
Integrated, many thanks !

PR #465, opened 2015-07-16
#463: OptionalAssert.contains does not use the specified comparator

### LINKED ISSUE #463: OptionalAssert.contains does not use the specified comparator
The following does not work

```java
assertThat(Optional.of(new DummyObject(value))).usingComparator(new FieldByFieldComparator()).contains(new DummyObject(value));
```

Looking at OptionalAssert it appears the contains does not look to see if a comparator was specified and instead relies on the actual objects equals method (line 105):

```java
if (!actual.get().equals(expectedValue)) throw failure(shouldContain(actual, expectedValue));
```

lucasweb78:
Added ```usingDefaultValueComparator```, ```usingFieldByFieldValueComparator``` and ```usingValueComparator``` to OptionalAssert. This allows you to specify the comparator that should be used when asserting an Optional contains a value.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractOptionalAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractOptionalAssert.java
@@ -12,12 +12,18 @@
  */
 package org.assertj.core.api;
 
+import org.assertj.core.internal.ComparatorBasedComparisonStrategy;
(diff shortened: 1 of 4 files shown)


joel-costigliola:
cool, thanks !

PR #468, opened 2015-07-17
Path's assertions should not rely on java.nio.Path.toFile() (fixes #439)

### LINKED ISSUE #439: Cannot assert Path content when using Jimfs file system.
Jimfs is an in memory file system from google for testing purposes. Its implementation of java.nio.Path.toFile() throws an UnsupportedOperationException. This is actually completely legit (implementations are not required to support toFile() according to javadoc)

The assertion on a Path's contents, however, appears to use toFile() to get to the contents. So my assertion on this file fails.

<b>Path's assertions should not rely on java.nio.Path.toFile()</b>

<pre>
java.lang.UnsupportedOperationException
	at com.google.common.jimfs.JimfsPath.toFile(JimfsPath.java:397)
	at org.assertj.core.internal.Paths.assertHasContent(Paths.java:277)
	at org.assertj.core.api.AbstractPathAssert.hasContent(AbstractPathAssert.java:257)
    ...
</pre>

As a workaround I read the contents of the file myself, and use an assertion on String equality.

PascalSchumacher:
After these changes the memoryfilesystem dependecy can be replaced with jimfs and all tests still succeed.

### DIFF
--- a/src/main/java/org/assertj/core/error/ShouldHaveBinaryContent.java
+++ b/src/main/java/org/assertj/core/error/ShouldHaveBinaryContent.java
@@ -13,12 +13,13 @@
 package org.assertj.core.error;
 
 import java.io.File;
+import java.nio.file.Path;
 
 import org.assertj.core.internal.BinaryDiffResult;
 
 
 /**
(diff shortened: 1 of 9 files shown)


joel-costigliola:
Thx

PR #476, opened 2015-08-07
Added assertThatException 

glhez:
Add `assertThatException` in complement of `assertThatThrownBy`.

This patches add two classes: `ExpectThrowableAssert` and `ChainedThrowableAssert`. The first delegate to the second and will check that the lambda actually throw an exception of expected type.

Also, this allow for chaining:

```
IntConsumer lambda = int n -> {
  if (n < 0) throw new IllegalArgumentException(n + " is negative");
  if (n > 100) throw new IllegalArgumentException(n + " must be less or equals to 100");
}
assertThatException(IllegalArgumentException.class)
  .isThrownBy(() -> lambda.accept(-1))
    .hasMessage("-1 is negative")
  .isThrownBy(() -> lambda.accept(101))
    .hasMessage("101 must be less or equals to 100")
;
```

To be compared with the existing `assertThatThrownBy`:

```
assertThatThrownBy(() -> lambda.accept(-1))
  .isInstanceOf(IllegalArgumentException.class)
  .hasMessage("-1 is negative")
;
assertThatThrownBy(() -> lambda.accept(101))
  .isInstanceOf(IllegalArgumentException.class)
  .hasMessage("101 must be less or equals to 100")
;
```

As said in the javadoc, `assertThatException` is named like that because of `assertThat(Class)` and type erasure.

### DIFF
--- a/src/main/java/org/assertj/core/api/ChainedThrowableAssert.java
+++ b/src/main/java/org/assertj/core/api/ChainedThrowableAssert.java
@@ -1,0 +1,30 @@
+package org.assertj.core.api;
+
+import org.assertj.core.api.ThrowableAssert.ThrowingCallable;
+
+/**
+ * Assertion methods for {@link java.lang.Throwable}.
+ * <p>
+ * This class is linked with the {@link ExpectThrowableAssert} and allow to check that an exception
+ * type is thrown by several lambda.
+ */
+public class ChainedThrowableAssert<T extends Throwable>
+    extends AbstractThrowableAssert<ChainedThrowableAssert<T>, T> {
+  private final ExpectThrowableAssert<T> parent;
+
+  ChainedThrowableAssert(final ExpectThrowableAssert<T> parent, final T actual) {
+    super(actual, ChainedThrowableAssert.class);
+    this.parent = parent;
+  }
+
+  /**
+   * Assert that an exception of type T is actually thrown by the {@code throwingCallable}.
+   *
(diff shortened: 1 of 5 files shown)


joel-costigliola:
Thanks, that's interesting, I would rename `hasMessage` to `withMessage` to make things more fluent:

```java
assertThatException(IllegalArgumentException.class)
  .isThrownBy(() -> lambda.accept(-1)).withMessage("-1 is negative")
  .isThrownBy(() -> lambda.accept(101)).withMessage("101 must be less or equals to 100");
```

I will review that in few weeks (pretty busy right now)

glhez:
Good idea. 
I pushed another commit bringing a lots of with method, and deprecated (in the Chained class) the hasXXX methods (I can't rename them, otherwise `assertThat(Throwable).hasMessage(String)` would lose its meaning).

joel-costigliola:
closed after deleting java8 branch

joel-costigliola:
I'm about to integrate this PR with few internal changes, I'm also thinking to change the entry point method to `assertThatExceptionOfType` instad of `assertThatException`, with that change assertion code would look like: 

```java
assertThatExceptionOfType(IOException.class).isThrownBy(() -> { throw new IOException("boom!") });
                                            .withMessage("boom!"); 
```

WDYT ?

glhez:
I'm not against it, and even find it more fluent.

joel-costigliola:
change done - PR integrated.

PR #480, opened 2015-08-19
Fix introspection error message in extractor

yurloc:
Names of the exception variables didn't match their meaning.


joel-costigliola:
good catch ! thanks

yurloc:
Btw, there was a broader change on master that also addressed this issue, so no need for cherry-picking.

joel-costigliola:
Right, I just realized you were looking at the 1.7.1 branch.
Since this is a minor issue and I'm not maintaining actively the 1.x branch, I'm closing it

PR #496, opened 2015-10-03
Fixes #493 : Add partial regex match assertion

### LINKED ISSUE #493: Add partial regex match assertion
Any chance a partial pattern match (i.e. typical non-Java version without implied `^` and `$`) assert could be added? I cobbled together the wrapper below to use in the meantime.

[Here is Hamcrest-flavoured example](https://github.com/derari/cthul/blob/master/matchers/src/main/java/org/cthul/matchers/object/ContainsPattern.java), with `matchesPattern` and `containsPattern`.
```java
import java.util.regex.Pattern;

public class Assertions extends org.assertj.core.api.Assertions {
    public static StringAssert assertThat(String actual) {
        return new StringAssert(actual);
    }

    public static class StringAssert extends org.assertj.core.api.StringAssert {
        protected StringAssert(String actual) {
            super(actual);
        }

        /*@Override
        public org.assertj.core.api.StringAssert matches(Pattern pattern) {
            return super.matches(pattern);
        }

        @Override
        public org.assertj.core.api.StringAssert matches(CharSequence regex) {
            return super.matches(regex);
        }*/

        public StringAssert contains(Pattern pattern) {
            if (pattern == null)
                throw new NullPointerException("The regular expression pattern to find should not be null");

            isNotNull();

            if (!pattern.matcher(actual).find())
                failWithMessage("%nExpecting:%n %s%nto contain pattern:%n %s", actual, pattern.pattern());

            return this;
        }

        public StringAssert containsPattern(CharSequence regex) {
            if (regex == null)
                throw new NullPointerException("The regular expression pattern to find should not be null");
            return contains(Pattern.compile(regex.toString()));
        }
    }
}
```

ptemplier:


joel-costigliola:
Will review it shortly, thanks.

joel-costigliola:
Thanks Pierre, merged into 2.x and will be available in master soon.

PR #500, opened 2015-10-07
Support Java 1.6 and thus Android applications under test.

felixvf:
The current assertj-core 2.x code is nearly Java 1.6 compliant. 

Being Java 1.6 compliant is necessary for using assert-core 2.x to test Android apps which in turn may both target newer and older Android devices. Currently, testing under older Android devices (e.g. Android Gingerbread) fails due to the Java 1.6 incompatibilities.

This code makes the assertj-core 2.x code Java 1.6 compliant.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractPathAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractPathAssert.java
@@ -131,7 +131,7 @@ public abstract class AbstractPathAssert<S extends AbstractPathAssert<S>> extend
    * Path xFileTurkish = Files.write(Paths.get("xfile.turk"), Collections.singleton("Gerçek Başka bir yerde mi"), turkishCharset);
    * 
    * // The following assertion succeeds:
-   * String expectedContent = "Gerçek Başka bir yerde mi" + System.lineSeparator();
+   * String expectedContent = "Gerçek Başka bir yerde mi" + org.assertj.core.util.Compatibility.System.lineSeparator();
    * byte[] binaryContent = expectedContent.getBytes(turkishCharset.name());
    * assertThat(xFileTurkish).hasBinaryContent(binaryContent);
    * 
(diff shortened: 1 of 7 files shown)


joel-costigliola:
I'm happy to integrate this PR ! I'm just surprised this is all the changes needed, I thought Android could not use `Path` at all.

felixvf:
Sure, if `Path` is not available, then `Path` specific methods do not work. But Android does not reject applications as a whole which refer non-existing classes, it apparently just rejects specific method calls which refer non-existing classes. So calling other methods (which do not refer `Path`) actually works (as soon as this PR is merged). :-)

Turbo87:
@felixvf @joel-costigliola  I've just tested this and it doesn't seem to work:

```java
assertj-android-test/app/src/test/java/io/github/turbo87/assertjtest/ExampleUnitTest.java:13: error: cannot access Path
        assertThat(2 + 2).isEqualTo(4);
        ^
  class file for java.nio.file.Path not found
```

I suspect that since the `Assertions` class imports the `Path` interface this fails. What we might be able to do is use a `Java6Assertions` class that only has `assertThat()` methods for Java 1.6 compliant classes and derive the `Assertions` class from that. I'm not sure if that would actually work, but it might be worth a try.

Turbo87:
I did a little experiment with my theory and it seems to work. I've extracted a `Java6Assertions` superclass without the Path assertion (see https://github.com/Turbo87/assertj-core/commit/9f3f61a70f9eb1c29e9abbb5d025dde01404752d), used that for the import instead of `Assertions` (https://github.com/Turbo87/assertj-android-test/commit/449713915bf04e821cd68f825a355c3612ea54dc) and now everything compiles and tests successfully.

joel-costigliola:
That's good news, thanks !

joel-costigliola:
Should be addressed by https://github.com/joel-costigliola/assertj-core/pull/521 thanks @felixvf for the work.

PR #506, opened 2015-10-14
Added containing-semantics for asserting content in OptionalAsserter

ovstetun:
I wish to do a more functional style for deeper asserting on the content of an optional. The implementation allows the following style for asserting on Optionals:
```java
assertThat(Optional.of(someString)).containing(s -> {
  assertThat(s).isEqualTo("something");
  assertThat(s).startsWith("some");
  assertThat(s).endsWith("thing");
});
```

Above will fail for any ```Optional.empty```.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractOptionalAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractOptionalAssert.java
@@ -20,6 +20,7 @@ import static org.assertj.core.error.OptionalShouldContainInstanceOf.shouldConta
 
 import java.util.Comparator;
 import java.util.Optional;
+import java.util.function.Consumer;
 
 import org.assertj.core.internal.ComparatorBasedComparisonStrategy;
 import org.assertj.core.internal.ComparisonStrategy;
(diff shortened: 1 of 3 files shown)


joel-costigliola:
Hi,
IMO it is simpler/lighter/more readable to do this :
```java
assertThat(Optional.of(someString).get()).isEqualTo("something")
                                         .startsWith("some")
                                         .endsWith("thing");
```

Is there any other use case for your PR ? 
Cheers

ovstetun:
For me the primary use case would be when working with a large nested object structure.

Think:
```
Person -> Optional<Address> -> Streetaddress, Optional<Country>
       -> Optional<PhoneNumber>
       -> Optional<Name> -> Firstname, Lastname
```

Asserting in this structure would be nice and very readable to do nested (example with address only, could make more complex with more nested attributes):
```java
Optional<Person> john = repository.find(42);
assertThat(john).containing( p -> {
  assertThat(p.address()).containing( a -> {
    assertThat(a.streetAddress()).isEqualTo("some street 42");
    assertThat(a.country()).containing( c -> {
      assertThat(c.name()).isEqualTo("Norway");
      assertThat(c.countryCode()).isEqualTo("NO");
    });
  });
});
```

Today I would write a lot of code like the following:
```java
assertThat(john).isPresent();
john.ifPresent(p -> {
  assertThat ........
});
```
where joining the ```isPresent```-check with the following ```ifPresent```-lambda being tightly coupled. My proposal joins them together explicitly to a single expression.

ovstetun:
Hi again!
I just merged master to make the build pass in Travis.
Any more feedback on this PR?

/trondmarius

joel-costigliola:
I like the idea but I find `containing` not very intuitive but I can't find a better name ...

ovstetun:
I have also tried to think of a different name, and will happily refactor naming. A couple suggestions could be

```public S satisfies(Consumer<T> requirement) {...}```

* ```assertThat(o).satisfies(s -> {```
* ```assertThat(o).satisfiesRequirement(s -> {```

joel-costigliola:
I think `satisfies` works well, I would be happy with it.

joel-costigliola:
Integrated, thanks !

joel-costigliola:
@ovstetun I just had a (better ?) idea for naming this assertion : `hasValueSatisfying`, it makes it clear we get the optional value, otherwise it's hard to see the difference between `matches` and `satisfies` unless you read the javadoc. WDYT ?

ovstetun:
```hasValueSatisfying``` sounds like a good name to me. Perhaps keep both with one as an alias of the other? Do you want me to make the change on my branch - the PR is closed?

joel-costigliola:
I will just use `hasValueSatisfying` to avoid API clutter.
I will make the change but thanks for proposing.

PR #510, opened 2015-10-29
fixed FindBugs warning for string concatenation in loop

PascalSchumacher:
of course performance is not a focus of AssertJ, but this fix does not make the code less readable imho


joel-costigliola:
integrated in 2.x, thanks

PR #511, opened 2015-10-30
Fixes #440: Assertions related to entries on AbstractMapAssert should…

### LINKED ISSUE #440: assertions related to entries on AbstractMapAssert should support java.util.Map.Entry as arguments
Hi,

It seems to me that AbstractMapAssert would be more flexible if AbstractMapAssert.contains (and other assertions related to MapEntry) were taking java.util.Map.Entry as arguments instead of MapEntry.
For backward compatibility MapEntry would need to implement java.util.Map.Entry interface.

What do you think?

PascalSchumacher:
… support java.util.Map.Entry as arguments

Enhancement seems to work fine. :)

I guess the test are wrong (especially the new test classes I added). Please give me some pointers what kind of test should be added where? And of course any other feedback. I will update the pull request accordingly.

Thanks!

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractMapAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractMapAssert.java
@@ -18,7 +18,6 @@ import static org.assertj.core.util.Arrays.array;
 import java.util.Comparator;
 import java.util.Map;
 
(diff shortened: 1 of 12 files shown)


joel-costigliola:
Good work, thx.

PR #513, opened 2015-10-31
Context-sensitive diff engine for unordered collections

gabga:
I have created a pull request as asked, although the issue is not completely solved. Five things to keep in mind:

1) I am not an experienced GitHub user, so if you notice some problems, please let me know
2) There is the mentioned issue with soft assertions so the tests don't pass
3) The Maps_assertEqual_Test does not really test the string representation; only that a failure is thrown (as it is done in other similar tests). I wanted to add a test to make sure that the provided MapRepresentation to the shouldBeEqual() is properly created and that it creates correct output but it is not really straight forward. I am still thinking on how to do this so if you have any ideas, they are more than welcome.
4) The output doesn't highlight the contents of the expected map. This will be solved.
5) I focused on map assertions. Probably a similar solution should be done for Set as well.


gabga:
Points 3 and 4 in the list above should be solved now.

joel-costigliola:
Thanks for the PR, here's my comments:
- don't use `[]` when building the error message, they are added automatically by JUnit to highlight differences
- test with the data from https://github.com/joel-costigliola/assertj-core/issues/479#issuecomment-130970834, you will see that the current solution is not optimal
- I think that we should sort the compared maps when possible, users will more easily see the differences and JUnit does good at highlighting differences in that case.

The last comment is the most important, let's try sorting maps if keys are comparable (which I think might be the case most of the time), if we can't I don't think we can easily (or should) replace JUnit highlighting differences by our own, it means that we don't need to compute map diff but just display sorted map - in that case you only need to change `org.assertj.core.util.Maps#format` method to sort the map before constructing its representation.

Thinking out loud : computing map diff is only useful if you can control map string representation used by JUnit as JUnit highlight String differences, it would mean putting displaying the identical key/value first then, the key differences and then value differences so that JUnit only highlight relevant part. Anyway I think map reordering is the easiest solution and it will work for the majority of maps.

gabga:
Oh, sorry, I completely misinterpreted the issue. I focused on the AssertJ formatting (which appears in my IDE) instead of the JUnit output.

You are correct, it should be sufficient to simply order the maps (if possible). I'll look into it and create another PR.

joel-costigliola:
no problem, you could have just push a new version of the code in this PR but creating a new one is also ok.

PR #515, opened 2015-11-01
Fixes #178: Add isEqualToComparingFieldByFieldRecursively

### LINKED ISSUE #178: Feature request: isEqualToComparingFieldByField - recursive
I like the idea of isEqualToComparingFieldByField, but it is of limited use when some of the fields are objects without Equals being overriden. 

I would like to have an option to use isEqualToComparingFieldByField recursively, instead of falling back to org.assertj.core.util.Objects.areEqual.

PascalSchumacher:
I created the prototype based on a slightly modified https://github.com/jdereg/java-util/blob/master/src/main/java/com/cedarsoftware/util/DeepEquals.java 

It's obviously not ready for merge yet, but I want to get some feedback if the approach is o.k. etc., before investing more time.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
@@ -282,4 +282,9 @@ public abstract class AbstractObjectAssert<S extends AbstractObjectAssert<S, A>,
     objects.assertHasFieldOrPropertyWithValue(info, actual, name, value);
     return myself;
   }
+
+    public S isEqualToComparingFieldByFieldRecursively(Object other) {
+      objects.assertIsEqualToComparingFieldByFieldRecursively(info, actual, other);
+      return myself;
+    }
 }
--- a/src/main/java/org/assertj/core/internal/Objects.java
+++ b/src/main/java/org/assertj/core/internal/Objects.java
@@ -47,6 +47,8 @@ import java.util.List;
 import java.util.Set;
 
 import org.assertj.core.api.AssertionInfo;
+import org.assertj.core.presentation.StandardRepresentation;
+import org.assertj.core.util.DeepEquals;
 import org.assertj.core.util.VisibleForTesting;
 import org.assertj.core.util.introspection.FieldSupport;
 import org.assertj.core.util.introspection.IntrospectionError;
@@ -650,6 +652,23 @@ public class Objects {
     return fieldSupport.isAllowedToRead(field) || propertySupport.publicGetterExistsFor(field.getName(), actual);
   }
 
+  /**
+   * Assert that the given object is "deeply" equals to the other by comparing all fields recursively.
+   *
+   * @param info contains information about the assertion.
+   * @param actual the given object.
+   * @param other the object to compare {@code actual} to.
+   * @throws AssertionError if actual is {@code null}.
+   * @throws AssertionError if the other object is not an instance of the actual type.
+   * @throws AssertionError if the actual and the given object are not "deeply" equal.
+   */
+  public <A> void assertIsEqualToComparingFieldByFieldRecursively(AssertionInfo info, A actual, A other) {
+    assertNotNull(info, actual);
+    assertHasSameClassAs(info, actual, other);
+    if (!DeepEquals.deepEquals(actual, other))
+        throw failures.failure(info, shouldBeEqual(actual, other, StandardRepresentation.STANDARD_REPRESENTATION));
+  }
+  
   /**
    * Get property value first and in case of error try field value.
    * <p>
--- a/src/main/java/org/assertj/core/internal/Objects.java
+++ b/src/main/java/org/assertj/core/internal/Objects.java
@@ -47,8 +47,9 @@ import java.util.List;
 import java.util.Set;
 
 import org.assertj.core.api.AssertionInfo;
-import org.assertj.core.presentation.StandardRepresentation;
-import org.assertj.core.util.DeepEquals;
+import org.assertj.core.error.ShouldBeEqualByComparingFieldByFieldRecursive;
+import org.assertj.core.util.DeepDifference;
+import org.assertj.core.util.DeepDifference.Difference;
 import org.assertj.core.util.VisibleForTesting;
 import org.assertj.core.util.introspection.FieldSupport;
 import org.assertj.core.util.introspection.IntrospectionError;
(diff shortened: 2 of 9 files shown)


src/main/java/org/assertj/core/api/AbstractObjectAssert.java
```diff
@@ -282,4 +282,9 @@ public S hasFieldOrPropertyWithValue(String name, Object value) {
     objects.assertHasFieldOrPropertyWithValue(info, actual, name, value);
     return myself;
   }
+
```
joel-costigliola: Javadoc needed.

src/main/java/org/assertj/core/internal/Objects.java
```diff
@@ -651,6 +653,23 @@ private boolean isActualIn(Object actual, Iterable<?> values) {
...
+   * @param info contains information about the assertion.
+   * @param actual the given object.
+   * @param other the object to compare {@code actual} to.
+   * @throws AssertionError if actual is {@code null}.
+   * @throws AssertionError if the other object is not an instance of the actual type.
+   * @throws AssertionError if the actual and the given object are not "deeply" equal.
+   */
+  public <A> void assertIsEqualToComparingFieldByFieldRecursively(AssertionInfo info, A actual, A other) {
+    assertNotNull(info, actual);
+    assertHasSameClassAs(info, actual, other);
```
joel-costigliola: remove the class check, this constraint being sometimes too strong - see [this guy complaining](https://github.com/joel-costigliola/assertj-core/issues/412) ;-)

src/main/java/org/assertj/core/internal/Objects.java
```diff
@@ -651,6 +653,23 @@ private boolean isActualIn(Object actual, Iterable<?> values) {
...
+    if (!DeepEquals.deepEquals(actual, other))
+        throw failures.failure(info, shouldBeEqual(actual, other, StandardRepresentation.STANDARD_REPRESENTATION));
```
joel-costigliola: we will need to call `ShouldBeEqualByComparingOnlyGivenFields` and give it the fields differences.

src/main/java/org/assertj/core/util/DeepEquals.java
```diff
@@ -0,0 +1,653 @@
...
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.Set;
+import java.util.SortedMap;
+import java.util.SortedSet;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Tests two objects for equivalence with a 'deep' comparison.
+ *
```
joel-costigliola: add a link to the original project (even if the licence does not require it)

src/test/java/org/assertj/core/internal/objects/Objects_assertIsEqualToComparingFieldByFieldRecursive_Test.java
```diff
@@ -0,0 +1,94 @@
...
+    actual.name = "John";
+    actual.home.address.number = 1;
+    
+    Person other = new Person();
+    other.name = "John";
+    other.home.address.number = 1;
+    
+    objects.assertIsEqualToComparingFieldByFieldRecursively(someInfo(), actual, other); 
+  }
+
```
joel-costigliola: would be nice to have a test with objects cycle - adding a `neighbour` field to `Person` should be enough.

src/test/java/org/assertj/core/util/TestDeepEquals.java
```diff
@@ -0,0 +1,375 @@
...
+import java.util.concurrent.ConcurrentSkipListMap;
+
+import org.junit.Test;
+
+/**
+ * @author John DeRegnaucourt
+ * @author sapradhan8
+ * @author Pascal Schumacher
+ */
+public class TestDeepEquals {
```
joel-costigliola: let's use AssertJ assertions by adding `assertAreDeepEquals(x,y)` that simply calls `assertThat(DeepEquals.deepEquals(x, y)).isTrue()`

src/test/java/org/assertj/core/util/TestDeepEquals.java
```diff
@@ -0,0 +1,375 @@
...
+        col.add("sierra");
+        col.add("tango");
+        col.add("uniform");
+        col.add("victor");
+        col.add("whiskey");
+        col.add("xray");
+        col.add("yankee");
+    }
+
+    @SuppressWarnings("unchecked")
```
joel-costigliola: let's use our `Sets.newLinkedHashSet`

joel-costigliola:
From what I see, the approach is ok.

PascalSchumacher:
Thanks for the review. :)

I will update the pull request.

src/main/java/org/assertj/core/internal/Objects.java
PascalSchumacher: I know ;), but the current implementation of ´deepEquals´ stops the comparison once it reaches non-Collection, non-Map objects which have different classes.

src/main/java/org/assertj/core/internal/Objects.java
PascalSchumacher: The `deepEquals` implementation currently does not support showing the path to the differences. It also does not consider properties.

I don't think adding robust support for this can be easily done (at least by me :().

Maybe we should change the name to `isDeepEqualTo` and add is `isEqualToComparingFieldByFieldRecursively` with all the features implies later on?

joel-costigliola: Sorry for the late feedback :(
It seems `deepEquals` only helps partially, not considering properties is a blocking issue as we want to have a consistent behaviour between whether the field by field comparison is recursive or not.
Moreover we need an implementation that tracks field/property path, if we only report that objects are not equal but not show why it would not be helpful for users.

I will give it a shot by the end of the year, first by looking for other `deep` equals implementation.
If you want to continue working on this one, feel free to do so, your work is always appreciated.

PascalSchumacher: No problem. :)

Time permitting I will try to add the missing features.

If you or somebody else comes up with a better solution meanwhile, I will close the pull request. I'm doing this to learn and to be able to use the functionality when testing. I'm not expecting to have ever pull request merged.

joel-costigliola:
Next on my todo list review.

PascalSchumacher:
There are no real updates, so no need to review. There still is no support for properties and no support to show the actual difference.

joel-costigliola:
I'm moving this to 2.4.0 for the following reasons:
* no way to compare different types having the same structure
* the reported error message should highlight which field differs
* more unit tests needed
* not to block 2.3.0 release

PascalSchumacher:
I agree. This needs serious work. Maybe it's better to base this feature on the existing assertj field/property support.

PascalSchumacher:
@joel-costigliola I was not able to rebase this successfully, so I created #615 instead.

PR #516, opened 2015-11-03
Issue479

gabga:
Fixed the issue with the solution suggested by Joel.

If you notice any problems, let me know.


joel-costigliola:
Thanks, I just simplified the tests and cleaned up the `Maps` code a little bit.

PR #521, opened 2015-11-13
Android support for v2.x

Turbo87:
This PR continues #500. Please doublecheck if this change has any undesired side effects.


joel-costigliola:
Sadly, some projects (including mine at work) have rules to prevent using inherited static methods, that would prevent using the inherited `Java6Assertions`. The only option is duplicating the class and removing `Path` assertions which makes me sad but I don't see any other options.

I'll make the change if you are busy ...

Turbo87:
that is exactly the side effect I was afraid of. IntelliJ didn't complain about that sort of thing by default but I can see how that would be a no-go for some people... I'm wondering if there is a way to automatically generate such a class instead of all the code duplication

joel-costigliola:
Well, generating the `Assertions` class would require two things:
* introspect everything the implements `Assert` 
* annotate all the utility methods exposed (like `offset`) 
but I'm not sure if we can deal easily with generic method like `offset`, how do you know you have to generate versions for `Double` and `Float` ?

It is simpler to duplicate stuff, if we miss something, users will complain and we will fix it in the next version - not ideal but pragmatic.

joel-costigliola:
Integrated, thanks @Turbo87

kxfang:
Hi, does the work done here cover SoftAssertions? With Android projects I'm seeing the same old nio Path issue when SoftAssertions are used.

joel-costigliola:
No :( but contributions are welcomed ;-)

kxfang:
Guess I'll do something similar to the Java6Assertions patch.

Issue #514, opened 2015-10-31
Add 'isNotEqualToIgnoringWhitespace' assertion


mkordas:
There is `isEqualToIgnoringCase`/`isNotEqualToIgnoringCase` pair, but `isEqualToIgnoringWhitespace` doesn't have corresponding `isNotEqualToIgnoringWhitespace`. Missing assertion should be added.

kryger:
What situation could this be useful in? Can you give an example of a pair of strings that are not equal ONLY if ignoring whitespace? This would imply that they are equal if whitespace *is* taken into account, and by definition trimming whitespace wouldn't change the outcome of the comparison.

mkordas:
@kryger 
My understanding of this assertion is that it would pass if strings are not equal even if all whitespaces are removed.

Example:
```java
assertThat("word").isNotEqualTo("    word    \t\r\n"); // pass - strings contain the same word, but difference is in whitespace
assertThat("word").isNotEqualTo("    word    \t\r\n"); // fail - strings are the same if whitespaces are not taken into account
```
To sum up, this is just stronger version of `isNotEqualTo`. And IMO negative assertions should be as strong as possible, so it may be useful in some contexts.

joel-costigliola:
if someone needs it, I'm ok to add it.

joel-costigliola:
Integrated in 2.x and soonish in master.

PR #528, opened 2015-12-01
Fix for #514

DanCorder:
Changes to implement 'isNotEqualToIgnoringWhitespace' assertion.


joel-costigliola:
Thanks, will have a look at it soon.

DanCorder:
I'm not very happy about the way that JUnit handles parameterised tests. Would you be OK with adding something like https://github.com/Pragmatists/junitparams or https://github.com/piotrturski/zohhak as a test dependency to the project?

joel-costigliola:
Yes  no problem, I kinda like https://github.com/piotrturski/zohhak better.

PascalSchumacher:
https://github.com/TNG/junit-dataprovider is another possibility

DanCorder:
Ok, I'll do a separate patch for parameterised tests, as they are used in other places too.

joel-costigliola:
Thanks Dan, integrated.

PR #530, opened 2015-12-03
make hasBeenThrown public accessible

qoomon:
can be public without a doubt. I need this to have a test code like this:

   @Test
    public void expectExceptionLambda() throws Exception {
        // give
        Thrower thrower = new Thrower();

        // when
        Throwable exception = catchThrowable(() -> thrower.throwNullPointerException("boom"));

        // then
        assertThat(exception).hasBeenThrown().hasMessage("boom");
    }

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
@@ -39,7 +39,7 @@ public abstract class AbstractThrowableAssert<S extends AbstractThrowableAssert<
     super(actual, selfType);
   }
 
-  protected S hasBeenThrown() {
+  public S hasBeenThrown() {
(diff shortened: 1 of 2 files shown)


joel-costigliola:
You actually don't need to call `hasBeenThrown()` at all since `hasMessage` checks for `exception` not to be null.
The only difference would be the error message.

qoomon:
Yes however i just need to check if there is an exception, the message does not matter.

joel-costigliola:
you can use `isNotNull()` to check there is "something"

qoomon:
maybe your right :-) but i will create an other PR for adding a better catchThrowable() Method to only catch explicit exceptions.

Thanks for response

joel-costigliola:
Ok, looking forward to your PR.

PR #531, opened 2015-12-04
+ catchThrowable for expected Exceptions only

qoomon:
so you can differ between test who have unexpected exception (Test Flag is typically Red) and those who throws normal assertionsErrors  (Test Flag is typically Yellow/Orange)


joel-costigliola:
I'm not sure to understand this PR, you have just added a private method that is not used (?)
What is the problem you are trying to solve ? 
Could you add a unit test showing how this works ?
Thanks

Issue #210, opened 2014-05-30
Diff.diff to produce a more concise diff output


kriegfrj:
The built-in text-based diff in internal.Diff can produce very verbose output for very simple changes. For example, consider comparing two Strings/text files:

String A: 1000 lines of text.
String B: An newline character followed by String A.

Diff.diff will output 1000 lines of error message to describe a single character difference. If the strings additionally differed on line 500 then it is likely that this difference would get lost in the message.

I stumbled into this issue when working on the [assertions generator](/joel-costigliola/assertj-assertions-generator). Though not essential, it would be nice if Diff.diff tried harder to produce a more concise representation of the actual differences (akin to Unix diff and other similar utilities).

joel-costigliola:
Done !

PR #532, opened 2015-12-04
#210: Use com.googlecode.java-diff-utils library to produce better diffs

ChrisA89:
I've added a dependency here. The library is on Google Code (so no longer maintained), but I couldn't find an official version on github.

The tests have changed ever so slightly as well, as EOF is now reported as empty string.

The other thing of note, is reading all of the buffered reader into memory, but I'm not sure this is avoidable with a better diffing algorithm.

Happy to make any markups if there are things in this you'd prefer done another way.


joel-costigliola:
Thanks for the PR.
By design, assertj-core does not have any dependencies, would it be easy to include the classes directly (and mention the original authors) ? 
I would also add a test where two files content are the same but one has an extra lines at the beginning, with the old algorithm, it leads to a lot of differences whereas there is really only one, with this PR I hope it only reports one diff.

ChrisA89:
I've pulled the files in. It's all under the Apache license (which is the same one you're using). I think it would be good to credit the original project, but I'm not sure where to put that.

The pull request now includes:
 - The diff library
 - Tests for the diff library
 - Updating the Diff assertion to use the diff library
 - An extra test to check that a newline at the start of a stream gives a useful output.

I've also just seen that the build didn't pass, so I'll go fix that.

PascalSchumacher:
There is also a (no longer maintained) fork on github https://github.com/KengoTODA/java-diff-utils which published three releases with bugfixes and new features. I do not know if any of them are relevant in this context.

joel-costigliola:
@PascalSchumacher I haven't seen anything relevant in the fork, moreover it depends on Guava so let's try without it.

joel-costigliola:
Thanks for your work, I have made it Java 7 compatible and integrated in 2.x after improving the diff output to make it more readable, example:

```java
// writer create files
writer.write(actual,   "line1",                     "line2", "line3", "line4", "line5", "line 9", "line 10", "line 11");
writer.write(expected, "line1", "line1a", "line1b", "line2", "line3", "line7", "line5");
```
gives the following diff output when executing `diff.diff(actual, expected)`
```
[
Missing content at line 2:
  ["line1a",
   "line1b"]
, 
Changed content at line 6:
expecting:
  ["line7"]
but was:
  ["line4"]
, 
Extra content at line 8:
  ["line 9",
   "line 10",
   "line 11"]
]
```

PascalSchumacher:
Great news :) I recently tried to to use java-diff-utils to improve text comparison in tests, but did not get very far.

PR #535, opened 2015-12-07
Use JUnitParams for parameterised tests

DanCorder:
I tried zohhak, but I couldn't see a way of nicely passing in array parameters to tests for ArrayWrapperList_size_Test so I switched to JUnitParams.
Hopefully this is the right branch this time :)


piotrturski:
i've just seen your attempt to use zohhak. coercing to an array (as to any other type) has to be programmed. samples: https://github.com/piotrturski/codility-starter/blob/803f7045b2ec82db7eb0a3d014a02c9f81a27d3f/src/test/java/SolutionTest.java#L45

or a bit better:
```
// "a b c" -> ["a", "b", "c"]
    @Coercion
    public String[] toArray2(String input) {
		return spaceSeparatedStream(input).toArray(String[]::new);
    }

// "1 2 -3" -> [1,2,-3]
    @Coercion
    public int[] toArray(String input) {
    	return spaceSeparatedStream(input).mapToInt(Integer::parseInt).toArray();
    }

 private Stream<String> spaceSeparatedStream(String input) {
    	return Arrays.stream(
    			StringUtils.isEmpty(input)? EMPTY_STRING_ARRAY : input.split(" ")
    			);
    }
```

DanCorder:
Thanks, I didn't fancy writing string parsing code to create the arrays (although your spaceSeparatedStream is quite neat). I also wasn't keen on needing to use a magic string to be able to use null in a coercion method. One the other hand, whilst the JUnitParams allows me to use objects directly they are in the generator method rather than right there on the test.
If there are strong opinions that zohhak is the way to go I can make the switch.

PascalSchumacher:
With https://github.com/TNG/junit-dataprovider/wiki/Features you could do it like this:

```java
@RunWith(DataProviderRunner.class)
public class ArrayWrapperList_size_Test {

    @Test
    @DataProvider({
            "0, 1, 2",
            "0",
            ""
        })
    public void should_return_size_of_array(int... array) {
        System.out.println(java.util.Arrays.toString(array));
        ArrayWrapperList list = new ArrayWrapperList(array);
        assertThat(list.size()).isEqualTo(array.length);
    }
}
```

joel-costigliola:
I have integrated Dan's code (thanks Dan).

After a second look I find junit-dataprovider quite neat as you don't have to worry that `@Parameters method` value does not match the actual method called.

I'll might give it a go if I have time ... (which I realize is unlikely)

PascalSchumacher:
I will submit a pull request for the switch to junit-dataprovider.

Issue #505, opened 2015-10-11
Add offset support to float/double arrays check methods


victor-zelinskiy:
Please add offset support to float/double arrays check, like you did with one double value check, without this I can't contains check double arrays, I get errors like this:
java.lang.AssertionError: 
Expecting:
  <[7.199999999999999,
    3.5999999999999996,
    -12.0,
    30.599999999999998,
    0.30000000000000004]>
to contain exactly (and in same order):
  <[7.2, 3.6, -12.0, 30.6, 0.3]>

joel-costigliola:
I'm ok with the idea but I have trouble finding a good API, i.e. how would you name the assertion ? 

This type of use case are nicely handled with an element comparator:
```java
@Test
public void comparing_array_of_real_numbers() {
  Comparator<Double> closeToComparator = new Comparator<Double>() {
    @Override
    public int compare(Double o1, Double o2) {
      return Math.abs(o1.doubleValue() - o2.doubleValue()) < 0.001 ? 0 : -1;
    }
  };
  assertThat(new double[] { 7.2, 3.6, -12.0 }).usingElementComparator(closeToComparator)
                                              .containsExactly(7.2000001, 3.5999999, -12.000001);
}
```

zilidev:
Why not just keep current names but add overloaded versions with offset argument, it's a good practice to use offset wherever is a double comparison, for example starndart jUnit assert library even don't have methods to check double arrays without offset, they have only method org.junit.Assert.assertArrayEquals(double[] expecteds, double[] actuals, double delta)
I'd like to see somthing like this in your library. Btw thank you for AssertJ library, it is very useful.

joel-costigliola:
fair enough, but won't be able to change assertions using varargs, will look like:
```java
assertThat(new double[] { 1.0, 2.0 }).containsExactly(new double[] { 0.999, 2.001 }, 0.001);
```

nbogdan:
Hi, I've been trying to work on this. (it looked like good practice for a beginner). For containsExactly I've obtained:

```java
public S containsExactly(double[] values, double precision) {
    usingElementComparator(getComparatorWithPrecision(precision));
    arrays.assertContainsExactly(info, actual, values);
    return myself;
  }
```

But it seems to fail the tests with the actual array being seen as empty. Any suggestions?

PascalSchumacher:
I think this issues should not be closed yet, because the requested offset support for float arrays is not implemented yet. What do you think?

[...]

PR #536, opened 2015-12-07
Added support for specifying a precision with which to compare elements of AbstractDoubleArrays

nbogdan:
This request is in relation with issue #505 . I attempted modify the containXXX and doesNotContainXXX methods from AbstractDoubleArrays.
Awaiting a review and possible recommendations on how to write the tests.


joel-costigliola:
Thanks.

I would only replace the last double by an `Offset<Double>` to clarify the intent of this last parameter and to be consistent with `DoubleAssert#isCloseTo(final double other, final Offset<Double> offset)`.

Add unit test following what was done in `DoubleArrays_assertContains_Test` and `DoubleArrayAssert_contains_Test`.

I'm also thinking to deprecate the existing `containsXXX` methods as it is best practice to give an offset.

joel-costigliola:
Finished

PR #541, opened 2015-12-13
ShouldHaveContent_create_Test should succeed when defaultCharset is n…

PascalSchumacher:
…ot UTF-8


joel-costigliola:
Raah, good catch

joel-costigliola:
done thx.

PR #549, opened 2015-12-20
Add Assertion to check that a map contains all entries of another map…

### LINKED ISSUE #545: Assertion to check that a map contains all entries of another map
I think it would be nice to have an Assertion to check that a map contains all entries of another map. What do you think?

PascalSchumacher:
… (fixes #545)

Not sure about the name and please suggest any other improvements (as usual :)).


joel-costigliola:
Integrated in 2.x (nothong Java 8 specific), renamed the assertion to `containsAllEntriesOf` for consistency and fix javadoc for all methods that was not escaping `<>` properly.

PascalSchumacher:
Thanks. :)

I will create non-java8 specific pull request from the 2.x branch next time.

joel-costigliola:
:+1:

PR #553, opened 2015-12-23
extracting method on objects (fixes #411)

### LINKED ISSUE #411: Feature Request: extracting method on objects
What I would like to do:

```java
public class Person {
     String firstname, lastname;
     Address adress;
}

then(new Person()).extracting("firstname", "lastname").doesNotContainNull();
```

Maybe this is already possible?

PascalSchumacher:
Maybe extracting to a `List` instead of an `array` would be better?


joel-costigliola:
Integrated, I really like this feature BTW.

PascalSchumacher:
Thanks!  :)

PR #557, opened 2015-12-26
simplify code by using try-with-resource statements

PascalSchumacher:


PascalSchumacher:
Just noticed that this will not work on older android versions: http://stackoverflow.com/questions/24290830/try-with-resources-are-not-supported-at-this-language-level-android

So merging this would break the support of android versions supporting only java 6 added with https://github.com/joel-costigliola/assertj-core/pull/521

joel-costigliola:
I can live without try-with-resource, although useful I'm not fond of how it looks in the code (that is very personal) and there is not a lot of places where it can be used anyway.

joel-costigliola:
Thanks for the investigation, btw.

PR #558, opened 2015-12-26
refactor Diff to reduce duplication

PascalSchumacher:


joel-costigliola:
Much cleaner indeed, thx.

PR #559, opened 2015-12-26
use Preconditions#checkNotNull instead of throwing a NullPointerException

PascalSchumacher:
If you prefer the usage without a static import I will update the pull request accordingly.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractFileAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractFileAssert.java
@@ -12,6 +12,8 @@
  */
 package org.assertj.core.api;
 
+import static org.assertj.core.util.Preconditions.checkNotNull;
+
 import org.assertj.core.api.exception.RuntimeIOException;
 import org.assertj.core.internal.Files;
 import org.assertj.core.util.VisibleForTesting;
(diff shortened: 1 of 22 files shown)


joel-costigliola:
No it's ok, static import are fine when they make the code more readable.

PascalSchumacher:
Just noticed that this will not work on older android versions: http://stackoverflow.com/questions/28613784/noclassdeffounderror-java-util-objects-android

So merging this would break the support of android versions supporting only java 6 added with https://github.com/joel-costigliola/assertj-core/pull/521

joel-costigliola:
Let's not use it then, I'm gonna remove the two places where `java.util.Objects` is used.

joel-costigliola:
I actually like the idea, let's introduce our `requireNonNull` in `org.assertj.core.util.Objects`.

PascalSchumacher:
O.k. I will update the pull request.

PascalSchumacher:
I changed the pull request to use `Preconditions#checkNotNull` instead of `Objects#requireNonNull`.

joel-costigliola:
:+1:

joel-costigliola:
Integrated, thanks.

PR #565, opened 2015-12-30
add missing examples to javadoc

PascalSchumacher:
add missing casts to short in javadoc examples


joel-costigliola:
good catch :)

PR #574, opened 2016-01-04
Check if binary files are identical before checking "readable" line by line

MichaelBitard:
When you had non UTF-8 (defaultCharset) characters in the files, you ended with a "Non readable file" caused by a  MalformedUrlException.

Now the `assertSameContentAs` method first check if the files have no differences "bit-wise" before trying to compare "readable line" by "readable line".

What Could be improved:
 * the unMockedFiles I created

It seems strange to have mocked Files under test, I had to change 2 tests (`should_pass_if_files_have_equal_content` and `should_throw_error_wrapping_catched_IOException`) because the internal method changed, whereas the behavior of the tested method didn't (it seems to me that these tests are testing the internal implementation)

This PR is part of https://github.com/joel-costigliola/assertj-core/issues/504 but seems independent enough to have its own PR.


joel-costigliola:
I'm not sure what you mean, `Files` is not mocked, its dependencies are, the rationale being to isolate it, I tend to think it's a good thing it one wants to write a true unit test. (btw, test was written in Fest 2.x that AssertJ is a fork of).

src/main/java/org/assertj/core/internal/Files.java
```diff
@@ -88,10 +89,15 @@ public static Files instance() {
...
     assertIsFile(info, actual);
+    BinaryDiffResult binaryDiffResult = null;
     try {
+      binaryDiffResult = binaryDiff.diff(actual, java.nio.file.Files.readAllBytes(expected.toPath()));
+      if (binaryDiffResult.hasNoDiff()) return;
       List<Delta<String>> diffs = diff.diff(actual, expected);
       if (diffs.isEmpty()) return;
       throw failures.failure(info, shouldHaveSameContent(actual, expected, diffs));
+    } catch (MalformedInputException e) {
+      throw failures.failure(info, shouldHaveBinaryContent(actual, binaryDiffResult));
```
joel-costigliola: I don't think it is correct to return an `shouldHaveBinaryContent` failure, it we can't read the file you can't say anything about its content, I think it's should be treated as an `IOException`. WDYT ?

src/main/java/org/assertj/core/internal/Files.java
```diff
@@ -88,10 +89,15 @@ public static Files instance() {
   public void assertSameContentAs(AssertionInfo info, File actual, File expected) {
     verifyIsFile(expected);
     assertIsFile(info, actual);
+    BinaryDiffResult binaryDiffResult = null;
     try {
+      binaryDiffResult = binaryDiff.diff(actual, java.nio.file.Files.readAllBytes(expected.toPath()));
```
joel-costigliola: `hasSameContentAs` should be updated to make ti clear that binary comparison is supported.

src/test/java/org/assertj/core/internal/files/Files_assertSameContentAs_Test.java
```diff
@@ -115,4 +122,29 @@ public void should_fail_if_files_do_not_have_equal_content() throws IOException
...
   }
+
+  @Test
+  public void should_pass_if_files_are_binary_identical() throws IOException {
+    unMockedFiles.assertSameContentAs(someInfo(), createFileWithNonUTF8Character(), createFileWithNonUTF8Character());
+  }
+
+  @Test
+  public void should_fail_if_files_are_not_binary_identical() throws IOException {
+    Throwable throwable = catchThrowable(() -> unMockedFiles.assertSameContentAs(someInfo(), createFileWithNonUTF8Character(), expected));
```
joel-costigliola: Because there is nothing Java 8 specific in comparing binary content, the code should be merged in assertj core 2.x   which means you shoudn't use lambdas (yes it sucks I know).

src/main/java/org/assertj/core/internal/Files.java
joel-costigliola: In our case, I'm not sure it is possible to have different binary content but equal text content as we end comparing text content with the default charset - which makes me think that users should probably be able to set the Charset for comparing text files (not possible for the time being), that would mean adding this to `AbstractFileAssert`:
```java
public S hasSameContentAs(File expected, Charset charset)
```

src/main/java/org/assertj/core/internal/Files.java
MichaelBitard: I did that because when you have a binary file which cannot be readLine(), you get an ugly exception : 
`Unable to compare contents of files:</tmp/junit6310584610408013618/junit1823075884829217683.tmp> and:</tmp/junit6310584610408013618/junit174924993450801969.tmp>` whereas the binary diff is able to say where the file differ, bit-wise (I think it's better to have an error saying that the files are binary differents than en error saying that we cannot compare the contents of the files). If you dont agree I can remove this part, just keep in mind that MalformedInpotException is only thrown by the `reader.readLine()` which is used only in the diff part (not the binaryDiff)

src/main/java/org/assertj/core/internal/Files.java
MichaelBitard: I agree with you, It's not possible to have different binary content and equal text content.

The global algorithm is: 
```
If 
 binary equivalent: return ok
else
 compare line by line
 if line by line comparison is not possible, return binary comparison
 return the differences, human readable
```

So in this case the `diff` is only used to return the most readable thing possible:
 1. the line by line diff
 2. the binary diff
 3. the `Unable to compare contents of files` error

For the ```java
public S hasSameContentAs(File expected, Charset charset)
```
I can add it but I'd rather do it in another PR, what do you think ?

src/test/java/org/assertj/core/internal/files/Files_assertSameContentAs_Test.java
MichaelBitard: Ok, I change that.

I make the PR for the 2.x branch or you will cherry-pick the commits ?

MichaelBitard:
You have any preferences on the PR lifeCycle ?
I know some like to have only one commit to merge (so I push force my branch until it is correct) but other prefer multiple commits (which can be more cumbersome to merge with the 2.x branch

MichaelBitard:
I assume the travis error is not related to my PR, right ? https://travis-ci.org/joel-costigliola/assertj-core/builds/100338477

src/main/java/org/assertj/core/internal/Files.java
joel-costigliola: ok, fine with that if the error is more relevant.

joel-costigliola:
for the PR lifecyle, the best would be a 2.x PR but a master is also ok, I usually never merge PR from github but rebase the branch and squash everything to one logical commit, I find it makes the history cleaner. 

the error is not related to your PR, I had similar issues depending on the JDK 8 version (I will check if I can set a specific JDK 8 version on Travis), as long as it can be built on my machine, I'm ok with that.

MichaelBitard:
Ok, I will make a "one commit" PR for the 2.x branch, so I can do it for this PR too.
Let me know if you think that this PR is complete and I'll do it.

joel-costigliola:
Integrated in 2.x, thanks.

PascalSchumacher:
Sorry for the being late,

but I'm not sure this change is for the best. 

Before `assertSameContentAs` was basically `assertSameTextContentsAs`. Now the assertion succeeds if the files are binary equal, but files using  different encodings can be binary equal but have still different text content. I do not know how likely that is, but I thought it worth mentioning.

If this new behavior is kept, I think there should be some mention in the javadoc (examples) that the file contents are binary compared first. Also the change should also be done for `AbstractPathAssert.hasSameContentAs`.

joel-costigliola:
I think it's highly improbable but I agree it's not correct

joel-costigliola:
I'll change the behaviour and only do the binary diff when a `MalformedInputException` occurs

MichaelBitard:
Good idea!

joel-costigliola:
pushed force to get rid of the previous commit message now irrelevant.

PascalSchumacher:
Thanks!

PR #578, opened 2016-01-08
#assertHasSameContentAs should use the charset set by #usingCharset t…

### LINKED ISSUE #576: #assertHasSameContentAs should use the charset set by #usingCharset to read the actual file/path
I guess these methods should use the charset set by #usingCharset to read the actual file (like #hasContent does)

@joel-costigliola If you agree I will provide a pull request for this change.

PascalSchumacher:
…o read the actual file/path (fixes #576)


joel-costigliola:
Good work, thanks.

PR #591, opened 2016-01-21
Ignore synthetic fields when comparing fields on an object

petergphillips:
Patch for ignoring synthetic fields - see https://github.com/joel-costigliola/assertj-core/issues/590.  I've targetted 2.x branch as the issue was tagged with 2.4.

Note that this could be potentially breaking change in that someone who was relying on the synthetic fields being compared (and being different) could then have test failures as the behaviour has now changed.  This change only affects the `isEqualToComparingFieldByField`, `isEqualToIgnoringGivenFields` and `isEqualToIgnoringNullFields` methods though, so for example explicitly specifying the field in `isEqualToComparingOnlyGivenFields` will still pass.


joel-costigliola:
Thanks ! PR integrated.

PR #599, opened 2016-01-28
try to keep 2.x as android compatible as possible by removing referen…

PascalSchumacher:
…ces to java.nio.file.Path in Files#assertSameContentAs


PascalSchumacher:
I did not find any existing helper method I could use, please point me to it if I missed it.

PascalSchumacher:
Just noticed that `Diff` also uses `java.nio.file.Path`, so this does not really help.

PR #601, opened 2016-01-29
MWF nexient patch 1

### LINKED ISSUE #598: SoftProxies package access hinders custom collectors
Wrote a custom SoftAssertion collector, but hit a snag while trying to access the `proxies` field inherited from AbstractStandardSoftAssertions and AbstractSoftAssertions. The field is protected, but SoftProxies's methods are package-private, meaning any custom collectors either need to have the same package (org.assertj.core.api) or use reflection to access either errorsCollected() or the collector field.

I don't know if this if this clashes with #339 or not.

MWFNexient:
This contribution attempts to close #598 by

1. adding errorsCollected() to AbstractSoftAssertions
2. removing the need from the various *SoftAssertion classes from accessing their proxies field directly
3. Testing that custom collectors with non-package access can access the errors they've collected

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
@@ -12,6 +12,8 @@
  */
 package org.assertj.core.api;
 
+import java.util.List;
+
 public class AbstractSoftAssertions {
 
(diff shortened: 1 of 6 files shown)


joel-costigliola:
Integrated, thanks

PR #616, opened 2016-02-23
Allow to set a percentage greater than 100% for isCloseToPercentage a…

PascalSchumacher:
…ssertions.

make javadoc reflect this change


joel-costigliola:
thanks

PR #618, opened 2016-02-25
Changed constructor of AbstractZonedDateTimeAssert to protected

ntroutman:
Changed constructor of AbstractZonedDateTimeAssert to protected from package-private to allow custom extensions.


joel-costigliola:
Integrated, will be in 3.4.0

PR #625, opened 2016-03-08
add more tests for assertIsCloseTo

PascalSchumacher:
just the result of a quick experiment, so no problem if you reject this


joel-costigliola:
In 2.x, cheers.

Issue #455, opened 2015-07-04
Allow to register custom object formatter in error messages


joel-costigliola:
Needed to format `Multimap` on multiple lines in assertj-guava but probably useful if users want to describe objects in a specific way (different from their `toString` representation).

PR #627, opened 2016-03-11
Allow to register custom object formatter in error messages #455

PascalSchumacher:
I would like to use this feature, so I put together a quick prototype. If you rather not expose `Representation`, I'm open to every other approach.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -23,6 +23,7 @@ import org.assertj.core.internal.ComparatorBasedComparisonStrategy;
 import org.assertj.core.internal.Conditions;
 import org.assertj.core.internal.Failures;
 import org.assertj.core.internal.Objects;
+import org.assertj.core.presentation.Representation;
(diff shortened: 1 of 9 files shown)


joel-costigliola:
Cool, will look into it soon but first I'm preparing 2.4.0/3.4.0 releases

PascalSchumacher:
No need to hurry.

Nice to hear that 2.4.0/3.4.0 will be released soon. :smiley:

joel-costigliola:
I think this is on the right track, I'm going to continue it, I want to refactor the `Representation` hierarchy and introduce a few things:
* a way to register a representation globally (not having to specify it for each assertions)
* a way to register in the current representation the description of a specific type without having to create a new representation (I think it is only possible using Java 8 lambda though)

PascalSchumacher:
Great, these enhancements will make the feature much more powerful. :+1:

PascalSchumacher:
Was merged and greatly improved upon by @joel-costigliola ! Thanks!

See https://github.com/joel-costigliola/assertj-core/issues/455 for details.

PR #629, opened 2016-03-12
Reuse empty Optional error message in other situations

pivovarit:
When we are comparing Optional's content against something else and Optional turns out to be empty, we get:

`Expecting:
  <Optional.empty>
to contain:
  <10>
but did not."`

I made it reuse already defined error message for the empty Optional case:

`Expecting Optional to contain:
  <10>
but was empty."`


joel-costigliola:
@pivovarit thanks for the PR !

pivovarit:
:+1:

Issue #589, opened 2016-01-19
Add an assertion to check that list has at least/at most 'x' elements


vanta:
Currently there is only a method to check concrete size of collection. It would be helpful to be able to assert for minimal/maximal size of list that one is interested in.

joel-costigliola:
I would just get the size of the collection and use `isBetween` assertion, WDYT ?

vanta:
Of course, it can be done this way, but with such approach hasSize() method
so it's not needed :) I always though that assertJ goal is to simplify
devs' lifes by allowing to assert in method chainging.

On Wed, Jan 20, 2016 at 7:17 AM, Joel Costigliola <notifications@github.com>
wrote:

> I would just get the size of the collection and use isBetween assertion,
> WDYT ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/joel-costigliola/assertj-core/issues/589#issuecomment-173107237>
> .
>



-- 
Krzysztof Wolny

joel-costigliola:
ok, would you be happy with `hasSizeBetween()` ? :)

vanta:
I would :)

On Wed, Jan 20, 2016 at 8:58 AM, Joel Costigliola <notifications@github.com>
wrote:

> ok, would you be happy with hasSizeBetween() ? :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/joel-costigliola/assertj-core/issues/589#issuecomment-173122395>
> .
>



-- 
Krzysztof Wolny

joel-costigliola:
Fixed with #301.

Issue #301, opened 2014-12-18
more assertions on collection sizes


jstrachan:
I figured on a ListAssert<T> I'd be able to do things like...

```
assertThat(someList).size().isGreaterThan(4);
```
i.e. that a method size() would be available returning the IntegerAssert for further more expressive assertions on a List size.

I can submit a PR if you like?

joel-costigliola:
Fixed.

PR #630, opened 2016-03-13
enable fluent assertions on size of list #301, #589

PascalSchumacher:
I adopted the navigable assertion idea of assertj-db. 

It's different from assertj's current api, but having the option to use the following assertion chain:

```java
List<Integer> ints = Arrays.asList(1, 2, 3);
assertThat(ints).size().isGreaterThan(0).isLessThanOrEqualTo(3).returnToList().contains(1).doesNotContain(4);
```

is interesting I think.


PascalSchumacher:
Only works for lists of integers. :(

I got confused by all the generics.

PR #631, opened 2016-03-13
enable fluent assertions on size of list (#301, #589) and map

PascalSchumacher:
I adopted the navigable assertion idea of assertj-db. 

It's different from assertj's current api, but having the option to use the following assertion chain:

```java
List<Integer> ints = Arrays.asList(1, 2, 3);
assertThat(ints).size().isGreaterThan(0).isLessThanOrEqualTo(3).returnToList().contains(1).doesNotContain(4);
```

is interesting I think. (This is a fixed version of #630. I had to create a new pull request as I could not reopen the old one.)

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
@@ -44,6 +44,8 @@ import org.assertj.core.internal.Iterables;
 import org.assertj.core.internal.ObjectArrays;
 import org.assertj.core.internal.Objects;
 import org.assertj.core.internal.OnFieldsComparator;
+import org.assertj.core.util.IterableUtil;
+import org.assertj.core.util.Preconditions;
 import org.assertj.core.util.VisibleForTesting;
 import org.assertj.core.util.introspection.IntrospectionError;
 
@@ -1514,5 +1516,11 @@ public abstract class AbstractIterableAssert<S extends AbstractIterableAssert<S,
   @Override
   public S withThreadDumpOnError() {
     return super.withThreadDumpOnError();
+  }
+
+  @SuppressWarnings({ "rawtypes", "unchecked" })
+  public AbstractIterableSizeAssert<S, A, T> size() {
+    Preconditions.checkNotNull(actual, "Can not assert on size of iterable which is null.");
+    return new IterableSizeAssert(this, IterableUtil.sizeOf(actual));
   }
 }
--- a/src/main/java/org/assertj/core/api/IterableSizeAssert.java
+++ b/src/main/java/org/assertj/core/api/IterableSizeAssert.java
@@ -1,0 +1,27 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2016 the original author or authors.
+ */
+package org.assertj.core.api;
+
+public class IterableSizeAssert<T> extends AbstractIterableSizeAssert<IterableAssert<T>, Iterable<? extends T>, T> {
+
+  private AbstractIterableAssert<IterableAssert<T>, Iterable<? extends T>, T> source;
+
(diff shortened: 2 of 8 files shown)


src/main/java/org/assertj/core/api/AbstractSizeAssert.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.util.List;
+
+public abstract class AbstractSizeAssert<S extends AbstractListAssert<S, A, T>, A extends List<? extends T>, T> extends AbstractIntegerAssert<SizeAssert<T, Integer>>{
+
+  protected AbstractSizeAssert(Integer actual, Class<?> selfType) {
+    super(actual, selfType);
+  }
+
+  public abstract AbstractListAssert<S, A, T> returnToList();
```
joel-costigliola: would it be possible to simply return `S` ?

joel-costigliola:
I think this should not be limited to `List`, we should provide the same feature for `Iterable` (although it does not provide a `size()` it is possible to get it with `IterableUtil.sizeOf`.

This would involve changing `returnToList` to `returnToIterable`.

src/main/java/org/assertj/core/api/AbstractSizeAssert.java
PascalSchumacher: Seems possible. :)

PascalSchumacher:
I agree that this would also be nice for `Iterable`s, but when I do the change I get:

>java.lang.ClassCastException: org.assertj.core.api.ListAssert cannot be cast to org.assertj.core.api.IterableAssert
	at org.assertj.core.api.IterableSizeAssert.returnToIterable(IterableSizeAssert.java:27)
	at org.assertj.core.api.list.ListAssert_size_Test.should_be_able_to_use_integer_assertions_on_size_of_list(ListAssert_size_Test.java:34)

Any ideas how to make this possible are greatly appreciated.

PascalSchumacher:
But even if the exception can be prevented, there is a major disadvantage to this approach:

You can only use `IterableAssert` (not `ListAssert`) on lists.

I think a way to solve this would be to make both `AbstractIterableAssert` and `AbstractListAssert` implement `#size`, but that would mean `AbstractListAssert` can not extend `AbstractIterableAssert`.

src/main/java/org/assertj/core/api/AbstractIterableAssert.java
```diff
@@ -1515,4 +1517,10 @@ public S withFailMessage(String newErrorMessage, Object... args) {
   public S withThreadDumpOnError() {
     return super.withThreadDumpOnError();
   }
+  
+  @SuppressWarnings({ "rawtypes", "unchecked" })
+  public IterableSizeAssert<T, Integer> size() {
```
joel-costigliola: try to declare returning an `AbstractIterableAssert`, I don't see a need to return an `IterableSizeAssert` (but keep building an `IterableSizeAssert`) and see if it solved the cast pb...

src/main/java/org/assertj/core/api/IterableSizeAssert.java
```diff
@@ -0,0 +1,29 @@
...
+  
+  private AbstractIterableAssert<IterableAssert<T>, Iterable<? extends T>, T> source;
+
+  public IterableSizeAssert(AbstractIterableAssert<IterableAssert<T>, Iterable<? extends T>, T> source, Integer i) {
+    super(i, IterableSizeAssert.class);
+    this.source = source;
+  }
+
+  @Override
+  public IterableAssert<T> returnToIterable() {
```
joel-costigliola: as per my other comment, return `AbstractIterableAssert`

src/main/java/org/assertj/core/api/ListSizeAssert.java
```diff
@@ -0,0 +1,30 @@
...
+  
+  private AbstractListAssert<ListAssert<T>, List<? extends T>, T> source;
+
+  public ListSizeAssert(AbstractListAssert<ListAssert<T>, List<? extends T>, T> source, Integer i) {
+    super(i, ListSizeAssert.class);
+    this.source = source;
+  }
+  
+  @Override
+  public ListAssert<T> returnToList() {
```
joel-costigliola: return `AbstractListAssert`

PascalSchumacher:
@joel-costigliola Thanks a lot for the help! Now it works. :)

I guess there is no easy way around this limitation:

>You can only use IterableAssert (not ListAssert) on lists.

(after calling `size`)?

PascalSchumacher:
Updated according to @joel-costigliola great suggestions and added the feature for maps as well (as suggested by @jstrachan in https://github.com/joel-costigliola/assertj-assertions-generator/issues/67).

joel-costigliola:
Integrated (just added javadoc).
Thanks Pascal.

PascalSchumacher:
Thanks for merging and for adding the javadoc. :)

PR #633, opened 2016-03-15
update dependencies to most recent versions

PascalSchumacher:


joel-costigliola:
Partly integrated, cglib 3.2.1 has some regressions I'm wary to use

PR #642, opened 2016-04-12
fixes #641 to add a NavigationListAssert for being able to navigate into elements in a list and chain assertions

### LINKED ISSUE #641: add a NavigationListAssert<T,EA> extends ListAssert<T> for easier navigation through generated Assert models
this is an implementation class used to provide generated navigation methods through generated Assert classes. 

For background see this whole thread for https://github.com/joel-costigliola/assertj-assertions-generator/issues/67

In particular for [here's an example navigating through lists](https://github.com/joel-costigliola/assertj-assertions-generator/issues/67#issuecomment-208756239)

So these classes are really intended to be reused by folks generating assertion classes using navigable list properties.

PR on its way shortly...

jstrachan:
fixes #641 to add a NavigationListAssert for being able to navigate into elements in a list and chain assertions

### DIFF
--- a/src/main/java/org/assertj/core/api/AssertFactory.java
+++ b/src/main/java/org/assertj/core/api/AssertFactory.java
@@ -1,0 +1,32 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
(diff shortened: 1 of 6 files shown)


joel-costigliola:
@jstrachan I'm not fond of the `Descriptions` class, I'd like to meld it into `NavigationListAssert`, can you give me a reason not to ?

jstrachan:
TBH I'm not solid on the Descriptions class either ;). Given its only one simple method I'm totally happy with your approach of using NavigationListAssert

joel-costigliola:
cool :)
I have made good progress on #641 based on your work, I'm not finished though, I'm gonna try allowing navigation on `Iterable` instead of being limited to `List`.

jstrachan:
Great stuff thanks! Yeah working on Iterable/Collection/Set would be handy too.

joel-costigliola:
Done ... finally, thanks @jstrachan I have modified few things from the original PR, don't hesitate to have a look and tell me if I screwed up !

PR #645, opened 2016-04-17
add #usingRecursiveFieldByFieldElementComparator to AbstractIterableA…

### LINKED ISSUE #636: Better support comparing lists / iterable objects
There is new feature in 3.4.0 version:
`assertThat(...).isEqualToComparingFieldByFieldRecursively(...)`

but comparing lists is still poor, because we have to use equals(...) method when comparing two objects - it is used by default?

We would like to have methods like:

```
assertThat(<some iterable>).containsExactlyElementsOf(<some iterable>) - already provided.
assertThat(<some iterable>).containsExactlyElementsOfComparingFieldByField(<some iterable>) - really missing
assertThat(<some iterable>).containsExactlyElementsOfComparingFieldByFieldRecursively(<some iterable>) - missing too 
```

As suggested by @PascalSchumacher we can use:
```
assertThat((Object) Arrays.asList("a", "c")).isEqualToComparingFieldByFieldRecursively(Arrays.asList("a", "c"));
assertThat((Object) Arrays.asList("a", "c")).isEqualToComparingFieldByField(Arrays.asList("a", "c"));
```

But this approach works only for identical Lists - internal state of lists such as private fields other than compared objects will be compared too.

PascalSchumacher:
…ssert and AbstractObjectArrayAssert (fixes #636)

javadoc is still missing, but I like to have the approach review before investing further effort

@joel-costigliola no need to hurry with the review

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractObjectArrayAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractObjectArrayAssert.java
@@ -42,6 +42,7 @@ import org.assertj.core.internal.ObjectArrayElementComparisonStrategy;
 import org.assertj.core.internal.ObjectArrays;
 import org.assertj.core.internal.Objects;
 import org.assertj.core.internal.OnFieldsComparator;
+import org.assertj.core.internal.RecursiveFieldByFieldComparator;
 import org.assertj.core.util.IterableUtil;
 import org.assertj.core.util.VisibleForTesting;
 import org.assertj.core.util.introspection.IntrospectionError;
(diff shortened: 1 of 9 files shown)


joel-costigliola:
@PascalSchumacher I'll try to review by the end of the month, please remind me if I fail to do so.

PascalSchumacher:
@joel-costigliola Thanks, no need to hurry. I'm fine as long as you as you review before the next release. I know you are a very busy man.  :)

PascalSchumacher:
rebase onto current 2.x

joel-costigliola:
Thanks, I'll review it tomorrow.

src/main/java/org/assertj/core/api/AbstractObjectArrayAssert.java
```diff
@@ -1105,6 +1106,10 @@ public S usingFieldByFieldElementComparator() {
     return usingElementComparator(new FieldByFieldComparator());
   }
 
+  public S usingRecursiveFieldByFieldElementComparator() {
```
joel-costigliola: Javadoc needed with a code example

src/test/java/org/assertj/core/api/iterable/IterableAssert_usingRecursiveFieldByFieldElementComparator_Test.java
```diff
@@ -0,0 +1,128 @@
...
+
+  @Override
+  protected ConcreteIterableAssert<Object> invoke_api_method() {
+    return assertions.usingRecursiveFieldByFieldElementComparator();
+  }
+
+  @Override
+  protected void verify_internal_effects() {
+    assertThat(iterablesBefore).isNotSameAs(getIterables(assertions));
+    assertThat(getIterables(assertions).getComparisonStrategy() instanceof ComparatorBasedComparisonStrategy).isTrue();
```
joel-costigliola: can use a more specific assertion:
```
assertThat(getIterables(assertions).getComparisonStrategy()).isInstanceOf(ComparatorBasedComparisonStrategy.class);
```

src/test/java/org/assertj/core/api/iterable/IterableAssert_usingRecursiveFieldByFieldElementComparator_Test.java
```diff
@@ -0,0 +1,128 @@
...
+    assertThat(getObjects(assertions).getComparisonStrategy() instanceof IterableElementComparisonStrategy).isTrue();
+  }
+
+  @Test
+  public void successful_isEqualTo_assertion_using_field_by_field_element_comparator() {
```
joel-costigliola: I would add `recursive` in the test name

src/test/java/org/assertj/core/internal/RecursiveFieldByFieldComparator_compareTo_Test.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+package org.assertj.core.internal;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Test;
+
+public class RecursiveFieldByFieldComparator_compareTo_Test {
+
+  private static final RecursiveFieldByFieldComparator RECURSIVE_FIELD_BY_FIELD_COMPERATOR = new RecursiveFieldByFieldComparator();
```
joel-costigliola: typo `COMPERATOR` -> `COMPARATOR`

src/test/java/org/assertj/core/internal/FieldByFieldComparator_toString_Test.java
```diff
@@ -20,7 +20,7 @@
 
   @Test
   public void should_return_description_of_FieldByFieldComparator() {
-	assertThat(new FieldByFieldComparator().toString()).isEqualTo("field by field comparator on all fields");
+    assertThat(new FieldByFieldComparator()).hasToString("field/property by field/property comparator on all fields/properties");
```
joel-costigliola: good one !

PascalSchumacher:
Thanks for the review :smile: 

I have updated the pull request with your suggestions.

joel-costigliola:
Thanks for the quick PR update ! 

I was reviewing the javadoc and realized that there is no support for registering specific field/type comparators (as mentioned in the existing javadoc), for completion and consistency we should add it.  

Adding `usingComparatorElementForType` and `usingComparatorForElementField` methods in `AbstractIterableAssert` and using the registered comparators in `FieldByFieldComparator` should solve this problem I believe . 

What is your opinion about the solution and the methods names ?

PascalSchumacher:
I agree there should be support for this registering field/typ comparators.

I think this can be added in a different issue/pull request, as this also concerns `usingElementComparatorOnFields`. I have opened https://github.com/joel-costigliola/assertj-core/issues/676 to track this and removed the paragraph from the javadoc.

I'm not sure the names make it clear enough that the comparators are not used for the elements itself, what about something like `usingComparatorForElementFieldsWithType` and `usingComparatorForElementFieldsWithName`?

I can start working on a pull request for this, unless you like to tackle this yourself?

joel-costigliola:
Good idea, I will integrate this PR and let you work on #676 if you have time to do so.

joel-costigliola:
Done :)

PR #651, opened 2016-04-19
FailureMessages should use String.format to replace the %n with corre…

timmikk:
The %n is not properly replaced with the system specific line separator if the string is not formatted with String.format method.

### DIFF
--- a/src/main/java/org/assertj/core/util/FailureMessages.java
+++ b/src/main/java/org/assertj/core/util/FailureMessages.java
@@ -18,11 +18,11 @@ package org.assertj.core.util;
 public final class FailureMessages {
 
   public static String actualIsEmpty() {
-    return "%nExpecting actual not to be empty";
+    return String.format("%nExpecting actual not to be empty");
   }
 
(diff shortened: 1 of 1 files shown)


PascalSchumacher:
Hi timmikk,

thanks for the pull request. :)

I did not check every possible path, but at least in some execution parts the String is formatted in ExpectedException#expectMessage.

Cheers,
Pascal

PascalSchumacher:
Of course formatting it directly does not hurt and makes it clearer to the reader, so it's probably a good idea to apply your pull request.

joel-costigliola:
I agree, it does not hurt.
Thanks

timmikk:
Hi,
Thanks for quick response.
I just came by this when solving failing AssertJ-Swing JUnit test:
org.assertj.swing.internal.assertions.images.Images_assertHasSize_Test#should_Fail_If_Actual_Is_Null.
That test uses directly the FailureMessages#actualIsNull without proper string formatting.

joel-costigliola:
Thanks @timmikk.

Issue #492, opened 2015-09-29
soft assertions should output line numbers or stack traces.


christophsturm:
when soft assertions fail it's hard to see which assertion was not met. I think the best way would be to capture the linenumber or stacktrace of each soft assertion and then output them with the failure message.

joel-costigliola:
Not sure what you mean, soft assertions shows a report of all failed assertions.
Can you give a test example ?

joel-costigliola:
closing the issue for lack of feedback.

fduminy:
I agree with @christophsturm since I have the same need.

Let's take the example of [SoftAssertionsTest.should_be_able_to_catch_exceptions_thrown_by_all_proxied_methods\(\)](https://github.com/joel-costigliola/assertj-core/blob/b5e448af3b363526f25633005608a912ba51bdb2/src/test/java/org/assertj/core/api/SoftAssertionsTest.java#L109).

At [line 111](https://github.com/joel-costigliola/assertj-core/blob/b5e448af3b363526f25633005608a912ba51bdb2/src/test/java/org/assertj/core/api/SoftAssertionsTest.java#L111)
```java
softly.assertThat(BigDecimal.ZERO).isEqualTo(BigDecimal.ONE);
```
should trigger an error message containing
```java
at org.assertj.core.api.SoftAssertionsTest:111
```

So, [line 210](https://github.com/joel-costigliola/assertj-core/blob/b5e448af3b363526f25633005608a912ba51bdb2/src/test/java/org/assertj/core/api/SoftAssertionsTest.java#L210)
```java
assertThat(errors.get(0)).isEqualTo("expected:<[1]> but was:<[0]>");
```
would become
```java
assertThat(errors.get(0)).isEqualTo("expected:<[1]> but was:<[0]> (at org.assertj.core.api.SoftAssertionsTest:111)");
```

joel-costigliola:
That would be an improvement indeed, I prefer the line number over the stack trace to keep the report readable.

joel-costigliola:
Thanks @PascalSchumacher  and @fduminy for your work.

PR #654, opened 2016-05-01
soft assertions should output line numbers of failed assertions (clos…

PascalSchumacher:
Fixes #

…es #492)

A problem is writing a unit test. Because assertj removes it internal lines from stacktraces the feature does not work when the test is added to org.assertj.core.api.SoftAssertionsTest.


joel-costigliola:
What if you moved the test in a package that does not contain `org.assert` ?

src/main/java/org/assertj/core/api/SoftAssertions.java
```diff
@@ -105,16 +106,43 @@
...
+    return error.getMessage() + "at (" + stackTraceElement.getClassName() + ":"
+           + stackTraceElement.getLineNumber() + ")";
+  }
+
+  private void removeReflectionAndCglibProxyElementsFromStackTrace(Throwable throwable) {
+    List<StackTraceElement> filtered = newArrayList(throwable.getStackTrace());
+    for (StackTraceElement element : throwable.getStackTrace()) {
+      String className = element.getClassName();
+      if (className.startsWith("sun.reflect") || className.startsWith("java.lang.reflect")
+          || className.startsWith("net.sf.cglib.proxy")) {
```
joel-costigliola: not sure the condition `className.startsWith("net.sf.cglib.proxy")` will ever be met as cglib is repackages inside assertj, use `className.startsWith("org.assertj.core.internal.cglib.proxy")` instead - you can verify that in assertj-examples.

PascalSchumacher:
> What if you moved the test in a package that does not contain org.assert ?

Then it would work. The pull request contains a test which is in a separate package.

joel-costigliola:
true, I did not check the test package, my bad.

PascalSchumacher:
Just noticed that there are side-effects with overriding errorMessage etc. I will have to update the pull request.

src/main/java/org/assertj/core/api/SoftAssertions.java
PascalSchumacher: Thanks! I will add a test in assertj-examples.

PascalSchumacher:
So it works when I run it in eclipse, but when run from maven both assertions seem to be on the same line?

PascalSchumacher:
If I change:
```java
softly.assertThat(1).isLessThan(0)
                    .isLessThan(1); 
```
to:
```java
softly.assertThat(1).isLessThan(0)
softly.assertThat(1).isLessThan(1); 
```
it works from maven too.

Not sure if the feature is worth adding, if it does not work correctly when assertions are chained.

PascalSchumacher:
@joel-costigliola @fduminy What do you think, is the feature still worth it (taking into consideration the limitation mentioned above)?

fduminy:
@PascalSchumacher For me, the feature still worth it, especially because I rarely chain more that one assertion per line. Maybe SoftAssertionError's error message should start with a warning like "line numbers might not be accurate".
I have also fixed the error message to have an hyperlink to source code in Intellij.
I have attached [a patch](https://github.com/joel-costigliola/assertj-core/files/249661/line_numbers_patch.txt) to this comment.

joel-costigliola:
I disagree, either we support it properly or we don't, if it works for non chained assertions users will expect it to work for chained ones.

fduminy:
It's obviously a bug in javac.
I have written following test : 
```java
public class Builder {
    public Builder a() {
        trace();
        return this;
    }

    public Builder b() {
        trace();
        return this;
    }

    private void trace() {
        StackTraceElement[] stackTrace = Thread.currentThread().getStackTrace();
//        StackTraceElement[] stackTrace = new Throwable().getStackTrace();
        System.out.println(stackTrace[0].toString());
        System.out.println(stackTrace[1].toString());
        System.out.println(stackTrace[2].toString());
        System.out.println(stackTrace[3].toString());
        System.out.println(stackTrace[4].toString());
    }

        new Builder().a()
                                .b();
    }
}
```
Line numbers are wrong with javac 1.6.0_45 and javac 1.7.0_80.
They are good with javac 1.8.0_66 and javac build 9-ea+116.

@joel-costigliola Can we move (or copy) that issue to master branch (java 8) ?

joel-costigliola:
Yes, let's move it to master (3.x versions)
Thanks for the investigation

PascalSchumacher:
@fduminy Thanks for looking into this. :+1: 

I will rebase the pull request onto master, extend it with your patch (minus the warning) and make sure that it works with the other soft assertion classes like BDDSoftAssertions.

PascalSchumacher:
I does not seem possible to change the branch the pull request is targeting, so I had to open an new one: https://github.com/joel-costigliola/assertj-core/pull/659

joel-costigliola:
:+1:

PR #657, opened 2016-05-05
Removed </p> as not supported in javadoc.

tommyshem:
#### Check List:
* Fixes # JavaDoc
* Unit tests : NA

Ide used InteliJ to build the javadoc. 
Also tried Netbeans ide but still getting errors on building the javadocs.

Removed &lt;p/&gt; as not supported in javadoc and erroring on build.
Changed < to &lt as error when building javadoc.
Changed > to &gt as error when building javadoc.
changed &lt;br/&gt; to &lt;br&gt; as error when building javadoc.

If it is something I am doing wrong then let me know as I am only a hobby programmer.


joel-costigliola:
Integrated, thanks !

PR #659, opened 2016-05-07
soft assertions should output line numbers of failed assertions (clos…

PascalSchumacher:
…es #492)

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
@@ -12,6 +12,9 @@
  */
 package org.assertj.core.api;
 
+import static java.lang.String.format;
+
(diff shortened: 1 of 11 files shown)


PascalSchumacher:
Extended the pull request to `BDDSoftAssertions`.

To make it work with `JUnit*SoftAssertions` it would be necessary to either change the message of the collected `Throwables` or somehow clone them. The only idea I have to use reflection to change the error message, but have not tried it yet. Any ideas?

PascalSchumacher:
Added this feature to JUnitSoftAssertions and JUnitBDDSoftAssertions, but this required a dirty hack, so I'm not sure this is worth it.

Feel free to cherry-pick just the first commit or direct me to a cleaner approach.

joel-costigliola:
Thanks Pascal.

PR #660, opened 2016-05-07
 Allow to check result of last soft assertion. Fix #653

### LINKED ISSUE #653: Allow to check result of last soft assertion
#### Summary
I would like to be able to check the result of the last soft assertion.

#### Real life use case or example
```java
Person person = ...
SoftAssertion soft = new SoftAssertions();
if (soft.assertThat(person.getAddress()).isNotNull().wasSuccess()) {
    soft.assertThat(person.getAddress().getStreet()).isNotNull();
}
```

Actual workaround is :
```java
soft.assertThat(person.getAddress()).isNotNull();
if (person.getAddress() != null) {
    soft.assertThat(person.getAddress().getStreet()).isNotNull();
}
```
but for more complex assertion, I must also write the equivalent boolean expression for the ```if``` statement.

The implementation is easy and need to set a flag to ```false``` at begin of ```ErrorCollector.intercept``` and, on failure, set the flag to ```true```.

#### Java 8 specific ?
NO

fduminy:
#### Check List:
* Fixes #653 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
@@ -29,4 +29,20 @@ public class AbstractSoftAssertions {
   protected List<Throwable> errorsCollected(){
     return proxies.errorsCollected();
   }
+
+  /**
+   * Returns the result of last assertion.<br/>
+   * Example :<br/>
+   * <pre>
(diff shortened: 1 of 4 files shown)


joel-costigliola:
That was a tricky one, thanks !
I just have refactored and added more comments as the code is not obvious.

PR #669, opened 2016-05-31
use "@since 2.5.0 / 3.5.0" instead of "@since 2.5.0" because otherwis…

PascalSchumacher:
I think using just 2.5.0 can confuse users who are not familiar with the AssertJ versioning scheme. They may think that e.g. 3.0.0 contains methods introduced in 2.5.0, because 3.0.0 is greater that 2.5.0

If you disagree, just close the pull request. :smile:


joel-costigliola:
I guess I agree as I merged the PR ;-)

PR #670, opened 2016-06-02
fixes #658 add support for java 8 Predicates

### LINKED ISSUE #658: Support for java 8 Predicates
#### Summary
It would be nice to have support for java 8 predicates.
Assertions should allow for testing them instead of testing everyone separetly.

#### Real life use case / example
Good:
```
//given
acceptedValues = Lists.newArrayList("a","b")

//when
final Predicate<String> predicate = Predicate.create(acceptedValues);

//then
assertThat(predicate).accepts("a")
                     .accepts("b")
                     .doesNotAccept("c")
                     .accepts(Lists.newArrayList("a","b"))
                     .doesNotAccept(Lists.newArrayList("c","d"));
```

Not good:
```
//given
acceptedValues = Lists.newArrayList("a","b")

//when
final Predicate<String> predicate = Predicate.create(acceptedValues);
boolean result1 = predicate.test("a");
boolean result2 = predicate.test("b");
boolean result3 = predicate.test("c");
boolean result4 = predicate.test("d");

//then
assertThat(result1).isTrue();
assertThat(result2).isTrue();
assertThat(result3).isFalse();
assertThat(result4).isFalse();
```
#### Java 8 specific ?
YES

filiphr:
Fixes #658 

@joel-costigliola I tried to use a wrapper Predicate so I can reuse the matchers in the ```AbstractPredicateLikeAssert```


joel-costigliola:
this is a partial review :  we don't to have description parameter, AssertJ provides `as()` methods for that.

filiphr:
I will remove the description parameter. I added it since I saw that the ```AbstractCompletableFutureAssert``` had a method ```isCompletedWithValueMatching(Predicate, String)```

filiphr:
@joel-costigliola Have a look at the latest commit with the new error factory for the ```accepts``` and ```doesNotAccept``` asserts. 

Do we need a different message factory for the ```acceptsAll``` and ```noneAccepted``` methods? I am reusing ```Iterables#assertAllMatch``` and ```Iterables#assertNoneMatch``` consequently.

joel-costigliola:
@filiphr `noneAccepted` does not make a natural sentence, I would rather have `doesNotAcceptAny` but I'm also thinking we could use `rejects` instead of `doesNotAccept` and `rejectsAll` instead of `noneAccepted`.

joel-costigliola:
Hem ... sorry for all the naming changes, I realize that there is a bunch of methods and files to rename, feel free to let me do it if you had enough.

Anyway I'm still interested in your opinion about renaming the negative assertions.

filiphr:
@joel-costigliola You are right about the naming of the negative assertions. ```rejects``` and ``rejectsAll``` are much better and are clear negation of the positive assertions ```accepts``` and ```acceptsAll```. 

If you can wait until this afternoon, I can do them, I do not mind. If you are in a hurry then feel free to do the changes.

joel-costigliola:
Thanks, take your time, it's getting late for me anyway (I live in New Zealand).

I'm happy with the error messages, I would just rename `ElementsShouldNotMatch` to `NoElementsShouldMatch` and remove the `ADVICE` fields in the error message classes.

filiphr:
I will rename the ```ElementsShouldNotMatch```.

As for the ```ADVICE```, I used it since it was there in the ```PredicateDescription```, so I thought it is something that should be general. I will remove it.

joel-costigliola:
I almost finished integrated the PR to realise that using `assertThat` predicate methods leads to a compilation error, Java 8 complaining they are ambiguous when passing a lambda parameter:
```java
// The method assertThat(Predicate<Object>) is ambiguous ... :(
assertThat(value -> value.contains("foot")).accepts("football");
```
The reason is that `assertThat(Predicate<T> actual)` and `assertThat(IntPredicate actual)` both take  interface parameters, in Java 8 this is problematic because it could decide which method to use for a class implementing both interfaces (Java 7 was more permissive in that regard).

Users can avoid this by casting their lambdas for declaring them:
```java
assertThat((Predicate<String>) value -> value.contains("foot")).accepts("football");
```
or
```java
Predicate<String> footPredicate = value -> value.contains("foot");
assertThat(footPredicate).accepts("football")
```


I think we have multiple options from here:
 1. don't change anything and force users to declare their predicate type
 2. rename `assertThat(IntPredicate actual)` to `assertThatIntPredicate(actual)`
 3. simply remove `assertThat(IntPredicate actual)`

WDYT ?

filiphr:
I think we should go with option 1, leave it as is and force users to declare their predicates.

I also think that this won't be a problem. Since when you do testing you do not really test lambdas per se. You would have some util class or something similar which generates predicates for you and you would want to test this class. At least this is why I would use the ```assertThat(Predicate)``` possibility.

If you want I can squash all my commits this evening and modify the pull request so it has only one commit.

joel-costigliola:
Ok option 1 was also my preferred choice.
Don't do anything I have already integrated the PR locally.

joel-costigliola:
Integrated, thanks for the work @filiphr ! I have simplified the API a bit by chaning `accepts` and `rejects` parameter to varargs (but keeping the nice error message is only one value is provided).

filiphr:
@joel-costigliola no problem, it was fun working on this. I saw the simplification, it looks much better :)

Issue #671, opened 2016-06-02
Methods within normal and BDD Asertions


filiphr:
#### Summary
This is more like a question then a problem. I was having a look at the ```BDDAssertions``` and ```Assertions```, also the soft types. Is the general idea that all of the ```assertThat``` from the ```Assertions``` exist as ```then``` method in ```BDDAssertions```?

If yes then I would propose some way of guaranteeing this, by either always generating the ```BDDAssertions``` from the ```Assertions```, or by creating a test that would test all the methods from the ```Assertions```. 

Additionally about the methods between the soft assertions and the assertions are not consistent.

#### Example

I did a small analysis on the (2.x branch) and there are some difference between the methods:

Assertions|BDDAssertions
---|---
public static <K, V> MapAssert<K, V> assertThat|public static <K, V> AbstractMapAssert<K, V> then
public static <T> T assertThat(final AssertProvider<T> component)| missing
public static <T extends AssertDelegateTarget> T assertThat(T assertion)| missing
public static AbstractPathAssert<?> assertThat(Path actual)| missing

#### Java 8 specific ?

NO

joel-costigliola:
That would be great indeed.

Thanks for the analysis, I'll fix that right away.

filiphr:
I also noticed that the soft assertions use the implemented classes of the assertions, is there some reason for this?

I am asking since I can also make a test which checks if the ```BDDAssertions``` have the same methods as the ```AbstractBDDSoftAssertions```. If you say yes, then we need to decide how to fix it. Should we always return our own implementations of the ```Abstract*Assert``` or the abstract ones. If we return the abstract we would introduce breaking changes. This same thing applies for the difference in the ```MapAssert``` vs the ```AbstractMapAssert```, and the methods in the ```Assertions``` vs the ```AbstractSoftAssertions```

joel-costigliola:
Soft assertions could actually return AbstractXXXAssert classes but need to proxy the concrete class (CGLIB needs a concrete class to instantiate to be able to create a proxy).

filiphr:
If it is ok to return the ```AbstractXXXAssert``` then I will do that. Shortly I would create a pull request with the test I have in mind and also the "fixes" for the test

filiphr:
@joel-costigliola Do you think that we could somehow use something from the assertj-assertions-generator to generate the different ```Assertions``` classes. 

[...]

PR #674, opened 2016-06-04
add tests for validating that there are consistent methods within the assertions

filiphr:
This is are proposal tests for validating that the assertions have the correct methods.
related to #671

@joel-costigliola I would appreciate if you can have a look at it and let me know what you think. There are some asserts which are in the static assertions but not in the soft.

The ```XXXNavigableXXXAssert```(s) and some others are missing. 

I can start on "fixing" the tests so they do not fail, but I would need some guidance.

### DIFF
--- a/src/test/java/org/assertj/core/api/AssertionsTest.java
+++ b/src/test/java/org/assertj/core/api/AssertionsTest.java
@@ -18,6 +18,8 @@ import java.util.List;
 
 import org.junit.Test;
 
+import static org.assertj.core.api.Assertions.assertThat;
+
 /**
  * @author Filip Hrisafov
  */
@@ -29,12 +31,11 @@ public class AssertionsTest extends BaseAssertionsTest {
     String methodA = "assertThat";
     Class<BDDAssertions> classB = BDDAssertions.class;
     String methodB = "then";
-    List<Method> assertThatMethods = findMethodsWithName(classA, methodA);
-    List<Method> thenMethods = findMethodsWithName(classB, methodB);
+    Method[] assertThatMethods = findMethodsWithName(classA, methodA);
+    Method[] thenMethods = findMethodsWithName(classB, methodB);
 
     Comparator<Method> methodComparator = ignoringDeclaringClassAndMethodName();
-    Assertions.assertThat(assertThatMethods).usingElementComparator(methodComparator)
-              .containsExactlyInAnyOrder(thenMethods.toArray(new Method[thenMethods.size()]));
+    assertThat(assertThatMethods).usingElementComparator(methodComparator).containsExactlyInAnyOrder(thenMethods);
   }
 
   @Test
@@ -43,12 +44,12 @@ public class AssertionsTest extends BaseAssertionsTest {
     String methodA = "assertThat";
     Class<AbstractStandardSoftAssertions> classB = AbstractStandardSoftAssertions.class;
     String methodB = "assertThat";
-    List<Method> assertThatMethods = findMethodsWithName(classA, methodA);
-    List<Method> assertThatSoftMethods = findMethodsWithName(classB, methodB);
+    Method[] assertThatMethods = findMethodsWithName(classA, methodA);
+    Method[] assertThatSoftMethods = findMethodsWithName(classB, methodB);
 
     Comparator<Method> methodComparator = ignoringDeclaringClassAndMethodName();
-    Assertions.assertThat(assertThatMethods).usingElementComparator(methodComparator)
-              .containsExactlyInAnyOrder(assertThatSoftMethods.toArray(new Method[assertThatSoftMethods.size()]));
+    assertThat(assertThatMethods).usingElementComparator(methodComparator)
+              .containsExactlyInAnyOrder(assertThatSoftMethods);
 
   }
 }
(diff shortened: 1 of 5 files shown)


joel-costigliola:
Nice work ! I like the approach, it showed an inconsistency between `Assertions` and `BDDAssertions` (the `Map` one that I'm fixing right now).

Getting the test on soft assertions working is gonna be more challenging, the `XXXNavigableXXXAssert` are not yet in soft assertions, I tried to add them but failed due to generic problems, I might give another try later (aah generic hell !).

I suggest to exclude `XXXNavigableXXXAssert` from the analysis and to define another method comparator that would ignore return types until we change soft assertions to return the same type as standard assertions.

When checking `Assertions` vs `AbstractStandardSoftAssertions`, no need to ignore the method names as they should identical.

How about splitting this PR in two, first part for checking `Assertions` and `BDDAssertions` and create another PR for soft assertions (would also include testing `AbstractBDDSoftAssertions`) ?

src/test/java/org/assertj/core/api/AssertionsTest.java
```diff
@@ -0,0 +1,54 @@
...
+  public void should_have_the_same_methods_as_in_bdd_assertions() {
+    Class<Assertions> classA = Assertions.class;
+    String methodA = "assertThat";
+    Class<BDDAssertions> classB = BDDAssertions.class;
+    String methodB = "then";
+    List<Method> assertThatMethods = findMethodsWithName(classA, methodA);
+    List<Method> thenMethods = findMethodsWithName(classB, methodB);
+
+    Comparator<Method> methodComparator = ignoringDeclaringClassAndMethodName();
+    Assertions.assertThat(assertThatMethods).usingElementComparator(methodComparator)
```
joel-costigliola: I would make `findMethodsWithName` to return `Method[]`, that would avoid the array conversion in the tests.

joel-costigliola: I would inverse `assertThatMethods` and `thenMethods` in the assertion, we are testing `BDDAssertions` methods not `Assertions` ones.
And also, let's static import `assertThat`, so much nicer to the eye ;-)

src/test/java/org/assertj/core/api/AssertionsTest.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+public class AssertionsTest extends BaseAssertionsTest {
+
+  @Test
+  public void should_have_the_same_methods_as_in_bdd_assertions() {
+    Class<Assertions> classA = Assertions.class;
+    String methodA = "assertThat";
+    Class<BDDAssertions> classB = BDDAssertions.class;
+    String methodB = "then";
+    List<Method> assertThatMethods = findMethodsWithName(classA, methodA);
```
joel-costigliola: inline local variables `classA`, `classB`, `methodA` and `methodB`

src/test/java/org/assertj/core/api/BaseAssertionsTest.java
```diff
@@ -0,0 +1,60 @@
...
+        matchingMethods.add(method);
+      }
+    }
+    return matchingMethods;
+  }
+
+  protected static Comparator<Method> ignoringDeclaringClassAndMethodName() {
+    return new Comparator<Method>() {
+      @Override
+      public int compare(Method o1, Method o2) {
```
joel-costigliola: rename `o1` to `method1` and `o2` to `method2` for better code readability.

filiphr:
Regarding the inherited methods, don't we also need to have ```Java6AbstractBDDSoftAssertions```. I will modify the ```findMethodWithName``` to search for declared methods up the inheritance tree.

I will add the possibility to ```findMethodWithName``` to ignore methods when the return type is instance of something passed. Will also add the second method comparator ignoring the return type.

I have no problem splitting the PR. I will remove the soft assertions tests from this one and do a force push.

src/test/java/org/assertj/core/api/AssertionsTest.java
filiphr: There is an inverse test in the ```BDDAssertionsTest```.  I am not sure if we need to test in both places though.  

Will do the static import, I am not sure how it slipped :).

joel-costigliola:
:+1: 

Indeed, I have reopened https://github.com/joel-costigliola/assertj-core/issues/672 for `Java6AbstractBDDSoftAssertions`.

filiphr:
So do we make ```findMethodWithName``` go up the inheritance tree or we should test each level separately?

joel-costigliola:
I prefer going up the inheritance tree as it ends up testing the concrete classes users use.

filiphr:
Did that. I think that this covers the ```Assertions``` vs ```BDDAssertions``` method validation.

Once this is merged, I will make a pull request with the ```SoftAssertions``` vs the ```BDDSoftAssertions``` methods validations test.

joel-costigliola:
Merged, I just haven't included `BDDSoftAssertionsTest` and `SoftAssertionsTest` as they should be part of the next PR.

filiphr:
@joel-costigliola did you maybe forgot to push the merge? I cannot see it.

PascalSchumacher:
@filiphr joel just pushed to the 2.x branch: https://github.com/joel-costigliola/assertj-core/commit/c8dff5be837c439aea97d7acf1f01f0afe7812a3 he did not merge the commit to master yet

filiphr:
@PascalSchumacher I didn't see that. Thanks for the information.

filiphr:
@joel-costigliola Maybe we need to keep this open until we merge it into master. I didn't check the tests against master. 

I can do this tonight and give you an analysis.

joel-costigliola:
Thanks but no need, I usually merge 2.x branch in master and fix/update whatever breaks.

PR #675, opened 2016-06-05
use more specific assertions for some tests

PascalSchumacher:
hasToString() instead of "toString().isEqualTo()"
isInstanceOf() instead of "instanceof isTrue()"


joel-costigliola:
cool :+1:

PR #677, opened 2016-06-05
add tests for validating the soft assertions methods consistency 

filiphr:
related to #671

currently ignoring some special assertions like the XXXBasedNavigableXXXAssert(s), AssertDelegateTarget and the special assert with AssertProvider


joel-costigliola:
Integrated, many thanks.

PR #679, opened 2016-06-06
Add Java6JUnitSoftAssertions to enable its usages in Andr…

kxfang:


kxfang:
Seems like we also need Java6BDDAssertions and Java6BDDSoftAssertions, right?

PascalSchumacher:
I guess they are not strictly required, but it would be nice to have then.

joel-costigliola:
Thanks @kxfang !

PR #685, opened 2016-06-12
add missing since javadoc tags for version 2.5.0

PascalSchumacher:


joel-costigliola:
Lucky that you are here spotting these omissions. :+1:

PR #695, opened 2016-06-27
CheckReturnValue for assertThat statements for Findbugs

### LINKED ISSUE #157: Detect incorrect API usage
Yesterday in one of our projects I found an incorrect AssertJ API usage:

```java
assertThat(a1.equals(b1));
assertThat(a2.equals(b2));
assertThat(a3.equals(b3));
```

Probably it can be usefull to try to detect such cases. For example, Mockito tries to detect wrong method call order.

In the example above I think it is possible to detect that `assertThat()` method was called but no methods on the returned object were executed. This can be detected in the consecutive call. State can be held in a static thread local variable. Also it makes sense to look at how Mockito does this.

WDYT?

karlicoss:
#### Check List:
* Fixes #157
* Unit tests : NA

I've found [this](https://github.com/joel-costigliola/assertj-core/pull/436), dug into Findbugs code, and [apparently](https://github.com/findbugsproject/findbugs/blob/264ae7baf890d2b347d91805c90057062b5dcb1e/findbugs/src/java/edu/umd/cs/findbugs/detect/BuildCheckReturnAnnotationDatabase.java#L120), Findbugs handles **any** annotation with name `CheckReturnValue` in its return value usage check. So, if we supply our own annotation, we can avoid being dependent on `jsr305` package.

If you approve it, I'll search for other classes in assertJ where it's reasonable to put this annotation and also will backport to 2.x version as well.

I also checked it against my build of this branch, it totally works! See https://github.com/karlicoss/assertj-return-value-demo

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractBDDSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractBDDSoftAssertions.java
@@ -12,6 +12,8 @@
  */
 package org.assertj.core.api;
 
+import org.assertj.core.util.CheckReturnValue;
+
 import java.nio.file.Path;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
@@ -39,6 +41,7 @@ public abstract class AbstractBDDSoftAssertions extends Java6AbstractBDDSoftAsse
    * @param actual the path
    * @return the created assertion object
    */
+  @CheckReturnValue
   public PathAssert then(Path actual) {
     return proxy(PathAssert.class, Path.class, actual);
   }
@@ -52,6 +55,7 @@ public abstract class AbstractBDDSoftAssertions extends Java6AbstractBDDSoftAsse
    * @return the created assertion object.
    */
   @SuppressWarnings("unchecked")
+  @CheckReturnValue
   public <T> OptionalAssert<T> then(Optional<T> actual) {
     return proxy(OptionalAssert.class, Optional.class, actual);
   }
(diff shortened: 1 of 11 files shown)


joel-costigliola:
That is awesome ! 

Do you think it is possible to customise the report ? I was looking at your project [sample report](https://github.com/karlicoss/assertj-return-value-demo/blob/master/findbugsXml.xml) but it does not explain what was wrong with the code, reporting a problem with an unrelevant explanation might confuse people.

karlicoss:
@joel-costigliola yeah, I thought about that too (although, it's a waaay smaller problem than useless assertions :) ). It might be possible to customize it, I'll take a look tomorrow.

joel-costigliola:
:+1:

karlicoss:
@joel-costigliola Looks like there is no way to customize error message :( Anyway, I feel like people who use findbugs are smart enough to figure that out or google :)

joel-costigliola:
Thanks for the investigation.

Can you also annotate : `BDDAssertions` and `AbstractStandardSoftAssertions`

karlicoss:
@joel-costigliola updated the PR.
I actually annotated every method returning assertion rather than the whole classes because not all of these methods return assertions and should be chain-called. For instance, `Assertions.contentOf` or `Assertions.withinPercentage`. 

In case someone is curious, I used a script to add the annotations:

    find src/main/java/org/assertj/core/api -type f -iname "*Assertions*.java" -exec sed -i 's/^[ tab]\+public .*Assert.* [a-zA-Z0-9]\+(.*).*$/  @CheckReturnValue\n&/' {} \;

joel-costigliola:
@karlicoss thanks !

If you can do the backport PR for this week end as I'm planning to relase 2.5.0/3.5.0 this sunday.
If you can't, that's ok, I'll do it

karlicoss:
Yeah, I'll open the PR today once I get to work!

Dima Gerasimov, ITMO University, study group A4200

joel-costigliola:
sweet :)

karlicoss:
@joel-costigliola done https://github.com/joel-costigliola/assertj-core/pull/697

PR #697, opened 2016-07-01
CheckReturnValue for assertThat statements for Findbugs

### LINKED ISSUE #157: Detect incorrect API usage

karlicoss:
#### Check List:
* Fixes #157
* Unit tests : NA

Backports https://github.com/joel-costigliola/assertj-core/pull/695

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractBDDSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractBDDSoftAssertions.java
@@ -12,6 +12,8 @@
  */
 package org.assertj.core.api;
 
+import org.assertj.core.util.CheckReturnValue;
+
 import java.nio.file.Path;
 
 public abstract class AbstractBDDSoftAssertions extends Java6AbstractBDDSoftAssertions {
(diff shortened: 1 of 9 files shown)


karlicoss:
I also guess that the check is worth mentioning this on wiki or something, @joel-costigliola what do you think? I could document it and open a PR on wiki as well.

joel-costigliola:
Good idea,  I will add a FAQ for that.

joel-costigliola:
@karlicoss I'll do it as part of the 2.5.0/3.5.0 documentation I'm currently writing. You can't really create a PR on something that does not exists ;-)

PR #704, opened 2016-07-06
Use @SafeVarargs to avoid warning on caller' side

glhez:
Hello

When user wants to test a predicate or check the content of a collection, such as:

```java
    Predicate<Map.Entry<String, String>> predicate = ...
    assertThat(predicate)
      .rejects(Pair.of("A", "B"))
      .accepts(Pair.of("C", "D"))
    ;
```
Or, with a map:

```java
   Map<String, String> map = ...
   assertThat(map.entrySet())
     .containsExactly(Pair.of("A", "B"))
   ;
```

The compiler will emits the following warning (under Eclipse, javac generates more or less the same kind of message): 

> Type safety: A generic array of Map.Entry<String,String> is created for a varargs parameter

To avoid this, the [`@SafeVarargs`](https://docs.oracle.com/javase/7/docs/api/java/lang/SafeVarargs.html) may be used. 
This implies that assertj does not do unsafe thing with the var args, which should be the case (especially in `IterableAssert `where `Iterables `methods mostly takes Object[] as source).

The alternative version is to use the iterables based methods (`rejectAll`, ...) but that produce more in-digest code:

```java
    assertThatPredicate(pred)
      .acceptsAll(asList(Pair.of("test", 1)))
      .rejectsAll(asList(Pair.of("test", 0)))
    ;
```

**Warning** : this commit will break the unit test code (see test in error here : [test_unit_error.txt](https://github.com/joel-costigliola/assertj-core/files/351084/test_unit_error.txt) ) because some use proxies and because `final` needs to be added (see javadoc of @SafeVarargs), the proxy fail to generate the method. 

Since I don't know what's the intent of that proxy, I committed it without it.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractBDDSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractBDDSoftAssertions.java
@@ -188,8 +188,8 @@ public abstract class AbstractBDDSoftAssertions extends Java6AbstractBDDSoftAsse
    */
   @SuppressWarnings("unchecked")
   @CheckReturnValue
-  public <T> PredicateAssert<T> then(Predicate<T> actual) {
-    return proxy(PredicateAssert.class, Predicate.class, actual);
+  public <T> SoftAssertionPredicateAssert<T> then(Predicate<T> actual) {
+    return proxy(SoftAssertionPredicateAssert.class, Predicate.class, actual);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


joel-costigliola:
The proxy is used for soft assertions.

glhez:
If you mean _to override the method using var args_, that is impossible: @SafeVarargs is mostly incompatible with overriding methods because of the `final` (because the overriding method could temper with the var args and make the whole not safe).

joel-costigliola:
you don't override anything in `AbstractPredicateAssert` only in `PredicateAssert`, does that make sense ?

glhez:
Okay, that should be good for PredicateAssert.
I'll do the same for List and Iterable assert.

joel-costigliola:
Integrated, thanks @glhez !

PR #705, opened 2016-07-07
Provide @SafeVarargs for ListAssert and IterableAssert

glhez:
Hello,

To continue my PR #704 , here the version for Assertj 2.5.1 which handle `ListAssert` and `IterableAssert`. I'll try to add it in 3.5.1 tomorrow so except a new PR.

The commit is a little tricky: 

- Some test failed because some method in `AbstractIterableAssert` returned a `ListAssert` while it should have been a `SoftAssertionIterableAssert`. 
- Some methods in `AbstractIterableAssert` returned `ListAssert` with a cast to `SELF`. It probably works (at runtime) because Java replace `SELF` by `AbstractIterableAssert` but it is wrong: in some place, `SELF` might mean `IterableAssert` and I don't think that `(IterableAssert) new ListAssert<>()` is valid especially if you try to call a method specific to `IterableAssert`.

So I added a new method `newListAssertInstance` which create an `AbstractListAssert` letting the implementation overriding it with the best choice - the default being `ListAssert`.

It is probably too complicated for "just" some warnings but the other alternatives à la `EnumSet.of` is no better unless there is a code generator that could silently generate those.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
@@ -31,6 +31,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
+import org.assertj.core.api.IterableAssert.LazyIterable;
 import org.assertj.core.api.filter.FilterOperator;
 import org.assertj.core.api.filter.Filters;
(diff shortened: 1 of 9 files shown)


joel-costigliola:
Integrated, thanks @glhez !

PR #724, opened 2016-07-27
#723 Exclude static fields from field based comparisons

### LINKED ISSUE #723: Field comparison not ignoring static fields
#### Summary
Similar to #590, but this time for statics. 
A static element won't ever be a field, so it shouldn't be comparing them.

Field has an isStatic() method to simplify this

### LINKED ISSUE #793: String.matches() fails on strings with parenthesis or brackets
#### Summary

See title; The below example fails using `org.assertj:assertj-core:3.5.2`; Similar, if using square or curly brackets.

#### Example

```java
    @Test
    public void example() {
        assertThat("this(string)here").matches("this(string)here");
    }
```

#### Java 8 specific ?

(I'm using Java 8)

mikezx6r:
#### Check List:
* Fixes #723 
* Unit tests : YES 

Static fields are never fields of an object as they're class fields. Therefore should be ignored for field by field comparisons.

This is only done on 2.x branch. Will need to be migrated to 3.x branch or let me know how I should tackle it, and I'll create additional pull requests, or update this one.


joel-costigliola:
Everything that goes in 2.x ends in master (3.x version) branch, don't worry about it.

joel-costigliola:
Integrated thanks !

PR #728, opened 2016-07-29
add AbstractCharSequenceAssert#isBlank, #isJavaBlank, #isNotBlank and…

### LINKED ISSUE #726: Add CharSequenceAssert#isNotBlank?
e.g. check that a string is not `Null`, not empty and contains not only white-space.

My main concern is that Javas definition of white-space is broken (some characters like non-breaking space are not considered white-space characters). I would follow Guavas definition of white-space.

PascalSchumacher:
… #isNotJavaBlank (closes #726)

I'm not sure what `isBlank` should check. I implemented the method to similar to commons-langs StringUtils#isBlank https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isBlank(java.lang.CharSequence) Then I noticed that `isEmpty` fails for `Null` so should `isBlank` fail for `Null` and empty Strings?

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
@@ -122,6 +122,100 @@ public abstract class AbstractCharSequenceAssert<S extends AbstractCharSequenceA
   }
 
   /**
+   * Verifies that the actual {@code CharSequence} is blank, i.e. is {@code null}, has a length of 0 or
+   * contains only whitespace characters.
+   * <p>
+   * These assertions will succeed:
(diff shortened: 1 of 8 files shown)


joel-costigliola:
I'm currently integrating your work, I have clarified a bit what whitespace meant in the javadoc (by referencing Guava's javadoc).
I think isBlank should fail for null or empty string, after all they don't contains anything (I disagree with apache commons on this one).

joel-costigliola:
done. thanks

PascalSchumacher:
>I think isBlank should fail for null or empty strings, after all they don't contains anything (I disagree with apache commons on this one).

o.k., this implementation is consistent with `isEmpty`

Thanks a lot for finishing this! :)

PR #733, opened 2016-08-16
remove path dependency from assertions

darko1002001:
* Fixes Android couldn't use this version due to issue with not finding `java.nio.file.Path`


PascalSchumacher:
On Android you can use the https://github.com/joel-costigliola/assertj-core/blob/master/src/main/java/org/assertj/core/api/Java6Assertions.java entry point, which has no `Path` dependency.

PR #736, opened 2016-08-18
add integer-based method overloads to ByteArrayAssert (fixes #729)

### LINKED ISSUE #729: Add integer-based overloads to ByteArrayAssert
#### Summary
ByteArrayAssert has a number of methods that take byte-parameters. However, since there is no byte-constant in java, you will have to cast each and every parameter value.

It would be useful to have integer-overloads of these methods.

#### Example

The first example does not work. The two next ones are workarounds.

```java
// Verify the magic number of a pdf-file by checking four first bytes
assertThat(fileContent).startsWith(0x25, 0x50, 0x44, 0x46);

// Java does not have byte constants, and ByteArrayAssert does not support integers, but...
assertThat(fileContent).startsWith((byte) 0x25, (byte) 0x50, (byte) 0x44, (byte) 0x46);
assertThat(fileContent).startsWith(new byte[] { 0x25, 0x50, 0x44, 0x46 });
```

#### Java 8 specific ?

PascalSchumacher:

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractByteArrayAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractByteArrayAssert.java
@@ -110,6 +110,34 @@ public abstract class AbstractByteArrayAssert<S extends AbstractByteArrayAssert<
   }
 
   /**
+   * Verifies that the actual array contains the given values, in any order.
+   * <p>
+   * Example:
+   * <pre><code class='java'> // assertion will pass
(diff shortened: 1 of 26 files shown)


joel-costigliola:
Awesome, I'll review it in the next few weeks but not right now - too busy :(

PascalSchumacher:
No problem. Take your time. I know you will get to it eventually. :)

joel-costigliola:
Thanks Pascal !

PR #754, opened 2016-09-08
add and improve since java doc tags of methods introduced in 2.6.0 / …

PascalSchumacher:
…3.6.0


joel-costigliola:
cool, thanks.

PR #759, opened 2016-09-19
add internal wrapper for the the comparator by types;

### LINKED ISSUE #758: usingComparatorForType is not used with registered type's subclasses
#### Summary

`usingComparatorForType` should work with registered type subclasses too.

http://stackoverflow.com/questions/39490785/using-assertj-usingcomparatorfortype-method-with-inheritance

filiphr:
Add new method to the ClassUtils, so all super classes can be retrieved

@joel-costigliola This is just my first iteration of the fix. I am not entirely sure about the change of modifier and the new method in the `ClassUtils`. Maybe we need to do this a bit differently.

* Fixes #758


joel-costigliola:
Integrated, thanks @filiphr !

PR #760, opened 2016-09-19
Fix platform specific test

filiphr:
While doing https://github.com/joel-costigliola/assertj-core/pull/759, this test was failing on my side (I am using Windows :)). This fixes the test (it makes sure that it uses `%s` instead of `\n`


PascalSchumacher:
:+1:

PR #764, opened 2016-09-23
isEqualToComparingFieldByFieldRecursively: increased priority of cust…

PascalSchumacher:
…om comparators so custom comparators can be used for collection types and maps

some work into the direction of https://github.com/joel-costigliola/assertj-core/issues/737


joel-costigliola:
Thanks Pascal !

Issue #748, opened 2016-08-27
Add extractingValue() for `Optional` assertions


mkobit:
#### Summary

It looks like this is already available in [`assertj-guava`](https://github.com/joel-costigliola/assertj-guava/blob/master/src/main/java/org/assertj/guava/api/OptionalAssert.java#L137).

Would like to see the same functionality available for Java 8 `Optional`s.
#### Example

``` java
final Path myPath = Paths.get("myPath");
assertThat(myPath).isPresent()
                               .extractingValue()
                               .endsWith(Paths.get("myPath"));
```
#### Java 8 specific ?
- YES : create Pull Request from the `master` branch

mkobit:
This seems like it would be easy to implement, but I'd like to get your thoughts on the right direction for the API for `Optional` with current differences between the:

- [Guava `OptionalAssert`](https://github.com/joel-costigliola/assertj-guava/blob/master/src/main/java/org/assertj/guava/api/OptionalAssert.java)
- [Core `OptionalAssert`](https://github.com/joel-costigliola/assertj-core/blob/master/src/main/java/org/assertj/core/api/OptionalAssert.java)

joel-costigliola:
@mkobit due to Java generic limitations it is not possible to chain specific type assertions, only Object assertions, in your example it means you can't chain the `endsWith` call . 
Is the feature still valuable without specific type assertions ?

mkobit:
That's a bummer.

I don't think the feature would be that valuable then because the fluent chaining is the valuable part.

Can you think of a different way to achieve similar functionality?

joel-costigliola:
there are two ways : 

1 / 
pass the assert type as a generic parameter : `extractingValue(T)`  where `T` is `StringAssert` to get `String` assertions, not elegant IMO as users need to know the assert types.

2 / 
add a bunch of methods like `extractingXXX()` where XXX is the value type, ex `extractingString()`, problem is it bloats the API and is not generic, it does not work with your custom assert type.

joel-costigliola:
Closing this issue until we have an acceptable way of implementing it.

PR #770, opened 2016-10-05
Add map, flatMap and extractingValue to OptionalAssert (#751, #748)

### LINKED ISSUE #751: Add map(…) and flatMap(…) to OptionalAssert
#### Summary

If you're dealing with `Optional`s whose values return `Optional`s, you end u with a lot of nesting. It would be cool if these could be flattened out like this:

#### Example

```java
Optiona<Foo> foo = Optional.of(foo);

assertThat(foo).hasValueSatisfying(it -> {
  assertThat(it.methodReturningAnOptional()).hasValueSatisfying(inner -> {
    assertThat(inner.methodReturningNoOptional()).isEqualTo(reference);
  }
}
```

Note, how you constantly have to nest expressions and also rewrap the values obtained from the method call into new asserts.

It would be cool if one could rather write this instead:

```java
assertThat(foo)
  .flatMap(it -> it.methodReturningAnOptional())
  .map(it -> it.methodReturningNoOptional())
  .isEqualTo(reference);
```

I currently have this custom assert class in my code to achieve this:

```java
public class OptionalAssert<T> extends AbstractAssert<OptionalAssert<T>, Optional<T>> {

  public OptionalAssert(Optional<T> actual) {
    super(actual, OptionalAssert.class);
  }

  public static <T> OptionalAssert<T> optionalAssertThat(Optional<T> optional) {
    return new OptionalAssert<T>(optional);
  }

  public <S> OptionalAssert<S> flatMap(Function<T, Optional<S>> function) {

    Assertions.assertThat(actual).isPresent();

    return optionalAssertThat(actual.flatMap(function));
  }

  public <S> AbstractObjectAssert<?, S> map(Function<T, S> function) {

    Assertions.assertThat(actual).isPresent();

    return Assertions.assertThat(actual.map(function).orElseThrow(() -> new IllegalStateException()));
  }
}
```

I guess that can be improved to report good error messages in case of a flat map step producing an empty `Optional`, but it does the trick for me now.

fiery-phoenix:
#### Check List:
* Fixes #751, #748
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractOptionalAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractOptionalAssert.java
@@ -20,7 +20,9 @@ import org.assertj.core.internal.StandardComparisonStrategy;
 import java.util.Comparator;
 import java.util.Optional;
 import java.util.function.Consumer;
+import java.util.function.Function;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.error.OptionalShouldBeEmpty.shouldBeEmpty;
 import static org.assertj.core.error.OptionalShouldBePresent.shouldBePresent;
 import static org.assertj.core.error.OptionalShouldContain.shouldContain;
@@ -51,7 +53,7 @@ public abstract class AbstractOptionalAssert<S extends AbstractOptionalAssert<S,
    * </p>
    * Assertion will pass :
    * <pre><code class='java'> assertThat(Optional.of("something")).isPresent();</code></pre>
-   * 
+   *
    * Assertion will fail :
    * <pre><code class='java'> assertThat(Optional.empty()).isPresent();</code></pre>
    *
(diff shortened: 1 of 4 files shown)


fiery-phoenix:
Implementation is a little bit different from what was originally proposed at #751. flatMap and map don't throw AssertionError if actual optional is empty. From my point of view it would be more expected behavior. As map() doesn't extract mapped value, have added extractingValue() method as it was proposed at #748.

joel-costigliola:
Sweet, will review it this we.

joel-costigliola:
Thanks @fiery-phoenix ! 
I have modified a bit your PR, I have left out `extractingValue()` as I'm not sure it is totally useful (waiting for feedback on that) and I have added support for soft assertions (a tricky part of AssertJ code).

fiery-phoenix:
Ok, thank you!

PR #773, opened 2016-10-11
add static assertSoftly method

jillesvangurp:
This makes SoftAssertions a bit nicer to use since you can simply accept the instance in a lamda function instead of having to assign it and then call assertAll() at the end.

#### Check List:
* Fixes: convenience method for making it easier to use SoftAssertions
* Unit tests : YES, SoftAssertionsTest.should_assert_using_assertSoftly

### DIFF
--- a/src/main/java/org/assertj/core/api/SoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/SoftAssertions.java
@@ -15,6 +15,7 @@ package org.assertj.core.api;
 import static org.assertj.core.api.Assertions.extractProperty;
 
 import java.util.List;
+import java.util.function.Consumer;
 
 /**
  * <p>
@@ -66,6 +67,7 @@ import java.util.List;
  *   softly.assertAll();
  * }</code></pre>
  *
+ *
  * <p>
  * Now upon running the test our JUnit exception message is far more detailed:
  * <pre><code class='java'> org.assertj.core.api.SoftAssertionError: The following 4 assertions failed:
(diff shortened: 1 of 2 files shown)


joel-costigliola:
@jillesvangurp are you aware of `JUnitSoftAssertions` ? If so I'm curious why do you prefer `assertSoftly` ?

```java
@Rule
public final JUnitSoftAssertions softly = new JUnitSoftAssertions();

@Test
public void host_dinner_party_where_nobody_dies() {
   Mansion mansion = new Mansion();
   mansion.hostPotentiallyMurderousDinnerParty();
   // use SoftAssertions instead of direct assertThat methods
   softly.assertThat(mansion.guests()).as("Living Guests").isEqualTo(7);
   softly.assertThat(mansion.kitchen()).as("Kitchen").isEqualTo("clean");
   softly.assertThat(mansion.library()).as("Library").isEqualTo("clean");
   softly.assertThat(mansion.revolverAmmo()).as("Revolver Ammo").isEqualTo(6);
   softly.assertThat(mansion.candlestick()).as("Candlestick").isEqualTo("pristine");
   softly.assertThat(mansion.colonel()).as("Colonel").isEqualTo("well kempt");
   softly.assertThat(mansion.professor()).as("Professor").isEqualTo("well kempt");
   // No need to call softly.assertAll(), it is done by JUnitSoftAssertions rule
}
```

jillesvangurp:
I use testng. So, no @Rule annotation. Having to call assertAll() manually is kind of clumsy. A lambda solves that. Also has the advantage that you can nest them if for whatever reason you'd want to achieve some grouping of assertions.

joel-costigliola:
What about `AutoCloseableSoftAssertions` ?

```java
@Test
 public void host_dinner_party_where_nobody_dies() {
   Mansion mansion = new Mansion();
   mansion.hostPotentiallyMurderousDinnerParty();
   try (AutoCloseableSoftAssertions softly = new AutoCloseableSoftAssertions()) {
 	softly.assertThat(mansion.guests()).as("Living Guests").isEqualTo(7);
 	softly.assertThat(mansion.kitchen()).as("Kitchen").isEqualTo("clean");
 	softly.assertThat(mansion.library()).as("Library").isEqualTo("clean");
 	softly.assertThat(mansion.revolverAmmo()).as("Revolver Ammo").isEqualTo(6);
 	softly.assertThat(mansion.candlestick()).as("Candlestick").isEqualTo("pristine");
 	softly.assertThat(mansion.colonel()).as("Colonel").isEqualTo("well kempt");
 	softly.assertThat(mansion.professor()).as("Professor").isEqualTo("well kempt");
 	// no need to call assertAll, it is done when softly is closed.
   }
 }
```

joel-costigliola:
@jillesvangurp WDYT ?

jillesvangurp:
Sorry, completely forgot to respond to this. I indeed overlooked AutoCloseableSoftAssertions, which I did not know existed. 

It achieves the same goal and I actually considered something similar as well. Only downside, I see is the warnings about unclosed resources you get when you can't use the try with resources for whatever reason (e.g. passing it around to other methods). This is why I went for a lambda. I think both approaches are valid.

Anyway, given that there is AutoCloseableSoftAssertions already, I don't feel strongly about leaving this pull request out. I leave this to you decide. You may want to add a note on this in the documentation.

I do think there is some value in leveraging lambda's a bit more like I attempted here. Whether this is the best way is of course up for debate. I'm not that familiar with all corners of this framework.

joel-costigliola:
Fair enough.

PR #775, opened 2016-10-12
Add Path assertions to WithAssertions

cykl:
#### Check List:
* Unit tests : YES

Hi Joel,

I just noticed that Path assertions are absent from the `WithAssertions` interface. Is it an inadvertent omission or a deliberate omission ?

### DIFF
--- a/src/main/java/org/assertj/core/api/WithAssertions.java
+++ b/src/main/java/org/assertj/core/api/WithAssertions.java
@@ -15,6 +15,8 @@ package org.assertj.core.api;
 import java.io.File;
 import java.io.InputStream;
 import java.math.BigDecimal;
+import java.net.URI;
+import java.net.URL;
 import java.nio.charset.Charset;
 import java.nio.file.Path;
 import java.text.DateFormat;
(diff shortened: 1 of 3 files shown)


joel-costigliola:
I would go for inadvertent ;-) 
We have a test to check that `Assertions` and `BDDAssertions` are in sync, we should use the same strategy for `WithAssertions`.

cykl:
I can give it a look if you like. Maybe another PR, though.

cykl:
A quick test showed a lot of discrepancies. I will try to fix them and update this PR or fill a new one.

joel-costigliola:
you can update this PR, don't bother creating a new one.

filiphr:
@cykl there are 12 `assertThat` methods which are missing. 

@joel-costigliola, should we use the same strategy as with the `SoftAssertions` and the `BddAssertions`. Shouldn't we check for the other methods? Shouldn't `WithAssertions` be exactly like the `Assertions`?

cykl:
@filiphr I added a test to `AssertionsTest`  and fixed all the issues. The only remaining discrepancies are Iterable and Iterator. Methods were present in both classes but with different return types. Surprisingly, compiler got confused when I replaced  `AbstractIterableAssert` by `IterableAssert`, but I haven' t yet investigated the issue. Is this a known issue ?

Today was a busy day, I will push the patch over the weekend.

filiphr:
@cykl That is strange. There shouldn't be an issue in that. The same thing is working on the `Assertions`. I also just tried it locally on my side and there are no problems.

joel-costigliola:
Integrated thanks Clément.

PR #777, opened 2016-10-16
Add missing tests for various classes

filiphr:
I've been playing a bit with [pitest](http://pitest.org/) and I saw that there are some tests which were missing.

@joel-costigliola, if you are interested I can setup a maven profile that will allow anyone to run pitest analysis over AssertJ.


joel-costigliola:
Awesome ! I would be interested in pitest setup (in a gist), that's something I had in my todo list for a while.

filiphr:
That's cool. I created #778 so we can have a dedicated discussion there. Setting pitest for AssertJ is pretty easy.

joel-costigliola:
Thanks, will have a look at #778 later this week.

PR #781, opened 2016-10-21
Expanded Optional assertion api with hasValueSatisfying Condition (#774)

### LINKED ISSUE #774: Expand Optional assertion api to include Conditions
#### Summary

The existing `Optional` assertion api includes `hasValue` and `hasValueSatisfying`. It would be nice to expand it to include `Condition`s.

This second method does allow a lamba or method reference (e.g. `Condition::matches`) to be written that asserts `Condition`s against the optional value, but I think not needing to do this is more readable.

#### Example

Assumes a method `jsonProperty` that returns a `Condition` checking a particular value in some JSON.

```java
Optional<JSON> json = maybeCreateSomeJson();
assertThat(json).hasValue(jsonProperty("$.property", "value"));
```

#### Java 8 specific ?

fiery-phoenix:
#### Check List:
* Fixes #774
* Unit tests : YES
* Javadoc with a code example (API only) : YES / NO

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractOptionalAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractOptionalAssert.java
@@ -60,8 +60,7 @@ public abstract class AbstractOptionalAssert<S extends AbstractOptionalAssert<S,
    * @return this assertion object.
    */
   public S isPresent() {
-    isNotNull();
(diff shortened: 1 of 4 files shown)


joel-costigliola:
In master, thanks !

PR #782, opened 2016-10-21
remove usage of java.beans.Introspection to improve android compatibi…

### LINKED ISSUE #247: Use openbeans to support assertions that use properties reflection
If you use assertions like isEqualToComparingFieldByField in Android they always return true and fail silently because the ClassNotFound errors are silently catched.

This behaivour is very error prone and I lost some time looking on what was really happening. I think that best solution will be to change the usage on java.beans with openbeans or at least don't catch silently the exceptions or do an explicit check for java.beans availability

PascalSchumacher:
…lity (fixes #247)

Work in progress.

The effort seems justified to me, what do you think Joel?

Three tests fail at the moment:

Two of them use inner classes whose properties can not be read at the moment, because I removed the `JavaBeanDescriptor` and its `readMethod.setAccessible(true);` call.

PropertySupport_publicGetterExistsFor_Test#should_return_false_if_public_getter_does_not_exist fails because the package private getter method is successfully invoked by the new implementation.


joel-costigliola:
I think it is worth a try, if things work without `JavaBeanDescriptor` fine by me.

PascalSchumacher:
O.k. I have updated the pull request. All tests pass.

joel-costigliola:
Sweet, integrated ! Thanks Pascal.

PascalSchumacher:
Just noticed that with this change read-only properties considered properties. I guess we need to add a check that a setter exists. What do you think?

PR #783, opened 2016-10-27
Fix for excluding fields for deep difference

fiery-phoenix:
Fix for the case when field is both synthetic and static.


joel-costigliola:
Good catch, integrated in 2.x.

PR #788, opened 2016-11-01
Support classes of java.util.concurrent.atomic

### LINKED ISSUE #746: Support classes of java.util.concurrent.atomic
When using the `Atomic*` classes, we always have to write `assertThat(myAtomic.get()).isEqualTo(42)`. I think it would be better to write `assertThat(myAtomic).isEqualTo(42)` instead (without the need of calling `get()`).

epeee:
#### Check List:
* Fixes #746 
* Unit tests : **YES** / NO / NA
* Javadoc with a code example (API only) : **YES** / NO / NA


src/main/java/org/assertj/core/api/AtomicMarkableReferenceAssert.java
```diff
@@ -26,6 +34,12 @@ protected VALUE getActualValue() {
     return actual.getReference();
   }
 
+  /**
+   * Verifies that the actual {@link AtomicMarkableReference} contains the given marked value.
```
joel-costigliola: This javadoc is not correct, a code example would be welcome.

joel-costigliola:
@epeee don't forget to add code example in assertions method javadoc, see [AbstractPathAssert#hasNoParent](https://github.com/joel-costigliola/assertj-core/blob/0ba3fa5d0d7f3efefdc9e0e71b1ea9fccf2cfab1/src/main/java/org/assertj/core/api/AbstractPathAssert.java#L938) for a good example.

epeee:
@joel-costigliola thx for pointing that out. I just added some examples.

epeee:
@joel-costigliola it looks like we do cover all classes of java.util.concurrent.atomic for java < 8.
I will start working on java8-only ones once this pr is merged.
sounds good?

joel-costigliola:
sounds good :)

joel-costigliola:
@epeee while I was reviewing your PR, I started changing a few things: 
- replace `isMarked(boolean)` `AtomicMarkableReference assertion `by `isMarked()` / `isNotMarked()`  
- atomic arrays are now treated like arrays, you can use all array assertions applied to the atomic internal array, ex:
```java
AtomicIntegerArray actual = new AtomicIntegerArray(new int[] { 1, 2, 3, 4 });
assertThat(actual).startsWith(1, 2)
                  .contains(3, atIndex(2))
                  .endsWith(4);
```
I'm thinking to do the same for `AtomicInteger`, `AtomicLong` and `AtomicReference`, that is reusing all int, long existing assertions. 
WDYT ?

epeee:
Yes, let's go for the changes mentioned.
Especially the 2nd one makes it way easier to use.

joel-costigliola:
Heads up on this PR: I'm moving this to 2.7.0 as it requires more work to get it right and I want to focus on the 2.6.0 release, I have pushed the current state of the work in the `atomic` branch.

I have removed `AtomicLongAssert` (and co) and make `assertThat(AtomicLong)` return `LongAssert`, the good is that you can use any `long` assertions, the bad is that `assertThat(myAtomicLong).isEqualTo(myAtomicLong)` fails as it compares a `long` to an `AtomicLong`. The same applies for atomic array assertions.

What needs to be done exactly ? Well, reintroduce the removed `AtomicLongAssert` (and co) and delegate to their underlying type assertions (long, int, Object, ...) and have two overloaded `isEqualTo` assertions: one taking an atomic and another with the atomic underlying type.

Sorry about leaving that out 2.6.0, it was more complex than I expected.

I'll do that by the end of the year but if you feel like it, you can jump back to this work (I totally get it if you don't).

Anyway thanks for the initial PR even if the actual state is quite far from it.

epeee:
ok.

I might be able to spend some time on this one during the next few days and provide a new PR (maybe just AtomicInteger for now).

PR #797, opened 2016-11-18
Added shortcut for assertThatExceptionOfType to exception:

glhez:
Hello,

I've pushed some short cut method for "common" exception such as NullPointerException: I though it was rather sad to always write something like:

```
assertThatExceptionOfType(NullPointerException.class)
  .isThrownBy(() -> file.save((Path) null)).withMessage("path");
```

While it could be:

```
assertThatNullPointerException().isThrownBy(() -> file.save((Path) null)).withMessage("path");
```

I did not add API example as there are purely alias to existing methods (eg: `assertThatExceptionOfType(NullPointerException.class)`).


joel-costigliola:
Thanks @glhez !

glhez:
You're welcome !

PR #798, opened 2016-11-18
Fixed two compile error with Eclipse Compiler.

glhez:
Eclipse (tested in Oxygen) compiler fails to compile some test because of type inference. 
This fix those compile error.


joel-costigliola:
Not having them in my eclipse (neon under linux) but merging this PR should not hurt.

glhez:
That's because I use Eclipse Oxygen (in fact, I used Neon with snapshot update site... So it updated the EJC).

PR #801, opened 2016-11-19
reintroduce AtomicIntegerAssert

### LINKED ISSUE #746: Support classes of java.util.concurrent.atomic

epeee:
#### Check List:
* Fixes  #746
* Unit tests : YES / NO / NA
* Javadoc with a code example (API only) : YES / NO / NA

### DIFF
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -793,8 +793,8 @@ public class Assertions {
    * @return the created assertion object.
    */
   @CheckReturnValue
-  public static AbstractIntegerAssert<?> assertThat(AtomicInteger actual) {
-    return new IntegerAssert(actual);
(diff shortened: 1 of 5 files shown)


epeee:
@joel-costigliola I reintroduced `AtomicIntegerAssert`, added some methods which are delegating to the underlying type assertions. Furthermore, I added a `hasValue` which is taking the atomic underlying type (we can use the implementation from the initial pull request for this one if we'd like to). The `isEqualTo` is provided via AbstractAssert, which I think should work.

joel-costigliola:
This has been integrated (with a bunch of changes), thanks @epeee !

PR #805, opened 2016-11-20
add some missing since javadoc tags to some methods added in recent p…

PascalSchumacher:
…ull requests


joel-costigliola:
I'm always forgetting them, thanks

PR #813, opened 2016-11-28
Improve error message for "optional.containsInstanceOf()"

### LINKED ISSUE #812: Improve error message for "optional.containsInstanceOf(clazz)"
#### Summary

The error message for OptionalAssert.containsInstanceOf() is less descriptive than the similar message for AbstractAssert.isInstanceOf. Pull request is in the works.

#### Example

Before change:

```java
Expecting Optional to contain a value of type Number.
at ...
```

After change:

```java
Expecting:
 <Optional>
to contain a value that is an instance of:
 <class java.lang.Number>
but did contain instance of:
 <class java.lang.String>
at ...
```

Existing isInstanceOf-message:

```java
Expecting:
 <"StringValue">
to be an instance of:
 <java.lang.Number>
but was instance of:
 <java.lang.String>
```

#### Java 8 specific ?

elygre:
#### Check List:
* Fixes #812
* Unit tests : YES (existing tests)

### DIFF
--- a/src/main/java/org/assertj/core/error/OptionalShouldContainInstanceOf.java
+++ b/src/main/java/org/assertj/core/error/OptionalShouldContainInstanceOf.java
@@ -1,16 +1,18 @@
 /**
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
- *
+ * <p>
  * http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
  * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
Hi elygre,

you are probably already aware of this, but you would have to change https://github.com/joel-costigliola/assertj-core/blob/fe84371353fa3f74ea574e04a9480c986f77e6fb/src/test/java/org/assertj/core/error/OptionalShouldContainInstanceOf_create_Test.java and add a test for the new behavior.

elygre:
I was not aware of that, no :-). My maven build succeeded, but I wonder if that is perhaps because it did not see the changes, 'cause it sure as heck fails right now! I'll take this tomorrow.

joel-costigliola:
Thanks @elygre !

PR #822, opened 2016-12-08
increase ExpectedException usage

PascalSchumacher:
There are many more tests which could be updated to use `ExpectedException`, but maybe we should rather wait till assertj is java 8 only and use `assertThatThrownBy`?


joel-costigliola:
:+1:

PR #831, opened 2016-12-10
allMatch prints all the elements that do not match

### LINKED ISSUE #816: allMatch should report all non matching elements
#### Summary

`allMatch` assertion only reports the first non matching elements, although correct it would be better to report all non matching elements.

#### Example

The following assertions should fail and report all characters that are not hobbits.
```java
assertThat(fellowshipOfTheRing).allMatch(hero -> hero.getRace() == HOBBIT, "hobbits");
```

mauricioaniche:
The `allMatch` assert now prints all the elements that do not match in the message. Currently, it only prints the first one that doesn't match.

#### Check List:
* Fixes #816 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/internal/Iterables.java
+++ b/src/main/java/org/assertj/core/internal/Iterables.java
@@ -68,6 +68,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 import org.assertj.core.api.AssertionInfo;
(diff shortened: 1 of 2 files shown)


joel-costigliola:
Thanks a lot @mauricioaniche, I have just improved the error message to use the plural form when multiple were not matching the predicate, ex:
```
java.lang.AssertionError: 
Expecting all elements of:
  <["123", "1234", "12345"]>
to match given predicate but these elements did not:
  <["1234", "12345"]>
```

PR #832, opened 2016-12-10
Introducing isEqualToNormalizingNewlines

### LINKED ISSUE #818: Add isEqualToNormalizingNewlines assertion
#### Summary

Sometimes it is useful to compare two strings and ignore os-specific newline differences to make test passing both on Windows and Linux. It would be enough to perform `.replaceAll("\\r\\n", "\n")` on both assertion arguments and then delegating to `isEqualTo`.

#### Example

```java
// pass
assertThat("a\r\nb").isEqualToNormalizingNewlines("a\nb")
assertThat("\n").isEqualToNormalizingNewlines("\r\n")

// fail
assertThat("\n").isEqualToNormalizingNewlines("")
assertThat("\r").isEqualToNormalizingNewlines("\n")
assertThat("a").isEqualToNormalizingNewlines("a\n")
```

#### Java 8 specific ?

mauricioaniche:
Introducing `isEqualToNormalizingNewlines`. It compares whether two strings are equals to each other after normalizing newlines.

#### Check List:
* Fixes #818
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
@@ -1010,4 +1010,26 @@ public abstract class AbstractCharSequenceAssert<S extends AbstractCharSequenceA
     strings.assertContainsPattern(info, actual, pattern);
     return myself;
   }
+  
+  /**
+   * Verifies that the actual {@code CharSequence} is equals to another
+   * {@code CharSequence} after normalizing new line characters
+   * (e.g. '\r\n' == '\n').
+   * <p>
+   * This assertion will succeed:
+   * <pre><code class='java'> String bookName = &quot;Lord of the Rings\r\n&quot;;
+   * assertThat(bookName).isEqualToNormalizingNewlines(&quot;Lord of the Rings\n&quot;);</code></pre>
+   * 
+   * Whereas this assertion will fail:
+   * <pre><code class='java'> String singleLine = &quot;\n&quot;;
+   * assertThat(singleLine).isEqualToNormalizingNewlines(&quot;&quot;);</code></pre>
+   *
+   * @param expectedLineCount the expected line count of the actual {@code CharSequence}.
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual line count is not equal to the expected one.
+   */
+  public S isEqualToNormalizingNewlines(CharSequence expected) {
+    strings.assertIsEqualToNormalizingNewlines(info, actual, expected);
+    return myself;
+  }
 }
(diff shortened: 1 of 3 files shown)


joel-costigliola:
Thanks, I will review that by the end of the year.

src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
```diff
@@ -1010,4 +1010,26 @@ public S containsPattern(Pattern pattern) {
     strings.assertContainsPattern(info, actual, pattern);
     return myself;
   }
+  
+  /**
+   * Verifies that the actual {@code CharSequence} is equals to another
+   * {@code CharSequence} after normalizing new line characters
+   * (e.g. '\r\n' == '\n').
```
mkordas: @mauricioaniche shouldn't this be rather `i.e.` than `e.g.`?

src/main/java/org/assertj/core/internal/Strings.java
```diff
@@ -412,6 +412,21 @@ private boolean areEqualIgnoringCase(CharSequence actual, CharSequence expected)
...
+   * Verifies that two {@code CharSequence}s are not equal, normalizing newlines.
+   *
+   * @param info contains information about the assertion.
+   * @param actual the actual {@code CharSequence} (newlines will be normalized).
+   * @param expected the expected {@code CharSequence} (newlines will be normalized)..
+   * @throws AssertionError if the given {@code CharSequence}s are equal after normalizing newlines.
+   */
+  public void assertIsEqualToNormalizingNewlines(AssertionInfo info, CharSequence actual, CharSequence expected) {
+    String actualNormalized = actual.toString().replaceAll("\\r\\n", "\n");
+    String expectedNormalized = expected.toString().replaceAll("\\r\\n", "\n");
```
mkordas: @mauricioaniche we need to use here compiled patterns  to improve performance

mauricioaniche:
Your suggestions (that I very much agree) are implemented!

PascalSchumacher:
Using `String#replace(CharSequence target, CharSequence replacement)` should suffice. Less code and even more efficient.

PascalSchumacher:
I would be nice if you could add `@since 2.7.0 / 3.7.0` to the javadoc. Thanks!

mauricioaniche:
Just implemented your two suggestions.

About only the `\r`, I actually believe that if you are dealing with a system that only uses `\r` nowadays, you write the tests in a way that the expected also contains `\r`. Does that make sense?

joel-costigliola:
Thanks @mauricioaniche ! I finally got time to integrate your PR.

PR #833, opened 2016-12-11
Convert tests to use ExpectedException and fix bug in ExpectedException#expectMessage

PascalSchumacher:
This fixes some tests with would have passed when the expected exception was not thrown.


PascalSchumacher:
The two commits https://github.com/joel-costigliola/assertj-core/pull/833/commits/482a44c1ae48f47e516c8319dc61a6012c4874df and https://github.com/joel-costigliola/assertj-core/pull/833/commits/fe139a070f6fc73ddc30fa10055b8dac01caa277 should be reviewed separately.

Normally I would have made the bugfix https://github.com/joel-costigliola/assertj-core/pull/833/commits/fe139a070f6fc73ddc30fa10055b8dac01caa277 a separate pull request.

src/test/java/org/assertj/core/api/Assertions_assertThat_asList.java
```diff
@@ -47,7 +48,7 @@ public void should_pass_list_asserts_on_list_strings_with_asList() {
   public void should_fail_list_asserts_on_non_list_objects_even_with_asList() {
     Object nonList = new Object();
 
-    thrown.expectAssertionError("an instance of:%n <java.util.List>%nbut was instance of:%n <java.lang.Object>");
+    thrown.expectAssertionErrorWithMessageContaining(format("an instance of:%n <java.util.List>%nbut was instance of:%n <java.lang.Object>"));
```
PascalSchumacher: https://github.com/joel-costigliola/assertj-core/blob/2.x/src/test/java/org/assertj/core/test/ExpectedException.java#L108 currently does not format the `String` parameters. Should I add string formatting to the method?

joel-costigliola: not sure as you would not be able to use `String.format` parameters substitution as `expectAssertionErrorWithMessageContaining` already defines varargs, it could simply use `String.format` without params substitution to address the `%n` problem on different OS (if that what `format` usage was solving).
Long story short, I let you decide what is best ;-)

PascalSchumacher:
Sorry for adding even more commits to this pull request. :(

I will stop doing this now.

joel-costigliola:
Don't worry, I won;t review all the changes, I will just take a sample and trust you, these are only tests changes so no big risks.

PascalSchumacher:
Thanks. :) I have rebased the pull request.

joel-costigliola:
Many thanks for this one !

PascalSchumacher:
Thanks for merging! :)

PR #835, opened 2016-12-11
Convert tests to use ExpectedException (3.x only part).

PascalSchumacher:


PascalSchumacher:
removed unnecessary `String.format` calls

joel-costigliola:
:+1:

PR #838, opened 2016-12-12
Added missing @SafeVarargs overrides

joeyb:
#### Check List:
* Unit tests : NA (Methods were already covered)

`IterableAssert` and `ListAssert` are missing a few overrides of varargs methods. Without those overrides, consumers get warnings about unchecked operations.


joel-costigliola:
Good catch, thanks.

joel-costigliola:
Thanks @joeyb !

PR #840, opened 2016-12-12
delete eclipse project files

PascalSchumacher:
I know somebody explicitly asked to have these files checked in, but I still think this a bad practice due to:

- they have to be excluded from every commit
- I have to run `reset --hard` every time when I want to switch branches

So please delete them again. Creating the files is an non-issue for everybody who know how to use maven or the maven eclipse plugin. Sorry for the rant. :(


joel-costigliola:
I never switch branch, I actually have 2 assertj-core projects in eclipse, one for 2.x and one for 3.x.
is that something that would work for you ?

PascalSchumacher:
That would solve the branch switching problem, but I would still have to excluded the (locally modified) files from every commit.

PascalSchumacher:
Generated files like these should not be checked in. Take a look at other popular open source java projects: guava, spring, mockito, commons-lang, jenkins etc. None of these repos contain eclipse/editor specific files.

Recent distributions of eclipse already include the eclipse maven plugin by default, so everybody should be aware of it.

joel-costigliola:
Fair enough, I want to keep my main contributor happy ;-)

PascalSchumacher:
Thanks!

PR #846, opened 2016-12-14
Add ExpectedException#expectWithCause methods and updates test to use…

PascalSchumacher:
… these methods.


PascalSchumacher:
commit depends on changes in #833 so I moved there

PR #851, opened 2016-12-18
remove unused methods and constructors in diff package

PascalSchumacher:
This change breaks binary compatibility if somebody uses these classes and there methods/constructors directly, so feel free to reject.


joel-costigliola:
This is not part of our API so we are not tied by any backward compatibility contract.

PR #858, opened 2016-12-23
convert some tests to use ExpectedException (2.x)

PascalSchumacher:


src/test/java/org/assertj/core/api/Assertions_useRepresentation_Test.java
```diff
@@ -32,21 +38,8 @@ public void afterTest() {
   @Test
   public void should_use_given_representation_in_assertion_error_messages() {
     Assertions.useRepresentation(customRepresentation);
-    try {
-      assertThat("foo").startsWith("bar");
-    } catch (AssertionError e) {
-      assertThat(e).hasMessageContaining("$foo$")
-                   .hasMessageContaining("$bar$");
-      Assertions.useDefaultRepresentation();
```
PascalSchumacher: It is safe to remove this part of the test, because the test below already test exactly this.

PR #878, opened 2017-01-11
extended stack trace filtering (#873)

### LINKED ISSUE #873: Suggested Improvements for AbstractAssert.failWithMessage(...)
### Summary
Currently, the error messages generated in custom assertions via `failWithMessage()` are less clear compared to similar messages from built in assertions.

	MyTuple actual = new MyTuple(new int[] { 1, 2, 3 });
	assertThat(actual).hasValue1("");

This test fails with the following stack trace:
```
java.lang.AssertionError: Expected the first value of <FailWithMessageImprovements$MyTuple@6d21714c> to be <> but was <[I@108c4c35>
	at org.assertj.core.api.AbstractAssert.failWithMessage(AbstractAssert.java:111)
	at FailWithMessageImprovements$MyTupleAssert.hasValue1(FailWithMessageImprovements.java:43)
	at FailWithMessageImprovements.customAssert(FailWithMessageImprovements.java:12)
```

While a built in assertion:

	MyTuple actual = new MyTuple(new int[] { 1, 2, 3 });
	assertThat(actual.getValue1()).isSameAs(""); // 

fails with
```
java.lang.AssertionError: 
Expecting:
 <"">
and actual:
 <[1, 2, 3]>
to refer to the same object
	at FailWithMessageImprovements.builtInAssert(FailWithMessageImprovements.java:24)
```

In the message from my (trivial) custom assert
- it's unclear what the values of `actual` and `expected` are
- the stack trace contains elements from AssertJ and my assert

### Suggestions

1. ~~Add an alternative to `failWithMessage()`, that uses the same formatting as AssertJ (i.e. the same behavior as before #320, but as opt-in).
The custom assert could have pre-formatted the values via `Representation.toStringOf(Object)`, but I think creating asserts should be as convenient as possible.~~
Never mind, I found `throwAssertionError(ErrorMessageFactory)`, which is exactly what I need.
2. Also filter the error stack trace in `failWithMessage()`, exactly like `Failures.failure()`.
3. Extend this filtering by also removing elements of the current assert class.

I could contribute this, if you agree with my suggestions.

### Example

	public static class MyTuple {
		// ...

		public Object getValue1() {
			return value1;
		}
	}

	public static class MyTupleAssert extends AbstractObjectAssert<MyTupleAssert, MyTuple> {
		// ...

		public MyTupleAssert hasValue1(Object expected) {
			isNotNull();

			if (!Objects.equals(actual.getValue1(), expected)) {
				failWithMessage("Expected the first value of <%s> to be <%s> but was <%s>", actual, expected, actual.getValue1());
			}

			return this;
		}
	}

#### Java 8 specific ?

Gubbel:
- remove assertj related elements from stack trace of error thrown by
`failWithMessage(...)`
- also remove all elements of custom assert subclasses

#### Check List:
* Fixes #873
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -12,6 +12,7 @@
  */
 package org.assertj.core.api;
 
+import static org.assertj.core.util.Lists.newArrayList;
 import static org.assertj.core.util.Strings.formatIfArgs;
 
 import java.util.Comparator;
@@ -42,6 +43,8 @@ import org.assertj.core.util.VisibleForTesting;
  * @author Nicolas François
  */
 public abstract class AbstractAssert<S extends AbstractAssert<S, A>, A> implements Assert<S, A> {
+
+  private static final String ORG_ASSERTJ = "org.assert";
 
   @VisibleForTesting
   Objects objects = Objects.instance();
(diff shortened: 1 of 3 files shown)


joel-costigliola:
Many thanks @Gubbel !

PR #882, opened 2017-01-12
Modify the way equals() is tested by checking for equals in either di…

### LINKED ISSUE #881: Timestamp and Date equals are not symmetric.
#### Summary
When making tests with hibernate and using the assertIsEqualToComparingFieldByFieldRecursively to compare the entity returned from Hibernate with the expected entity, there are problems with unexpected failures with Dates. 
While the input field is a java.util.Date the resulting value is java.sql.Timestamp and the equals() methods are not symmetric, so the objects ae not considered equal.
See http://stackoverflow.com/questions/8929242/compare-date-object-with-a-timestamp-in-java

#### Example

The following will be written as a date but be returned as a Timestamp.
```java
    @Temporal(TemporalType.TIMESTAMP)
    @Column(name = "created_on")
    private Date createdOn;

```

#### Java 8 specific ?
It is not java8 specific and it is a small fix, I will make a pull request for this.

mjeffrey:
…rection a.equals(b) OR b.equals(a). Solves an issue with Hibernate returning java.sql.Timestamps when the original field was a java.util.Date  (#881)

#### Check List:
* Fixes #881 
* Unit tests : YES 

### DIFF
--- a/src/main/java/org/assertj/core/internal/DeepDifference.java
+++ b/src/main/java/org/assertj/core/internal/DeepDifference.java
@@ -280,7 +280,7 @@ public class DeepDifference {
       }
 
       if (hasCustomEquals(key1.getClass())) {
-        if (!key1.equals(key2)) {
+        if (!(key1.equals(key2) || key2.equals(key1))) {
           differences.add(new Difference(currentPath, key1, key2));
           continue;
         }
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
Thanks for the pull request, but I non-symmetric equals are an anti-pattern and I think assert-j should support/hide this.

joel-costigliola:
Won't be merged for the reason explained in https://github.com/joel-costigliola/assertj-core/issues/881#issuecomment-323571951

PR #887, opened 2017-01-15
Mark JUnit dependency of assertj-core as optional

adessaigne:
The JUnit dependency of assertj-core is mark as provided but not as optional. The unfortunate consequence is that integration tests using AssertJ requires to have a dependency towards JUnit even if you're not using it at all and using another test framework (TestNG for instance). Since it's not marked as optional, any OSGi framework will refuse to start because it misses the JUnit packages.

Marking this dependency as optional will allow running integration tests using OSGi and TestNG.

#### Check List:
* Unit tests : NA

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,6 +40,7 @@
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <scope>provided</scope>
+      <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>


joel-costigliola:
True, sorry about that.

adessaigne:
No problem. Thank you for merging it !

PascalSchumacher:
@joel-costigliola I guess this change should also be applied to 2.x?

joel-costigliola:
yes indeed

PR #892, opened 2017-01-21
typos

epeee:
some typos I stumbled over.


joel-costigliola:
Thank you very much, this is hard to catch when one has written it.

PR #894, opened 2017-01-24
Adds support for other stream shapes : IntStream, LongStream, DoubleStream

bric3:
AssertJ currently lacks support for the primitive variants like `IntStream`. While Stream methods are defined in the specific interfaces, `Stream` and other stream shapes extends `BaseStream`. This interfaces defines a _little_ set of common methods that are useful enough for assertion as the stream is transformed into a `ListAssert` internally.

From outside it basically changes the signature of `assertThat(Stream<T>)` to `assertThat(BaseStream<T, S extends BaseStream<T, S>>)`.

This PR only adds tests for the external API since ListAssert is already tested.

One thing that still bothers me right now, is that I had to suppress warning for unchecked cast, I'm not sure how to resolve that now.

I'm open for comment and modifications.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractBDDSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractBDDSoftAssertions.java
@@ -12,8 +12,6 @@
  */
 package org.assertj.core.api;
 
-import org.assertj.core.util.CheckReturnValue;
-
 import java.nio.file.Path;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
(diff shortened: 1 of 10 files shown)


joel-costigliola:
AssertJ should support primitive streams for sure.
I'll see if I can get rid of the user's facing warning (the internal ones are not as important).

bric3:
Actually the proposed implementation does not trigger compiler or IDE warning. E.g.

```
assertThat(IntStream.of(823952, 1947230585));
assertThat(Stream.of("Luke", "Leia"));
```

dont raise any warnings.

> One thing that still bothers me right now, is that I had to suppress warning for unchecked cast, I'm not sure how to resolve that now.

In this sentence I was mentioning internal warnings.

joel-costigliola:
Integrated thanks Brice !

bric3:
Cool Thanks

PR #898, opened 2017-01-27
get rid of unused AtomicShouldContain

epeee:
Due to some refactorings we are not using `AtomicShouldContain` any longer.


joel-costigliola:
Good catch !

PR #899, opened 2017-01-27
Long literal ending with 'l' instead of 'L' (since l might look prett…

epeee:
…y similar to 1 in certain fonts)

just a minor one but some tools are generating warnings for this one.


joel-costigliola:
fair enough :)

PascalSchumacher:
:+1:

PR #901, opened 2017-01-27
Propose a new assertThat(ThrowingCallable) signature with hasNotThrownException() assertion

bric3:
The signature can be used to verify that call **don't** raise an exception, instead of relying on other technics. This allows to have a finer control over statement(s) that should or should not raise an exception.

The idea of this API is to be able to write assertions like that : 

```java
assertThat(() -> foo.bar()).hasNotThrownException();
assertThat(() -> jedi.saber()).hasNotThrownException();
```

Ideas on how to improve this PR are welcome ; API signature, name alternative, documentation, ...

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
@@ -17,6 +17,8 @@ import org.assertj.core.internal.Failures;
 import org.assertj.core.internal.Throwables;
 import org.assertj.core.util.VisibleForTesting;
 
+import static java.lang.String.format;
+
 /**
  * Base class for all implementations of assertions for {@link Throwable}s.
  * 
@@ -356,7 +358,21 @@ public abstract class AbstractThrowableAssert<SELF extends AbstractThrowableAsse
     return myself;
   }
 
-  public void hasNotThrownException() {
-    if (actual != null) throw Failures.instance().failure("Expecting code not to raise a throwable.");
+  /**
+   * Verifies that the {@link org.assertj.core.api.ThrowableAssert.ThrowingCallable} didn't raise a throwable.
+   *
+   * Example :
+   * <pre><code class='java'>assertThat(() -> foo.bar()).didNotThrowAnyException();
+   * assertThat(() -> jedi.saber()).didNotThrowAnyException();</code></pre>
+   *
+   * @throws AssertionError if the actual statement raised a {@code Throwable}.
+   */
+  public void didNotThrowAnyException() {
+    if (actual != null) {
+      throw Failures.instance()
+                    .failure(format("Expecting code not to raise a throwable but caught '%s' with message : %s",
+                                    actual.getClass().getName(),
+                                    actual.getMessage()));
+    }
   }
 }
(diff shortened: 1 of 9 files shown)


joel-costigliola:
I have a slight preference for `didNotThrowAnyException` but your wording works ok too, I'm fine keeping it if you feel strongly about it, ex:
```java
assertThat(() -> jedi.saber()).didNotThrowAnyException();
```
Improvements : 
- add javadoc to `hasNotThrownException` or `didNotThrowAnyException`, we try to have some useful javadoc for every user facing methods (greatly inspired by ... Mockito, just saying  :wink:).
- `hasNotThrownException` should also display the thrown exception in the error message, that will help understand what went wrong.

Build failing (`AssertionsTest`), whenever an `assertThat` method is added to `Assertions`, it checks that: 
- the same method exists in `AbstractStandardSoftAssertions` and `WithAssertion`
- the corresponding `then` method exists in `BDDAssertions`.

joel-costigliola:
I have rebased this PR and fixed the tests and improved the javadoc but I'm facing a usability problem, the following code from your original javadoc won't compile:
```java
assertThat(() -> { throw new Exception("boom!"); }).didNotThrowAnyException();
```
the compiler reports the following error (a bit reformatted):
```
reference to assertThat is ambiguous both method 
  <T>assertThat(T) in org.assertj.core.api.Assertions 
and method 
  <ELEMENT>assertThat(java.lang.Iterable<? extends ELEMENT>) in org.assertj.core.api.Assertions
match
incompatible types: cannot infer type-variable(s) T
(argument mismatch; T is not a functional interface)
```
To fix that, one needs to extract the lambda to a local variable that will fix the type inference ambiguity, this is possible but not great from a usability point of view (I already imagine users complaining about that ...)

I'm not sure there is a way to disambiguate the initial code, anyway I'd like to have your thoughts about this.

bric3:
Didn't see your comment, yesterday I was pushing in a hurry yesterday while on the go ;)
I'm still  exploring the idea though, the compilation issue of the javadoc seems to be a real issue though. I'd look into it.

bric3:
> not great from a usability point of view (I already imagine users complaining about that ...)

Indeed I agree. The ambiguity didn't exist before because the API was named differently. That could be the option here.

joel-costigliola:
I'll see if I can find a good name, I have the feeling that if we don't use `assertThat` it's not gonna be easy to discover.

PS : I haven't pushed the rebased commits on any branch

joel-costigliola:
First, I'm gonna use the present tense instead of the past so replacing `didNotThrowAnyException` by `doesNotThrowAnyException` since AssertJ uses the present tense.

Second, at the moment I thinking of renaming `assertThat(ThrowingCallable tc)` to
- `assertThatCode(ThrowingCallable tc)`
- `assertThatThrowingCallable(ThrowingCallable tc)`

```java
assertThatCode(() -> { throw new Exception("boom!"); }).doesNotThrowAnyException();
assertThatThrowingCallable(() -> { throw new Exception("boom!"); }).doesNotThrowAnyException();
```
I kinda prefer `assertThatCode` I must say, any ideas welcome.

bric3:
I kinda like `asserThatCode`, other alternatives : `assertThatBlock` ?

Do you think we can add an `@Incubating` annotation, to indicate possible renaming ?

joel-costigliola:
Let's go for `assertThatCode`, good idea to use an incubating annotation.

bric3:
OK, let's go for `assertThatCode` :)

joel-costigliola:
done.

PR #902, opened 2017-01-28
increase version to 3.6.3-SNAPSHOT

epeee:
... since 3.6.2 is already released.


joel-costigliola:
I will update that to 3.7.0-SNAPSHOT since I don't think we will have another bugfix release.
Thanks anyway.

PR #904, opened 2017-01-29
add ExpectedException#expectAssertionError(ErrorMessageFactory errorM…

PascalSchumacher:
…essageFactory) to reduce boilerplate code in tests

Just an idea (I wish I would have picked an example with less usages ;)).

Of course this is not as isolated as using mocks and arguably still tests implementation not behavior.

### DIFF
--- a/src/main/java/org/assertj/core/internal/DoubleArrays.java
+++ b/src/main/java/org/assertj/core/internal/DoubleArrays.java
@@ -41,11 +41,9 @@ public class DoubleArrays {
 
   private Arrays arrays = Arrays.instance();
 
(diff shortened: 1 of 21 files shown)


joel-costigliola:
I'm fine with the approach, the only downsides are that I sometimes find difficult to identify what went wrong in `thrown.xxx` error message and that I don't like having the `THEN` clause before the `WHEN` one in the test. Anyway the PR is has more positive than negative, I can live with that.

PascalSchumacher:
I agree the error message can be hard to read. I will check if I can find a way to improve this.

PR #906, opened 2017-01-30
Add a more type-safe way to assert object's property with `returns` API

Mura-Mi:
We can verify object's multiple properties using `extracting` API to generate `ObjectArrayAssert` instance, but it does not provide type-driven code assist because the type of test subject is changed to `Object[]`.
Though custom assertion generator is provided in AssertJ project, I think more easy and flexible way to assert method's return value (including getter method) of a test subject helps writing test code.

#### Check List:
* Fixes N/A
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
@@ -12,6 +12,7 @@
  */
 package org.assertj.core.api;
 
+import static java.util.Objects.requireNonNull;
 import static org.assertj.core.extractor.Extractors.byName;
 
 import java.util.function.Function;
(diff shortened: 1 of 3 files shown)


joel-costigliola:
Interesting ... my first thoughts is the proposed API readability is a bit clumsy because you have the method/lambda before the result, thining out loud I would written something like

```java
assertThat(yoda).returns("Yoda", from(Jedi::getName))
                .returns(150, from(Jedi::getWeight));
```
`from` is just syntactic sugar to wrap the lambda but it does not do anything, this means you can also write the assertions without `from`.

Mura-Mi:
Awesome! The better argument order is the point i've looking for but good idea didn't come up with. The idea wrapping with `from` syntax sugar looks nice to me because it's not the mandatory API so that it's easy to understand the way to use. I'll improve the API. Thanks! 😉

joel-costigliola:
As a side note `from` method should be added in `Assertions` to make it easier to discover it.

Mura-Mi:
Thanks! I've amended the original commit.

joel-costigliola:
Thanks @Mura-Mi ! PR Integrated.

PR #908, opened 2017-02-01
Add containsOnlyInstancesOf() to ObjectEnumerableAssert

pmwmedia:
Add a method to verify that all elements of an iterable or array are instances of given classes or interfaces.

Example:
```
Iterable objects = Arrays.asList(new String(), new StringBuilder());
assertThat(objects).containsOnlyInstancesOf(CharSequence.class);
```

#### Check List:
* Fixes N/A
* Unit tests : YES


pmwmedia:
See #909 for branch 2.x

PR #910, opened 2017-02-01
Add containsOnlyInstancesOf() to ObjectEnumerableAssert

pmwmedia:
Add a method to verify that all elements of an iterable or array are instances of given classes or interfaces.

Example:
```
Iterable objects = Arrays.asList(new String(), new StringBuilder());
assertThat(objects).containsOnlyInstancesOf(CharSequence.class);
```

#### Check List:
* Fixes N/A
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/error/ShouldContainOnlyInstances.java
+++ b/src/main/java/org/assertj/core/error/ShouldContainOnlyInstances.java
@@ -1,0 +1,47 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2017 the original author or authors.
+ */
+package org.assertj.core.error;
+
+/**
+ * Creates an error message indicating that an assertion that verifies a group of elements contains elements that
+ * are not an instance of one of the given types. A group of elements can be an iterable or an array of objects.
+ * 
+ * @author Martin Winandy
+ */
+public class ShouldContainOnlyInstances extends BasicErrorMessageFactory {
+
+  /**
+   * Creates a new </code>{@link shouldContainOnlyInstances}</code>.
+   * 
+   * @param object the object value in the failed assertion.
+   * @param types the expected classes and interfaces.
+   * @param dismatches elements that are not an instance of one of the given types.
+   * @return the created {@code ErrorMessageFactory}.
+   */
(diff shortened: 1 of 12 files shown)


PascalSchumacher:
Thanks for the pull request! :+1: 

I think it would be nice if the failure message would also display the class of non-matching elements. What do you think?

pmwmedia:
I agree with you and have implemented that the class of non-matching elements will be displayed if the class name is not part of the `toString()` method.

src/main/java/org/assertj/core/error/ShouldContainOnlyInstances.java
```diff
@@ -0,0 +1,68 @@
...
+    return new ShouldContainOnlyInstances(actual, types, dismatches);
+  }
+
+  private ShouldContainOnlyInstances(Object actual, Class<?>[] types, Iterable<?> dismatches) {
+    super("%n" +
+        "Expecting:%n" +
+        "  <%s>%n" +
+        "to contain only instances of:%n" +
+        "  <%s>%n" +
+        "elements are not instances:%n" +
```
joel-costigliola: I would rephrase `elements are not instances:` to `but these elements are not:`

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -846,6 +847,27 @@ public void assertContainsExactlyInAnyOrder(AssertionInfo info, Iterable<?> actu
                            shouldContainExactlyInAnyOrder(actual, values, notFound, notExpected, comparisonStrategy));
   }
 
+  public void assertContainsOnlyInstancesOf(AssertionInfo info, Iterable<?> actual, Class<?>[] types) {
+    checkIsNotNull(types);
+    assertNotNull(info, actual);
+
+    List<Object> dismatches = newArrayList();
```
joel-costigliola: I would rename `dismatches` to `mismatches` it feels clearer to me

joel-costigliola:
@pmwmedia thanks for the PR ! 
There are a few important missing tests: the ones that test the actual assertion logic in `Iterables` and `Arrays`, what should be added are tests like [Iterables_assertContains_Test](https://github.com/joel-costigliola/assertj-core/blob/d3dab0d22cae3028dd4c73a89a5f8bca1fd21a6d/src/test/java/org/assertj/core/internal/iterables/Iterables_assertContains_Test.java) but for `assertContainsOnlyInstancesOf `.

Would you have the time to add such tests ?

joel-costigliola:
and sorry for the late feedback, too busy these days ...

joel-costigliola:
For API consistency I'm renaming this assertion to `hasOnlyElementsOfTypeIn` as we already provide `hasOnlyElementsOfType`.

pmwmedia:
What is about merging both to `hasOnlyElementsOfType(Class<?>... types)` or rename my new assertion to `hasOnlyElementsOfTypes`?

joel-costigliola:
merging with `hasOnlyElementsOfType` would not work as `Type` is not plural.
I might go for `hasOnlyElementsOfTypes` indeed

joel-costigliola:
Renamed to `hasOnlyElementsOfTypes`.

Thanks @pmwmedia !

PR #911, opened 2017-02-01
2.x meaningful generic names

epeee:
What about using meaningful generic names?
Currently, we are using meaningful generics just in some parts of the codebase.
I will do another iteration if we decide to go for meaningful generic names.


joel-costigliola:
We want meaningful generic names, yes ! 
I did not do it because I was lazy and I was thinking of the conflicts for the other PR I would have to integrate - but non of them were good reasons not to do this refactoring, thanks.

joel-costigliola:
Thanks a lot @epeee !

PR #912, opened 2017-02-04
Classes_assertIs*Final_Test: call assertIs*Final

epeee:
call isFinal/isNotFinal in isFinal/isNotFinalTest


PascalSchumacher:
Thanks! Nice catch! 👍

PR #916, opened 2017-02-08
2.x class method visibility

### LINKED ISSUE #888: test class/method field visibility
#### Summary

because I tend to write libraries, I've noticed that sometimes I take the IDE's suggestion of "may be package private" on something that should be a public class/method whatever. Below I've included the first API that comes to mind, but I'm sure there are others.

#### Example

```
    @Test
    public void test() {
        assertThat( My.class ).isPublic().hasPublicMethods( "getFoo", "getBar" );
        assertThat( Constants.class ).isPublic().hasPublicFields( "FOO", "BAR" );

    }
```

alternatively (would lead to less method explosion)

```
    @Test
    public void test() {
        assertThat( My.class ).isPublic().hasMethods( "getFoo", "getBar" ).withModifer(Modifier::isPublic);

    }
```

also worth noting  that hasDeclaredFields/hasFields don't seem to pick up static fields (not surprising)

epeee:
#### Check List:
* Fixes #888 (at least partially)
* Unit tests : YES

a very first draft for class and method visibility.

e.g.

```
  @Test
  public void test1() {
    Assertions.assertThat(ArrayList.class).hasDeclaredPublicMethods("size", "grow", "add" , "fastRemove");
  }
```

would end up in:

```
java.lang.AssertionError: 
Expecting
  <java.util.ArrayList>
to have public methods:
  <["size", "grow", "add", "fastRemove"]>
but the following are not public:
  <{"fastRemove"="private", "grow"="private"}>
```

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractClassAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractClassAssert.java
@@ -331,7 +331,7 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
   }
 
   /**
-   * Verifies that the actual {@code Class} has the given {@code methodNames}.
+   * Verifies that the actual {@code Class} has the given methods.
    *
    * <pre><code class='java'> class MyClass {
    *     public void methodOne() {}
@@ -357,7 +357,7 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
   }
 
   /**
-   * Verifies that the actual {@code Class} has the given declared {@code methodNames}.
+   * Verifies that the actual {@code Class} has the given declared methods.
    *
    * <pre><code class='java'> class MySuperClass {
    *     public void superMethod() {}
@@ -387,7 +387,7 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
   }
 
   /**
-   * Verifies that the actual {@code Class} has the given public {@code methodNames}.
+   * Verifies that the actual {@code Class} has the given public methods.
    *
    * <pre><code class='java'> class MyClass {
    *     public void methodOne() {}
@@ -415,7 +415,7 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
   }
 
   /**
-   * Verifies that the actual {@code Class} has the given declared public {@code methodNames}.
+   * Verifies that the actual {@code Class} has the given declared public methods.
    *
    * <pre><code class='java'> class MySuperClass {
    *     public void superMethod() {}
--- a/src/main/java/org/assertj/core/api/AbstractClassAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractClassAssert.java
@@ -179,6 +179,51 @@ public abstract class AbstractClassAssert<S extends AbstractClassAssert<S>> exte
   }
 
   /**
+   * Verifies that the actual {@code Class} is public (has {@code public} modifier).
+   *
+   * <pre><code class='java'> protected class MyClass { }
+   *
+   * // These assertions succeed:
+   * assertThat(String.class).isPublic();
+   * assertThat(Math.class).isPublic();
+   *
+   * // This assertion fail:
+   * assertThat(MyClass.class).isPublic();</code></pre>
+   *
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} is not public.
+   *
+   * @since 2.7.0 / 3.7.0
+   */
+  public S isPublic() {
+    classes.assertIsPublic(info, actual);
+    return myself;
+  }
+
+  /**
+   * Verifies that the actual {@code Class} is protected (has {@code protected} modifier).
+   *
+   * <pre><code class='java'> public class MyClass { }
+   *
+   * // This assertion succeed:
+   * assertThat(MyClass.class).isProtected();
+   *
+   * // These assertions fail:
+   * assertThat(String.class).isProtected();
+   * assertThat(Math.class).isProtected();</code></pre>
+   *
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} is not protected.
+   *
+   * @since 2.7.0 / 3.7.0
+   */
+  public S isProtected() {
+    classes.assertIsProtected(info, actual);
+    return myself;
+  }
+
+
+  /**
    * Verifies that the actual {@code Class} has the given {@code Annotation}s.
    * 
    * <pre><code class='java'> &#64;Target(ElementType.TYPE)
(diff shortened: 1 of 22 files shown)


joel-costigliola:
@epeee, just a quick comment (I haven't reviewed the PR yet) can you put `to have declared public methods` instead of `to have public methods` for `hasDeclaredPublicMethods` assertion error message ?

src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -179,6 +179,51 @@ public SELF isNotFinal() {
...
   /**
+   * Verifies that the actual {@code Class} is public (has {@code public} modifier).
+   *
+   * <pre><code class='java'> protected class MyClass { }
+   *
+   * // These assertions succeed:
+   * assertThat(String.class).isPublic();
+   * assertThat(Math.class).isPublic();
+   *
+   * // This assertion fail:
```
joel-costigliola: typo: `fails`

src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -179,6 +179,51 @@ public SELF isNotFinal() {
...
+    classes.assertIsPublic(info, actual);
+    return myself;
+  }
+
+  /**
+   * Verifies that the actual {@code Class} is protected (has {@code protected} modifier).
+   *
+   * <pre><code class='java'> public class MyClass { }
+   *
+   * // This assertion succeed:
```
joel-costigliola: `succeeds`

src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -284,4 +329,111 @@ public SELF hasDeclaredFields(String... fields) {
     classes.assertHasDeclaredFields(info, actual, fields);
     return myself;
   }
+
+  /**
+   * Verifies that the actual {@code Class} has the {@code methods}.
```
joel-costigliola: reword: `has the given methods.` I use `{@code` for something that denotes a java element like {@code true} or {@code equals}

src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -284,4 +329,111 @@ public SELF hasDeclaredFields(String... fields) {
...
+   * assertThat(MyClass.class).hasMethods("methodTwo");
+   * assertThat(MyClass.class).hasMethods("methodThree");</code></pre>
+   *
+   * @param methods the methods which must be in the class.
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} doesn't contains all of the methods.
+   *
+   * @since 2.7.0 / 3.7.0
+   */
+  public SELF hasMethods(String... methods) {
```
joel-costigliola: minor refactoring: I think we should rename `methods` to `methodNames` so that it is clear that we are not dealing with the `Method` type.

src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -284,4 +329,111 @@ public SELF hasDeclaredFields(String... fields) {
...
+   * <pre><code class='java'> class MyClass {
+   *     public void methodOne() {}
+   *     private void methodTwo() {}
+   * }
+   *
+   * // This one should pass :
+   * assertThat(MyClass.class).hasDeclaredMethods("methodOne", "methodTwo");
+   *
+   * // This one should fail :
+   * assertThat(MyClass.class).hasDeclaredMethods("methodThree");</code></pre>
```
joel-costigliola: to make it clear, define a superclass for `MyClass` with a public method and use it in the failed assertion example (keep the example of the non existent method)

src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -284,4 +329,111 @@ public SELF hasDeclaredFields(String... fields) {
...
+   *     public void methodOne() {}
+   *     public void methodTwo() {}
+   *     protected void methodThree() {}
+   * }
+   *
+   * // This one should pass :
+   * assertThat(MyClass.class).hasPublicMethods("methodOne");
+   * assertThat(MyClass.class).hasPublicMethods("methodOne", "methodTwo");
+   *
+   * // This one should fail :
```
joel-costigliola: plural

src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -284,4 +329,111 @@ public SELF hasDeclaredFields(String... fields) {
...
+   */
+  public SELF hasPublicMethods(String... methods) {
+    classes.assertHasPublicMethods(info, actual, methods);
+    return myself;
+  }
+
+  /**
+   * Verifies that the actual {@code Class} has the declared public {@code methods}.
+   *
+   * <pre><code class='java'> class MyClass {
```
joel-costigliola: same comment as before: introduce a superclass

src/main/java/org/assertj/core/error/ClassModifierShouldBe.java
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Creates an error message indicating that an assertion that verifies that a class is (or is not) final.
+ *
+ * @author Michal Kordas
+ */
+public class ClassModifierShouldBe extends BasicErrorMessageFactory {
+
+  private ClassModifierShouldBe(Class<?> actual, boolean positive, String modifier) {
+    super("%nExpecting:%n  <%s>%n" + (positive ? "to" : "not to") + " be a %s class.", actual, modifier);
```
joel-costigliola: suggestion: finish the error message with `but was %` and add actual's modifier value ?

src/main/java/org/assertj/core/internal/Classes.java
```diff
@@ -231,6 +263,115 @@ public void assertHasDeclaredFields(AssertionInfo info, Class<?> actual, String.
...
+   *
+   * @param info contains information about the assertion.
+   * @param actual the "actual" {@code Class}.
+   * @param methods the methods who must be declared in the class.
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} doesn't contains all of the methods.
+   */
+  public void assertHasDeclaredMethods(AssertionInfo info, Class<?> actual, String... methods) {
+    assertNotNull(info, actual);
+    doAssertHasMethods(info, actual, actual.getDeclaredMethods(), methods);
```
joel-costigliola: how do you now in `doAssertHasMethods` that the error message is about declared methods and not methods ?

src/main/java/org/assertj/core/internal/Classes.java
```diff
@@ -231,6 +263,115 @@ public void assertHasDeclaredFields(AssertionInfo info, Class<?> actual, String.
...
+    if(!noNonMatchingModifier(expectedMethodNames, actualMethods, nonMatchingModifiers, Modifier.PUBLIC)) {
+
+      throw failures.failure(info, shouldHaveMethods(actual, expectedMethodNames, Modifier.toString(Modifier.PUBLIC), nonMatchingModifiers));
+    }
+  }
+
+  private static boolean noNonMatchingModifier(Set<String> expectedMethodNames, Map<String, Integer> actualMethods, Map<String, String> nonMatchingModifiers, int modifier) {
+    for (String method : actualMethods.keySet()) {
+      if(expectedMethodNames.contains(method)) {
+        if ((actualMethods.get(method) & modifier) == 0) {
```
joel-costigliola: I believe you can combine `if` conditions

src/main/java/org/assertj/core/internal/Classes.java
```diff
@@ -231,6 +263,115 @@ public void assertHasDeclaredFields(AssertionInfo info, Class<?> actual, String.
...
+      throw failures.failure(info, shouldHaveMethods(actual, expectedMethodNames, missingMethodNames));
+    }
+    Map<String,String> nonMatchingModifiers= new LinkedHashMap<>();
+    if(!noNonMatchingModifier(expectedMethodNames, actualMethods, nonMatchingModifiers, Modifier.PUBLIC)) {
+
+      throw failures.failure(info, shouldHaveMethods(actual, expectedMethodNames, Modifier.toString(Modifier.PUBLIC), nonMatchingModifiers));
+    }
+  }
+
+  private static boolean noNonMatchingModifier(Set<String> expectedMethodNames, Map<String, Integer> actualMethods, Map<String, String> nonMatchingModifiers, int modifier) {
```
joel-costigliola: rename `actualMethods` to better express what this map contains (`methodsModifer` ?)

src/test/java/org/assertj/core/internal/classes/Classes_assertHasPublicMethods_Test.java
```diff
@@ -0,0 +1,70 @@
...
+
+  @Test
+  public void should_fail_if_actual_is_null() {
+    actual = null;
+    thrown.expectAssertionError(actualIsNull());
+    classes.assertHasPublicMethods(someInfo(), actual);
+  }
+
+  @Test
+  public void should_pass_if_no_methods_are_expected() {
```
epeee: k, thx. I will add this one.

epeee: To sum it up:
When no methods are given
 - `hasPublicMethods` will always fail because of methods inherit from `Object`
 - `hasDeclaredPublicMethods` will pass if no public methods are declared in the given class
 - `hasMethods` also will always fail because of methods inherit from `Object`

Am I right?

src/main/java/org/assertj/core/internal/Classes.java
epeee: thx for pointing that out. You are right, this one was still missing but should be fixed now.

epeee:
@joel-costigliola thx for the feedback. I have integrated your suggestions.

joel-costigliola:
Thanks @epeee, I'm currently polishing the code, will be integrated soonish.

`hasMethods` only checks public methods, I believe this is an error as we have an `hasPublicMethods` assertion, have I missed something ?

joel-costigliola:
Another thing, can you (in the future) rebase your branches instead of merging the main branch in yours ? The reason is that it is easier for me to keep a linear history with one commit per feature (most of the time at least). I took care of this branch but that involved a  lot a cherry pick and few conflicts to solve.

epeee:
@joel-costigliola k, I will do so in the future (rebase).

ad `hasMethods` only checks public methods:
good catch, the `isEmptyAndHasNoMethodsWithModifier` check is currently done just for public methods (wrongly).

joel-costigliola:
@epeee thanks for the additional commit.

I was trying to run the tests and found some were flaky because `ShouldHaveMethods` does not enforce a predictable order in the set of methods it receives, for example this test fails sometimes:
```java
    thrown.expectAssertionError(shouldNotHaveMethods(actual, true,
             newLinkedHashSet("publicMethod", "privateMethod", "protectedMethod")));
    classes.assertHasDeclaredMethods(someInfo(), actual);
```
because the order of methods is not always the same (a LinkedHashSet ensure).

The fix is simple, use `SortedSet` instead of `Set` in `ShouldNotHaveMethods` (it will also sort methods alphabetically which is gonna be helpful for users).

One more thing, could you add some tests for `ShouldNotHaveMethods`, you can follow  `ShouldHaveFields_create_Test` as an example.

One last thing (promised !) could you format the constructors in `ShouldHaveMethods` so that it goes to a new line after a `%n` like in `ShouldHaveSuppressedException`:

```java
  private ShouldHaveSuppressedException(Throwable actual, Throwable expectedSuppressedException) {
    super("%n" +
          "Expecting:%n" +
          "  <%s>%n" +
          "to have a suppressed exception with the following type and message:%n" +
          "  <%s> / <%s>%n" +
          "but could not find any in actual's suppressed exceptions:%n" +
          "  <%s>.",
          actual, expectedSuppressedException.getClass().getName(), expectedSuppressedException.getMessage(),
          actual.getSuppressed());
  }
```
it makes the error message easier to read.

thanks !

epeee:
@joel-costigliola thx for your feedback. I've already updated the pr.

joel-costigliola:
Integrated thanks !

joel-costigliola:
@epeee I think it would be better to align fields and methods assertions, I'm thinking of reducing methods assertions to: 
- `hasMethods(String... methodNames)` : check public accessible methods
- `hasDeclaredMethods(String... methodNames)` :  check declared methods (includes non public ones)
I want to do so to be consistent with the fields assertions
- `hasFields(String... fieldNames)` : check public accessible fields
- `hasDeclaredFields(String... fieldNames)` :  check declared fields (includes non public ones)

I might consider renaming `hasMethods` to `hasPublicMethods` and `hasFields` to `hasPublicFields` to avoid ambiguity.

WDYT ?

WDYT ?

epeee:
@joel-costigliola: :+1:
actually, it looks like the javadoc for `org.assertj.core.api.AbstractClassAssert#hasMethods` is wrong (the sample using the private method).

PR #918, opened 2017-02-10
When extracting a property using a String expression, I would find it…

adesmet:
… very useful if that String expression was immediately shown in the description


joel-costigliola:
good idea.

I can't accept this PR, it should be targeting 2.x and lacks tests.

adesmet:
Yeah sorry, I didn't have time for the overhead, just consider it an idea pitch. Thank you for considering it.

Kind regards,
Anthony

PR #923, opened 2017-02-12
meaningful generic names  + get rid of TODO

epeee:
- using meaningful generic names in `AbstractBigIntegerAssert`
- get rid of `//TODO add to WithAssertions in 3.x` since `AtomicBooleanAssert` is already added to `WithAssertions` in 3.x


joel-costigliola:
Thanks for this PR

PR #924, opened 2017-02-13
add license badge and reorganize badges

PascalSchumacher:
No ideal, as there seems to be no way to include branch information in  a travis badge, but the available space is put to better use.

For a test library the license is probably not that important, but it is still nice to know that assertj is apache licensed.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,6 @@
 # AssertJ - Fluent assertions for java
 
-Build status:
-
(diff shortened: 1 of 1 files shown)


PascalSchumacher:
@joel-costigliola What do you think?

joel-costigliola:
the licence is already available in the github project page as in https://help.github.com/articles/adding-a-license-to-a-repository/

PascalSchumacher:
Sorry, I never noticed that before. :( Maybe this was added recently. I will remove the license badge.

PascalSchumacher:
License display was added to the project page five months ago: https://github.com/blog/2252-license-now-displayed-on-repository-overview

PR #925, opened 2017-02-14
Meaningful generic names

epeee:
A very first bunch of master-only meaningful generic names.


joel-costigliola:
Sweet, thank you @epeee.

PR #931, opened 2017-02-20
Meaningful generic names master only

epeee:
some more meaningful generic names: Optional and CompletableFuture


joel-costigliola:
Thanks @epeee.

PR #935, opened 2017-02-25
Add a new AssertionMatcher that reuses AssertJ assertions

### LINKED ISSUE #926: AssertionMatcher proposal
#### Summary

I have a proposal to add a generic Hamcrest Matcher implementation that would fit assertj library:

```java
    public abstract class AssertionMatcher<T> extends BaseMatcher<T> {
        private AssertionError lastError;

        @Override
        public boolean matches(Object argument) {
            T actual = (T) argument;
            try {
                assertion(actual);
                return true;
            } catch (AssertionError e) {
                lastError = e;
                return false;
            }
        }

        public abstract void assertion(T actual) throws AssertionError;

        @Override
        public void describeTo(Description description) {
            //Describe lastError and its stacktrace
        }
    }
```

Advantages as I see it:

- don't create complex boolean expressions inside matches method
- override only assertion method
- inside assertion method use assertj default, custom or any other assertions that throw AssertionError
- failure description is provided with detailed description get from assertion and with a stacktrace that points exactly to line where assertion has failed.

#### Example

Basic usage is simple, for example with Mockito library:

```java
    verify(customerRepository).save(argThat(new AssertionMatcher<Customer>() {
            @Override
            public void assertion(Customer actual) throws AssertionError {
                assertThat(actual).hasName("Tom")
                    .hasAge(30);
            }
    }));
```

#### Java 8 specific ?

I this this proposal has generic use case and fits well with assertj library. Please review it and if you agree I would provide a PR.

SpOOnman:
#### Check List:
* Fixes #926 - AssertionMatcher implementation proposal
* Unit tests : YES

I'm not quite happy with a test AssertionMatcher_matches_Test#should_fill_description_when_assertion_fails. I've explained in code comments. I feel like I can't alter static Failures instance in this test because it would alter other tests.

Please review my contribution. Thanks!

### DIFF
--- a/src/main/java/org/assertj/core/matcher/AssertionMatcher.java
+++ b/src/main/java/org/assertj/core/matcher/AssertionMatcher.java
@@ -27,7 +27,7 @@ import org.hamcrest.Matcher;
  *
  * <p>
  * Example with Mockito use case:
- * <pre><code class='java'>verify(customerRepository).save(argThat(new AssertionMatcher&lt;Customer&gt;() {
+ * <pre><code class='java'> verify(customerRepository).save(argThat(new AssertionMatcher&lt;Customer&gt;() {
  *   &#64;Override
  *   public void assertion(Customer actual) throws AssertionError {
  *     assertThat(actual)
@@ -76,7 +76,7 @@ public abstract class AssertionMatcher<T> extends BaseMatcher<T> {
     if (lastError != null) {
       description.appendText("AssertionError with message: ");
       description.appendText(lastError.getMessage());
-      description.appendText("\n\nStacktrace was: ");
+      description.appendText(String.format("%n%nStacktrace was: "));
       description.appendText(Throwables.getStackTrace(lastError));
     }
   }
--- a/src/main/java/org/assertj/core/matcher/AssertionMatcher.java
+++ b/src/main/java/org/assertj/core/matcher/AssertionMatcher.java
@@ -1,0 +1,83 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2017 the original author or authors.
+ */
+package org.assertj.core.matcher;
+
+import org.assertj.core.util.Throwables;
+import org.hamcrest.BaseMatcher;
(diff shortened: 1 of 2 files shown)


src/main/java/org/assertj/core/matcher/AssertionMatcher.java
```diff
@@ -0,0 +1,83 @@
...
+ * Generic Hamcrest {@link Matcher} that reuses AssertJ assertions.
+ *
+ * Overriding classes should only override {@link AssertionMatcher#assertion(Object)}
+ * method. {@link Matcher#matches(Object)} and {@link Matcher#describeTo(Description)}
+ * are provided. If matcher fails it will provide description that contains
+ * stacktrace of a failed assertion.
+ *
+ * <p>
+ * Example with Mockito use case:
+ * <pre><code class='java'>verify(customerRepository).save(argThat(new AssertionMatcher&lt;Customer&gt;() {
```
joel-costigliola: add a space before `verify` to make javadoc indentation perfect

src/main/java/org/assertj/core/matcher/AssertionMatcher.java
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void describeTo(Description description) {
+    if (lastError != null) {
+      description.appendText("AssertionError with message: ");
+      description.appendText(lastError.getMessage());
+      description.appendText("\n\nStacktrace was: ");
```
joel-costigliola: Prefer `String.format("%n%nStacktrace was: ")` to be OS agnostic.

joel-costigliola:
You can alter the stack trace filtering boolean as long as you set it back to its original value (using `Failures` methods). 
AssertJ tests are not run in parallel, it should not cause any failures.

src/main/java/org/assertj/core/matcher/AssertionMatcher.java
SpOOnman: Done.

src/main/java/org/assertj/core/matcher/AssertionMatcher.java
SpOOnman: Done.

SpOOnman:
I use setUp and tearDown methods to ensure that removeAssertJRelatedElementsFromStackTrace is set back to its original value.

joel-costigliola:
AssertionMatcher_matches_Test.should_fill_description_when_assertion_fails failed in travis : https://travis-ci.org/joel-costigliola/assertj-core/builds/205524692

SpOOnman:
I don't know why but line numbers were different. I've removed line numbers. Still, I don't know why there were different. I've amended my commit and now it passes.

SpOOnman:
Hi, any news on this? Is there anything I can improve?

joel-costigliola:
Will do a final review in the we

joel-costigliola:
Integrated thanks  @SpOOnman !

SpOOnman:
Thank you!

PR #942, opened 2017-02-28
meaningful generic name + get rid of unnecessary semicolon

epeee:
a minor one


joel-costigliola:
Thanks for all this code cleaning work @epeee  !

epeee:
You're welcome!

PR #945, opened 2017-03-03
added biginteger to java6assertions

epeee:
It looks like I missed Java6*Assertions in the initial pr.

### DIFF
--- a/src/main/java/org/assertj/core/api/Java6Assertions.java
+++ b/src/main/java/org/assertj/core/api/Java6Assertions.java
@@ -17,6 +17,7 @@ import static org.assertj.core.data.Percentage.withPercentage;
 import java.io.File;
 import java.io.InputStream;
 import java.math.BigDecimal;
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
Strange. There are tests which compare the methods of the different assertion classes. I guess they buggy/incomplete as the missed this difference.

joel-costigliola:
Thanks @epeee.

@PascalSchumacher there are no tests to synchronize `Assertions` with `Java6Assertions` as the latter is a subset of the former. There is no easy way (that I know of) to select the subset of `Assertions` `assertThat` methods to have in `Java6Assertions` (unless you can ask Java in which version a type was added).

joel-costigliola:
FYI I have added tests to check that Java 6 assertions are in sync with Java 6 BDD and Soft assertions.

PascalSchumacher:
Note to self: Do not post late at night, it makes one look stupid. Sorry for the noise.

PR #955, opened 2017-03-10
When extracting a using a string expression use the expression in the…

PascalSchumacher:
… description (closes #918).

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
@@ -12,6 +12,7 @@
  */
 package org.assertj.core.api;
 
+import static java.lang.String.format;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.filter.Filters.filter;
 import static org.assertj.core.extractor.Extractors.byName;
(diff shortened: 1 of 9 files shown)


PascalSchumacher:
Too much copy&paste but where should the code reside?

joel-costigliola:
There is one problem, if I set a description it is replaced by the one provided by the `extracting` method.
I suggest to only describe the extracted fields if no description was set, I was thinking to add the extracted fields description to the existing but that would likely mess up the user's custom description - we don't want to mess up with fine-crafted user description, right ? :D

WDYT ?

@PascalSchumacher  I have refactored a bit your PR by moving the extracted fields descritption in `Extractors`, I have pushed the work in progress in the [provide_extraction_description](https://github.com/joel-costigliola/assertj-core/commit/35b51b3e24a74589ad517383d7d7d51735f42bdc) branch where I only have changed `AbstractIterableAssert` to keep the existing description.

One option to avoid duplicating the logic of keeping the existing description would be to add a static method in `Description` or maybe to pass the existing description in `extractedPropertiesOrFieldsDescription` or something like that ;-) 

WDYT again ?

PascalSchumacher:
I agree, it is a good idea to only add a description if it is currently empty.

Adding a static method like `Description#newIfEmpty`  or `Description#newIfNullOrEmpty` sounds like a good idea.

joel-costigliola:
Cool, I will finish the work, not much left to do.

joel-costigliola:
Thanks @PascalSchumacher  !

PR #959, opened 2017-03-15
using proper methods in some Tests

epeee:


joel-costigliola:
Good catch ! (facepalm) to me

PR #961, opened 2017-03-22
Recursive equals improvements for unordered sets comparison

### LINKED ISSUE #841: FieldByFieldRecursively comparator does not properly use comparatorForType when comparing Sets
#### Summary
Consider the test case in the example below. The last assertion fails though it should succeed. I have debugged the issue and it seems to boil down to an issue in ```org.assertj.core.internal.DeepDifference#compareUnorderedCollection```. The method uses a hashCode created using the ```deepHashCode``` method, which does not consider the ```comparatorByPropertyOrField```or ```comparatorByType``` maps.
Two BigDecimals with same value but different scale are not equal and have different hash codes, but compareTo returns 0 for them. Thus, a comparison using the hash code will fail, while a comparison using a BigDecimal::compareTo comparator will succeed.

#### Example

```java
@Test
public void bigDecimal() {
	BigDecimal value1 = BigDecimal.valueOf(100);
	BigDecimal value2 = value1.setScale(2);

	assertThat(value1).isNotEqualTo(value2);
	assertThat(value1.hashCode()).isNotEqualTo(value2.hashCode());
	assertThat(value1).isEqualByComparingTo(value2);
	
	class Foo {
		private final BigDecimal value;

		Foo(BigDecimal value) {
			this.value = value;
		}

		BigDecimal getValue() {
			return value;
		}
	}

	Foo foo1 = new Foo(value1);
	Foo foo2 = new Foo(value2);

	assertThat(foo1)
			.usingComparatorForType(BigDecimal::compareTo, BigDecimal.class)
			.isEqualToComparingFieldByFieldRecursively(foo2);

	class Bar {
		private final Set<Foo> foos;

		Bar(Set<Foo> foos) {
			this.foos = foos;
		}

		Set<Foo> getFoos() {
			return foos;
		}
	}

	assertThat(new Bar(singleton(foo1)))
			.usingComparatorForType(BigDecimal::compareTo, BigDecimal.class)
			.isEqualToComparingFieldByFieldRecursively(new Bar(singleton(foo2)));
}
```

#### Java 8 specific ?
Example is Java 8 specific but the problem should be the same in older versions.

fiery-phoenix:
#### Check List:
* Fixes #841
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/internal/DeepDifference.java
+++ b/src/main/java/org/assertj/core/internal/DeepDifference.java
@@ -20,7 +20,6 @@ import static org.assertj.core.util.introspection.PropertyOrFieldSupport.COMPARI
 import java.lang.reflect.Array;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.Deque;
@@ -58,10 +57,6 @@ public class DeepDifference {
       this.path = path;
       this.key1 = key1;
       this.key2 = key2;
-    }
-
-    private DualKey(Object key1, Object key2) {
-      this(new ArrayList<String>(), key1, key2);
     }
 
     @Override
@@ -140,7 +135,8 @@ public class DeepDifference {
    * 
    * @param a Object one to compare
    * @param b Object two to compare
-   * @param comparatorByType
+   * @param comparatorByPropertyOrField comparators to compare properties or fields with the given names
+   * @param comparatorByType comparators to compare properties or fields with the given types
    * @return the list of differences found or an empty list if objects are equivalent.
    *         Equivalent means that all field values of both subgraphs are the same,
    *         either at the field level or via the respectively encountered overridden
@@ -149,8 +145,14 @@ public class DeepDifference {
   public static List<Difference> determineDifferences(Object a, Object b,
                                                       Map<String, Comparator<?>> comparatorByPropertyOrField,
                                                       TypeComparators comparatorByType) {
+    return determineDifferences(a, b, null, comparatorByPropertyOrField, comparatorByType);
+  }
+
+  private static List<Difference> determineDifferences(Object a, Object b, List<String> parentPath,
+                                                      Map<String, Comparator<?>> comparatorByPropertyOrField,
+                                                      TypeComparators comparatorByType) {
     final Set<DualKey> visited = new HashSet<>();
-    final Deque<DualKey> toCompare = initStack(a, b, visited);
+    final Deque<DualKey> toCompare = initStack(a, b, parentPath, comparatorByPropertyOrField, comparatorByType);
     final List<Difference> differences = new ArrayList<>();
 
     while (!toCompare.isEmpty()) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/assertj/core/internal/DeepDifference.java
```diff
@@ -307,37 +313,58 @@ private static boolean hasCustomComparator(DualKey dualKey, Map<String, Comparat
...
-          String fieldName = field.getName();
-          DualKey dk = new DualKey(Arrays.asList(fieldName),
-                                   COMPARISON.getSimpleValue(fieldName, a),
-                                   COMPARISON.getSimpleValue(fieldName, b));
-          if (!visited.contains(dk)) {
+      Collection<Field> aFields = getDeclaredFieldsIncludingInherited(a.getClass());
+      if (!aFields.isEmpty()) {
+        Set<String> aFieldsNames = getFieldsNames(aFields);
+        Set<String> bFieldsNames = getFieldsNames(getDeclaredFieldsIncludingInherited(b.getClass()));
+        if (!bFieldsNames.containsAll(aFieldsNames)) {
```
fiery-phoenix: Probably here instead of `contansAll` could have been done check on equals of fileds sets, but it corresponds more to how it is working now.

src/main/java/org/assertj/core/internal/DeepDifference.java
```diff
@@ -443,6 +464,44 @@ private static boolean compareArrays(Object array1, Object array2, List<String>
...
+      // this comparison is used for performance optimization reasons
+      return compareUnorderedCollectionByHashCodes(col1, col2, path, toCompare, visited);
+    }
+
+    Collection<V> col2Copy = new LinkedList<>(col2);
+    for (Object o1 : col1) {
+      Iterator<V> iterator = col2Copy.iterator();
+      while (iterator.hasNext()) {
+        Object o2 = iterator.next();
+        if (determineDifferences(o1, o2, path, comparatorByPropertyOrField, comparatorByType).isEmpty()) {
```
fiery-phoenix: probably here (and some other places) could have been added something to the path, to distinguish collection element from collection itself. (May be needed in case there are custom comparators by field names for collection elements)

joel-costigliola: the idea is good but I'm not sure you would identify a specific element, it works for List as you can use an index but not for a Set

joel-costigliola:
Many thanks @fiery-phoenix, `DeepDifference` is complex code !

src/main/java/org/assertj/core/internal/DeepDifference.java
fiery-phoenix: I was thinking about adding smth like `.item` to the path. For example, at `#testUnorderedCollectionWithCustomComparatorsByFieldName `

    SetWrapper a = new SetWrapper(newLinkedHashSet(new Wrapper("one"), new Wrapper("two")));
    fieldComparators.put("set.o", ALWAY_EQUALS);

`"set.o"` points to `Wrapper` field values "one" and "two", but there is no way, to specify path to `Wrapper` itself, as path `"set"`, points to both `set` field from `SetWrapper` and its elements of type `Wrapper`.

But as there is a workaround for this, and this change is to break currently used paths (`"set.o"`  should be replaced with `"set.item.o"`), I just mentioned it as a comment.

fiery-phoenix:
You are welcome :)

PR #970, opened 2017-04-10
Add doesNotContainPattern to String assertions as counterpart to containsPattern

michal-szkutnik:
The containsPattern String assertion lacks a negated version, which this pull requests adds.
Rationale - users tend to expect having both negative and positive assertion versions,
I was surprised to not find a matching negative version myself and had to use 
doesNotMatch(".*<pattern>.*"), which does the job of course, but is not as descriptive
and elegant.
Based on Pierre Templier's original work on containsPattern.

#### Check List:
* Fixes N/A
* Unit tests : YES
* Javadoc with a code example (API only) : YES /

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
@@ -570,6 +570,51 @@ public abstract class AbstractCharSequenceAssert<SELF extends AbstractCharSequen
   }
 
   /**
+   * Verifies that the actual {@code CharSequence} does not contain the given regular expression.
+   * <p>
+   * Example :
+   * <pre><code class='java'>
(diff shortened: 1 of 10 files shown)


joel-costigliola:
We don't always add negative assertions because we are not sure they are always useful, if some users ask for it then we add them as we are sure there is a use case for it.

michal-szkutnik:
Then I would be one user for such an assertion. If there needs to be an additional broader discussion/voting, let's say in an issue (don't know the workflow), I can create one.
If not, then thanks in advance for considering the pull request.

joel-costigliola:
No need, you needing it is enough for me, I was just explaining why we don't always add negative assertions.

joel-costigliola:
Integrated thanks @michal-szkutnik !

PR #978, opened 2017-04-23
get rid of deprecated Tuple#addData

epeee:
What about getting rid of deprecated Tuple#addData?
It is not used by the assertj-core project and seems to be deprecated for > 1.5 years now.


joel-costigliola:
Thanks a lot !

Issue #984, opened 2017-05-08
Double.NaN isCloseTo always true


gn6996:
#### Summary

AssertJ (3.6.2) returns true, if checking double is close to NaN. 
[StackOverflow](http://stackoverflow.com/questions/43848512/double-nan-iscloseto-double-value-returns-true)
#### Example

```java
assertThat(Double.NaN).isCloseTo(0.007, Percentage.withPercentage(0.1));
```

PascalSchumacher:
Thanks for reporting this bug also occurs in 2.7.0/3.7.0!

joel-costigliola:
Good catch ! Thanks for reporting it.

joel-costigliola:
We will do a bugfix release in the next weeks to address that.

PascalSchumacher:
I have started to work on a pull request for this. Looks like there are also bugs related to `Double.NEGATIVE_INFINITY` and `Double.POSITIVE_INFINITY`.

PascalSchumacher:
@gn6996 AssertJ versions 2.8.0 / 3.8.0 with the fix for this have just been released. :smile:

PR #987, opened 2017-05-12
fix bugs in Numbers#assertIsCloseToPercentage, #assertIsNotCloseToPer…

PascalSchumacher:
…centage and #assertIsNotCloseTo related to Float/Double#NaN, #POSITIVE_INFINITY and #NEGATIVE_INFINITY (#984)

No sure if it is the best way to fix these problems.


joel-costigliola:
Thanks Pascal !

PR #989, opened 2017-05-13
class based navigable assert: do not swallow reflection problems

cal101:


joel-costigliola:
If this is the same as changes #990 then there is no need for this PR, the 2.x branch is regularly merged in master.

PR #992, opened 2017-05-13
class based navigable assert: do not swallow reflection problems

cal101:

### DIFF
--- a/src/main/java/org/assertj/core/api/ClassBasedNavigableIterableAssert.java
+++ b/src/main/java/org/assertj/core/api/ClassBasedNavigableIterableAssert.java
@@ -53,7 +53,7 @@ public class ClassBasedNavigableIterableAssert<SELF extends ClassBasedNavigableI
       throw new RuntimeException("Fail to find a constructor matching " + value
                                  + " class to build the expected Assert class");
     } catch (Exception e) {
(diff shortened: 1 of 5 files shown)


cal101:
Stand by, I seem to have made a git mess.

cal101:
OK, looks good, just a local problem.

src/main/java/org/assertj/core/api/ClassBasedNavigableIterableAssert.java
```diff
@@ -53,7 +53,7 @@ public ELEMENT_ASSERT toAssert(ELEMENT value, String description) {
       throw new RuntimeException("Fail to find a constructor matching " + value
                                  + " class to build the expected Assert class");
     } catch (Exception e) {
-      throw new RuntimeException("Fail to build an assert object with " + value);
+      throw new RuntimeException("Fail to build an assert object with " + value + ": " + e.getMessage());
```
PascalSchumacher: I guess it would be even more helpful to add the whole exception instead of just the message? (e.g. `new RuntimeException("Failed to build an assert object with " + value, e)`)

cal101: Feel free ...

PascalSchumacher:
Thanks! :+1:

PR #996, opened 2017-05-21
Check start <= end in isBetween

### LINKED ISSUE #994: Check start <= end in isBetween and isStrictlyBetween assertions
For the time being, it is possible to pass an end < start :(.

AssertJ should throw an `IllegalArgumentException` in that case

epeee:
#### Check List:
* Fixes #994 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/internal/Comparables.java
+++ b/src/main/java/org/assertj/core/internal/Comparables.java
@@ -19,6 +19,7 @@ import static org.assertj.core.error.ShouldBeGreaterOrEqual.shouldBeGreaterOrEqu
 import static org.assertj.core.error.ShouldBeLess.shouldBeLess;
 import static org.assertj.core.error.ShouldBeLessOrEqual.shouldBeLessOrEqual;
 import static org.assertj.core.error.ShouldNotBeEqual.shouldNotBeEqual;
+import static org.assertj.core.util.Preconditions.checkArgument;
 import static org.assertj.core.util.Preconditions.checkNotNull;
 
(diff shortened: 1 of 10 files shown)


joel-costigliola:
Thanks @epeee, @PascalSchumacher or I will review that soonish.

src/main/java/org/assertj/core/internal/Comparables.java
```diff
@@ -255,12 +256,16 @@ private boolean isLessThan(Object actual, Object other) {
...
    * @throws NullPointerException if start value is {@code null}.
    * @throws NullPointerException if end value is {@code null}.
+   * @throws IllegalArgumentException if end value is less than start value.
    */
   public <T extends Comparable<? super T>> void assertIsBetween(AssertionInfo info, T actual, T start, T end,
       boolean inclusiveStart, boolean inclusiveEnd) {
     assertNotNull(info, actual);
     checkNotNull(start, "The start range to compare actual with should not be null");
     checkNotNull(end, "The end range to compare actual with should not be null");
+    checkArgument(comparisonStrategy.isLessThanOrEqualTo(start, end),
```
joel-costigliola: if we use `isStrictlyBetween` then we should not accept start == end.

epeee: @joel-costigliola thx, currently we end up in a `AssertionError` in such a case. If this change is ok I can change the implementation to also throw an `IllegalArgumentException` in this case.

PascalSchumacher:
Thanks! :+1:

PR #998, opened 2017-05-22
[ci] Fix relativePath from pom

hazendaz:
Restores a fix added previously.  RelativePath needs to be set unless this project is nexted within parent.  It is not so it needs set.  I believe I did this once before but some merged from origin/2.x wrote over it.  The same fix would need over on 2.x branch as well if you agree I can submit a PR there too.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,6 +12,7 @@
     <groupId>org.assertj</groupId>
     <artifactId>assertj-parent-pom</artifactId>
     <version>2.1.6</version>
+    <relativePath />
   </parent>
   <mailingLists>
     <mailingList>


joel-costigliola:
That's weird, I thought Maven would resolve the parent pom like a dependency.
The CI build seems to work, where do you see the problem ?

hazendaz:
It just spits out a warning during build.  It resolves ok, it just notes that it expects pom.xml one folder up which it won't be thus the warning.  I'll paste the warning here in a moment.

hazendaz:
```xml
[INFO] Scanning for projects...
[WARNING]
[WARNING] Some problems were encountered while building the effective model for org.assertj:assertj-core:bundle:3.8.1-SNAPSHOT
[WARNING] 'parent.relativePath' of POM org.assertj:assertj-core:3.8.1-SNAPSHOT (C:\Users\Jeremy\Documents\GitHub\assertj-core\pom.xml) points at com.github.hazendaz:clean-target instead of org.assertj:assertj-parent-pom, please verify your project structure @ line 11, column 11
[WARNING]
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING]
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING]
```

joel-costigliola:
Thanks for the explanation !

hazendaz:
So, most likely noticed because I have a pom above in front of all my projects that it sees.  It might not complain if one isn't present in that location.  Generally though this speeds the build up slightly since it will simply not look there and instead go to the local repo first.

hazendaz:
And that further would explain why I see this fairly often.  Both at home and at work, I drop all my projects in a nexted mavn structure so I can run a massive clean against everything to save space ;)

joel-costigliola:
Can you do the fix on 2.x only ? we are regularly merging 2.x into master as master = 2.x + java 8 stuff so the fix will eventually get in master (eventual consistency ... lol)

hazendaz:
Yup I'll do that.  Closing this one.

PR #999, opened 2017-05-23
[ci] Fix relativePath from pom

hazendaz:
In reference to #998, making adjustment here which should hit 3.x (master) eventually.


joel-costigliola:
thanks @hazendaz !

Issue #1011, opened 2017-06-16
Some CompletableFuture assertion error message are confusing


joel-costigliola:
#### Summary

In any given unit test, the given `Future` might not have completed yet, causing the test to be brittle. The exception report can be quite confusing because between the test failure and the printing of the test error, the future might complete. So even a test asserting `isDone()` can fail with a message showing a done future.

#### Java 8 specific ?

* No : create Pull Request from the `2.x` branch

tkruse:
I created a PR with test to reproduce. Will think on how to improve. 

For the suggested improvement, still thinking. The current report example: 
```
[ERROR]   CompletableFutureAssert_isCompletedExceptionally_Test.should_pass_if_completable_future_will_complete_exceptionally[89] 
Expecting
  <CompletableFuture[Failed: java.lang.RuntimeException]>
to be completed exceptionally
```
should indicate that because this is an assert on a Future, the value shown may be different from the value that failed the assert. And this could include a suggestion to make the test stable by waiting for the async result before asserting.

joel-costigliola:
I think one of the problem here is `StandardRepresentation#toStringOf(CompletableFuture<?> future)` that tries to call `join()` on the future. definitely bad.

I'm thinking that the current assertions check the future at at point in time and the error should represent the future at this same point in time, if that is not possible it should avoid to display the future state later on as it would not be relevant. 

@tkruse thoughts ?

tkruse:
I have not looked at the code causing this yet. Your analysis sounds fine. For the design of the message, I generally think the more information the better, if it can be presented in a non-confusing way.
Not showing the state would hide useful information in most cases. The race condition is an exceptional case.

A workaround would be to call toStringOf() before and after the check, and if the values differ, display them both somehow.

Else a hint that the displayed value might be updated should be good enough.

joel-costigliola:
Interesting, I need to try the workaround to see if it can read nicely, if it does not then adding a warning like "be aware that the description of the future displayed here can different from the one at the moment the assertion was performed" as a the last line of  the end of the error message (if you have a better wording ... mine could certainly be improved)

joel-costigliola:
I don't think it is possible to capture the future description and evaluate it atomically so the best that we can do is to add a warning in the error messages pertaining to future assertions.

tkruse:
I'd prefer the information and a warning / hint.

joel-costigliola:
[...]

PR #1012, opened 2017-06-16
add brittle test with race condition

tkruse:
reproducing #1011 

Run the changed test. It repeats the same test 100 times. This fails for me like this:
```
mvn test -Dtest=CompletableFutureAssert_isCompletedExceptionally_Test

[ERROR]   CompletableFutureAssert_isCompletedExceptionally_Test.should_pass_if_completable_future_will_complete_exceptionally[88] 
Expecting
  <CompletableFuture[Incomplete]>
to be completed exceptionally


[ERROR]   CompletableFutureAssert_isCompletedExceptionally_Test.should_pass_if_completable_future_will_complete_exceptionally[89] 
Expecting
  <CompletableFuture[Failed: java.lang.RuntimeException]>
to be completed exceptionally
```

As you can see the first message makes sense, the future is incomplete. In the second case, the error says "expecting ... to be completed exceptionally, but it shows [Failed: java.lang.RuntimeException]"

Not sure if or how this can best be fixed, probably just a hint in the message might work.


tkruse:
needless to say, this is not to be merged, and can be closed without further comment

PR #1019, opened 2017-06-26
Add assertion for completion stage

### LINKED ISSUE #1010: Add CompletionStage assertion
#### Summary
AbstractCompletableFutureAssert has methods to run assertions on CompletableFutures.
However, in any given unit test, the given CompletableFuture might not have completed yet, causing the test to be brittle. The exception report can be quite confusing because between the test  failure and the printing of the test error, the future might complete. So even a test asserting isDone() can fail with a message showing a done future.

Coding around this is ugly. Would be real nice if assertj had some waiting (with timeout, with default timeout) function embedded.

Also would be nice if this would work with CompletionStages as well.

Something like
```
CompletionStage<Foo> fooStage = getFoo();
assertThat(fooStage).completesWithin(1000).hasFailedWithThrowableThat()
```

Note: I think the awaitility library could also help, but so far did not use it. Could be this is better solved by using both awaitility and assertj.

tkruse:
#### Check List:
* Fixes #1010
* Unit tests :  NO

Note I was not creative, so far only Copy&Paste from CompletableFuture classes (and javadoc is still unchanged)

### DIFF
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -43,6 +43,7 @@ import java.util.OptionalDouble;
 import java.util.OptionalInt;
 import java.util.OptionalLong;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -190,6 +191,19 @@ public class Assertions {
    */
   @CheckReturnValue
   public static <RESULT> CompletableFutureAssert<RESULT> assertThat(CompletableFuture<RESULT> actual) {
+    return AssertionsForClassTypes.assertThat(actual);
+  }
+
+  /**
+   * Create assertion for {@link java.util.concurrent.CompletionStage}.
+   *
+   * @param actual the actual value.
+   * @param <RESULT> the type of the value contained in the {@link java.util.concurrent.CompletionStage}.
+   *
+   * @return the created assertion object.
+   */
+  @CheckReturnValue
+  public static <RESULT> CompletionStageAssert<RESULT> assertThat(CompletionStage<RESULT> actual) {
     return AssertionsForClassTypes.assertThat(actual);
   }
 
--- a/src/main/java/org/assertj/core/api/AssertionsForClassTypes.java
+++ b/src/main/java/org/assertj/core/api/AssertionsForClassTypes.java
@@ -36,6 +36,7 @@ import java.util.OptionalDouble;
 import java.util.OptionalInt;
 import java.util.OptionalLong;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 
 import org.assertj.core.api.ThrowableAssert.ThrowingCallable;
 import org.assertj.core.api.exception.RuntimeIOException;
@@ -82,6 +83,19 @@ public class AssertionsForClassTypes {
   @CheckReturnValue
   public static <RESULT> CompletableFutureAssert<RESULT> assertThat(CompletableFuture<RESULT> actual) {
     return new CompletableFutureAssert<>(actual);
+  }
+
+  /**
+   * Create assertion for {@link java.util.concurrent.CompletionStage}.
+   *
+   * @param actual the actual value.
+   * @param <RESULT> the type of the value contained in the {@link java.util.concurrent.CompletionStage}.
+   *
+   * @return the created assertion object.
+   */
+  @CheckReturnValue
+  public static <RESULT> CompletionStageAssert<RESULT> assertThat(CompletionStage<RESULT> actual) {
+    return new CompletionStageAssert<>(actual);
   }
 
   /**
--- a/src/main/java/org/assertj/core/api/BDDAssertions.java
+++ b/src/main/java/org/assertj/core/api/BDDAssertions.java
@@ -35,6 +35,7 @@ import java.util.OptionalDouble;
 import java.util.OptionalInt;
 import java.util.OptionalLong;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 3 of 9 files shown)


tkruse:
Javadoc and Test still TODO.
This currently fails in Maven/oracle-java8/Ubuntu for me, but passes in IntelliJ.

Failure is
```
 [ERROR] Failures: 
[ERROR]   Assertions_sync_assertThat_with_BDD_and_Soft_variants_Test.bdd_assertions_and_bdd_soft_assertions_should_have_the_same_assertions_methods 
...
but the following elements were unexpected:
  <[public static org.assertj.core.api.CompletionStageAssert org.assertj.core.api.BDDAssertions.then(java.util.concurrent.CompletionStage)]>
```

joel-costigliola:
Thanks @tkruse , I will review your PR in the next days !

joel-costigliola:
@tkruse I can build your branch with maven on ubuntu 16.04, more precisely:
```
Maven home: /home/joel/prog/apache-maven-3.5.0
Java version: 1.8.0_112, vendor: Oracle Corporation
Java home: /home/joel/prog/java/jdk1.8.0_112/jre
Default locale: en_NZ, platform encoding: UTF-8
OS name: "linux", version: "4.4.0-83-generic", arch: "amd64", family: "unix"
```

tkruse:
Ok, well, I cannot reproduce today. Maybe some IntelliJ interference. Any feedback on the codig approach? SHould I start writing unit tests, or would you want some redesign first?

joel-costigliola:
I'll give some feedback by the end of the week.

src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -194,6 +195,19 @@ public static DoublePredicateAssert assertThat(DoublePredicate actual) {
...
+   * Create assertion for {@link java.util.concurrent.CompletionStage}.
+   *
+   * @param actual the actual value.
+   * @param <RESULT> the type of the value contained in the {@link java.util.concurrent.CompletionStage}.
+   *
+   * @return the created assertion object.
+   */
+  @CheckReturnValue
+  public static <RESULT> CompletionStageAssert<RESULT> assertThat(CompletionStage<RESULT> actual) {
+    return AssertionsForClassTypes.assertThat(actual);
```
joel-costigliola: should be in `AssertionsForInterfaceTypes`

src/main/java/org/assertj/core/api/AssertionsForClassTypes.java
```diff
@@ -85,6 +86,19 @@
...
   /**
+   * Create assertion for {@link java.util.concurrent.CompletionStage}.
+   *
+   * @param actual the actual value.
+   * @param <RESULT> the type of the value contained in the {@link java.util.concurrent.CompletionStage}.
+   *
+   * @return the created assertion object.
+   */
+  @CheckReturnValue
+  public static <RESULT> CompletionStageAssert<RESULT> assertThat(CompletionStage<RESULT> actual) {
```
joel-costigliola: move this method should be in `AssertionsForInterfaceTypes` as `CompletionStage` is an interface

src/main/java/org/assertj/core/api/BDDAssertions.java
```diff
@@ -1132,6 +1133,19 @@ public static AtomicLongArrayAssert then(AtomicLongArray actual) {
...
   /**
+   * Create assertion for {@link java.util.concurrent.CompletionStage}.
+   *
+   * @param future the actual value.
+   * @param <RESULT> the type of the value contained in the {@link java.util.concurrent.CompletionStage}.
+   *
+   * @return the created assertion object.
+   */
+  @CheckReturnValue
+  public static <RESULT> CompletionStageAssert<RESULT> then(CompletionStage<RESULT> future) {
```
joel-costigliola: rename `future` to `completionStage`

joel-costigliola:
@tkruse `CompletionStage` API does not expose data, I don't see which assertions we could perform on a `CompletionStage`.
Is there a value to convert a `CompletionStage` to a `CompletableFuture` on behalf of the user ? (I mean they could do when they need to).

tkruse:
There is no other value to it than to reduce the code for users dealing with APIs based on CompletionStage.

joel-costigliola:
If I understand you correctly, all you need is are assertions factory method to get a `CompletableFutureAssert` from a `CompletionStage` ? 

```java
CompletableFutureAssert<RESULT> assertThat(CompletionStage<RESULT> actual) {
  return assertThat(actual.toCompletableFuture());
}
```

tkruse:
Yes, looks much easier than what I tried.

tkruse:
So should this PR be closed (for users to create such a method themselves), or modified adding this factory Method to some class?
The code you give is not null-safe, and also not corner-case safe (weird custom implementations of CompletionStage might exist that do undesired things on toCompletableFuture()).

tkruse:
@joel-costigliola : Any progress? This can be closed, but still it would be nice to have those  assertions factory methods you mentioned

joel-costigliola:
I will integrate the factory method for the next release, super busy right now so if you feel like revamping this PR to only contain the factory method + unit tests then that would help.

joel-costigliola:
Finished in 5badc3775065ad4041a8b974198b73040d3ad4f6 by adding a factory method to return `CompletableFuture` assertion from a `CompletionStage`.

tkruse:
Hi,

so the sad thing about this is that the compilation breaks for any Objects that extend both Future and CompletionStage:

    private static interface Both<V> extends CompletionStage<V>, Future<V> {
        
    }
    
    @Test
    public void testAssertJ() {
        Both future2 = mock(Both.class);
        // ambiguous method call
        assertThat(future2).isNotDone();
        // Workaround
        assertThat((Future<?>) future2).isNotDone();
    }

This hapens e.g. for RFuture of Redisson. Maybe not a frequent use-case. A simple workaround is to disambiguate with a cast. But I wonder if this could be avoided entirely, somthing like this:

    static <T, I extends Future<T> & CompletionStage<T>> void assertThat(I x) {
    }

joel-costigliola:
this is related to https://stackoverflow.com/questions/29499847/ambiguous-method-in-java-8-why I believe which was sadly introduced in java 8.

tkruse:
I guess another workaround would be to implement a custom static assertThat(RFuture future) {...} method.

PR #1026, opened 2017-07-02
Extract ObjectAssert for value from OptionalAssert (#1021)

### LINKED ISSUE #1021: OptionalAssert should have all the methods that the AbstractObjectAssert has
#### Summary
It would be good if I can do assertions on the internal value of an `Optional` in the same way as we are doing assertions for objects.

Possible solutions:
* Provide a method that checks if a value is present and returns an `AbstractObjectAssert` back
* Copy all the methods from `AbstractObjectAssert` with names that would make meaningful for `Optional`
* Make sure that `AbstractOptionalAssert` extends `AbstractObjectAssert` and override all methods with a new implementation
* Extract an interface for all the "interesting" methods and make sure that the `AbstractObjectAssert` and `AbstractOptionalAssert` implement the interface. 

#### Example

```java
Optional<MyObject> myObject = new MyObject();

Assertions.assertThat(myObject).isEqualToComparingFieldByField(new MyObject());
```

#### Java 8 specific ?

filiphr:
#### Check List:
* Fixes #1021
* Unit tests : YES 

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractOptionalAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractOptionalAssert.java
@@ -405,6 +405,34 @@ public abstract class AbstractOptionalAssert<SELF extends AbstractOptionalAssert
     return assertThat(actual.map(mapper));
   }
 
+  /**
+   * Verifies that the actual {@link Optional} is not {@code null} and not empty,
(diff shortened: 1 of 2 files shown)


joel-costigliola:
Thanks @filiphr

PR #1027, opened 2017-07-02
Use an SPI for registering a custom Representation (#1016)

### LINKED ISSUE #1016: Provide a way to plugin a custom toStringOf
#### Summary
I know that one can provide a custom representation and I know about `Assertions#registerFormatterForType(Class, Function)`. However, for both of them one has to manually set them statically in each test.

Wouldn't it be good to have some sort of Service Provider Interface (SPI) that the users can implement and AssertJ can pick it up?

That way one can use something like [`ReflectionToStringBuilder`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/builder/ReflectionToStringBuilder.html) from `commons-lang`
Maybe there is some way to do it globally, but I have missed it.

#### Java 8 specific ?

filiphr:
#### Check List:
* Fixes #1016 
* Unit tests : YES
* Javadoc with a code example (API only):  NA

I added `ConfigurationProvider` that could potentially be a single entry place for SPI / default configuration within AssertJ. 

We need to find a way to test the `ConfigurationProvider` better. We need to test that we use the `StandardRepresentation` when nothing is registered or we use the registered configuration when something is registered. In [mapstruct](https://github.com/mapstruct/mapstruct) another project I am working on we have a nice way of testing such SPIs. There is a custom JUnit 4 runner that does some "magic" with the classloaders and registers SPIs if needed (via annotation). We want to extract this testing infrastructure into a separate module / project, but we haven't had the time yet.

I am open to discussion for everything in the PR.

### DIFF
--- a/src/main/java/org/assertj/core/configuration/ConfigurationProvider.java
+++ b/src/main/java/org/assertj/core/configuration/ConfigurationProvider.java
@@ -1,0 +1,43 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ * <p>
+ * Copyright 2012-2017 the original author or authors.
+ */
+package org.assertj.core.configuration;
+
+import org.assertj.core.presentation.Representation;
+import org.assertj.core.presentation.StandardRepresentation;
+
+/**
+ * Provider for all the configuration settings / parameters within AssertJ.
+ * <p>
+ * All the configuration possibilities are registered via an SPI.
+ *
+ * @author Filip Hrisafov
+ * @since 2.9.0 / 3.9.0
+ */
+public final class ConfigurationProvider {
+
+  public static final ConfigurationProvider CONFIGURATION_PROVIDER = new ConfigurationProvider();
+
+  private final Representation defaultRepresentation;
+
+  public ConfigurationProvider() {
+    defaultRepresentation = Services.get(Representation.class, new StandardRepresentation());
+  }
+
+  /**
+   * @return the default {@link Representation} that needs to be used within AssertJ
+   * @since 2.9.0 / 3.9.0
+   */
+  public Representation representation() {
+    return defaultRepresentation;
+  }
+}
--- a/src/main/java/org/assertj/core/configuration/ConfigurationProvider.java
+++ b/src/main/java/org/assertj/core/configuration/ConfigurationProvider.java
@@ -1,13 +1,13 @@
 /**
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
- * <p>
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * <p>
+ *
  * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
  * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
  * specific language governing permissions and limitations under the License.
- * <p>
+ *
  * Copyright 2012-2017 the original author or authors.
  */
 package org.assertj.core.configuration;
--- a/src/main/java/org/assertj/core/util/diff/Delta.java
+++ b/src/main/java/org/assertj/core/util/diff/Delta.java
@@ -136,6 +136,7 @@ public abstract class Delta<T> {
   }
 
   String formatLines(List<T> lines) {
+    //TODO see what we should use here
     return STANDARD_REPRESENTATION.format(lines, DEFAULT_START, DEFAULT_END, ELEMENT_SEPARATOR_WITH_NEWLINE, "   ");
   }
 
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -12,8 +12,8 @@
  */
 package org.assertj.core.api;
 
+import static org.assertj.core.configuration.ConfigurationProvider.CONFIGURATION_PROVIDER;
 import static org.assertj.core.data.Percentage.withPercentage;
-import static org.assertj.core.presentation.StandardRepresentation.STANDARD_REPRESENTATION;
 
 import java.io.File;
 import java.io.InputStream;
@@ -2126,7 +2126,7 @@ public class Assertions {
    * @since 2.5.0 / 3.5.0
    */
   public static void useDefaultRepresentation() {
-    AbstractAssert.setCustomRepresentation(STANDARD_REPRESENTATION);
+    AbstractAssert.setCustomRepresentation(CONFIGURATION_PROVIDER.representation());
   }
 
   /**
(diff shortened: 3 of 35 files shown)


src/main/java/org/assertj/core/util/diff/Delta.java
```diff
@@ -136,6 +136,7 @@ int lineNumber() {
   }
 
   String formatLines(List<T> lines) {
+    //TODO see what we should use here
     return STANDARD_REPRESENTATION.format(lines, DEFAULT_START, DEFAULT_END, ELEMENT_SEPARATOR_WITH_NEWLINE, "   ");
```
filiphr: This was the only place that I could not remove the `STANDARD_REPRESENTATION` usage. Maybe we need to move this formatting to some other class instead.

filiphr:
I think that we need to remove all the usage of `StandardRepresentation.STANDARD_REPRESENTATION` within the entire codebase.

src/main/java/org/assertj/core/configuration/ConfigurationProvider.java
```diff
@@ -0,0 +1,43 @@
...
+ * @since 2.9.0 / 3.9.0
+ */
+public final class ConfigurationProvider {
+
+  public static final ConfigurationProvider CONFIGURATION_PROVIDER = new ConfigurationProvider();
+
+  private final Representation defaultRepresentation;
+
+  public ConfigurationProvider() {
+    defaultRepresentation = Services.get(Representation.class, new StandardRepresentation());
```
joel-costigliola: Can you clarify something ? if I have my own `Representation` implementation, wouldn't `Services.get` fail as it would find two (`StandardRepresentation` and mine) ?

filiphr: It won't, we (assertj) won't register the `StandardRepresentation` as an SPI. 

In `Services#get(Class, Object) ` the second parameter is a fallback, in case no SPI has been registered. There are already multiple `Representation` implementations in assertj, and it is not a problem. 

However, if you have 2 dependencies registering a `Representation` then it'll fail. What is the desired behaviour here?

joel-costigliola: got it, thanks for the explanation !
I think throwing an exception is a bit to harsh as it does not affect the tests, for the time being I would prefer to log an error  with `System.err.println();` and inform users which representation is going to be used (which one I don't know, first ? last ? found or switchin gto the default one ?)

filiphr: Just logging is fine for me. I just don't know whether it is OK to log to `System.err`. I'll update the PR. 

I am away from normal Internet until next week, I can do it then. I hope it can make it in 2.9.0 / 3.9.0

joel-costigliola: no internet = real holidays :)
Don't worry about missing the release, there are many more things not yet done for the next release, for the logging I would have preferred to use a warning level but that does not exist, I picked `System.err` over `System.out` to make sure users are aware of this.

filiphr: I update the PR. I saw that the `java.util.logging` is used in `Closeables`. Do you think it is OK to use it, or you prefer `System.err`?

Thinking about it, with the logger the users might miss the warning, with `System.err` it will definitely be logged

filiphr:
I also have another idea for this, and the idea is to provide a way for other frameworks that provide their own assertions with the help of assertj-core to plug-in representation for their own objects. 

My idea is that the `StandardRepresentation` can get a list of SPIs and try to find a string for them if it finds it returns immediately, other wise it just goes to the next one. This way you can have dependency bringing in implementations to their own objects. Maybe this should be in a separate issue as an improvement of this

joel-costigliola:
@filiphr let's move that to another issue

src/main/java/org/assertj/core/configuration/ConfigurationProvider.java
joel-costigliola: I would go with `System.err` to be sure it is not missed.

src/main/java/org/assertj/core/util/diff/Delta.java
joel-costigliola: why can't you use `CONFIGURATION_PROVIDER.representation()` here exactly ?

src/main/java/org/assertj/core/configuration/ConfigurationProvider.java
joel-costigliola: I'm addressing this one

src/main/java/org/assertj/core/util/diff/Delta.java
filiphr: Because the method is part of the `StandardRepresentation` API and not part of the `Representation`. If it is OK for you we can use `CONFIGURATION_PROVIDER.representation().toStringOf(lines)` then I can change that as well and deprecate the `StandardRepresentation.STANDARD_REPRESENTATION` and remove it in the version after this one.

joel-costigliola: Thanks for the explanation, I'll see if it makes sense to move `toStringOf(lines)` in the interface, gut feeling is we should not but I'll think about it. 

I was registering a `Representation` in assertj-examples, it worked but we really need to have some doc on how to do this, I will update the javadoc for `Assertion.useStandardRepresentation()` as it does not necessarily switch to `StandardRepresentation` anymore.

joel-costigliola: I moved the `toString` logic from the `Delta` and its subclasses in `StandardRepresentation`, now `Delta.toString()` is simply a call to `ConfigurationProvider.CONFIGURATION_PROVIDER.representation().toStringOf(this);`

joel-costigliola:
Thanks a lot @filiphr ! I finally closed the PR after doing a bit of refactoring and documentation. Cheers !

filiphr:
Thanks for accepting the PR @joel-costigliola, I think that I saw some types and some thinks that I don't agree (I think they are misleading) in the documentation. I'll provide a small PR that will change the documentation. Also currently there is nothing in the examples, would you like me to create one? I can also write a small highlight feature for the website as well (is it also on GitHub?)

Great job for the deltas. And thanks for cleaning up some of the things I did 😄 

Also I saw that you left the `public static final` field in the `StandardRepresentation`. Don't you want it deprecated and removed later on? It has a public constructor, so if someone really needs to use it then they can instantiate it themselves

filiphr:
I just saw that you are using the `public static final` field in the `ConfigurationProvider`, I was creating an instance because I wanted to get rid of the field. In any case as you wish.

I also saw that I forgot to make the `ConfigurationProvider` constructor private. Do you think that makes sense?

joel-costigliola:
I'm happy to discuss documentation enhancement, looking forward to your PR ! :+1: 

I added an example is in [`java-7-with-snapshots`](https://github.com/joel-costigliola/assertj-examples/tree/java-7-with-snapshots) branch here https://github.com/joel-costigliola/assertj-examples/commit/f20e09ddd5ae2b206eb11dbc9201dbddb916f851.

If you can a section in http://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html that would be great.

I'm using `STANDARD_REPRESENTATION` in assertj-examples, I can remove it but keeping it is not a big deal as it is stateless, I can go either way.

I had to make the `CONFIGURATION_PROVIDER` to reset the representation to the default one in `Assertions.useDefaultRepresentation()`.

I agree `ConfigurationProvider` should not be public.

filiphr:
The example looks great 😄 .

I'll try to wrap something about the `Representation` in the documentation. 

Regarding the `STANDARD_REPRESENTATION`, my goal was just not to encourage users to use it, as there is a better way of doing it now 😄 , and there is one less thing to worry about the the AssertJ public API (if you remove it).

 >I had to make the CONFIGURATION_PROVIDER to reset the representation to the default one in Assertions.useDefaultRepresentation().

Here I meant that you are using the `public static final STANDARD_REPRESENTATION ` in the `ConfigurationProvider`, not about the `ConfigurationProvider` instance.

PR #1032, opened 2017-07-12
Introducing assumptions (draft, #975)

### LINKED ISSUE #975: Implement assumeThat() static method to replace JUnit implementation
#### Summary
I'd like to use all of the benefits of fluent assertJ style assertThat() method with assumeThat() method, with as it's couterpart in JUnit would through AssumptionViolatedException and allow to skip test execution on failed assumptions.

The reason is that I use that kind of assumptions pretty heavily and don't want to depend on JUnit specifically. There is great power in AssertJ already, and I'd like it to be available not just for assertions, but also for the assumptions.

Looks like this work was already initiated at `org.assertj.swing.junit.util.Assumptions.assumeThat`, but as for me it's proper place is in assertj-core library.

#### Example

```java
assumeThat(someString).isNotBlank();
```

#### Java 8 specific ?

fiery-phoenix:
I have added implementation of `assumeThat()` for string and general object types. If this draft looks ok, can work at adding other assumption methods.

#### Check List:
* Fixes #975
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/Assumptions.java
+++ b/src/main/java/org/assertj/core/api/Assumptions.java
@@ -45,6 +45,7 @@ import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import java.util.concurrent.atomic.AtomicStampedReference;
 
 import static org.assertj.core.api.Assertions.catchThrowable;
+import static org.assertj.core.util.Arrays.array;
 
 /**
  * Entry point for assumption methods for different types, which allow to skip test execution on failed assumptions.
@@ -62,7 +63,7 @@ public class Assumptions {
   @CheckReturnValue
   @SuppressWarnings("unchecked")
   public static <T> AbstractObjectAssert<?, T> assumeThat(T actual) {
-    return enhanceAsAssumption(ObjectAssert.class, Object.class, actual);
+    return asAssumption(ObjectAssert.class, Object.class, actual);
   }
 
   /**
@@ -73,7 +74,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractCharSequenceAssert<?, String> assumeThat(String actual) {
-    return enhanceAsAssumption(StringAssert.class, String.class, actual);
+    return asAssumption(StringAssert.class, String.class, actual);
   }
 
   /**
@@ -84,7 +85,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractBigDecimalAssert<?> assumeThat(BigDecimal actual) {
-    return enhanceAsAssumption(BigDecimalAssert.class, BigDecimal.class, actual);
+    return asAssumption(BigDecimalAssert.class, BigDecimal.class, actual);
   }
 
   /**
@@ -95,7 +96,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractBigIntegerAssert<?> assumeThat(BigInteger actual) {
-    return enhanceAsAssumption(BigIntegerAssert.class, BigInteger.class, actual);
+    return asAssumption(BigIntegerAssert.class, BigInteger.class, actual);
   }
 
   /**
@@ -106,7 +107,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractUriAssert<?> assumeThat(URI actual) {
-    return enhanceAsAssumption(UriAssert.class, URI.class, actual);
+    return asAssumption(UriAssert.class, URI.class, actual);
   }
 
   /**
@@ -117,7 +118,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractUrlAssert<?> assumeThat(URL actual) {
-    return enhanceAsAssumption(UrlAssert.class, URL.class, actual);
+    return asAssumption(UrlAssert.class, URL.class, actual);
   }
 
   /**
@@ -128,7 +129,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractBooleanAssert<?> assumeThat(boolean actual) {
-    return enhanceAsAssumption(BooleanAssert.class, Boolean.class, actual);
+    return asAssumption(BooleanAssert.class, Boolean.class, actual);
   }
 
   /**
@@ -139,7 +140,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractBooleanAssert<?> assumeThat(Boolean actual) {
-    return enhanceAsAssumption(BooleanAssert.class, Boolean.class, actual);
+    return asAssumption(BooleanAssert.class, Boolean.class, actual);
   }
 
   /**
@@ -150,7 +151,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractBooleanArrayAssert<?> assumeThat(boolean[] actual) {
-    return enhanceAsAssumption(BooleanArrayAssert.class, boolean[].class, actual);
+    return asAssumption(BooleanArrayAssert.class, boolean[].class, actual);
   }
 
   /**
@@ -161,7 +162,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractByteAssert<?> assumeThat(byte actual) {
-    return enhanceAsAssumption(ByteAssert.class, Byte.class, actual);
+    return asAssumption(ByteAssert.class, Byte.class, actual);
   }
 
   /**
@@ -172,7 +173,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractByteAssert<?> assumeThat(Byte actual) {
-    return enhanceAsAssumption(ByteAssert.class, Byte.class, actual);
+    return asAssumption(ByteAssert.class, Byte.class, actual);
   }
 
   /**
@@ -183,7 +184,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractByteArrayAssert<?> assumeThat(byte[] actual) {
-    return enhanceAsAssumption(ByteArrayAssert.class, byte[].class, actual);
+    return asAssumption(ByteArrayAssert.class, byte[].class, actual);
   }
 
   /**
@@ -194,7 +195,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractCharacterAssert<?> assumeThat(char actual) {
-    return enhanceAsAssumption(CharacterAssert.class, Character.class, actual);
+    return asAssumption(CharacterAssert.class, Character.class, actual);
   }
 
   /**
@@ -205,7 +206,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractCharacterAssert<?> assumeThat(Character actual) {
-    return enhanceAsAssumption(CharacterAssert.class, Character.class, actual);
+    return asAssumption(CharacterAssert.class, Character.class, actual);
   }
 
   /**
@@ -216,7 +217,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractCharArrayAssert<?> assumeThat(char[] actual) {
-    return enhanceAsAssumption(CharArrayAssert.class, char[].class, actual);
+    return asAssumption(CharArrayAssert.class, char[].class, actual);
   }
 
   /**
@@ -227,7 +228,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractCharSequenceAssert<?, ? extends CharSequence> assumeThat(CharSequence actual) {
-    return enhanceAsAssumption(CharSequenceAssert.class, CharSequence.class, actual);
+    return asAssumption(CharSequenceAssert.class, CharSequence.class, actual);
   }
 
   /**
@@ -238,7 +239,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractShortAssert<?> assumeThat(short actual) {
-    return enhanceAsAssumption(ShortAssert.class, Short.class, actual);
+    return asAssumption(ShortAssert.class, Short.class, actual);
   }
 
   /**
@@ -249,7 +250,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractShortAssert<?> assumeThat(Short actual) {
-    return enhanceAsAssumption(ShortAssert.class, Short.class, actual);
+    return asAssumption(ShortAssert.class, Short.class, actual);
   }
 
   /**
@@ -260,7 +261,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractShortArrayAssert<?> assumeThat(short[] actual) {
-    return enhanceAsAssumption(ShortArrayAssert.class, short[].class, actual);
+    return asAssumption(ShortArrayAssert.class, short[].class, actual);
   }
 
   /**
@@ -271,7 +272,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractIntegerAssert<?> assumeThat(int actual) {
-    return enhanceAsAssumption(IntegerAssert.class, Integer.class, actual);
+    return asAssumption(IntegerAssert.class, Integer.class, actual);
   }
 
   /**
@@ -282,7 +283,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractIntegerAssert<?> assumeThat(Integer actual) {
-    return enhanceAsAssumption(IntegerAssert.class, Integer.class, actual);
+    return asAssumption(IntegerAssert.class, Integer.class, actual);
   }
 
   /**
@@ -293,7 +294,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractIntArrayAssert<?> assumeThat(int[] actual) {
-    return enhanceAsAssumption(IntArrayAssert.class, int[].class, actual);
+    return asAssumption(IntArrayAssert.class, int[].class, actual);
   }
 
   /**
@@ -304,7 +305,7 @@ public class Assumptions {
    */
   @CheckReturnValue
   public static AbstractLongAssert<?> assumeThat(long actual) {
-    return enhanceAsAssumption(LongAssert.class, Long.class, actual);
+    return asAssumption(LongAssert.class, Long.class, actual);
   }
 
   /**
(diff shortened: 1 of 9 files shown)


src/main/java/org/assertj/core/api/Assumptions.java
```diff
@@ -0,0 +1,84 @@
...
+      public Object intercept(Object o, Method method, Object[] args, MethodProxy methodProxy) throws Throwable {
+        try {
+          Object result = methodProxy.invokeSuper(o, args);
+          if (result != o && result instanceof AbstractAssert) {
+            return enhanceAsAssumption(result);
+          }
+          return result;
+        } catch (AssertionError e) {
+          throw (RuntimeException) Class.forName("org.junit.AssumptionViolatedException")
+                                        .getConstructor(String.class, Throwable.class).newInstance("", e);
```
fiery-phoenix: Here I use reflection to create JUnit's `AssumptionViolatedException`. Probably it is an overdo, as JUnit dependency scope is `provided` at pom.xml. But as it was requested to avoid hard dependency on JUnit, implemented it this way.

src/main/java/org/assertj/core/api/Assumptions.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+      }
+    });
+
+    return (ASSERTION) enhancer.create(new Class[] { actualType }, new Object[] { actual });
+  }
+
+  private static Object enhanceAsAssumption(Object assertion) {
```
fiery-phoenix: Don't like this method implementation, used for the case when assertion returns new assertion, but did not find better implementation when using cglib. Ideas are welcome.

src/main/java/org/assertj/core/api/Assumptions.java
joel-costigliola: we already do a similar thing [to build a `org.junit.ComparisonFailure`](https://github.com/joel-costigliola/assertj-core/blob/21fe1d50f2ddb23f8b22cdb0c886099a489db77b/src/main/java/org/assertj/core/error/ShouldBeEqual.java#L179) instead of a basic `AssertionError`.

joel-costigliola: Not sure if it is actually used somewhere but it would be good to pass the assertion error message as the AssumptionViolatedException's message, maybe prefixed by `"assumption failed due to " + assertionErrorMessage `

src/main/java/org/assertj/core/api/Assumptions.java
```diff
@@ -0,0 +1,84 @@
...
+    if (assertion instanceof StringAssert) {
+      return enhanceAsAssumption(StringAssert.class, String.class, ((StringAssert) assertion).actual);
+    }
+    if (assertion instanceof ListAssert) {
+      return enhanceAsAssumption(ListAssert.class, List.class, ((ListAssert) assertion).actual);
+    }
+
+    throw new IllegalArgumentException("Unsupported assumption cretion for " + assertion.getClass());
```
joel-costigliola: typo `creation`

src/test/java/org/assertj/core/api/assumptions/Assumptions_assumeThat_Object_Test.java
```diff
@@ -0,0 +1,54 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assumptions.assumeThat;
+
+public class Assumptions_assumeThat_Object_Test {
+
+  private static int ranTests = 0;
+
+  @Test
+  public void should_ignore_test_when_assumption_fails() {
+    assumeThat((Object) "test").isNotNull().isEqualTo("other");
```
joel-costigliola: I would prefer to have a `Object objectUnderTest` field that is initialized in each test to avoid inelegant casts.

src/main/java/org/assertj/core/api/Assumptions.java
joel-costigliola: I agree that is not great ... but no idea yet on how to improve that (my gut feeling is you can't)

src/main/java/org/assertj/core/api/Assumptions.java
PascalSchumacher: I'm also not aware of any place where the error message is displayed, but at least it makes debugging easier.

src/main/java/org/assertj/core/api/Assumptions.java
fiery-phoenix: fixed

src/test/java/org/assertj/core/api/assumptions/Assumptions_assumeThat_Object_Test.java
fiery-phoenix: changed

src/main/java/org/assertj/core/api/Assumptions.java
fiery-phoenix: Assertion error message is present in passed to constructor throwable.

PascalSchumacher: Sorry, I missed that. By the way: It is great that you are working on this feature.

src/main/java/org/assertj/core/api/Assumptions.java
```diff
@@ -0,0 +1,86 @@
...
+      @Override
+      public Object intercept(Object o, Method method, Object[] args, MethodProxy methodProxy) throws Throwable {
+        try {
+          Object result = methodProxy.invokeSuper(o, args);
+          if (result != o && result instanceof AbstractAssert) {
+            return enhanceAsAssumption(result);
+          }
+          return result;
+        } catch (AssertionError e) {
+          throw (RuntimeException) Class.forName("org.junit.AssumptionViolatedException")
```
PascalSchumacher: Maybe we should catch the `ClassNoFoundException` and ask the user to add junit 4+ to the classpath in a custom exception message?

joel-costigliola: We should also try to throw a [`SkipException`](https://jitpack.io/com/github/cbeust/testng/master-6.11-gd61e30c-133/javadoc/org/testng/SkipException.html) for TestNG users, then as @PascalSchumacher  suggested warn the user that either JUnit or TestNG must be in the classpath for assumptions to work.

fiery-phoenix: Good idea, done

src/main/java/org/assertj/core/api/Assumptions.java
fiery-phoenix: :)

fiery-phoenix:
I didn't add assumptions for next assertions:
- `assertThat(Iterable<? extends ELEMENT>, AssertFactory<ELEMENT,ELEMENT_ASSERT>) `
- `assertThat(List<? extends ELEMENT>, AssertFactory<ELEMENT,ELEMENT_ASSERT>) `
- `assertThat(List<? extends ELEMENT>, Class<ELEMENT_ASSERT>) `
- `assertThat(T)  // <T extends Comparable<? super T>> `
- `assertThat(T)  // <T extends AssertDelegateTarget> `
- `assertThat(ACTUAL, java.lang.Class<ELEMENT_ASSERT>) `
- `assertThat(AssertProvider<T>)`
Please let me know if any of them is critical.

src/test/java/org/assertj/core/api/assumptions/Assumptions_assumeThat_Test.java
```diff
@@ -0,0 +1,354 @@
...
+import static org.assertj.core.api.Assumptions.assumeThatThrownBy;
+import static org.assertj.core.util.Arrays.array;
+import static org.mockito.Mockito.mock;
+
+@RunWith(Parameterized.class)
+public class Assumptions_assumeThat_Test {
+
+  private static int ranTests = 0;
+
+  private AssumptionRunner assumptionRunner;
```
joel-costigliola: Fix warning by generifying with `<?>`

src/main/java/org/assertj/core/api/Assumptions.java
```diff
@@ -0,0 +1,757 @@
...
+        throw new IllegalStateException("Assumptions require JUnit or TestNG in the classpath");
+      }
+    }
+  }
+
+  private static RuntimeException createAssumptionViolatedException(Class<?> exceptionClass,
+                                                                    AssertionError e) throws ReflectiveOperationException {
+    return (RuntimeException) exceptionClass.getConstructor(String.class, Throwable.class)
+                                            .newInstance("assumption failed due to assertion error", e);
+  }
```
joel-costigliola: keep the assertion error message to explain why the assumption was not met, I'm not sure we should mention assertion here as it is not about asserting but assuming, maybe something like `String.format("assumption was not met, error was:%n%s", e.getMessage())`

src/test/java/org/assertj/core/api/assumptions/Assumptions_assumeThat_Test.java
```diff
@@ -0,0 +1,354 @@
...
+
+  @Test
+  public void should_run_test_when_assumption_passes() {
+    assumptionRunner.runPassingAssumption();
+    ranTests++;
+  }
+
+  @AfterClass
+  public static void afterClass() {
+    assertThat(ranTests).isEqualTo(provideAssumptionsRunners().length);
```
joel-costigliola: add `.as("number of tests run")`, I was running a single test that succeeded but the junit report showed : `org.junit.ComparisonFailure: expected:<[1]> but was:<[0]>` which was really confusing until I realized the assertion here ;-)
also move this method just before the tests method, it will make it easier to understand the test behavior.

src/main/java/org/assertj/core/api/Assumptions.java
```diff
@@ -0,0 +1,757 @@
...
+        } catch (AssertionError e) {
+          throw createAssumptionViolatedException(e);
+        }
+      }
+    });
+
+    return (ASSERTION) enhancer.create(new Class[] { actualType }, new Object[] { actual });
+  }
+
+  private static RuntimeException createAssumptionViolatedException(AssertionError e) throws ReflectiveOperationException {
```
joel-costigliola: would be good to find a more accurate name as we can build a `SkipException`, maybe `failingAssumption` or `unmetAssumption` (not sure this is correct english) or `assumptionNotMet`, I'm not ending with `Exception` because it is clear it is one since it comes after `throw `, I think `throw assumptionNotMet(e);` reads well, thoughts ?

src/main/java/org/assertj/core/api/Assumptions.java
joel-costigliola: that also allows users to give a description of their assumption through `as()` methods although I wonder f we should introduce an assumption description separated from the assertion description, I'd like to hear your thoughts on that.

src/main/java/org/assertj/core/api/Assumptions.java
```diff
@@ -0,0 +1,757 @@
...
+          }
+          return result;
+        } catch (AssertionError e) {
+          throw createAssumptionViolatedException(e);
+        }
+      }
+    });
+
+    return (ASSERTION) enhancer.create(new Class[] { actualType }, new Object[] { actual });
+  }
```
joel-costigliola: can use `org.assertj.core.util.Arrays.array` static import to write the lighter: `enhancer.create(array(actualType), array(actual))`

src/main/java/org/assertj/core/api/Assumptions.java
```diff
@@ -0,0 +1,757 @@
...
+
+  @SuppressWarnings("unchecked")
+  private static <ASSERTION, ACTUAL> ASSERTION enhanceAsAssumption(Class<ASSERTION> assertionType,
+                                                                   Class<ACTUAL> actualType,
+                                                                   Object actual) {
+    Enhancer enhancer = new Enhancer();
+    enhancer.setSuperclass(assertionType);
+    enhancer.setCallback(new MethodInterceptor() {
+      @Override
+      public Object intercept(Object o, Method method, Object[] args, MethodProxy methodProxy) throws Throwable {
```
joel-costigliola: rename `o` to `assertion` to make the code easier to understand

src/main/java/org/assertj/core/api/Assumptions.java
```diff
@@ -0,0 +1,757 @@
...
+   * @param shouldRaiseThrowable The {@link ThrowingCallable} or lambda with the code that should raise the throwable.
+   * @return the created assumption for assertion object.
+   */
+  @CheckReturnValue
+  public static AbstractThrowableAssert<?, ? extends Throwable> assumeThatThrownBy(ThrowingCallable shouldRaiseThrowable) {
+    return enhanceAsAssumption(ThrowableAssert.class, Throwable.class, catchThrowable(shouldRaiseThrowable));
+  }
+
+  @SuppressWarnings("unchecked")
+  private static <ASSERTION, ACTUAL> ASSERTION enhanceAsAssumption(Class<ASSERTION> assertionType,
```
joel-costigliola: I'd like to rename `enhanceAsAssumption`, IMO it's not really an enhancement since we change the behavior of the called code, I was thinking of the simple `asAssumption` or `wrapAsAssumption`.

src/main/java/org/assertj/core/api/Assumptions.java
```diff
@@ -0,0 +1,757 @@
...
+import java.util.concurrent.atomic.AtomicStampedReference;
+
+import static org.assertj.core.api.Assertions.catchThrowable;
+
+/**
+ * Entry point for assumption methods for different types, which allow to skip test execution on failed assumptions.
+ * @since 2.9.0
+ */
+public class Assumptions {
+
```
joel-costigliola: would be great to have a test that makes sure we haven't forgotten to wrap all `assertThat` methods to `assumeThat`, it should be rather easy to write an `Assertions_sync_with_Assumptions_Test`  test inspired by `Assertions_sync_assertThat_with_BDD_and_Soft_variants_Test`.

joel-costigliola: If we add this test (I think we should for API consistency) it means the we need to add the assertions you mentioned as assumptions:
- `assertThat(Iterable<? extends ELEMENT>, AssertFactory<ELEMENT,ELEMENT_ASSERT>) `
- `assertThat(List<? extends ELEMENT>, AssertFactory<ELEMENT,ELEMENT_ASSERT>) `
- `assertThat(List<? extends ELEMENT>, Class<ELEMENT_ASSERT>) `
- `assertThat(T)  // <T extends Comparable<? super T>> `
- `assertThat(T)  // <T extends AssertDelegateTarget> `
- `assertThat(ACTUAL, java.lang.Class<ELEMENT_ASSERT>) `
- `assertThat(AssertProvider<T>)`

src/test/java/org/assertj/core/api/assumptions/Assumptions_assumeThat_Test.java
fiery-phoenix: done

src/test/java/org/assertj/core/api/assumptions/Assumptions_assumeThat_Test.java
fiery-phoenix: done (had corresponding check turned off)

src/main/java/org/assertj/core/api/Assumptions.java
fiery-phoenix: done

src/main/java/org/assertj/core/api/Assumptions.java
fiery-phoenix: done

src/main/java/org/assertj/core/api/Assumptions.java
fiery-phoenix: renamed to `asAssumption`.

src/main/java/org/assertj/core/api/Assumptions.java
fiery-phoenix: Agree, `throw assumptionNotMet(e);` looks good.

src/main/java/org/assertj/core/api/Assumptions.java
fiery-phoenix: Have a problem with implementing this one. Don't know if it is only on my computer, but when I write smth like:

`return (RuntimeException) exceptionClass.getConstructor(String.class, Throwable.class).newInstance("assumption was not met due to " + e.getMessage(), e);
`
Error message doesn't get concatenated, as if it was empty. Though it gets concatenated when I print it via `System.out`. As it looks weird, added screenshot.

screenshot

The only case when I can get it working as expected is:

`exceptionClass.getConstructor(String.class) .newInstance("assumption was not met due to: " + e.getMessage() + "\n")`
(removed Throwable argument and added empty line)

Could you please check how it works for you?

joel-costigliola: weird, it works for me with:
```java
private static RuntimeException assumptionNotMet(Class<?> exceptionClass,
                                                 AssertionError e) throws ReflectiveOperationException {
  return (RuntimeException) exceptionClass.getConstructor(String.class, Throwable.class)
                                          .newInstance("assumption was not met due to: " + e.getMessage(), e);
}
```
this test passes correctly:
```java
@Test
public void should_describe_the_reason_why_the_assumption_was_not_met() {
  try {
    assumeThat(STRING_OBJECT).isNotNull().isEqualTo("other");
  } catch (Exception e) {
    assertThat(e).hasMessage("assumption was not met due to: expected:<\"[other]\"> but was:<\"[test]\">");
    return;
  }
  fail("assumption should not have been met.");
}
```

fiery-phoenix: Message doesn't get truncated when running tests with maven, so it does look like Idea problem.
Is it ok to have error message duplicated in stacktrace, or I should pass only error message to not met assumption exception, or just leave it as part of stacktrace only?

joel-costigliola: I don't see any harm keeping the assertion error as the cause of the unmet assumption since it is really the cause of of it.

src/main/java/org/assertj/core/api/Assumptions.java
fiery-phoenix: Added the test. It was required to add only:

`assertThat(T)  // <T extends Comparable<? super T>>`

joel-costigliola:
I have integrated this PR, thanks a lot for the good work @fiery-phoenix, appreciated.

fiery-phoenix:
You are welcome.

PR #1033, opened 2017-07-13
Use method generic return type and generic parameters for testing the…

### LINKED ISSUE #1005: [WithAssertions] ListAssert is broken due to generics fluff
#### Summary

#775 fixed some discrepancies between `WithAssertions` and `Assertions` (shipped in 3.6.0). Unfortunately,   I'm afraid that I got some generics wrong leading unwanted regression 😞 

#### Example

```java
public class RemoveMe implements WithAssertions {

  @Test
  public void fixme() {
    List<String> l = new ArrayList<>();
    assertThat(l).contains("foo");
  }
}
```

Is broken since 3.6.0.

```
Error:(213, 18) java: no suitable method found for contains(java.lang.String)
    method org.assertj.core.api.AbstractIterableAssert.contains(capture#3 of ? extends java.lang.String...) is not applicable
      (varargs mismatch; java.lang.String cannot be converted to capture#3 of ? extends java.lang.String)
    method org.assertj.core.api.ListAssert.contains(capture#3 of ? extends java.lang.String...) is not applicable
      (varargs mismatch; java.lang.String cannot be converted to capture#3 of ? extends java.lang.String)
```

This is related to the addition of the following method:

```
default <T> ListAssert<? extends T> assertThat(final List<? extends T> actual)
```

I think that return type should be `ListAssert<T>`, or `ListAssert<? super T>`, rather than `ListAssert<? extends T>`.    

Do you want me to open a PR to fix it? Should we review the API to spot such mistakes?

#### Java 8 specific ?

filiphr:
#### Check List:
* Partially Fixes #1005 (A PR is needed for the master as well)
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
+++ b/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
@@ -23,6 +23,7 @@ import static org.assertj.core.util.Strings.quote;
 
 import java.io.File;
 import java.lang.reflect.Array;
+import java.lang.reflect.Method;
 import java.text.SimpleDateFormat;
 import java.util.Calendar;
 import java.util.Collection;
@@ -138,6 +139,7 @@ public class StandardRepresentation implements Representation {
     if (object instanceof Map<?, ?>) return toStringOf((Map<?, ?>) object);
     if (object instanceof Tuple) return toStringOf((Tuple) object);
     if (object instanceof MapEntry) return toStringOf((MapEntry<?, ?>) object);
+    if (object instanceof Method) { return ((Method) object).toGenericString(); }
     return object == null ? null : object.toString();
   }
 
--- a/src/test/java/org/assertj/core/api/BaseAssertionsTest.java
+++ b/src/test/java/org/assertj/core/api/BaseAssertionsTest.java
@@ -12,11 +12,13 @@
  */
 package org.assertj.core.api;
 
+import static org.assertj.core.test.TypeCanonizer.canonize;
 import static org.assertj.core.util.Arrays.array;
 import static org.assertj.core.util.Sets.newLinkedHashSet;
 
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
@@ -67,15 +69,16 @@ public class BaseAssertionsTest {
         // static vs not static is not important Soft vs Not Soft assertions
         boolean equal = (ACCESS_MODIFIERS & method1.getModifiers() & method2.getModifiers()) != 0;
         equal = equal && (ignoreDeclaringClass || method1.getDeclaringClass().equals(method2.getDeclaringClass()));
-        equal = equal && (ignoreReturnType || method1.getReturnType().equals(method2.getReturnType()));
+        equal = equal && (ignoreReturnType || canonize(method1.getGenericReturnType())
+          .equals(canonize(method2.getGenericReturnType())));
         equal = equal && (ignoreMethodName || method1.getName().equals(method2.getName()));
 
-        Class<?>[] pTypes1 = method1.getParameterTypes();
-        Class<?>[] pTypes2 = method2.getParameterTypes();
+        Type[] pTypes1 = method1.getGenericParameterTypes();
+        Type[] pTypes2 = method2.getGenericParameterTypes();
         equal = equal && pTypes1.length == pTypes2.length;
         if (equal) {
           for (int i = 0; i < pTypes1.length; i++) {
-            equal = equal && pTypes1[i].equals(pTypes2[i]);
+            equal = equal && canonize(pTypes1[i]).equals(canonize(pTypes2[i]));
           }
         }
         return equal ? 0 : 1;
--- a/src/test/java/org/assertj/core/api/BaseAssertionsTest.java
+++ b/src/test/java/org/assertj/core/api/BaseAssertionsTest.java
@@ -27,16 +27,14 @@ import java.util.Set;
 /**
  * @author Filip Hrisafov
  */
-public class BaseAssertionsTest {
+public abstract class BaseAssertionsTest {
 
   private static final int ACCESS_MODIFIERS = Modifier.PUBLIC | Modifier.PROTECTED | Modifier.PRIVATE;
 
-  protected static final Comparator<Method> IGNORING_DECLARING_CLASS_AND_METHOD_NAME = internalMethodComparator(true,
-                                                                                                                false,
+  protected static final Comparator<Method> IGNORING_DECLARING_CLASS_AND_METHOD_NAME = internalMethodComparator(false,
                                                                                                                 true);
 
   protected static final Comparator<Method> IGNORING_DECLARING_CLASS_AND_RETURN_TYPE = internalMethodComparator(true,
-                                                                                                                true,
                                                                                                                 false);
 
   // Object is ignored because of the AssertProvider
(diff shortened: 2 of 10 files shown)


src/main/java/org/assertj/core/presentation/StandardRepresentation.java
```diff
@@ -138,6 +139,7 @@ public String toStringOf(Object object) {
     if (object instanceof Map<?, ?>) return toStringOf((Map<?, ?>) object);
     if (object instanceof Tuple) return toStringOf((Tuple) object);
     if (object instanceof MapEntry) return toStringOf((MapEntry<?, ?>) object);
+    if (object instanceof Method) { return ((Method) object).toGenericString(); }
```
filiphr: I am not sure if this makes sense to be part of this PR or not. Does it even make sense to be included in AssertJ.

This one helped me a lot to figure out what exactly was the problem in the failing tests.

src/test/java/org/assertj/core/presentation/StandardRepresentation_toStringOf_Test.java
```diff
@@ -12,8 +12,6 @@
  */
 package org.assertj.core.presentation;
 
-import static java.lang.String.format;
```
filiphr: I usually try to avoid commits with unnecessary imports moves, but this one sneaked on me...

src/test/java/org/assertj/core/test/TypeCanonizer.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.Queue;
+import java.util.Set;
+
+/**
+ * A Type canonizar that helps with the resolving of a {@link Type} so it can be compared to a similar one
+ * considering generics.
+ * @author Filip Hrisafov
+ * @author Clement Mathieu
+ */
+public class TypeCanonizer {
```
filiphr: I am not sure if the other AssertJ projects need something like this. If they do then maybe we should add this to the parent pom so everyone can use it

src/main/java/org/assertj/core/presentation/StandardRepresentation.java
joel-costigliola: that a sufficient reason to me, especially if ones perform assertions on methods

src/test/java/org/assertj/core/presentation/StandardRepresentation_toStringOf_Test.java
joel-costigliola: I don't focus too much on imports reorganization, I let the IDE handle them, the only pain is when comparing commits but I can live with that.

src/test/java/org/assertj/core/api/BaseAssertionsTest.java
```diff
@@ -67,15 +69,16 @@ public int compare(Method method1, Method method2) {
         // static vs not static is not important Soft vs Not Soft assertions
         boolean equal = (ACCESS_MODIFIERS & method1.getModifiers() & method2.getModifiers()) != 0;
         equal = equal && (ignoreDeclaringClass || method1.getDeclaringClass().equals(method2.getDeclaringClass()));
-        equal = equal && (ignoreReturnType || method1.getReturnType().equals(method2.getReturnType()));
+        equal = equal && (ignoreReturnType || canonize(method1.getGenericReturnType())
+          .equals(canonize(method2.getGenericReturnType())));
         equal = equal && (ignoreMethodName || method1.getName().equals(method2.getName()));
```
joel-costigliola: for better readability, I would prefer to extract each return type comparison to a private static method, something like:
```
private boolean sameReturnType(Method method1, Method method2) {
  return canonize(method1.getGenericReturnType()).equals(canonize(method2.getGenericReturnType()));
}
```
I believe the logical condition will be more expressive: `ignoreReturnType || sameReturnType(method1, method2))`.
obviously we should do that we the other conditions.

src/test/java/org/assertj/core/test/TypeCanonizer.java
joel-costigliola: I'm not sure there is a need yet, the other assertions modules don't have bdd or soft assertions, let's keep it here for the time being.

src/test/java/org/assertj/core/test/TypeCanonizer.java
```diff
@@ -0,0 +1,127 @@
...
+
+    for (TypeVariable typeVariable : findAllTypeVariables(initialType)) {
+      typeResolver = typeResolver.where(typeVariable, replacementClassSupplier.get());
+    }
+
+    return typeResolver.resolveType(initialType);
+  }
+
+  private static boolean doesNotNeedCanonization(Type type) {
+    return !(type instanceof ParameterizedType || type instanceof GenericArrayType || type instanceof WildcardType
```
joel-costigliola: format: put `||` on a new line for readability

src/test/java/org/assertj/core/test/TypeCanonizer.java
```diff
@@ -0,0 +1,127 @@
...
+  public static Type canonize(Type initialType) {
+    if (doesNotNeedCanonization(initialType)) {
+      return initialType;
+    }
+
+    ReplacementClassSupplier replacementClassSupplier = new ReplacementClassSupplier();
+    TypeResolver typeResolver = new TypeResolver();
+
+    for (TypeVariable typeVariable : findAllTypeVariables(initialType)) {
+      typeResolver = typeResolver.where(typeVariable, replacementClassSupplier.get());
```
joel-costigliola: It's not clear to me how this works: 
1 - are we resolving type variable with some random classesbut always the sames (the order matters)
2 - is there a logic for choosing the classes in `new ReplacementClassSupplier()`, in any case this needs more comments to explain the strategy.

src/test/java/org/assertj/core/test/TypeCanonizer.java
```diff
@@ -0,0 +1,127 @@
...
+    return typeVariables;
+  }
+
+  /**
+   * Adds all {@code type}'s {@link TypeVariable} to {@code typeVariables}
+   *
+   * @param typeVariables that need to be populated
+   * @param types the types for which the {@link TypeVariable}(s) need to be extracted
+   */
+  private static void findAllTypeVariables(Set<TypeVariable> typeVariables, Type... types) {
```
joel-costigliola: rename to `populateAllTypeVariables`, when I read `find` my brain says "oh, no side effects" ;-)

src/test/java/org/assertj/core/api/BaseAssertionsTest.java
filiphr: Completely agree with you. I'll do that. I will also clean it up a bit. For example `ignoreDeclaringClass` is always `true`.

src/test/java/org/assertj/core/test/TypeCanonizer.java
filiphr: I'll write some more comments there.

In shor here:

src/test/java/org/assertj/core/test/TypeCanonizer.java
filiphr: 😄 you are completely right about this. Will do it

filiphr:
Thanks for the comments @joel-costigliola. I'll update the PR with your feedback. Most probably it will be by the end of this week, as I don't have good access to internet right now 😄

filiphr:
@joel-costigliola I have update the PR with your comments from the review.

joel-costigliola:
Integrated thanks a lot @filiphr  !

filiphr:
>Integrated thanks a lot @filiphr !

No problem.

I will prepare a PR for the mater branch. How should we proceed with the merge? Should I prepare a PR with a merge commit? The reason I am asking is that there might be test failures on the master branch.

joel-costigliola:
Tests should pass if you add the missing methods in `WithAssertions` but if this is too much work, just leave your PR as is, most of the time I rebase PR branches locally, make minor  changes (formatting and stuff) amend the commit and push when I'm happy with the results.

filiphr:
There was no sync test between `Assertions` and `WithAssertions`. PR is coming in few minutes 😄

PR #1038, opened 2017-07-19
Allow overriding collected error list modification

### LINKED ISSUE #1036: AbstractSoftAssertions.errorsCollected() modifies error list
The code in the public method AbstractSoftAssertions.errorsCollected() (https://github.com/joel-costigliola/assertj-core/blob/master/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java#L104) modifies error list by changing their messages, even though the list itself is immutable.

nebehr:
This is a suggestion of how to avoid error list modification for those who doesn't want it, and still retain this functionality for those who do. I am not sure about method names but this is the general idea anyway.

#### Check List:
* Fixes #1036
* Unit tests : NO


joel-costigliola:
@nebehr  Sorry it took so much time, it fell under my radar.

PR #1039, opened 2017-07-19
Custom error processor

nebehr:
Here is another PR with a possible implementation of pluggable and customizable error processing hook which hopefully does not break existing API and does not expose much of the internals. This is basically what I was trying to achieve when encountered issue #1036.

Unfortunately this PR it contains changes from my previous PR #1038 as well. I should have kept them separate.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
@@ -101,7 +101,16 @@ public class AbstractSoftAssertions {
    * @return a copy of list of soft assertions collected errors.
    */
   public List<Throwable> errorsCollected() {
-    return addLineNumberToErrorMessages(proxies.errorsCollected());
+    return decorateErrorsCollected(proxies.errorsCollected());
+  }
+  
+  /**
+   * Modifies collected errors. Override to customize modification.
+   * @param errors list of errors to decorate
+   * @return decorated list
+  */
+  protected List<Throwable> decorateErrorsCollected(List<Throwable> errors) {
+    return addLineNumberToErrorMessages(errors);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


filiphr:
I hope you don't mind if I pitch in a bit. I don't know exactly the use case of the pluggable error processor, do you want it to be for all the tests or just your custom soft assertions. 

An alternative approach could be using an SPI something like what I am proposing in PR #1027. With this one can continue using the default assertions entry points from assertj-core and plug in a service for all the classes in a test suite. However, this can't be used in single tests.

Just throwing an idea out there.

nebehr:
Thanks for the input Fil. In my particular case I am only interested in soft assertions because regular assertion failures can be caught and processed by testing framework itself. Besides, if entry point class for custom soft assertions extends ```AbstractSoftAssertions``` then overridden behaviour should be available for out-of-the-box assertions as well. I do agree that some unified approach would probably be better in the long run, but then it would also require more refactoring and possibly introduce breaking changes.

filiphr:
@nebehr I meant that only soft assertions, not the other ones. The change would be done in a same way with the only exception being that instead of using your own SoftAssertions that need to extend the base ones you can plug it via an SPI. Using an SPI would mean that you can use the assertj alsoft assertions and have the needed functionality. 

In any case, that might overcomplicate things. I just wanted to give an idea. Yours is pretty good as well

joel-costigliola:
@nebehr what about allowing to use your own `SoftProxies` ? 
I believe this could solve your issue and #1040 at the same time.

@filiphr my understanding is that having a SPI is nice if the usage is global but is that relevant if you want to have a different behavior for certain tests ? (thinking out loud I haven't yet played with #1027 to get a clear idea of what it entails).

nebehr:
@joel-costigliola, yes, I believe if there was a way to override SoftProxies instance and then also override its ErrorCollector it would solve both issues. It would expose more internals but I don't think anybody would object.

joel-costigliola:
This is addressed by https://github.com/joel-costigliola/assertj-core/issues/1970 and https://assertj.github.io/doc/#assertj-core-soft-assertions-collect-listener (please comment if that's was not the case).

PR #1049, opened 2017-07-30
Align WithAssertions and Assertions

### LINKED ISSUE #1005: [WithAssertions] ListAssert is broken due to generics fluff

filiphr:
This PR aligns the `WithAssertions` and `Assertions`. Additionally tests are added that will make sure that we keep the API consistent, i.e. `WithAssertions` should have all methods that `Assertions` has.

#### Check List:
* Fixes #1005 
* Unit tests : YES

I was not sure if we need to add tests that make sure `WithAssertions` delegates the calls.

### DIFF
--- a/src/main/java/org/assertj/core/api/WithAssertions.java
+++ b/src/main/java/org/assertj/core/api/WithAssertions.java
@@ -568,7 +568,7 @@ public interface WithAssertions {
    * @since 3.7.0
    */
   @CheckReturnValue
-  default <RESULT> AbstractFutureAssert<?, ? extends Future<? extends RESULT>, RESULT> assertThat(Future<RESULT> actual) {
+  default <RESULT> FutureAssert<RESULT> assertThat(Future<RESULT> actual) {
     return Assertions.assertThat(actual);
   }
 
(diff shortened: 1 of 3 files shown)


filiphr:
What about the `@since` tags? Should we add those to the methods? If yes, should they be the same as in `Assertions` or `2.9.0 / 3.9.0`?

joel-costigliola:
Put `2.9.0 / 3.9.0` in `WithAssertions` since they were not available before.

filiphr:
I added only `@since 3.9.0`, because `WithAssertions` is present only in the master branch

joel-costigliola:
in master, thanks !

PR #1050, opened 2017-08-01
Throw assertion error for assertThat.extracting when actual is null

### LINKED ISSUE #1041: NullPointerException in assertThat.extracting when actual object is null
#### Summary
When using `assertThat` followed by `extracting` and the actual array or collection is null, AssertJ throws a NullPointerException.

```
	at org.assertj.core.groups.FieldsOrPropertiesExtractor.extract(FieldsOrPropertiesExtractor.java:54)
	at org.assertj.core.groups.FieldsOrPropertiesExtractor.extract(FieldsOrPropertiesExtractor.java:44)
	at org.assertj.core.api.AbstractObjectArrayAssert.extracting(AbstractObjectArrayAssert.java:1225)
        ...
```
While I think this is not a major issue because the assertion would fail anyway, it is inconsistent compared to using `assertThat` without extracting. In this case, an AssertionError "Expecting actual not to be null" is raised.

#### Example
```java
  class Foo {
    Object bar;
  }

  @Test
  public void testAssertThatExctractedContains() {
    Foo expectedFoo = new Foo();
    Foo[] actualFoos = null;
    assertThat( actualFoos ).contains( expectedFoo );
  }
```

#### Java 8 specific ?
No

mle-enso:
#### Check List:
* Fixes #1041 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/groups/FieldsOrPropertiesExtractor.java
+++ b/src/main/java/org/assertj/core/groups/FieldsOrPropertiesExtractor.java
@@ -23,40 +23,46 @@ import org.assertj.core.api.iterable.Extractor;
 import org.assertj.core.util.Lists;
 
 /**
- * 
+ *
  * Understands how to retrieve fields or values from a collection/array of objects.
  * <p>
  * You just have to give the field/property name or an {@link Extractor} implementation, a collection/array of objects
  * and it will extract the list of field/values from the given objects.
- * 
+ *
  * @author Joel Costigliola
  * @author Mateusz Haligowski
- * 
+ *
  */
 public class FieldsOrPropertiesExtractor {
-  
   /**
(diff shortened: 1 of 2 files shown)


joel-costigliola:
@mle-enso thanks ! 
Can you recreate on the 2.x branch please ? This change is not java 8 specific so it must be in 2.x that is for java 7 (2.x is regularly merged in master so that master = 2.x + java 8 specific code)

mle-enso:
Ah I see, thanks for the explanation. Wasn't sure about the merge direction of these two branches. Further commit is coming soon.

mle-enso:
Just for documentation, the commit for the 2.x branch now resides in PR #1051.

joel-costigliola:
closed as duplicate of #1051

PR #1051, opened 2017-08-01
#1041 Throw assertion error for assertThat.extracting when actual is …

### LINKED ISSUE #1041: NullPointerException in assertThat.extracting when actual object is null

mle-enso:
…null – further commit for 2.x branch

#### Check List:
* Fixes #1041
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/groups/FieldsOrPropertiesExtractor.java
+++ b/src/main/java/org/assertj/core/groups/FieldsOrPropertiesExtractor.java
@@ -23,40 +23,48 @@ import org.assertj.core.api.iterable.Extractor;
 import org.assertj.core.util.Lists;
 
 /**
- * 
+ *
  * Understands how to retrieve fields or values from a collection/array of objects.
  * <p>
  * You just have to give the field/property name or an {@link Extractor} implementation, a collection/array of objects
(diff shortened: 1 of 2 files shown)


joel-costigliola:
Integrated thanks !

PR #1055, opened 2017-08-12
Refactor primitive array tests

### LINKED ISSUE #1043: Refactor primitive array tests
#### Summary

Primitive array tests can simplified, there is quite a bit of duplication that can be removed. 

#### Example

Let's take `IntArrays_assertContains_Test`, instead of verifying that `IntArrays` delegates to `Arrays` it tests all the possible `assertContains` use cases, these tests  should be moved to `Arrays` tests as it actually implement `assertContains` assertion logic.

The duplication comes from `LongArrays_assertContains_Test`, `FloatArrays_assertContains_Test` that do the same thing testing `assertContains` assertion logic  again and again.

A better approach is the one taken with `IntArrays_assertContainsAnyOf_Test` where the assertion logic tests are written in `Arrays_containsAnyOf_Test`, the only test in `IntArrays_assertContainsAnyOf_Test` being `should_delegate_to_internal_Arrays`

#### Java 8 specific ?

joaodelgado:
#### Check List:
* Fixes #1043 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/internal/Arrays.java
+++ b/src/main/java/org/assertj/core/internal/Arrays.java
@@ -155,7 +155,8 @@ public class Arrays {
     hasSameSizeAsCheck(info, array, other, sizeOf(array));
   }
 
-  void assertContains(AssertionInfo info, Failures failures, Object actual, Object values) {
+  @VisibleForTesting
+  public void assertContains(AssertionInfo info, Failures failures, Object actual, Object values) {
     if (commonChecks(info, actual, values)) return;
     Set<Object> notFound = new LinkedHashSet<>();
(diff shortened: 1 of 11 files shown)


joel-costigliola:
Thanks @joaodelgado for the PR.

There is one missing bit, can you change `ObjectArrays_assertContains_Test` so that is only test it delegates its work to  `Arrays` as in `ObjectArrays_assertContainsAnyOf_Test` ?

joaodelgado:
Sure thing.

I also noticed that boolean, byte, char, double and short arrays could use the same refactor.
I'll work on them too.

joel-costigliola:
Integrated thanks @joaodelgado  !

PR #1058, opened 2017-08-19
Small cleanups

filiphr:
This is some small Javadoc changes and cleanups:

* Add some missing since tags
* Make the ConfigurationProvider constructor private
* Do some rewording and fix some typos around the new SPI Representation

I think that the most important change here is that the `META-INF/services` folder needs to be in `src/test/resources` and not `src/main/resources`. Users usually have a test scoped dependency of AssertJ.

I've also fixed some small typos that I saw and I added the recommended way of writing a custom `Representation`. I think it is much easier for the users if they just extend the `StandardRepresentation` and override `fallbackToStringOf(Object)` method, this way they won't mess with all the defaults from AssertJ.

### DIFF
--- a/src/main/java/org/assertj/core/presentation/Representation.java
+++ b/src/main/java/org/assertj/core/presentation/Representation.java
@@ -28,9 +28,11 @@ import org.assertj.core.api.Assertions;
  * <p>
  * To register a {@link Representation}, you need to do several things:
  * <ul>
- * <li>create a file named in {@code org.assertj.core.presentation.Representation} file in META-INF/services directory</li>
+ * <li>create a file named {@code org.assertj.core.presentation.Representation} file in META-INF/services directory</li>
  * <li>put the fully qualified class name of your {@link Representation} in it</li>   
- * <li>make sure {@code META-INF/services/org.assertj.core.presentation.Representation} is in the runtime classpath, usually putting it in {@code src/main/resources} is enough</li>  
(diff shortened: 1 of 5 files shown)


src/main/java/org/assertj/core/presentation/Representation.java
```diff
@@ -28,9 +28,11 @@
  * <p>
  * To register a {@link Representation}, you need to do several things:
  * <ul>
- * <li>create a file named in {@code org.assertj.core.presentation.Representation} file in META-INF/services directory</li>
+ * <li>create a file named {@code org.assertj.core.presentation.Representation} file in META-INF/services directory</li>
  * <li>put the fully qualified class name of your {@link Representation} in it</li>   
- * <li>make sure {@code META-INF/services/org.assertj.core.presentation.Representation} is in the runtime classpath, usually putting it in {@code src/main/resources} is enough</li>  
+ * <li>make sure {@code META-INF/services/org.assertj.core.presentation.Representation} is in the runtime classpath, usually putting it in {@code src/test/resources} is enough</li>
+ * <li>we recommend that you extend from the {@link StandardRepresentation} and override the
+ * {@link StandardRepresentation#fallbackToStringOf(Object)}. By doing this all the defaults of AssertJ would be applied and you can apply your own customization</li>
```
joel-costigliola: good one, I will override  `fallbackToStringOf` in assertj-examples

joel-costigliola:
Good fixes thanks.

filiphr:
No problem. Glad that you liked them, and sorry for not making them earlier, I was supposed to do them in the initial PR 😄

PR #1060, opened 2017-08-20
Fixed issue #1059

### LINKED ISSUE #1059: Comparing equal binary files fails with java.nio.charset.MalformedInputException
Comparing equal binary files fails with a "java.nio.charset.MalformedInputException: Input length = 1" exception. Method that fails: "FileAssert.hasSameContentAs(java.io.File)".

This is actually caused by a wrong 'fall back' in the 'org.assertj.core.internal.Files' class:
```Java
BinaryDiffResult binaryDiffResult = binaryDiff.diff(actual, readAllBytes(expected.toPath()));
if (binaryDiffResult.hasNoDiff()) {
    // fall back to the RuntimeIOException : not throwing an error is wrong as there was one in the first place.
   throw e;
}
```
In fact this only means the both files are binary qual and it should be:
```Java
BinaryDiffResult binaryDiffResult = binaryDiff.diff(actual, readAllBytes(expected.toPath()));
if (binaryDiffResult.hasNoDiff()) {
   // Both files are binary equal
   return;
}
```
I added a [pull request](https://github.com/joel-costigliola/assertj-core/pull/1060) with the fix and added a test.

michael-schnell:
* Fixes #1059
* Unit tests : YES


michael-schnell:
As discussed in the issue there should be two new methods  "hasSameTextualContentAs" and "hasSameBinaryContentAs" that will replace the deprecated method "hasSameContentAs". Therefore I will cancel this pull request.

PR #1066, opened 2017-09-11
Make URI parameter checks use the raw query.

GrahamDennis:
Previously URI parameters would be parsed after being decoded, which led to some mishandling of escaped '&' and '='.

* Fixes: #1699
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/internal/Uris.java
+++ b/src/main/java/org/assertj/core/internal/Uris.java
@@ -135,7 +135,7 @@ public class Uris {
   public void assertHasParameter(AssertionInfo info, URI actual, String name) {
     assertNotNull(info, actual);
 
-    Map<String, List<String>> parameters = getParameters(actual.getQuery());
+    Map<String, List<String>> parameters = getParameters(actual.getRawQuery());
     if (!parameters.containsKey(name)) throw failures.failure(info, shouldHaveParameter(actual, name));
   }
 
@@ -143,7 +143,7 @@ public class Uris {
                                  String expectedParameterValue) {
     assertNotNull(info, actual);
 
-    Map<String, List<String>> parameters = getParameters(actual.getQuery());
+    Map<String, List<String>> parameters = getParameters(actual.getRawQuery());
 
     if (!parameters.containsKey(expectedParameterName))
       throw failures.failure(info, shouldHaveParameter(actual, expectedParameterName, expectedParameterValue));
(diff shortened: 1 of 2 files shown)


joel-costigliola:

joel-costigliola:
@GrahamDennis can you clarify what you expect with some examples ?

In the test you have added  `http://assertj.org/news?article=abc%26page%3D5` should give only one parameter `article` with value `abc&page=5` while I believe there are two parameters: `article` with value `abc` and `page` with value `5`.

GrahamDennis:
The `&` is escaped, so there is only one parameter, `article` with value `abc&page=5`. Previously the URI query was double-decoded, leading to this being incorrectly treated as having two parameters. 

At least the above is my expectation. I’ve done some searching and I can’t actually find a standard that defines how query string parameters should be extracted. From what I can see the URI RFC leaves the interpretation of query strings as unspecified.... in that case, is there a major implementation of URI parameter parsing that we can refer to as a “standard” and then make the behaviour of assertj whatever that does?

joel-costigliola:
I need a real life example because I find really confusing that in `article=abc%26page%3D5` the value of `article` is `abc%26page%3D5`.

Just thinking out loud but maybe we should provide an `hasRawParameter` assertion that will not decode anything ?

jjlharrison:
I have just hit this problem myself. @GrahamDennis is right, as long as the ampersand is encoded, it doesn't mark the start of a new parameter.

See [IETF RFC 2986](https://tools.ietf.org/html/rfc3986#section-2.2).

Parisfaye:
Hgdy

scordio:
As pointend by @jjlharrison, the RFC mentions:
>    If data for a URI component would conflict with a reserved
>   character's purpose as a delimiter, then the conflicting data must be
>   percent-encoded before the URI is formed.

Also, from [What every web developer must know about URL encoding](https://www.talisman.org/~erlkonig/misc/lunatech%5Ewhat-every-webdev-must-know-about-url-encoding/#Whichcharacterencoding%3F):

> The syntax of the URL is only meaningful before it is URL-decoded: after URL-decoding, reserved characters may appear. 

> Do not expect URI.getPath() to give you structured data

Overall, this seems to be a fair request.

The assertion should be performed without decoding the corresponding component of the URI. This means it would be better to use the raw variants for `assertHasAuthority`, `assertHasFragment`, `assertHasPath`, `assertHasQuery`, `assertHasUserInfo` and all the `assertHasParameter`/`assertHasNoParameter` methods.

@joel-costigliola if you agree, I can extend this PR unless @GrahamDennis is willing to continue on it.

joel-costigliola:
I also agree we should do double decoding.

scordio:
Integrated, thanks @GrahamDennis!

PR #1079, opened 2017-10-01
Add java 9 to master build

PascalSchumacher:
Filtering the "jdk.internal.reflect" classes from the stacktraces is required to fix SoftAssertion error messages. Otherwise e.g. using SoftAssertions#fail will display an ugly error message (see below):

```
[ERROR]   SoftAssertionsTest.should_return_failure_after_fail
Expecting message:
 <"Should not reach here">
but was:
 <"Should not reach here
at NativeMethodAccessorImpl.invoke0(NativeMethodAccessorImpl.java:-2)">
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,5 +9,6 @@ addons:
       - oracle-java8-installer
 
 jdk:  
-   - oraclejdk8
-   - openjdk8
+  - oraclejdk8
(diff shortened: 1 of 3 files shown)


joel-costigliola:
Thanks Pascal.

PR #1082, opened 2017-10-02
Update java docs for containsOnly and containsOnlyElementsOf (#1075)

### LINKED ISSUE #1075: containsOnlyElementsOf method is misleading
#### Summary
containsOnlyElementsOf method is misleading
#### Example

```java
        List<Integer> expectedValues = Arrays.asList(1, 2, 3);
        List<Integer> actualValues = Arrays.asList(1, 2);

        assertThat(actualValues).containsOnlyElementsOf(expectedValues);
```
You expect this to pass since actual values contains only elements of expectedValues values i.e.
actual values is a subset of expected values. But it fails:

java.lang.AssertionError: 
Expecting:
  <[1, 2]>
to contain only:
  <[1, 2, 3]>
  <[3]>

#### Java 8 specific ?

btrajkovski:
#### Check List:
* Fixes #1075
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/ObjectEnumerableAssert.java
+++ b/src/main/java/org/assertj/core/api/ObjectEnumerableAssert.java
@@ -57,6 +57,8 @@ public interface ObjectEnumerableAssert<SELF extends ObjectEnumerableAssert<SELF
   /**
    * Verifies that the actual group contains only the given values and nothing else, in any order.
    * <p>
+   * Use {@link #isSubsetOf(Iterable)} to check that actual is a subset of given iterable
+   * <p>
    * Example :
    * <pre><code class='java'> Iterable&lt;String&gt; abc = newArrayList("a", "b", "c");
    *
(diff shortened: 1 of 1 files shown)


src/main/java/org/assertj/core/api/ObjectEnumerableAssert.java
```diff
@@ -57,6 +57,8 @@
   /**
    * Verifies that the actual group contains only the given values and nothing else, in any order.
    * <p>
+   * Use {@link #isSubsetOf(Object[])} to check if actual is a subset of given iterable
```
joel-costigliola: `to check if` -> `to check that`, the link point to `isSubsetOf` that takes a vararg and not an iterable

joel-costigliola: add an example in the `//assertion will fail` section to show that this assertion fails because of the missing element `"d"`.
```
assertThat(abc).containsOnly("a", "b", "c", "d");
```

joel-costigliola:
Sweet, thanks Bojan!

Issue #928, opened 2017-02-19
Java 9 reflection changes


PascalSchumacher:
Java 9 is much more restrictive concerning reflective access to non-public fields/methods. For example `setAccessible` called by `org.assertj.core.util.introspectionFieldUtils#readField` on a private field fails with:
```
java.lang.reflect.InaccessibleObjectException: Unable to make field private transient long java.util.Date.fastTime accessible: module java.base does not "opens java.util" to unnamed module @66b72664
```

This means that assertj methods which use reflection on non-public fields/methods will fail/behave differently. For example the `isEqualTo*Fields` methods.

PascalSchumacher:
It still works if the user opens all required packages for assertj, e.g. `--add-opens java.base/java.util=ALL-UNNAMED` (ALL-UNNAMED can be replaced with the assertj module name once there is one).

For example `DeepDifference_Test` fails on java 9, but runs with these args:

```
mvn -Dtest=DeepDifference_Test test -DargLine="--add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.math=ALL-UNNAMED"
```

PascalSchumacher:
According to http://mail.openjdk.java.net/pipermail/jigsaw-dev/2017-March/011763.html --permit-illegal-access in a future version of jdk9 will simplify granting access.

joel-costigliola:
If I understand correctly we are going to get screwed in Java 10 ?

PascalSchumacher:
Users who like to use assertj features using reflection on private stuff will have to do some setup to allow assertj access. On java 9 they can use `--permit-illegal-access`, on java 10 they will have to do some more work by opening up individual modules. Does not look to bad to me.

joel-costigliola:
I think this solves the issue : https://www.theregister.co.uk/2017/05/12/oracle_loses_jpms_vote/ ...

sormuras:
Copied comment from https://github.com/junit-team/junit5/issues/1063 by @AlanBateman

> In the case of AssertJ then it looks like it has a re-packaged cglib that is hacking into ClassLoader.defineClass. They need to be steered towards using java.lang.invoke.MethodHandles.Lookup.defineClass instead.

So. If you re-package a cglib version that uses method handles... no screw in Java 10 or 18.3.

PascalSchumacher:
But sadly there is no such version of cglib yet. 

Assertj `3.8.0/2.8.0` already use the latest release of cglib (`3.2.5`).

[...]

PR #1093, opened 2017-10-16
Use byte buddy instead of cglib

filiphr:
#### Check List:
* Related to: #928 
* Unit tests : NA

With this PR I am dropping cglib and adding bytebuddy as a dependency. I still have not removed the shading. One test is failing due to `ClassCastException`, and I am not entirely sure how I can resolve it. Maybe @raphw can help out a bit 😄. The issue is in the `ErrorCollector#intercept` method.

I think that we should not be shading bytebuddy. Their recommendation is that you should shade if you use the exposed ASM, which I am not. Also the other major players such as Mockito are pulling it as a dependency (we could do it as an optional one or provide 2 modules, one with normal assertions and another with soft).

Something that I noticed with bytebuddy is that you can theoretically generate the bytecode and ship that as part of the AssertJ release (that way you have no dependency on anything during runtime)


PascalSchumacher:
Thanks of the pull request. 👍 

The relocation configuration of the shade plugin also has to be updated.

filiphr:
@PascalSchumacher I know about the shading. I first wanted to discuss with you before doing it. As I mentioned an alternative can be that instead of shading we package the generated bytecode by bytebuddy.

PascalSchumacher:
Packaging the generated bytecode is probably preferable if it is easily doable. @joel-costigliola What do you think?

raphw:
Just as a side note: The API you need has been quite stable for many releases, it is not normally necessary to shade Byte Buddy (e.g. Mockito and Hibernate do not shade the library).

filiphr:
I am also more for pulling in the dependency. However, that is up to @joel-costigliola and the team to decide.

For now I removed all the shading that was being done (as only cglib was being shaded).

I also fixed the class cast exception that was happening. The reason for it was the way ByteBuddy works (it generates byte code which contained the cast). 

For cglib the test was passing. However, if one did `softly.assertThat(asList()).first().isEqualTo(something);` there was going to be a `ClassCastException`. Now I am returning `null`, but I am not sure that this is correct.

@raphw I suppose it is not possible to make something like it was before?

joel-costigliola:
I can live with a **really really** stable dependency but any regression like `softly.assertThat(asList()).first().isEqualTo(something);` not working is not acceptable.

filiphr:
@joel-costigliola maybe I was not clear enough. `softly.assertThat(asList()).first().isEqualTo(something);` does not work at the moment with cglib. With ByteBuddy `softly.assertThat(asList()).first()` was throwing the `ClassCastException`. The reason why this is happening is because ByteBuddy generates the byte code and the cast is done in there. cglib works differently.

raphw:
Is there a unit test I can look at? If an interceptor returns an instance from a `@DynamicType` method that does not correspond to the factual type, a class cast exception is thrown. I doubt that is a bug in Byte Buddy, you are most likely returning an incompatible type.

filiphr:
Yes there is. Have a look at [`SoftAssertionsTest#should_propagate_AssertionError_from_nested_proxied_calls`] (https://github.com/joel-costigliola/assertj-core/blob/2.x/src/test/java/org/assertj/core/api/SoftAssertionsTest.java#L571) in branch 2.x.

I didn't mean that it is a bug in ByteBuddy, I was just asking if there is some way to make it work. 

Right now I am intercepting the `@SuperMethod` so I can get ahold of the return type and do the check if it can be assigned, if not return `null`, which I don't like as users will get an NPE if they chain. Is it a good idea to create another proxy in the intercept method that will just throw exception that chaining is not allowed to to an assertion error before? 

The entire idea for the proxies is to swallow the exceptions and allow chaining of assertions.

raphw:
The reason for this exception is an actual bug in assertj. cglib is not aware of generics and only compiles proxy methods to erasures. You can however trigger the same error using the 2.x branch by simply assigning the proxy to its generic type:

```java
ObjectAssert<?> oa = softly.assertThat(asList()).first();
```

Due to cglibs behavior, it causes a heap pollution. Byte Buddy complies generics similar to javac and inserts bridge methods just as the specification requires it. Due to these bridge methods, the class cast exception is already triggered during the method call.

With Java 9, this exception does however always happen. Even with cglib and without the assignment heap pollutiing assignment above. Since Java 9, javac recognizes the chance of a heap pollution and inserts a class cast statement to the type that the generic signature expects which is `ObjectAssert`.

The invalid type chain is:
class org.assertj.core.api.SoftAssertionListAssert
extends org.assertj.core.api.AbstractIterableAssert

where:
AbstractIterableAssert: ELEMENT_ASSERT extends AbstractAssert

The latter assert statement is however bound to `ObjectAssert` in the `SoftAssertionListAssert` which is not a super type of this same type. This is a non-valid generic reference.

filiphr:
Thanks a lot for your detailed explanation @raphw. This means that by introducing ByteBuddy we have the chance to avoid such bugs.

I would propose that for such cases, super call does not return and `this` does not match the return type of the super method, we create a proxy that would throw an illegal argument exception that would just say that the call is not valid because it comes from an invalid assertion (the one before is the culprit).

raphw:
I would suggest to fix these bugs at their root, hopefully not to many of them exist in the library. Fixing all of them would make this proxy mechanism obsolete.

filiphr:
@raphw I am not sure I follow. What would be fixing it at their root?

What do you mean by "this proxy mechanism"? My suggestion or the entire PR?

Sorry if I misunderstood something

raphw:
By fixing the root I mean fixing the generic signature of the assertion class. If I understand it correctly, I think it should return a different type.

By "proxy mechanim" I refer to your suggestion to return an instance that throws an exception. If I understood it correctly and the above generic signature is by mistake, then this mechanism would be obsolete once all such mistakes are fixed.

filiphr:
@raphw that particular case there is by design. What it does it allows the users to convert from an `Iterable` assert into an `Object` or some other more specific assert and do assertions over it. For example in the case I linked the use case is:

```
assertThat(asList("something", "is", "not", "working"))
    .first() // This gives you back ObjectAssert
    .isEqualTo("something");
```

Or
```
assertThat(asList("something", "is", "not", "working"), StringAssert.class).first()
    .startsWith("som") // This gives you back StringAssert
    .endsWith("ing");
```

This allows users to more easily assert the first element, last, or an arbitrary element at some index of an `Iterable`.

raphw:
I see, with this strategy in mind, returning `this` is however never a reliable approach. If this approach is desired, the interceptor should generally check if the return type is applicable as with Java 9, methods are no longer invokable if the return type is not compatible what is a threat to type-safety.

src/main/java/org/assertj/core/api/ErrorCollector.java
```diff
@@ -46,7 +50,12 @@ public Object intercept(Object obj, Method method, Object[] args, MethodProxy pr
       lastResult.setSuccess(false);
       errors.add(e);
     }
-    return result;
+    if (method != null && !method.getReturnType().isInstance(obj)) {
```
filiphr: >If this approach is desired, the interceptor should generally check if the return type is applicable as with Java 9, methods are no longer invokable if the return type is not compatible what is a threat to type-safety.

Some check like this @raphw?

This is the place where the return type of the method might not be the same as `this` and we need to do something. The `null` is not a good approach I know.

filiphr:
Thanks a lot @raphw for the comments and helping out with this.

Maybe the best solution would be to throw the assertion error on such places (when the return type is not the same as `this`) and handle the specifics like we already do with the `ProxifyExtractingResult` which by the way would bubble up the `AssertionError` if the call throws an exception.

joel-costigliola:
Many thanks @raphw for taking the time to look into this issue !

raphw:
Your most welcome. Ping me if you have any other things you want to discuss!

filiphr:
@joel-costigliola I think the issues from ByteBuddy side are resolved. 

We now need to decide how we are going to proceed with this. What are we going to do when the return type is not the assertion builder itself?

sormuras:
Any news on this topic?

filiphr:
I think we only need to resolve my last comment. How should we handle that specific case.

@joel-costigliola I think that it would be good if we can do this for 3.9.0, as there is not a lot that is left for this.

joel-costigliola:
I will get back to this issue after I release 2.9.0/3.9.0.
Stay tuned!

joel-costigliola:
@filiphr I'm not sure I fully understand everything, does that mean soft assertions will never be able to support methods that change the return type?

joel-costigliola:
@filiphr I pushed some commits on the PR branch after rebasing it to master

joel-costigliola:
I finally figured out why some methods are ignored, it's because they are `final` to be able to use `@SafeVarargs` so that users don't have a warning when using such methods.

raphw:
Yes, the only way to proxy a final method is by using a Java agent to redefine the actual code. You can use Byte Buddy's advice for doing so and the byte-buddy-agent project for attaching an agent to the current process. Based on this agent, you can redefine an existing method. If you want to go for this solution, I can assist you with details.

Fair warning: attaching an agent is an expensive procedure (1-2 seconds and 3-5 in Java 9 as it requires starting a helper process). Especially users that run simple tests often do often not appreciate this expensive one time cost.

joel-costigliola:
Thank you so much for your help, **really** appreciated.

I think I'll try to add subclasses like `SoftAssertionListAssert` but for assumptions, this is a class that does not override methods as final to put some `@SafeVarargs` so it should ok to use it for proxying.

Another approach I'm keen to evaluate for assumptions is to generate the assumption wrappers code, assumptions mechanism being stateless I believe it is a good candidate, soft assertions is a bit more complex in that regard and I'm not sure it is possible/handy to use code generation for that.

joel-costigliola:
I fixed the tests in https://github.com/joel-costigliola/assertj-core/pull/1093/commits/237257b2f33adddc4a4f6966bfcaad79e0ffffad but that uncovered areas that were not tested, for example soft assertions on streams, ex:

```java
  @Test
  public void should_work_with_stream() {
    try {
      softly.then(Stream.of("a", "b", "c")).contains("a", "b")
            .contains("d")
            .isEmpty();
      softly.assertAll();
    } catch (SoftAssertionError e) {
      List<String> errors = e.getErrors();
      assertThat(errors).hasSize(2);
    }
  }
```
this succeeds on master but fails in this branch for the same reason as before, i.e. `contains` is `final`.
to make it work, `then(Stream actual)` implementation must proxy `StreamAwareListAssert` instead of `ListAssert`.

raphw:
One way to get around the final class issue is to write adapters that delegate to some protected method that can still be proxied:

```java
interface Foo<T> {
  void bar(T... t);
}

abstract class FooAdapter<T> implements Foo<T> {
  @SafeVarArgs
  public final void bar(T... t) {
    doBar(t);
  }
  protected abstract void bar(T[] t);
}
```

Would that be an option?

joel-costigliola:
@raphw yes that's a good option too.

joel-costigliola:
Done, thanks @filiphr and @raphw!

filiphr:
No problem @joel-costigliola, I am really glad that we got this in. Thank you for picking it up abd finishing it.

sormuras:
Looking forward to [3.9.1](https://github.com/joel-costigliola/assertj-core/milestone/38). ✔️

PR #1096, opened 2017-10-21
Use correct SELF class for SoftAssertionPredicateAssert

filiphr:
While fiddling with the proxies I noticed that the `SoftAssertionPredicate` has wrong class in it's extends clause

#### Check List:
* Unit tests : NA


joel-costigliola:
Integrated thanks !

PR #1103, opened 2017-10-26
[#1102] Improve the method doesNotContain in CharSequenceAssert for accepting varargs

### LINKED ISSUE #1102: [API improvement] method doesNotContain in CharSequenceAssert should keep consistence with other assert methods
#### Summary
The current method `doesNotContain(CharSequence sequence)` in `AbstractCharSequenceAssert` only accepts **one CharSequence** as input param, it should be consistent with the method `contains(CharSequence... values)` in `AbstractCharSequenceAssert` and other methods like `doesNotContain(ELEMENT... values)` in `AbstractObjectArrayAssert` , so it may be like this `doesNotContain(CharSequence... values)` and accepts **varargs**.

#### Example

```java
// this should pass
    String str = "Practice makes perfect";
    assertThat(str).doesNotContain("practice", "made", "good");

// this should fail
    String str = "Practice makes perfect";
    assertThat(str).doesNotContain("practice", "make", "good");
```

#### Java 8 specific ?

BillyYccc:
#### Check List:
* Fixes #1102 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
@@ -603,24 +603,48 @@ public abstract class AbstractCharSequenceAssert<SELF extends AbstractCharSequen
   }
 
   /**
-   * Verifies that the actual {@code CharSequence} does not contain the given sequence.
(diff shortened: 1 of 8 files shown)


joel-costigliola:
Integrated thanks @BillyYccc!

PR #1108, opened 2017-11-06
Added catchThrowableOfType() assertion to do type-safe downcast.

kriegfrj:
#### Check List:
* Adds a new feature
* Unit tests : YES 

During my testing using the new catchThrowable() syntax, I frequently found myself wanting to catch a Throwable of a particular subtype, and then run further assertions on the specific subtype. The new function `catchThrowableOfType()` facilitates this.

Before:
````
Throwable t = catchThrowable(() -> { functionThatThrowsParseException(); });
assertThat(t).isInstanceOf(ParseException.class);
ParseException p = (ParseException)t;
assertThat(p.getErrorOffset()).isEqualTo(10);
````
Now:
````
ParseException p = catchThrowableOfType(() -> { functionThatThrowsParseException(); }, 
                                                                     ParseException.class );
assertThat(p.getErrorOffset()).isEqualTo(10);
````
`catchThrowableOfType()` does the `isInstanceOf()` assertion and down-casting for you. The resulting code is not only shorter, it also removes the need for ugly casts in your test code and enforces type safety at compile time.

### DIFF
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -1117,7 +1117,7 @@ public class Assertions {
   }
 
   /**
-   * Allows to catch an {@link Throwable} more easily when used with Java 8 lambdas.
+   * Allows catching a {@link Throwable} more easily when used with Java 8 lambdas.
    *
    * <p>
    * This caught {@link Throwable} can then be asserted.
@@ -1146,12 +1146,13 @@ public class Assertions {
   }
 
   /**
-   * Allows catching an {@link Throwable} of a specific type more easily when used with Java 8 lambdas.
+   * Allows catching a {@link Throwable} of a specific type more easily when used with Java 8 lambdas.
    *
    * <p>
-   * A call is made to {@link #catchThrowable(ThrowingCallable)}, and an assertion is made on the caught {@link Throwable} to ensure it is of the specified type. If
-   * the assertion passes, then the caught {@link Throwable} is cast to the correct subtype before
-   * being returned, making it convenient to perform subtype-specific assertions on the result.
+   * A call is made to {@link #catchThrowable(ThrowingCallable)}, and an assertion is made on the caught
+   * {@link Throwable} to ensure it is of the specified type. If the assertion passes, then the caught
+   * {@link Throwable} is cast to the correct subtype before being returned, making it convenient to
+   * perform subtype-specific assertions on the result.
    * </p>
    *
    * <p>
@@ -1159,15 +1160,39 @@ public class Assertions {
    * </p>
    *
    * <pre><code class='java'>{@literal @}Test
+   * 
+   * class CustomParseException extends Exception {
+   *   final private int line;
+   *   final private int column;
+   *   
+   *   public CustomParseException(String msg, int l, int c) {
+   *     super(msg);
+   *     line = l;
+   *     column = c;
+   *   }
+   *   
+   *   public int getLine() {return line;}
+   *   public int getColumn() {return column;}
+   * }
+   * 
    * public void testException() {
-   *   // when
+   *   // when (assertion will pass)
    *   CustomParseException e = catchThrowableOfType(() -&gt; { throw new CustomParseException("boom!", 1, 5); },
    *                                                 CustomParseException.class);
    *
    *   // then
-   *   assertThat(e).as("message").hasMessageContaining("boom");
-   *   assertThat(e.getLine()).as("line").isEqualTo(1);
-   *   assertThat(e.getColumn()).as("column").isEqualTo(5);
+   *   assertThat(e).hasMessageContaining("boom");
+   *   assertThat(e.getLine()).isEqualTo(1);
+   *   assertThat(e.getColumn()).isEqualTo(5);
+   * }
+   * 
+   * public void testRuntimeException() {
+   *   // when (assertion will fail)
+   *   RuntimeException e = catchThrowableOfType(() -&gt; { throw new IOException("boom!", 1, 5); },
+   *                                             RuntimeException.class);
+   *
+   *   // then
+   *   assertThat(e).hasMessageContaining("boom");
    * } </code></pre>
    *
    * @param shouldRaiseThrowable The lambda with the code that should raise the exception.
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -1139,9 +1139,45 @@ public class Assertions {
    *
    * @param shouldRaiseThrowable The lambda with the code that should raise the exception.
    * @return The captured exception or <code>null</code> if none was raised by the callable.
+   * @see #catchThrowableOfType(ThrowingCallable, Class)
    */
   public static Throwable catchThrowable(ThrowingCallable shouldRaiseThrowable) {
     return AssertionsForClassTypes.catchThrowable(shouldRaiseThrowable);
+  }
+
+  /**
+   * Allows catching an {@link Throwable} of a specific type more easily when used with Java 8 lambdas.
+   *
+   * <p>
+   * A call is made to {@link #catchThrowable(ThrowingCallable)}, and an assertion is made on the caught {@link Throwable} to ensure it is of the specified type. If
+   * the assertion passes, then the caught {@link Throwable} is cast to the correct subtype before
+   * being returned, making it convenient to perform subtype-specific assertions on the result.
+   * </p>
+   *
+   * <p>
+   * Example:
+   * </p>
+   *
+   * <pre><code class='java'>{@literal @}Test
+   * public void testException() {
+   *   // when
+   *   CustomParseException e = catchThrowableOfType(() -&gt; { throw new CustomParseException("boom!", 1, 5); },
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/main/java/org/assertj/core/api/Assertions.java [hunk 1]
```diff
@@ -1139,12 +1139,48 @@ public static AtomicLongArrayAssert assertThat(AtomicLongArray actual) {
...
+   *
+   *   // then
+   *   assertThat(e).as("message").hasMessageContaining("boom");
+   *   assertThat(e.getLine()).as("line").isEqualTo(1);
+   *   assertThat(e.getColumn()).as("column").isEqualTo(5);
+   * } </code></pre>
+   *
+   * @param shouldRaiseThrowable The lambda with the code that should raise the exception.
+   * @param type The type of exception that the code is expected to raise.
+   * @return The captured exception or <code>null</code> if none was raised by the callable.
```
joel-costigliola: I believe null is never returned as we try to cast the return to `E`

src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -1139,12 +1139,48 @@ public static AtomicLongArrayAssert assertThat(AtomicLongArray actual) {
...
    * @param shouldRaiseThrowable The lambda with the code that should raise the exception.
    * @return The captured exception or <code>null</code> if none was raised by the callable.
+   * @see #catchThrowableOfType(ThrowingCallable, Class)
    */
   public static Throwable catchThrowable(ThrowingCallable shouldRaiseThrowable) {
     return AssertionsForClassTypes.catchThrowable(shouldRaiseThrowable);
   }
 
   /**
+   * Allows catching an {@link Throwable} of a specific type more easily when used with Java 8 lambdas.
```
joel-costigliola: Typo `an` -> `a` (fix it in `catchThrowable` as you go)

src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -1139,12 +1139,48 @@ public static AtomicLongArrayAssert assertThat(AtomicLongArray actual) {
...
+   * A call is made to {@link #catchThrowable(ThrowingCallable)}, and an assertion is made on the caught {@link Throwable} to ensure it is of the specified type. If
+   * the assertion passes, then the caught {@link Throwable} is cast to the correct subtype before
+   * being returned, making it convenient to perform subtype-specific assertions on the result.
+   * </p>
+   *
+   * <p>
+   * Example:
+   * </p>
+   *
+   * <pre><code class='java'>{@literal @}Test
```
joel-costigliola: for clarify I would add `CustomParseException` with `line` and `column` as public final fields.

src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -1139,12 +1139,48 @@ public static AtomicLongArrayAssert assertThat(AtomicLongArray actual) {
...
+   * public void testException() {
+   *   // when
+   *   CustomParseException e = catchThrowableOfType(() -&gt; { throw new CustomParseException("boom!", 1, 5); },
+   *                                                 CustomParseException.class);
+   *
+   *   // then
+   *   assertThat(e).as("message").hasMessageContaining("boom");
+   *   assertThat(e.getLine()).as("line").isEqualTo(1);
+   *   assertThat(e.getColumn()).as("column").isEqualTo(5);
```
joel-costigliola: remove `as` usage, it does not add much value to the example IMO.

src/main/java/org/assertj/core/api/ThrowableAssert.java
```diff
@@ -63,4 +63,11 @@ public static Throwable catchThrowable(ThrowingCallable shouldRaiseThrowable) {
     }
     return null;
   }
+
+  @SuppressWarnings("unchecked")
+  public static <E extends Throwable> E catchThrowableOfType(ThrowingCallable shouldRaiseThrowable, Class<E> type) {
+    Throwable t = AssertionsForClassTypes.catchThrowable(shouldRaiseThrowable);
+    new ThrowableAssert(t).isInstanceOf(type);
```
joel-costigliola: following my other comment, add `.overridingErrorMessage("Expecting code to throw a <%s> but did not", type)` before `.isInstanceOf`

src/test/java/org/assertj/core/api/Assertions_assertThat_with_Throwable_Test.java
```diff
@@ -52,11 +53,39 @@ public void should_fail_if_no_throwable_was_thrown() {
...
+  }
+
+  @Test
+  public void catchThrowableOfType_should_succeed_and_return_actual_instance_with_correct_class() {
+    final Exception expected = new RuntimeException("boom!!");
+    Exception actual = null;
+    try {
+      actual = catchThrowableOfType(raisingException(expected), Exception.class);
+    } catch (AssertionError a) {
+      Fail.fail("catchThrowableOfType should not have asserted", a);
```
joel-costigliola: static import `fail`

src/test/java/org/assertj/core/api/Assertions_assertThat_with_Throwable_Test.java
```diff
@@ -52,11 +53,39 @@ public void should_fail_if_no_throwable_was_thrown() {
...
-    Throwable boom = catchThrowable(raisingException("boom!!!!"));
+    Exception e = new Exception("boom!!");
+    Throwable boom = catchThrowable(raisingException(e));
 
     // then
-    assertThat(boom).isInstanceOf(Exception.class)
-                    .hasMessageContaining("boom");
+    assertThat(boom).isSameAs(e);
+  }
+
```
joel-costigliola: add test when the code does not throw the expected throwable.

src/main/java/org/assertj/core/api/Assertions.java
kriegfrj: Yes, in fact I already noticed this in my own code - the isInstanceOf() will assert if the exception is null, and the error message is not very helpful. I already fixed the error message in my local repository.

However, one concern I have with going down that path is is that of consistency between `catchThrowable()` and `catchThrowableOfType()`. The former will happy return `null` if no exception is thrown. Will it confuse developers if `catchThrowableOfType()` behaves slightly differently? I suppose it's already behaving differently because it is doing an assertion anyway, so we might as well do that. But then perhaps renaming it to something else like `assertThatBlockThrowsThrowableOfType()` would make the difference more clear?

joel-costigliola: I was thinking the most likely use case would be for the code to raise the expected exception type but we actually can allow null and still have the return value to be cast to the expected exception.

basically it boils down to only check the instance type if not null
```java
if (t != null) {
  new ThrowableAssert(t).isInstanceOf(type);
}
return (E)t; // works if t is null
```

src/test/java/org/assertj/core/api/Assertions_assertThat_with_Throwable_Test.java
kriegfrj: This was a pre-existing test for `catchThrowable()`, which doesn't expect any particular throwable.

I modified it to make it stricter. The previous test could have passed if `catchThrowable()` returned a similar exception class to the one thrown; this one asserts that it must actually be the same exception.

Happy to add a test for `catchThrowable()` when the code doesn't throw an exception, but this is not strictly within the scope of my PR (though neither was this change, so I guess I brought it on myself!)

src/main/java/org/assertj/core/api/Assertions.java
kriegfrj: In practice I find it makes it easier to find the assertion that failed. But if you feel it makes the example too cluttered I'm happy to leave them out.

joel-costigliola: that was my point :+1:

src/main/java/org/assertj/core/api/Assertions.java
kriegfrj: Done.

src/main/java/org/assertj/core/api/Assertions.java
kriegfrj: Done.

src/main/java/org/assertj/core/api/Assertions.java
kriegfrj: Done.

src/main/java/org/assertj/core/api/ThrowableAssert.java
kriegfrj: Done.

src/test/java/org/assertj/core/api/Assertions_assertThat_with_Throwable_Test.java
kriegfrj: Done.

src/test/java/org/assertj/core/api/Assertions_assertThat_with_Throwable_Test.java
kriegfrj: Done.

joel-costigliola:
Integrated thanks @kriegfrj !

PR #1111, opened 2017-11-14
Have AbstractInstantAssertions implement ComparableAssert

jbarnettwomply:
Using an older assert/j, we have a bunch of assertions that
work using ComparableAssertions on Instant objects.  With the
addition of native Instant assertions, these currently fail to
compile.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractInstantAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractInstantAssert.java
@@ -32,7 +32,7 @@ import org.assertj.core.internal.Objects;
  * @since 3.7.0
  */
 public class AbstractInstantAssert<SELF extends AbstractInstantAssert<SELF>>
-    extends AbstractTemporalAssert<SELF, Instant> {
+    extends AbstractTemporalAssert<SELF, Instant> implements ComparableAssert<SELF, Instant> {
 
   /**
    * Creates a new <code>{@link org.assertj.core.api.AbstractInstantAssert}</code>.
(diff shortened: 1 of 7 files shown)


joel-costigliola:
Yeah, we should do the same for any Comparable type we provide assertions for.

src/main/java/org/assertj/core/api/AbstractInstantAssert.java
```diff
@@ -32,7 +32,7 @@
  * @since 3.7.0
  */
 public class AbstractInstantAssert<SELF extends AbstractInstantAssert<SELF>>
-    extends AbstractTemporalAssert<SELF, Instant> {
+    extends AbstractTemporalAssert<SELF, Instant> implements ComparableAssert<SELF, Instant> {
```
joel-costigliola: try extending `AbstractComparableAssert<SELF, Instant>` to get the comparable assertions implementation for free.

jbarnettwomply: I tried doing that at the `AbstractTemporalAssert` level, but since `Temporal` doesn't extend `Comparable` that didn't work.  Seemed better to inherit the Temporal functionality and duplicate the Comparable, but I guess it could be the other way around?  Or is there a better class hierarchy modification you can think of?

joel-costigliola: If a type does not implement `Comparable`, its assertion type should not implement `ComparableAssert`.

Moreover, I believe it is better to use `isAfter` instead of `isGreaterThan` for `Instant` assertions as `after` expresses time better than `greater` but I'm ok to add the comparable assertions to avoid breaking people code.

jbarnettwomply: I agree. Thanks.

joel-costigliola:
@jbarnettwomply how much work is it to migrate the `Comparable` assertions to `Instant` ? I feel that adding `isGreaterThan` to `Instant` assertion is actually polluting the API so I would rather see people using the time related assertions instead of the number related ones. 
Is there any assertion in `Comparable` that does not have an equivalent in `Instant` ?

jbarnettwomply:
it's not much work to change each `Comparable` assertion to the appropriate `Instant`/`Temporal` assertion, just a number of small changes scattered throughout our codebases and teams.  I agree it'd be better to be using time related assertions, the main reason for requesting this change is for backwards compatibility moreso than anything else.

joel-costigliola:
I think I'm going to pass on this one and https://github.com/joel-costigliola/assertj-core/issues/1112 because of API pollution. Feel free to come with additional to change my mind.

PR #1119, opened 2017-11-22
Print stack trace information for `AbstractThrowableAssert.doesNotThr…

mkobit:
…owAnyException()`

closes https://github.com/joel-costigliola/assertj-core/issues/1118

#### Check List:

* Fixes https://github.com/joel-costigliola/assertj-core/issues/1118
* Unit tests : YES
* Javadoc with a code example (API only) : YES / NO / NA

### DIFF
--- a/src/test/java/org/assertj/core/api/Assertions_assertThatCode_Test.java
+++ b/src/test/java/org/assertj/core/api/Assertions_assertThatCode_Test.java
@@ -17,6 +17,7 @@ import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.assertj.core.test.ExpectedException.none;
 
 import org.assertj.core.api.ThrowableAssert.ThrowingCallable;
+import org.assertj.core.error.ShouldNotHaveThrown;
 import org.assertj.core.test.ExpectedException;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


src/test/java/org/assertj/core/api/Assertions_assertThatCode_Test.java
```diff
@@ -39,17 +40,14 @@ public void can_invoke_late_assertion_on_assertThatCode() {
...
-    thrown.expectAssertionError(format("[Test] %n" +
-                                       "Expecting code not to raise a throwable but caught a%n" +
-                                       "  <java.lang.Exception>%n" +
-                                       "with message :%n" +
-                                       "  \"boom\""));
+    thrown.expectAssertionError(ShouldNotHaveThrown.shouldNotHaveThrown(exception));
 
     // When
-    assertThatCode(boom).as("Test").doesNotThrowAnyException();
+    assertThatCode(boom).doesNotThrowAnyException();
```
joel-costigliola: why removing `as("Test")` ? 
we want to make sure that the error allows a description set with `as`

src/test/java/org/assertj/core/api/Assertions_assertThatCode_Test.java
```diff
@@ -39,17 +40,14 @@ public void can_invoke_late_assertion_on_assertThatCode() {
...
+    Exception exception = new Exception("boom");
+    ThrowingCallable boom = raisingException(exception);
 
     // Expect
-    thrown.expectAssertionError(format("[Test] %n" +
-                                       "Expecting code not to raise a throwable but caught a%n" +
-                                       "  <java.lang.Exception>%n" +
-                                       "with message :%n" +
-                                       "  \"boom\""));
+    thrown.expectAssertionError(ShouldNotHaveThrown.shouldNotHaveThrown(exception));
```
joel-costigliola: static import `shouldNotHaveThrown`

joel-costigliola: add a separate test to verify the error message content as in https://github.com/joel-costigliola/assertj-core/blob/08c276844c55e74b5773248e7c5c9b6a741a4b4d/src/test/java/org/assertj/core/error/ShouldBeBefore_create_Test.java

mkobit: I wasn't sure of the best way to test both the description with `as()` and also testing the "stacktrace" way without making the test brittle - I'll take another look at the stacktrace case and maybe I can just assert that the message begins with the description in the `as()` test case? WDYT?

joel-costigliola: `startsWith` and `contains` String assertions should do the job.

joel-costigliola:
Integrated thanks @mkobit !

PR #1124, opened 2017-11-29
Fix `isEqualTo` typo.

pomle:
Remove a source of confusion when reading the docs.


joel-costigliola:
Good catch @pomle, thanks.

PR #1129, opened 2017-12-07
Feature/compare local date time with xml gregorian calendar

### LINKED ISSUE #1127: Compare LocalDateTime with XMLGregorianCalendar
#### Summary
When testing object mapping, it can be useful to compare `LocalDateTime` with `XMLGregorianCalendar` without having to compare field by field or via a filed extractor

#### Example

```java
LocalDate date = // ... 2017-01-01 02:03:04
XmlGregorinaCalendar calendar = // ... 2017-01-01 02:03:04
Assertions.assertThat(date).isEqualTo(calendar); // true
```

#### Java 8 specific ?

~* NO  : create Pull Request from the `2.x` branch~

gervaisb:
#### Check List:
* Fixes #1127 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractLocalDateAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractLocalDateAssert.java
@@ -26,6 +26,8 @@ import java.util.Arrays;
 import org.assertj.core.internal.Failures;
 import org.assertj.core.internal.Objects;
 
+import javax.xml.datatype.XMLGregorianCalendar;
+
 /**
  * Assertions for {@link LocalDate} type from new Date &amp; Time API introduced in Java 8.
  *
@@ -34,7 +36,8 @@ import org.assertj.core.internal.Objects;
 public abstract class AbstractLocalDateAssert<SELF extends AbstractLocalDateAssert<SELF>>
     extends AbstractTemporalAssert<SELF, LocalDate> {
 
-  public static final String NULL_LOCAL_DATE_TIME_PARAMETER_MESSAGE = "The LocalDate to compare actual with should not be null";
+  private static final String NULL_LOCAL_DATE_TIME_PARAMETER_MESSAGE = "The LocalDate to compare actual with should not be null";
+  private static final String NULL_XML_GREGORIAN_CALENDAR_PARAMETER_MESSAGE = "The XMLGregorianCalendar to compare actual with should not be null";
 
   /**
    * Creates a new <code>{@link org.assertj.core.api.AbstractLocalDateAssert}</code>.
@@ -265,6 +268,26 @@ public abstract class AbstractLocalDateAssert<SELF extends AbstractLocalDateAsse
   }
 
   /**
+   * Same assertion as {@link #isEqualTo(Object)} (where Object is expected to be {@link javax.xml.datatype.XMLGregorianCalendar})
+   * but here you pass {@link javax.xml.datatype.XMLGregorianCalendar}
+   * <p>
+   * Example :
+   * <pre><code class='java'> // use XMLGregorianCalendar  as-is without conversion
+   * assertThat(parse("2000-01-01")).isEqualTo(xmlGregorianCalendar);</code></pre>
+   *
+   * @param xmlGregorianCalendar to compare with actual {@link LocalDate}.
+   * @return this assertion object.
+   * @throws AssertionError if the actual {@code LocalDate} is {@code null}.
+   * @throws IllegalArgumentException if given calendar is null.
+   * @throws AssertionError if the actual {@code LocalDate} is not equal to the {@link XMLGregorianCalendar}
+   *  year, month and day. (Time and Timezone are ignored).
+   */
+  public SELF isEqualTo(XMLGregorianCalendar xmlGregorianCalendar) {
+    assertXMLGregorianCalendarParameterIsNotNull(xmlGregorianCalendar);
+    return isEqualTo(convert(xmlGregorianCalendar));
+  }
+
+  /**
    * Same assertion as {@link #isIn(Object...)} (where Objects are expected to be {@link LocalDate}) but here you
    * pass {@link LocalDate} String representations that must follow <a href=
    * "http://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_LOCAL_DATE"
@@ -462,6 +485,16 @@ public abstract class AbstractLocalDateAssert<SELF extends AbstractLocalDateAsse
     return LocalDate.parse(localDateAsString);
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  protected LocalDate convert(XMLGregorianCalendar xmlGregorianCalendar) {
+    return xmlGregorianCalendar.toGregorianCalendar()
+      .toZonedDateTime()
+      .toLocalDate();
+  }
+
   private static Object[] convertToLocalDateArray(String... localDatesAsString) {
     return Arrays.stream(localDatesAsString).map(LocalDate::parse).toArray();
   }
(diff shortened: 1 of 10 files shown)


src/main/java/org/assertj/core/api/AbstractTemporalAssert.java
```diff
@@ -113,4 +116,35 @@ public SELF isCloseTo(String otherAsString, TemporalOffset<? super TEMPORAL> off
...
+
+  /**
+   * Check that the {@link T} to compare actual {@link TEMPORAL} to is not null, in that case throws
+   * a {@link IllegalArgumentException} with an explicit message
+   *
+   * @param type the expected type of parameter. Used for message, not for validation
+   * @param parameter the {@link T} parameter to check
+   * @throws IllegalArgumentException with an explicit message if the given parameter is null
+   */
+  protected static <T> void assertParameterIsNotNull(Class<T> type, T parameter) {
```
joel-costigliola: is that used ?

src/main/java/org/assertj/core/api/AbstractTemporalAssert.java
```diff
@@ -113,4 +116,35 @@ public SELF isCloseTo(String otherAsString, TemporalOffset<? super TEMPORAL> off
...
+
+  /**
+   * Check that the {@link LocalDateTime} string representation to compare actual {@link LocalDateTime} to is not null,
+   * otherwise throws a {@link IllegalArgumentException} with an explicit message
+   *
+   * @param type the expected type of parameter. Used for message, not for validation
+   * @param string String representing the {@link TEMPORAL} to compare actual with
+   * @throws IllegalArgumentException with an explicit message if the given {@link String} is null
+   */
+  protected static void assertParameterIsNotNull(Class<?> type, String string) {
```
joel-costigliola: is that used ?

joel-costigliola:
This may be normal if the PR is not finished but the only tests added were in `LocalDateAssert_isEqualTo_Test`.
The same kind of tests need to be added for the other `isEqualTo(XMLGregorianCalendar)`

src/main/java/org/assertj/core/api/AbstractLocalDateAssert.java
```diff
@@ -264,6 +267,26 @@ public SELF isNotEqualTo(String localDateAsString) {
...
   }
 
+  /**
+   * Same assertion as {@link #isEqualTo(Object)} (where Object is expected to be {@link javax.xml.datatype.XMLGregorianCalendar})
+   * but here you pass {@link javax.xml.datatype.XMLGregorianCalendar}
+   * <p>
+   * Example :
+   * <pre><code class='java'> // use XMLGregorianCalendar  as-is without conversion
+   * assertThat(parse("2000-01-01")).isEqualTo(xmlGregorianCalendar);</code></pre>
+   *
```
joel-costigliola: add a note about time zone with an example showing the assertion behavior when the local time zone is not the same as the provided `XMLGregorianCalendar`.

src/main/java/org/assertj/core/api/AbstractLocalDateAssert.java
```diff
@@ -494,4 +527,15 @@ private static void assertLocalDateParameterIsNotNull(LocalDate other) {
...
   }
 
+  /**
+   * Check that the {@link XMLGregorianCalendar} to compare actual {@link LocalDate} to is not null, in that case throws
+   * a {@link IllegalArgumentException} with an explicit message
+   *
+   * @param xmlGregorianCalendar the {@link XMLGregorianCalendar} to check
+   * @throws IllegalArgumentException with an explicit message if the given {@link XMLGregorianCalendar} is null
+   */
+  private static void assertXMLGregorianCalendarParameterIsNotNull(XMLGregorianCalendar xmlGregorianCalendar) {
```
joel-costigliola: Should be in `AbstractTemporalAssert` as there is nothing specific to `LocalDate` assertions.

src/main/java/org/assertj/core/api/AbstractLocalDateAssert.java
```diff
@@ -264,6 +267,26 @@ public SELF isNotEqualTo(String localDateAsString) {
...
+   * assertThat(parse("2000-01-01")).isEqualTo(xmlGregorianCalendar);</code></pre>
+   *
+   * @param xmlGregorianCalendar to compare with actual {@link LocalDate}.
+   * @return this assertion object.
+   * @throws AssertionError if the actual {@code LocalDate} is {@code null}.
+   * @throws IllegalArgumentException if given calendar is null.
+   * @throws AssertionError if the actual {@code LocalDate} is not equal to the {@link XMLGregorianCalendar}
+   *  year, month and day. (Time and Timezone are ignored).
+   */
+  public SELF isEqualTo(XMLGregorianCalendar xmlGregorianCalendar) {
```
joel-costigliola: It should be possible to move this method (or at least its implementation) in `AbstractTemporalAssert` but still override the method in order to write the proper javadoc.

src/test/java/org/assertj/core/api/localdate/LocalDateAssertBaseTest.java
```diff
@@ -35,9 +43,29 @@
   @DataPoint
   public static LocalDate localDate3 = LocalDate.now().plusDays(10);
 
+  protected XMLGregorianCalendar xmlOfLocalDate1;
```
joel-costigliola: rename to `xmlGregorianCalendar1` or `xmlGregorianCalendarOfLocalDate1`

src/test/java/org/assertj/core/api/localdate/LocalDateAssert_isEqualTo_Test.java
```diff
@@ -39,6 +39,12 @@ public void test_isEqualTo_assertion(LocalDate referenceDate) {
     verify_that_isEqualTo_assertion_fails_and_throws_AssertionError(referenceDate);
   }
 
+  @Test
+  public void test_isEqualTo_assertion_with_XmlGregorianCalendar() {
+    assertThat(localDate1).isEqualTo(xmlOfLocalDate1);
+    assertThat(localDate1).isNotEqualTo(xmlOfLocalDate2);
```
joel-costigliola: `isNotEqualTo` assertion will always success whatever `XmlGregorianCalendar` is given (even xmlOfLocalDate1!) since it does not any conversion to LocalDate to `XmlGregorianCalendar`.
To be consistent, we need to add `isNotEqualTo (XMLGregorianCalendar xmlGregorianCalendar)`

joel-costigliola:
BTW, I'm planning to release 3.9.0 in two weeks (if I have the time to), if you want to see this assertion in it then you know what to do ... no pressure! :smile_cat:

filiphr:
Just one small remark from me. I am not sure what the stand on Java 9 for AssertJ this. However, this addition might break Java 9 support, as it uses a type that is not in `java.base`.

joel-costigliola:
@filiphr can you elaborate ? The  java9 build is passing on Travis

filiphr:
The build compiles because it performs the compilation on the classpath. Which means that all modules are available. However, if you want to run it on the module path you would need to add the `java.xml` module if you want to use the java time assertions because of this addition, as the `XmlGregorianCalendar` must be present when you want to use the `AbstractInstantAssert` for example, which makes the `java.xml` module mandatory.

Note: We are using classes from the `java.xml` module already however this is in util classes which are invoked only if you call an explicit assertion method and that is fine as then the module is optionally required.

It is up to you to decide what we want to do, I just wanted to bring this information up. If you do deeper analysis you'll see that there are some other modules that are being used, but none are in the public api package and it would be easier to change those things. However, this addition like this would be more difficult to be changed.

A way that would be to make this optional would be to change the API to something like:

```
assertThat(localDate).xml().isEqualTo(xmlGregorianCalendar);
```

`xml()` would return an assert that contains the `XmlGregorianCalendar`. With something like this users will be able to use the temporal asserts without the need of he `java.xml` module. That module will only be needed when one calls `xml()` explicitly.

joel-costigliola:
@filiphr thanks for the explanation, just to be clear, does that mean no instant assertions can be used if you don't have the `java.xml` module in the path or is it just this one ?

filiphr:
@joel-costigliola you won't be able to use all assertions that extend the classes that contain the `XmlGregorianCalendar` imports as you would be trying to load class which is not there. It would act exactly the same as if you would add joda time `DateTime` and joda is not present

gervaisb:
@filiphr and @joel-costigliola what should I do. Do you prefer a dedicated `XmlGregorianCalendarAssert` ?

filiphr:
@gervaisb I think it is up to @joel-costigliola to decide what would be the best approach to resolve this. I just wanted to bring this into attention in the PR.

joel-costigliola:
Closing due to lack of interest and the revamped date/time API in java but happy to reopen it if someones wants to finish it.

PR #1134, opened 2017-12-19
Make configurable if AbstractAssert#equals throws an exception

### LINKED ISSUE #1128: Must AbstractAssert#equals  throw UnsupportedOperationException
#### Summary
@joel-costigliola Is it necessary for the `AbstractAssert#equals` to throw the `UnsupportedOperationException`?
This is against the contract of `Object#equals` and brakes other libraries which are not handling this. I know that it should be the deal of the libraries, if they can handle this error but this situation is absolutely not obvious. Is is possible, to change this?

In fact, currently it breaks the coroutines of Kotlin.

#### Example

```java
org.assertj.core.api.AbstractAssert

  /**
   * {@inheritDoc}
   *
   * @deprecated use {@link #isEqualTo} instead
   *
   * @throws UnsupportedOperationException if this method is called.
   */
  @Override
  @Deprecated
  public boolean equals(Object obj) {
    throw new UnsupportedOperationException("'equals' is not supported...maybe you intended to call 'isEqualTo'");
  }

```

#### Java 8 specific ?

guenhter:
#### Check List:
* Fixes #1128 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -50,6 +50,9 @@ import org.assertj.core.util.VisibleForTesting;
  */
 public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, ACTUAL> implements Assert<SELF, ACTUAL> {
 
+  // https://github.com/joel-costigliola/assertj-core/issues/1128
+  public static boolean throwUnsupportedExceptionOnEquals = true;
+
   private static final String ORG_ASSERTJ = "org.assert";
 
   @VisibleForTesting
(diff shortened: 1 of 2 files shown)


src/main/java/org/assertj/core/api/AbstractAssert.java
```diff
@@ -50,6 +50,8 @@
  */
 public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, ACTUAL> implements Assert<SELF, ACTUAL> {
 
+  public static boolean throwUnsupportedExceptionOnEquals = true;
```
joel-costigliola: add a comment with a link to https://github.com/joel-costigliola/assertj-core/issues/1128 to make it clear why this field exists.

guenhter:
Done

joel-costigliola:
Thanks @guenhter.

PR #1137, opened 2017-12-23
Allow for % in fields when comparing recursively

### LINKED ISSUE #1136: `isEqualToComparingFieldByFieldRecursively()` throws Exception when using percentage '%' in string field
#### Summary

I have got a test example using the method `isEqualToComparingFieldByFieldRecursively()`. I expect this test to fail but a `MissingFormatArgumentException` is thrown instead.

#### Example

This is my test code:
```java

import org.junit.Test;

public class PercentageTest
{
    private static class Foo
    {
        String value;
    }

    @Test
    public void testPercentage() throws Exception
    {
        Foo foo1 = new Foo();
        foo1.value = "foo";

        Foo foo2 = new Foo();
        foo2.value = "%foo";

        assertThat(foo1).isEqualToComparingFieldByFieldRecursively(foo2);
    }
}
```

This is the exception with stacktrace:
```
java.util.MissingFormatArgumentException: Format specifier '%f'
	at java.util.Formatter.format(Formatter.java:2519)
	at java.util.Formatter.format(Formatter.java:2455)
	at java.lang.String.format(String.java:2940)
	at org.assertj.core.util.Strings.formatIfArgs(Strings.java:92)
	at org.assertj.core.error.MessageFormatter.format(MessageFormatter.java:61)
	at org.assertj.core.error.BasicErrorMessageFactory.create(BasicErrorMessageFactory.java:110)
	at org.assertj.core.internal.Failures.failure(Failures.java:116)
	at org.assertj.core.internal.Objects.assertIsEqualToComparingFieldByFieldRecursively(Objects.java:713)
	at org.assertj.core.api.AbstractObjectAssert.isEqualToComparingFieldByFieldRecursively(AbstractObjectAssert.java:647)
	at PercentageTest.testPercentage(PercentageTest.java:26)
```

#### Java 8 specific ?

I have tested this with AssertJ versions 2.6.0 and 3.8.0.

ChrisCanCompute:
#### Check List:
* Fixes #1136
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/error/ShouldBeEqualByComparingFieldByFieldRecursively.java
+++ b/src/main/java/org/assertj/core/error/ShouldBeEqualByComparingFieldByFieldRecursively.java
@@ -64,8 +64,8 @@ public class ShouldBeEqualByComparingFieldByFieldRecursively extends BasicErrorM
                   "- expected: <%s>%n" +
                   "- actual  : <%s>",
                   join(difference.getPath()).with("."),
-                  otherFieldValueRepresentation,
-                  actualFieldValueRepresentation);
+                  otherFieldValueRepresentation.replace("%", "%%"),
+                  actualFieldValueRepresentation.replace("%", "%%"));
   }
 
 }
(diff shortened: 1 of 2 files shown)


filiphr:
I think it would be interesting to see what happens when there is more than one `%`

ChrisCanCompute:
`    other.name = "%%foo";`
In the test gives:
```
Path to difference: <name>
- expected: <"%%foo">
- actual  : <"foo">
```

ChrisCanCompute:
"%%" just gets converted back to "%" (see the [docs](https://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html#syntax)), so it should handle any combination of them.

joel-costigliola:
Integrated thanks @ChrisCanCompute!

PR #1138, opened 2017-12-23
Add containsExactly/containsExactlyInAnyOrder String assertion

### LINKED ISSUE #1063: containsExactly and containsExactlyInAnyOrder should also work for substrings
#### Summary
AssertJ has very convenient methods ```containsExactly``` and ```containsExactlyInAnyOrder``` for testing Collections.
On the other hand, it can also test substrings using ```containsSequence``` but is limited only to the exact sequence of Strings

I would like to have the ability to assert, that given substrings contains exactly expected result (like ```containsSequence```) but contains it exactly in any order.

Maybe instead of ```containsSequence``` you could overload ```containsExactly``` and ```containsExactlyInAnyOrder``` to work for substrings as well?

#### Example

```java
 assertThat("Element1 Element2").containsExactlyInAnyOrder("Element2", "Element1");
```
#### Java 8 specific ?
* NO

BillyYccc:
#### Check List:
* Fixes #1063 
* Unit tests : YES

#### Something needed to be discussed
While I'm implementing the `assertContainsExactly` assertion, I have reused the `IterableDiff` in package `org.assertj.core.internal` like it is done in `Iterables#assertContainsExactly`.

But I find something unexpected, if you write an iterable assertion like this 
```
    List<String> myList = Arrays.asList("$","$","#");
    Assertions.assertThat(myList)
              .containsExactly("#", "#", "$");
```
AssertionError is this
```
java.lang.AssertionError: 
  <"$">
whereas expected element was:
  <"#">
```

method `differencesFound()` in `IterableDiff` will return false when calling `IterableDiff.diff(Arrays.asList("#","#","$"),Arrays.asList("$","$","#"),StandardComparisonStrategy.instance()).differencesFound()`

It looks like this may lead to give a **wrong** assertion message 

> Actual and expected have the same elements but not in the same order

in fact they **don't** have the same elements?

For now I'm following the same impl in `containsExactly` Iterable assertion when implementing `containsExactly` String assertion.

So the question is 
shall we change the `IterableDiff` or work this out in some other ways?

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
@@ -543,7 +543,7 @@ public abstract class AbstractCharSequenceAssert<SELF extends AbstractCharSequen
    *
    * @param values the charSequences to check.
    * @return {@code this} assertion object.
-   * @throws NullPointerException if the array of values is {@code null}.
+   * @throws NullPointerException if the array or one of its values is {@code null}.
    * @throws IllegalArgumentException if the array of values is empty.
    * @throws AssertionError if the given {@code CharSequence} is {@code null}.
    * @throws AssertionError if the given {@code CharSequence} does not contain exactly the values in a sequential order.
@@ -570,7 +570,7 @@ public abstract class AbstractCharSequenceAssert<SELF extends AbstractCharSequen
    *
    * @param values the charSequences to check.
    * @return {@code this} assertion object.
-   * @throws NullPointerException if the Iterable of values is {@code null}.
+   * @throws NullPointerException if the Iterable or one of its values is {@code null}.
    * @throws IllegalArgumentException if the Iterable of values is empty.
    * @throws AssertionError if the given {@code CharSequence} is {@code null}.
    * @throws AssertionError if the given {@code CharSequence} does not contain exactly the values in a sequential order.
@@ -597,7 +597,7 @@ public abstract class AbstractCharSequenceAssert<SELF extends AbstractCharSequen
    *
    * @param values the charSequences to check.
    * @return {@code this} assertion object.
-   * @throws NullPointerException if the array of values is {@code null}.
+   * @throws NullPointerException if the array or one of its values is {@code null}.
    * @throws IllegalArgumentException if the array of values is empty.
    * @throws AssertionError if the given {@code CharSequence} is {@code null}.
    * @throws AssertionError if the given {@code CharSequence} does not contain exactly the values in any order.
@@ -624,7 +624,7 @@ public abstract class AbstractCharSequenceAssert<SELF extends AbstractCharSequen
    *
    * @param values the charSequences to check.
    * @return {@code this} assertion object.
-   * @throws NullPointerException if the iterable of values is {@code null}.
+   * @throws NullPointerException if the iterable or one of its values is {@code null}.
    * @throws IllegalArgumentException if the iterable of values is empty.
    * @throws AssertionError if the given {@code CharSequence} is {@code null}.
    * @throws AssertionError if the given {@code CharSequence} does not contain exactly the values in any order.
--- a/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
@@ -527,6 +527,60 @@ public abstract class AbstractCharSequenceAssert<SELF extends AbstractCharSequen
   }
 
   /**
+   * Verifies that the actual {@code CharSequence} contains exactly all the values of the given charSequences <b>in a sequential order with no more or less elements</b>.
+   * <p>
+   * Example:
+   * <pre><code class='java'> String actual = &quot;foo, bar&quot;;
+   *
+   * // this assertion succeeds
+   * assertThat(actual).containsExactly(&quot;foo&quot;, &quot;, bar&quot;);
+   *
+   * // this assertion fails because of the wrong order
+   * assertThat(actual).containsExactly(&quot;, bar&quot;, &quot;foo&quot;);
+   *
+   * // this assertion fails because of unmatched elements
+   * assertThat(actual).containsExactly(&quot;foo&quot;);</code></pre>
+   *
+   * @param values the charSequences to check.
+   * @return {@code this} assertion object.
+   * @throws NullPointerException if the array of values is {@code null}.
+   * @throws IllegalArgumentException if the array of values is empty.
+   * @throws AssertionError if the given {@code CharSequence} is {@code null}.
+   * @throws AssertionError if the given {@code CharSequence} does not contain exactly the values in a sequential order.
+   */
+  public SELF containsExactly(CharSequence... values){
+    strings.assertContainsExactly(info, actual, values);
+    return myself;
+  }
+
+  /**
+   * Verifies that the actual {@code CharSequence} contains exactly all the values of the given Iterable <b>in a sequential order with no more or less elements</b>.
+   * <p>
+   * Example:
+   * <pre><code class='java'> String actual = &quot;foo, bar&quot;;
+   *
+   * // this assertion succeeds
+   * assertThat(actual).containsExactly(asList(&quot;foo&quot;, &quot;, bar&quot;));
+   *
+   * // this assertion fails because of the wrong order
+   * assertThat(actual).containsExactly(asList(&quot;, bar&quot;, &quot;foo&quot;));
+   *
+   * // this assertion fails because of unmatched elements
+   * assertThat(actual).containsExactly(asList(&quot;foo&quot;));</code></pre>
+   *
+   * @param values the charSequences to check.
+   * @return {@code this} assertion object.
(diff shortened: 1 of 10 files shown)


joel-costigliola:
I agree the error message is wrong, there is a bug in `IterableDiff` as it does not deal correctly with duplicate elements.

BillyYccc:
Hi! After issue #1140 is fixed, I did some changes and make the `containsExactly/containsExactlyInAnyOrder` String Assertion work implemented with `IterableDiff`.

In addition, I rework the Impl of `containsExactlyInAnyOrder` Array/Iterable Assertion with `IterableDiff`.

src/main/java/org/assertj/core/internal/Strings.java
```diff
@@ -767,6 +773,75 @@ private void assertNotNull(AssertionInfo info, CharSequence actual) {
...
+
+    if (actualAsList.size() != valuesAsList.size()) {
+      throw failures.failure(info, shouldHaveSameSizeAs(actual, actualAsList.size(), valuesAsList.size()));
+    }
+
+    IterableDiff diff = diff(actualAsList, valuesAsList, comparisonStrategy);
+
+    if (!diff.differencesFound()) {
+      // actual and values have the same elements but are they in the same order ?
+      for (int i = 0; i < actualAsList.size(); i++) {
```
GaspardPO: Maybe it could be cleaner to extract the for loop in a specific method, and it will allow to remove the comment and express it via the method name?

joel-costigliola: agreed

BillyYccc: Thanks for your advice! I have extracted the code into a new method.😄

BillyYccc:
@joel-costigliola any plan for this?

joel-costigliola:
for 3.10!

src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
```diff
@@ -526,6 +526,114 @@ public SELF contains(Iterable<? extends CharSequence> values) {
     return myself;
   }
 
+  /**
+   * Verifies that the actual {@code CharSequence} contains exactly all the values of the given charSequences <b>in a sequential order with no more or less elements</b>.
+   * <p>
```
joel-costigliola: add a note to inform the given values must not contain null elements

src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
```diff
@@ -526,6 +526,114 @@ public SELF contains(Iterable<? extends CharSequence> values) {
...
+   *
+   * // this assertion fails because of the wrong order
+   * assertThat(actual).containsExactly(&quot;, bar&quot;, &quot;foo&quot;);
+   *
+   * // this assertion fails because of unmatched elements
+   * assertThat(actual).containsExactly(&quot;foo&quot;);</code></pre>
+   *
+   * @param values the charSequences to check.
+   * @return {@code this} assertion object.
+   * @throws NullPointerException if the array of values is {@code null}.
```
joel-costigliola: `if the array or one of its values is {@code null}.`

src/main/java/org/assertj/core/internal/Strings.java
```diff
@@ -767,6 +774,67 @@ private void assertNotNull(AssertionInfo info, CharSequence actual) {
...
+   * @throws AssertionError if the given {@code CharSequence} does not contain exactly the values in a sequential order.
+   */
+  public void assertContainsExactly(AssertionInfo info, CharSequence actual, CharSequence[] values) {
+    doCommonCheckForCharSequence(info, actual, values);
+
+    List<String> actualAsList = splitCharSequencesToListOfElements(actual);
+    List<String> valuesAsList = splitCharSequencesToListOfElements(values);
+
+    if (actualAsList.size() != valuesAsList.size()) {
+      throw failures.failure(info, shouldHaveSameSizeAs(actual, actualAsList.size(), valuesAsList.size()));
```
joel-costigliola: we removed the check on sizes as the error message was not informative enough https://github.com/joel-costigliola/assertj-core/issues/1132

src/main/java/org/assertj/core/internal/Strings.java
```diff
@@ -767,6 +774,67 @@ private void assertNotNull(AssertionInfo info, CharSequence actual) {
...
+    }
+
+    IterableDiff diff = diff(actualAsList, valuesAsList, comparisonStrategy);
+
+    if (!diff.differencesFound()) {
+      checkIdenticalElementsAreInSameOrder(info, actualAsList, valuesAsList);
+      return;
+    }
+
```
joel-costigliola: Although the algorithm is correct, the error reported with `shouldContainExactly` is not user friendly as it shows an array of the missing/extra characters instead of missing/extra strings. 
By splitting strings into chars, we loose how users have grouped chars into strings parameters and not a collection of chars.

I would rather show the diff with the missing/extra strings instead, for example
```
assertThat("Practice makes perfect!").containsExactly("Practice ", "makes ", "!");
```
should show something lie
```
java.lang.AssertionError: 
Expected: <java.lang.AssertionError: 
Expecting:
  <"Practice makes perfect!">
to contain exactly in any order:
  <["Practice ", "makes ", "!"]>
but the following elements were not found:
  <["perfect"]>
>
```
I haven't thought of a nice way of doing that but let's give it a try.

src/test/java/org/assertj/core/internal/strings/Strings_assertContainsExactlyInAnyOrder_Test.java
```diff
@@ -0,0 +1,123 @@
...
+import static org.assertj.core.util.FailureMessages.*;
+
+/**
+ * Tests for <code>{@link Strings#assertContainsExactlyInAnyOrder(AssertionInfo, CharSequence, CharSequence[])}</code>.
+ *
+ * @author Billy Yuan
+ */
+public class Strings_assertContainsExactlyInAnyOrder_Test extends StringsBaseTest {
+  String actual = "Practice makes perfect!";
+
```
joel-costigliola: add some tests where actual and expected have duplicates (we had some bugs in iterable assertions related to duplicates)

BillyYccc:
@joel-costigliola I'm really sorry for the late reply, I have made some changes according to your feedback.
The algorithm is not yet finished, but I think it will be difficult to handle the string in words because we don't give a clear definition of the bound of words. 
Imagine you write an assertion like
`assertThat("Practice makes perfect!").containsExactly("Practisse ", "made ", "perffect!");` or 
`assertThat("Practicemakesperfect!").containsExactly("perfect!", "makes ", "Practice");`, what will be the output of the assertion error?

Maybe we should still use the size check and if size is different we throw a `shouldHaveSameSize` exception, and check the char one by one but (something different here from the previous impl) regard the chars as a string if they are next to each other, this way would show something meaningful when the minor typos appear but can do nothing about missing or unexpected elements(I just think it only makes sense to compare when the size is same).

for example, a typo scenario here:
```
assertThat("Practice madesperfect !").containsExactly("Practice ", "makes ", "perfect!");
```

shows something like: 
```
java.lang.AssertionError: 
Expected: <java.lang.AssertionError: 
Expecting:
  <"Practice madesperfect !">
to contain exactly and in order:
  <["Practice ", "makes ", "perfect!"]>
  <["k"," perfect!"]>
and others were not expected:
  <["d", "perfect !"]>
```

as for the different size scenario 
```
assertThat("Practice makes perfect!").containsExactly("Practice ", "makes ", "!");
```

we handle like this:
```
java.lang.AssertionError: 
Expected: <java.lang.AssertionError: 
Expecting:
  <"Practice makes perfect!">
to contain exactly and in order:
  <["Practice ", "makes ", "!"]>
but actual and expected don't have same numbers of characters.
```

WDYK?

joel-costigliola:
I need to think about this, I'll try to give you my opinion by the end of the week.

I wanted to thank for your work, don't worry about delays, we all have important things to do and not enough to do them all!

joel-costigliola:
I think if there are missing elements (even there is only character diff), the whole expected elements should be reported as missing and not only reporting the diff, on your example:
```java
assertThat("Practice madesperfect !").containsExactly("Practice ", "makes ", "perfect!");
```
should return an error like:
```
Expecting:
  <"Practice madesperfect !">
to contain exactly and in order:
  <["Practice ", "makes ", "perfect!"]>
  <["makes"," perfect!"]>
```
`"makes"` is reported as missing although there is only a one char diff.

joel-costigliola:
Another type of error you pointed out is when all the expected elements are found but actual is bigger:
```java
assertThat("Practice makes perfect!").containsExactly("Practice ", "makes ", "!");
```
would ideally report an error like:
```
Expecting:
  <"Practice makes perfect!">
to contain exactly and in order:
  <["Practice ", "makes ", "!"]>
but the following elements were not expected:
  <["perfect"]>
```
if that is not possible, I'm happy with the size check.

joel-costigliola:
Closing due to lack of interest and potential complexity of the assertion but happy to reopen it if someones is interested.

PR #1139, opened 2017-12-23
Show diff when containsExactlyElementsOf arrays have different sizes

### LINKED ISSUE #1132: containsExactlyElementsOf should display the diff when the size differ
Using containsExactlyElementsOf for a list with more than 1000 entries does not show the missing elements if they have an index > 1000 because the error message truncates after 1000 element entries with ...

```java
assertThat(List<String> a).containsExactlyElementsOf(List<String> b);
```

### LINKED ISSUE #1136: `isEqualToComparingFieldByFieldRecursively()` throws Exception when using percentage '%' in string field

ChrisCanCompute:
This is most useful for large lists (which now has a unit test).

#### Check List:
* Fixes #1132
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/error/AbstractShouldContainExactly.java
+++ b/src/main/java/org/assertj/core/error/AbstractShouldContainExactly.java
@@ -1,36 +1,17 @@
-/**
- * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
- * the License. You may obtain a copy of the License at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
- * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
- * specific language governing permissions and limitations under the License.
- *
- * Copyright 2012-2017 the original author or authors.
- */
 package org.assertj.core.error;
 
-import java.util.List;
-
 import org.assertj.core.description.Description;
-import org.assertj.core.presentation.Representation;
-import org.assertj.core.util.diff.Delta;
 
-/**
- * Base class for text content error.
- */
-public class AbstractShouldHaveTextContent extends BasicErrorMessageFactory {
+public class AbstractShouldContainExactly extends BasicErrorMessageFactory {
 
   protected String diffs;
 
-  public AbstractShouldHaveTextContent(String format, Object... arguments) {
+  public AbstractShouldContainExactly(String format, Object... arguments) {
     super(format, arguments);
   }
 
   @Override
-  public String create(Description d, Representation representation) {
+  public String create(Description d) {
     // we append diffs here as we can't add in super constructor call, see why below.
     //
     // case 1 - append diffs to String passed in super :
@@ -45,6 +26,6 @@ public class AbstractShouldHaveTextContent extends BasicErrorMessageFactory {
     //
     // The solution is to keep diffs as an attribute and append it after String.format has been applied on the error
     // message.
-    return super.create(d, representation) + diffs;
+    return super.create(d) + diffs;
   }
 }
--- a/src/main/java/org/assertj/core/error/AbstractShouldContainExactly.java
+++ b/src/main/java/org/assertj/core/error/AbstractShouldContainExactly.java
@@ -4,7 +4,7 @@ import org.assertj.core.description.Description;
 
 public class AbstractShouldContainExactly extends BasicErrorMessageFactory {
 
-  protected String diffs;
+  protected String diffs = "";
 
   public AbstractShouldContainExactly(String format, Object... arguments) {
     super(format, arguments);
--- a/src/main/java/org/assertj/core/error/ShouldContainExactly.java
+++ b/src/main/java/org/assertj/core/error/ShouldContainExactly.java
@@ -12,10 +12,21 @@
  */
 package org.assertj.core.error;
 
+import static java.lang.String.format;
+import static org.assertj.core.util.Arrays.isArray;
 import static org.assertj.core.util.IterableUtil.isNullOrEmpty;
+import static org.assertj.core.util.IterableUtil.toArray;
+import static org.assertj.core.util.diff.DiffUtils.diffsAsString;
 
+import org.assertj.core.api.exception.RuntimeIOException;
 import org.assertj.core.internal.ComparisonStrategy;
+import org.assertj.core.internal.Diff;
 import org.assertj.core.internal.StandardComparisonStrategy;
+import org.assertj.core.util.Arrays;
+import org.assertj.core.util.diff.Delta;
+
+import java.io.IOException;
+import java.util.List;
 
 /**
  * Creates an error message indicating that an assertion that verifies a group of elements contains exactly a given set
@@ -24,7 +35,9 @@ import org.assertj.core.internal.StandardComparisonStrategy;
  * 
  * @author Joel Costigliola
  */
-public class ShouldContainExactly extends BasicErrorMessageFactory {
+public class ShouldContainExactly extends AbstractShouldContainExactly {
+
+  private static final Diff DIFF = new Diff();
 
   /**
    * Creates a new <code>{@link ShouldContainExactly}</code>.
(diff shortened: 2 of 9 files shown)


src/main/java/org/assertj/core/error/AbstractShouldContainExactly.java
```diff
@@ -0,0 +1,31 @@
+package org.assertj.core.error;
```
ChrisCanCompute: This is an almost direct copy of AbstractShouldHaveTextContent.java

src/main/java/org/assertj/core/error/ShouldContainExactly.java
```diff
@@ -92,36 +105,49 @@ public static ErrorMessageFactory shouldContainExactly(Object actual, Object exp
 
   public static ErrorMessageFactory shouldHaveSameSize(Object actual, Object expected, int actualSize,
                                                        int expectedSize, ComparisonStrategy comparisonStrategy) {
-    return StandardComparisonStrategy.instance().equals(comparisonStrategy) ?
-        new ShouldContainExactly(actual, expected, actualSize, expectedSize) :
-        new ShouldContainExactly(actual, expected, actualSize, expectedSize, comparisonStrategy);
+
+    Object[] actualArray = arrayOrIterableToArray(actual);
+    Object[] expectedArray = arrayOrIterableToArray(expected);
```
ChrisCanCompute: These two can be Iterables or Arrays. I can make this neater if I change the method signature, but that doesn't seem ideal either.

src/main/java/org/assertj/core/error/ShouldContainExactly.java
```diff
@@ -92,36 +105,49 @@ public static ErrorMessageFactory shouldContainExactly(Object actual, Object exp
...
+    try {
+      List<Delta<String>> diffs = DIFF.diff(actualArray, expectedArray);
```
ChrisCanCompute: This can throw (because internally it uses a BufferedReader), but I don't think it ever will.

src/main/java/org/assertj/core/error/ShouldContainExactly.java
```diff
@@ -92,36 +105,49 @@ public static ErrorMessageFactory shouldContainExactly(Object actual, Object exp
...
+    } catch (IOException e) {
+      throw new RuntimeIOException(format("Unable to verify text contents of path:<%s>", actual), e);
+    }
   }
 
-  private ShouldContainExactly(Object actual, Object expected, int actualSize, int expectedSize,
+  private static Object[] arrayOrIterableToArray(Object input) {
+    if (isArray(input)) {
+      // Convert to List and back to cope with primative arrays.
+      return toArray(Arrays.asList(input));
```
ChrisCanCompute: Some of the tests are passing primative arrays.

src/main/java/org/assertj/core/internal/Diff.java
```diff
@@ -66,11 +67,29 @@
...
+      readerFor(arrayToStringWithDelimeter(actual, org.assertj.core.util.Compatibility.System.lineSeparator())),
+      readerFor(arrayToStringWithDelimeter(expected, org.assertj.core.util.Compatibility.System.lineSeparator())));
+  }
+
   @VisibleForTesting
   public List<Delta<String>> diff(Path actual, String expected, Charset charset) throws IOException {
     return diff(newBufferedReader(actual, charset), readerFor(expected));
   }
 
+  private String arrayToStringWithDelimeter(Object[] input, String delimeter) {
```
ChrisCanCompute: Just calling .toString() on the array, puts it all on one line, which removes any benefit of using the Diff functionality, so this is used to put the array onto multiple lines.

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -934,9 +935,10 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
     checkIsNotNull(values);
     assertNotNull(info, actual);
     int actualSize = sizeOf(actual);
-    if (values.length != actualSize)
-      throw failures.failure(info, shouldHaveSameSize(actual, values, actualSize, values.length, comparisonStrategy));
-    assertHasSameSizeAs(info, actual, values); // include check that actual is not null
```
ChrisCanCompute: I removed this, as it wasn't doing anything (the null check the comment mentioned is done a few lines up).

src/main/java/org/assertj/core/error/ShouldContainExactly.java
```diff
@@ -92,36 +105,49 @@ public static ErrorMessageFactory shouldContainExactly(Object actual, Object exp
...
+    Object[] actualArray = arrayOrIterableToArray(actual);
+    Object[] expectedArray = arrayOrIterableToArray(expected);
+    try {
+      List<Delta<String>> diffs = DIFF.diff(actualArray, expectedArray);
+      final String diffString = diffsAsString(diffs);
+      return StandardComparisonStrategy.instance().equals(comparisonStrategy) ?
+        new ShouldContainExactly(diffString, actualSize, expectedSize) :
+        new ShouldContainExactly(diffString, actualSize, expectedSize, comparisonStrategy);
+    } catch (IOException e) {
+      throw new RuntimeIOException(format("Unable to verify text contents of path:<%s>", actual), e);
```
joel-costigliola: not sure that `path` is relevant here since actual is not a file

src/main/java/org/assertj/core/error/ShouldContainExactly.java
joel-costigliola: I would prefer we catch the `IOException` and rethrow it as a `RuntimeIOException` in `Diff` so that we don't pollute the calling code that can't do anything with the `IOException` anyway.

src/main/java/org/assertj/core/error/ShouldContainExactly.java
```diff
@@ -92,36 +105,49 @@ public static ErrorMessageFactory shouldContainExactly(Object actual, Object exp
...
+    }
   }
 
-  private ShouldContainExactly(Object actual, Object expected, int actualSize, int expectedSize,
+  private static Object[] arrayOrIterableToArray(Object input) {
+    if (isArray(input)) {
+      // Convert to List and back to cope with primative arrays.
+      return toArray(Arrays.asList(input));
+    } else {
+      return toArray((Iterable<?>) input);
```
joel-costigliola: are we sure that we will always have an `Iterable` here? 
If is safer to check that input is one and throw an `IllegalArgumentException` saying that it expected an array or an iterable but got something else.

src/main/java/org/assertj/core/error/ShouldContainExactly.java
```diff
@@ -92,36 +105,49 @@ public static ErrorMessageFactory shouldContainExactly(Object actual, Object exp
...
+      return StandardComparisonStrategy.instance().equals(comparisonStrategy) ?
+        new ShouldContainExactly(diffString, actualSize, expectedSize) :
+        new ShouldContainExactly(diffString, actualSize, expectedSize, comparisonStrategy);
+    } catch (IOException e) {
+      throw new RuntimeIOException(format("Unable to verify text contents of path:<%s>", actual), e);
+    }
   }
 
-  private ShouldContainExactly(Object actual, Object expected, int actualSize, int expectedSize,
+  private static Object[] arrayOrIterableToArray(Object input) {
```
joel-costigliola: This method belongs naturally to `org.assertj.core.util.Arrays`, let's move it there.
This should have some unit tests too.

src/main/java/org/assertj/core/error/ShouldContainExactly.java
```diff
@@ -92,36 +105,49 @@ public static ErrorMessageFactory shouldContainExactly(Object actual, Object exp
...
+    if (isArray(input)) {
+      // Convert to List and back to cope with primative arrays.
+      return toArray(Arrays.asList(input));
+    } else {
+      return toArray((Iterable<?>) input);
+    }
+  }
+
+  private ShouldContainExactly(String diffString, int actualSize, int expectedSize,
```
joel-costigliola: Rename `diffString` to `diff`

src/main/java/org/assertj/core/internal/Diff.java
```diff
@@ -66,11 +67,29 @@
     return diff(actual.toPath(), expected, charset);
   }
 
+  @VisibleForTesting
+  public List<Delta<String>> diff(Object[] actual, Object[] expected) throws IOException {
+    return diff(
+      readerFor(arrayToStringWithDelimeter(actual, org.assertj.core.util.Compatibility.System.lineSeparator())),
```
joel-costigliola: extract `org.assertj.core.util.Compatibility.System.lineSeparator()` to a local variable for readability

src/main/java/org/assertj/core/internal/Diff.java
```diff
@@ -66,11 +67,29 @@
     return diff(actual.toPath(), expected, charset);
   }
 
+  @VisibleForTesting
+  public List<Delta<String>> diff(Object[] actual, Object[] expected) throws IOException {
```
joel-costigliola: Add unit tests for this method.

joel-costigliola: let's get replace the `IOException` with `RuntimeIOException` as per my other comment.

src/main/java/org/assertj/core/internal/Diff.java
joel-costigliola: See if you can reuse `org.assertj.core.util.Strings.StringsToJoin` to achive the same result.

src/main/java/org/assertj/core/util/diff/DiffUtils.java
```diff
@@ -344,4 +344,18 @@
     return buffer;
   }
 
+  /**
+   * Convert the output of a diff to a string that can be used in e.g. an error message.
+   *
+   * @param diffsList the output of the diff.
+   * @return A string showing the output of the diff.
+   */
+  public static String diffsAsString(List<Delta<String>> diffsList) {
```
joel-costigliola: Add unit tests

src/main/java/org/assertj/core/util/diff/DiffUtils.java
```diff
@@ -344,4 +344,18 @@
...
+    StringBuilder stringBuilder = new StringBuilder();
+    for (Delta<String> diff : diffsList) {
+      stringBuilder.append(org.assertj.core.util.Compatibility.System.lineSeparator()).append(diff.toString());
```
joel-costigliola: extract to a local variable outside the loop

joel-costigliola:
@ChrisCanCompute I have chosen to address this issue a bit differently by simply removing the size check and just displaying the diff all the time. Thanks for work anyway!

PR #1142, opened 2017-12-25
[Issue #1140]Modify the IterableDiff class and rework assertContainsOnly Iterable/Array assertion

### LINKED ISSUE #1140: IterableDiff does not deal correctly with duplicate elements
#### Summary

 Reported by @BillyYccc in https://github.com/joel-costigliola/assertj-core/pull/1138.

#### Example

```
List<String> myList = Arrays.asList("$","$","#");
assertThat(myList).containsExactly("#", "#", "$");
```
AssertionError is this
```
java.lang.AssertionError: 
  <"$">
whereas expected element was:
  <"#">
```

The problem is that the two lists have different elements so the message is confusing.

#### Java 8 specific ?

BillyYccc:
#### Check List:
* Fixes #1140 
* Unit tests : NO

Modify the `IterableDiff ` class to make it aware of the differences even if there are **duplicate** elements.

Besides, `assertContainsOnly` iterable/array assertion used to be implemented with the `IterableDiff ` class which has an unexpected intention, so I rework impl of both assertions.

### DIFF
--- a/src/main/java/org/assertj/core/internal/Arrays.java
+++ b/src/main/java/org/assertj/core/internal/Arrays.java
@@ -208,11 +208,22 @@ public class Arrays {
 
   void assertContainsOnly(AssertionInfo info, Failures failures, Object actual, Object values) {
     if (commonChecks(info, actual, values)) return;
-    IterableDiff diff = diff(asList(actual), asList(values), comparisonStrategy);
-    if (diff.differencesFound())
-      throw failures.failure(info, shouldContainOnly(actual, values,
-                                                     diff.missing, diff.unexpected,
-                                                     comparisonStrategy));
+    List<Object> notExpected = asList(actual);
+    List<Object> notFound = asList(values);
+
+    for (Object value : asList(values)) {
+      if (iterableContains(notExpected, value)) {
+        iterableRemoves(notExpected, value);
+        iterableRemoves(notFound, value);
+      }
+    }
+
+    if (notExpected.isEmpty() && notFound.isEmpty()) return;
+
+    throw failures.failure(info, shouldContainOnly(actual, values,
+                                                   notFound, notExpected,
+                                                   comparisonStrategy));
+
   }
 
   void assertContainsOnlyNulls(AssertionInfo info, Failures failures, Object[] actual) {
@@ -298,6 +309,13 @@ public class Arrays {
 
   private void iterablesRemoveFirst(Collection<?> actual, Object value) {
     comparisonStrategy.iterablesRemoveFirst(actual, value);
+  }
+
+  /**
+   * Delegates to {@link ComparisonStrategy#iterableRemoves(Iterable, Object)}
+   */
+  private void iterableRemoves(Collection<?> actual, Object value) {
+    comparisonStrategy.iterableRemoves(actual, value);
   }
 
   void assertContainsSequence(AssertionInfo info, Failures failures, Object actual, Object sequence) {
--- a/src/main/java/org/assertj/core/internal/IterableDiff.java
+++ b/src/main/java/org/assertj/core/internal/IterableDiff.java
@@ -12,47 +12,54 @@
  */
 package org.assertj.core.internal;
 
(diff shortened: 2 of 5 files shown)


src/main/java/org/assertj/core/internal/Arrays.java
```diff
@@ -208,11 +208,22 @@ void assertDoesNotContain(AssertionInfo info, Failures failures, Object array, O
 
   void assertContainsOnly(AssertionInfo info, Failures failures, Object actual, Object values) {
     if (commonChecks(info, actual, values)) return;
-    IterableDiff diff = diff(asList(actual), asList(values), comparisonStrategy);
-    if (diff.differencesFound())
-      throw failures.failure(info, shouldContainOnly(actual, values,
-                                                     diff.missing, diff.unexpected,
-                                                     comparisonStrategy));
```
joel-costigliola: I might miss a point but if `IterableDiff` is fixed, why do we need to change this method?

src/main/java/org/assertj/core/internal/IterableDiff.java
```diff
@@ -12,47 +12,54 @@
...
+    return new IterableDiff(actual, expected, comparisonStrategy);
+  }
+
   boolean differencesFound() {
     return !unexpected.isEmpty() || !missing.isEmpty();
   }
-  
-  private List<Object> missingElements(Iterable<Object> actual, Iterable<Object> expected) {
+
+  private <T> List<Object> missingElements(Iterable<T> actual, Iterable<T> expected) {
```
joel-costigliola: Add some unit tests covering the previous faulty behavior to show that it now works.

src/main/java/org/assertj/core/internal/IterableDiff.java
```diff
@@ -12,47 +12,54 @@
...
         missing.add(element);
+      } else {
+        iterablesRemoveFirst(copyOfActual, element);
       }
     }
     return missing;
   }
 
-  private List<Object> unexpectedElements(Iterable<Object> actual, Iterable<Object> expected) {
+  private <T> List<Object> unexpectedElements(Iterable<T> actual, Iterable<T> expected) {
```
joel-costigliola: Add some unit tests covering the previous faulty behavior to show that it now works.

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -268,11 +275,22 @@ private void iterablesRemoveFirst(Iterable<?> actual, Object value) {
   public void assertContainsOnly(AssertionInfo info, Iterable<?> actual, Object[] values) {
     if (commonCheckThatIterableAssertionSucceeds(info, actual, values)) return;
 
-    IterableDiff diff = diff(newArrayList(actual), asList(values), comparisonStrategy);
-    if (diff.differencesFound())
-      throw failures.failure(info, shouldContainOnly(actual, values,
-                                                     diff.missing, diff.unexpected,
-                                                     comparisonStrategy));
```
joel-costigliola: same comment as in `Arrays`

joel-costigliola:
Thanks @BillyYccc, I have done a quick review, I will take a second look after you address my comments.

BillyYccc:
@joel-costigliola Thanks for reviewing! I have added some unit tests for `IterableDiff`.

The intention to change `assertContainsOnly` methods both in Iterable/Array is: 
They used to be implemented by old `IterableDiff`, which just compares the elements but does not consider duplicate elements.

And the intention of `assertContainsOnly` is to verify whether it contains only the given values and nothing else, in any order.(Duplicate elements are allowed in the given values)

After changing `IterableDiff` class, new `IterableDiff` will consider duplicate elements when find differences between iterables, which does not make sense in `assertContainsOnly` assertion.

So I made some changes to the impl of `assertContainsOnly`. 

Apart from this, I think the impl `containsExactlyInAnyOrder` assertion in Array/Iterable can make use of  `IterableDiff` now.

joel-costigliola:
Integrated thanks @BillyYccc!

PR #1148, opened 2018-01-07
.travis.yml: run javadoc:javadoc to detect javadoc errors

PascalSchumacher:
Currently the javadoc is in pristine state (no warnings or errors). 😄 

Let's make sure it stays that way and let travis create javadoc so javadoc errors of pull requests are detected early on.

What do you think?


joel-costigliola:
Good idea!

PR #1155, opened 2018-01-13
Fixes: #1145 - `isEqualToComparingFieldByFieldRecursively` throws Nul…

### LINKED ISSUE #1145: `isEqualToComparingFieldByFieldRecursively` throws NullPointerException while comparing null values
#### Summary

If a nested field is null while comparing equality for two objects with `isEqualToComparingFieldByFieldRecursively` a NullPointerException is thrown.

### Stacktrace

```
Exception in thread "main" java.lang.NullPointerException
	at org.assertj.core.error.ShouldBeEqualByComparingFieldByFieldRecursively.describeDifference(ShouldBeEqualByComparingFieldByFieldRecursively.java:67)
	at org.assertj.core.error.ShouldBeEqualByComparingFieldByFieldRecursively.shouldBeEqualByComparingFieldByFieldRecursive(ShouldBeEqualByComparingFieldByFieldRecursively.java:32)
	at org.assertj.core.internal.Objects.assertIsEqualToComparingFieldByFieldRecursively(Objects.java:728)
	at org.assertj.core.api.AbstractObjectAssert.isEqualToComparingFieldByFieldRecursively(AbstractObjectAssert.java:653)
...
```

#### Example

```java
class Main {

    static class Value {
        private final String v;

        public Value(String v) {
            this.v = v;
        }
    }

        Value actual = new Value(null); //swapping actual and expected leads to the same error
        Value expected = new Value("something");
        Assertions.assertThat(expected).isEqualToComparingFieldByFieldRecursively(actual);
    }

}
```

#### Java 8 specific ?

* YES
Tested with Version 3.9.0

siladu:
…lPointerException while comparing null values

#### Check List:
* Fixes #1145
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/error/ShouldBeEqualByComparingFieldByFieldRecursively.java
+++ b/src/main/java/org/assertj/core/error/ShouldBeEqualByComparingFieldByFieldRecursively.java
@@ -17,6 +17,7 @@ import static org.assertj.core.util.Strings.join;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 
 import org.assertj.core.internal.DeepDifference.Difference;
 import org.assertj.core.presentation.Representation;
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
@joel-costigliola I think we should also fix this in 2.x, what do you think?

joel-costigliola:
yes it should but I can take of that, thanks for the PR  I'll review it shortly

joel-costigliola:
fixed in master for 3.9.1

Skylark95:
@joel-costigliola @PascalSchumacher Will a 2.x release be done for this. Issue is still present as of 2.9.0.

joel-costigliola:
@Skylark95 will do in couple weeks

Issue #1144, opened 2017-12-30
Feature request : upgrade to junit 5


fduminy:
#### Summary

Upgrade dependencies to junit 5.

As far as I know, it's only needed to support SoftAssertions rule.
I guess it implies changing almost all classes testing assertj itself, right ?

#### Java 8 specific ?

* YES, since it's required by junit 5 => create Pull Request from the `master` branch

joel-costigliola:
@fduminy except for JUnit rulesm have you found any blockers to use the current AssertJ with JUnit 5?

mkobit:
I have used AssertJ extensively with JUnit 5 and not ran into any issues. The only thing I that might be useful would be to throw OTA4J types (from https://github.com/ota4j-team/opentest4j) instead of the `java.lang.AssertionError`. This might be better reflected in test reporting or IntelliJ but I really don't know.

PascalSchumacher:
AssertJ 3.9.0 should throw the opentest4j types (see e.g.: https://github.com/joel-costigliola/assertj-core/blob/cc178abc5871bc34525113550373ffd90e303034/src/main/java/org/assertj/core/error/ShouldBeEqual.java#L115)

mkobit:
Thanks @PascalSchumacher I didn't see that mentioned in the change log and am only in the process of upgrading today, but that looks good!

joel-costigliola:
@mkobit this is the [2.9.0 release notes (improvements section)](http://joel-costigliola.github.io/assertj/assertj-core-news.html#assertj-core-2.9.0), but I don't blame  you not reading the fine prints, could be good new year resolution though :smile:

mkobit:
Hah maybe you should blame me. I searched the page for "ota" and not "opentest4j" 🤦.

joel-costigliola:
I blame you @mkobit!  :joy:

filiphr:
@mkobit just keep in mind that if JUnit 4 is on the classpath then it would use `ComparisonFailure`, as we don't know whether we are in a JUnit 4 or 5 test

uhafner:
Should throwing of opentest4j exceptions also work for custom assertions? Seems that this part is using some other technique. I have an assertion like the following:
```
    public IssueAssert hasDescription(final String description) {
        isNotNull();

        if (!Objects.equals(actual.getDescription(), description)) {
            failWithMessage(EXPECTED_BUT_WAS_MESSAGE, "description", actual, description, actual.getDescription());
        }
[...]

PR #1163, opened 2018-01-23
execute tests on junit platform

epeee:
a very first step towards #1144

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,6 +8,12 @@
   <name>AssertJ fluent assertions</name>
   <description>Rich and fluent assertions for testing for Java</description>
   <inceptionYear>2014</inceptionYear>
+  <properties>
+    <junit.version>4.12</junit.version>
+    <junit.jupiter.version>5.0.3</junit.jupiter.version>
+    <junit.vintage.version>${junit.version}.3</junit.vintage.version>
+    <junit.platform.version>1.0.3</junit.platform.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -91,6 +97,40 @@
...
+      <scope>test</scope>
+    </dependency>
+    <!-- Only required to run tests in an IDE that bundles an older version -->
+    <dependency>
+      <groupId>org.junit.vintage</groupId>
+      <artifactId>junit-vintage-engine</artifactId>
+      <version>${junit.vintage.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <!-- To avoid compiler warnings about @API annotations in JUnit code -->
```
filiphr: I think that they returned the transitive dependency on apiguardian with [5.0.3](http://junit.org/junit5/docs/current/user-guide/#release-notes-5.0.3-overall-improvements)

filiphr: Just adding that part as piece of information, not that it is something crucial. I am not member of the project, but I am excited for adding JUnit 5 :)

epeee: @filiphr thx for pointing that out!
I will update the pr accordingly.

joel-costigliola:
I believe this has been addressed in https://github.com/joel-costigliola/assertj-core/commit/393b9f36529746d646d6d5629354b5cb0455ad9e

PR #1170, opened 2018-02-01
Fixes: #1164 AbstractZonedDateTimeAssert.isEqualTo(ZonedDateTime expe…

### LINKED ISSUE #1164: AbstractZonedDateTimeAssert.isEqualTo(ZonedDateTime expected) throws NPE
#### Summary
Asserting equality between two ZonedDateTime instances that are null throws NullPointerException. This does not happen, for instance, with LocalDateTime.

#### Example

```java
        ZonedDateTime a = null;
        ZonedDateTime b = null;
        assertThat(a).isEqualTo(b);  // throws NPE
```

pswiatowski:
…cted) throws NPE

#### Check List:
* Fixes #1164 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractZonedDateTimeAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractZonedDateTimeAssert.java
@@ -414,10 +414,14 @@ public abstract class AbstractZonedDateTimeAssert<SELF extends AbstractZonedDate
    *
    * @param expected the given value to compare the actual value to.
    * @return {@code this} assertion object.
+   * @throws AssertionError if the actual {@code ZonedDateTime} is {@code null}.
+   * @throws IllegalArgumentException if other {@code ZonedDateTime} is {@code null}.
    * @throws AssertionError if the actual {@code ZonedDateTime} is not equal to the {@link ZonedDateTime} in the actual
    *           ZonedDateTime's java.time.ZoneId.
    */
   public SELF isEqualTo(ZonedDateTime expected) {
+    Objects.instance().assertNotNull(info, actual);
+    assertDateTimeParameterIsNotNull(expected);
     return super.isEqualTo(sameInstantInActualTimeZone(expected));
   }
 
@@ -447,6 +451,7 @@ public abstract class AbstractZonedDateTimeAssert<SELF extends AbstractZonedDate
    *           given String.
    */
   public SELF isEqualTo(String dateTimeAsString) {
+    Objects.instance().assertNotNull(info, actual);
     assertDateTimeAsStringParameterIsNotNull(dateTimeAsString);
     return super.isEqualTo(parse(dateTimeAsString));
   }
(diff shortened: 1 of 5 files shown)


pswiatowski:
@joel-costigliola ready for code review, thanks!

src/main/java/org/assertj/core/api/AbstractZonedDateTimeAssert.java
```diff
@@ -414,10 +414,14 @@ public SELF isEqualToIgnoringHours(ZonedDateTime other) {
...
    * @param expected the given value to compare the actual value to.
    * @return {@code this} assertion object.
+   * @throws AssertionError if the actual {@code ZonedDateTime} is {@code null}.
+   * @throws IllegalArgumentException if other {@code ZonedDateTime} is {@code null}.
    * @throws AssertionError if the actual {@code ZonedDateTime} is not equal to the {@link ZonedDateTime} in the actual
    *           ZonedDateTime's java.time.ZoneId.
    */
   public SELF isEqualTo(ZonedDateTime expected) {
+    Objects.instance().assertNotNull(info, actual);
+    assertDateTimeParameterIsNotNull(expected);
```
joel-costigliola: to be consistent with `AbstractAssert.isEqualTo`, this assertion should pass when both actual and expected are null.

src/main/java/org/assertj/core/api/AbstractZonedDateTimeAssert.java
```diff
@@ -459,10 +464,14 @@ public SELF isEqualTo(String dateTimeAsString) {
    *
    * @param expected the given value to compare the actual value to.
    * @return {@code this} assertion object.
+   * @throws AssertionError if the actual {@code ZonedDateTime} is {@code null}.
+   * @throws IllegalArgumentException if other {@code ZonedDateTime} is {@code null}.
    * @throws AssertionError if the actual {@code ZonedDateTime} is equal to the {@link ZonedDateTime} in the actual
    *           ZonedDateTime's java.time.ZoneId.
    */
   public SELF isNotEqualTo(ZonedDateTime expected) {
+    Objects.instance().assertNotNull(info, actual);
```
joel-costigliola: This should be consistent with `AbstractAssert.isNotEqualTo`, if actual is null and expected is not, the assertion should succeed and vice versa.

src/main/java/org/assertj/core/api/AbstractZonedDateTimeAssert.java
```diff
@@ -487,6 +496,7 @@ public SELF isNotEqualTo(ZonedDateTime expected) {
    *           String.
    */
   public SELF isNotEqualTo(String dateTimeAsString) {
+    Objects.instance().assertNotNull(info, actual);
```
joel-costigliola: I might miss something but I believe this is not needed as `super.isNotEqualTo` will fail is actual is null since expected is not.

src/main/java/org/assertj/core/api/AbstractZonedDateTimeAssert.java
pswiatowski: Right! I will work on this. Thanks for review and suggestion.

src/main/java/org/assertj/core/api/AbstractZonedDateTimeAssert.java
pswiatowski: Yes, same as above comment.

src/main/java/org/assertj/core/api/AbstractZonedDateTimeAssert.java
pswiatowski: Correct, it is not needed. I will add one more unit test for that case.

pswiatowski:
@joel-costigliola I applied the fixes and I added new unit tests to verify all tests cases with ZoneDateTime.

joel-costigliola:
Integrated thanks @pswiatowski !

pswiatowski:
@joel-costigliola that is great!

PR #1171, opened 2018-02-03
Added isEqualTo(long) for IntegerAssert

### LINKED ISSUE #1162: Add isEqualTo(long) for IntegerAssert
#### Summary

Add ```isEqualTo(int)``` method to ```AbstractLongAssert``` and a ```@Deprecated``` override of ```isEqualTo(Object)``` to avoid use by mistake.

#### Rationale

I may be missing something obvious, but I can't understand why there is no isEqualTo(long) overload in assertions for ```int``` type like there are for BigInteger.

A typical pitfall I see is:
```java
assertThat(foo.getBar()).isEqualTo(5L)
```

If ```getBar()``` returns ```long``` or ```BigInteger``` with a numeric value of 5, then this passes as expected. But if the return type is ```int``` that it will fail as the generic ```isEqualTo(Object)``` will be chosen (which will never succeed if the passed object is not an instance of Integer, but than isEqualTo(int) would be used so you never want to call it).

####

epeee:
#### Check List:
* Fixes #1162
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractIntegerAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractIntegerAssert.java
@@ -70,6 +70,33 @@ public abstract class AbstractIntegerAssert<SELF extends AbstractIntegerAssert<S
   }
 
   /**
+   * Verifies that the actual value is equal to the given one.
+   * <p>
+   * Example:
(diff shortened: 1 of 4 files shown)


joel-costigliola:
Integrated thanks !

PR #1179, opened 2018-02-08
Fixes: #1167 CompletionStage should print stacktrace on failed assert…

### LINKED ISSUE #1167: CompletionStage should print stacktrace on failed assertion with failed-future
#### Summary

When asserting non exception result about CompletionStage / Future like isNotDone or isCompleted, assertJ reports something like this:

```
Expecting
  <CompletableFuture[Failed: java.lang.NullPointerException]>
not to be done
```

In such cases, it would be useful for the Exception Stacktrace to be printed, IMO, as a good start for debugging the cause of the test failure.

pswiatowski:
…ion with failed-future

#### Check List:
* Fixes #1167 
* Unit tests : YES


pswiatowski:
@joel-costigliola As I mentioned before that the stacktrace is different and we cannot assert message 1to1 so I introduced stacktrace regex and I used hasMessageMatching method to assert the message for failed future.

pswiatowski:
@joel-costigliola can you take a look to that fix? thanks!

src/main/java/org/assertj/core/presentation/StandardRepresentation.java
```diff
@@ -289,7 +289,7 @@ protected String toStringOf(CompletableFuture<?> future) {
       Object joinResultRepresentation = joinResult instanceof CompletableFuture ? joinResult : toStringOf(joinResult);
       return concat(className, "[Completed: ", joinResultRepresentation, "]");
     } catch (CompletionException e) {
-      return concat(className, "[Failed: ", toStringOf(e.getCause()), "]");
+      return concat(className, "[Failed: ", toStringOf(e.getCause()), "]\n", toStringOf(e.getStackTrace()));
```
joel-costigliola: we already display stack trace in error like [`ShouldBeInstance`](https://github.com/joel-costigliola/assertj-core/blob/cc178abc5871bc34525113550373ffd90e303034/src/main/java/org/assertj/core/error/ShouldBeInstance.java#L68), let's keep on doing that for this issue.
Replace `\n` by `String.format("%n")` for consistency.
We might end up with something like `String.format("%n%s", getStackTrace(e))`  ...

src/test/java/org/assertj/core/util/StackTraceUtils.java
```diff
@@ -28,4 +28,10 @@ public static boolean hasStackTraceElementRelatedToAssertJ(Throwable throwable)
     return false;
   }
 
+  /**
+   * Returns regex which match stack trace.
+   */
+  public static String getStackTraceRegex() {
+    return "\\[(?s).*[\\n\\r].*\\]";
```
joel-costigliola: let's avoid regex as much as we can, I only use regex when everything else fails since they are really unreadable and easy to screw up.

src/test/java/org/assertj/core/presentation/StandardRepresentation_format_CompletableFuture_Test.java
```diff
@@ -38,7 +39,8 @@ public void should_format_complete_future() {
   public void should_format_failed_future() {
     CompletableFuture<Object> future = new CompletableFuture<>();
     future.completeExceptionally(new RuntimeException("some random error"));
-    assertThat(STANDARD_REPRESENTATION.toStringOf(future)).isEqualTo("CompletableFuture[Failed: java.lang.RuntimeException: some random error]");
+    assertThat(STANDARD_REPRESENTATION.toStringOf(future)).matches("CompletableFuture\\[Failed: java.lang.RuntimeException: some random error\\]\n"
```
joel-costigliola: avoid using `\n`, I remember windows contributors having tests failing because of such usage.

src/test/java/org/assertj/core/presentation/StandardRepresentation_format_CompletableFuture_Test.java
```diff
@@ -38,7 +39,8 @@ public void should_format_complete_future() {
...
+                                                                   + StackTraceUtils.getStackTraceRegex());
```
joel-costigliola: use `hasMessageContaining` if possible

src/test/java/org/assertj/core/api/future/CompletableFutureAssert_hasNotFailed_Test.java
```diff
@@ -55,6 +56,7 @@ public void should_fail_if_completable_future_has_failed() {
     future.completeExceptionally(new RuntimeException());
 
     assertThatThrownBy(() -> assertThat(future).hasNotFailed()).isInstanceOf(AssertionError.class)
-                                                               .hasMessage(shouldNotHaveFailed(future).create());
+                                                               .hasMessageMatching(shouldNotHaveFailed(future).create(emptyDescription(),
+                                                                                                              new FailedCompletableFutureRepresentation()));
```
joel-costigliola: I'm not too keen about having a `FailedCompletableFutureRepresentation` that kind of duplicate the production code.

I believe we should simply verify that some subpart of the stack trace is present  (a subpart that can only be in the stack trace) either with `hasMessageMatching` or `hasMessageContaining`, the rationale is that if such subpart exists then the whole stack trace should be there too.

joel-costigliola:
Integrated thanks @pswiatowski.

PR #1185, opened 2018-02-19
Fix 1149

### LINKED ISSUE #1149: CompletionStage asserts about failed state should fail when not Done, and only join() else
`hasFailed()` assertion currently passes when the Future/CompletionStage is not yet done, causing brittle tests and race conditions. `hasFailedWithThrowableThat()` on the other hand joins, so it's also an inconsistency.   hasFailedWithThrowableThat() should only join() when done. joining else may cause infinitely running tests, which is undesirable even if it's the original test cases fault.

* [x]  hasFailed() should fail when not done.
* [ ] hasNotFailed() should fail when not done.
* [x]  hasFailedWithThrowableThat() should only join() when done. 

Discussed in #1010

pswiatowski:
#### Check List:
* Fixes #1149 
* Unit tests : YES

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -117,12 +117,8 @@
               <createDependencyReducedPom>true</createDependencyReducedPom>
               <relocations>
                 <relocation>
-                  <pattern>net.sf.cglib</pattern>
-                  <shadedPattern>org.assertj.core.internal.cglib</shadedPattern>
-                </relocation>
-                <relocation>
-                  <pattern>org.objectweb.asm</pattern>
-                  <shadedPattern>org.assertj.core.internal.asm</shadedPattern>
+                  <pattern>net.bytebuddy</pattern>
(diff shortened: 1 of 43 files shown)


scordio:
Closing as mentioned in https://github.com/assertj/assertj-core/issues/1149#issuecomment-752016727:

> This issue is out of date as `hasFailed` has been deprecated in favor of  [`failsWithin`](https://www.javadoc.io/static/org.assertj/assertj-core/3.18.1/org/assertj/core/api/AbstractCompletableFutureAssert.html#failsWithin(java.time.Duration)). [hasNotFailed](https://www.javadoc.io/static/org.assertj/assertj-core/3.18.1/org/assertj/core/api/AbstractCompletableFutureAssert.html#hasNotFailed()) has been deprecated too.

PR #1200, opened 2018-03-10
#1197 fix of extracting/flatExtracting method not having proper description if set (for 2.x branch)

### LINKED ISSUE #1197: As doesn't seem to work correctly when using extracting with a method reference
It's entirely possible that I'm doing something wrong here but here's what I've noticed:

```java
assertThat(myListOfObjects)
                .as("Check that a list of things contains a thing with this property value")
                .extracting(MyClass::getAProperty).contains("Something");
```

The assertion itself works correctly but when the assertion fails the description text isn't printed. I just get:

```java
java.lang.AssertionError: 
Expecting:
 <["LoremIpsum", "dolor sit"]>
to contain:
 <["Something"]>
but could not find:
 <["Something"]>
```

If I change the assertion to: 

```java
assertThat(myListOfObjects)
                .as("Check that a list of things contains a thing with this property value")
                .extracting("propertyName").contains("Something");
```

then the assertion description is printed correctly:

```java
java.lang.AssertionError: [Check that a list of things contains a thing with this property value] 
Expecting:
 <["LoremIpsum", "dolor sit"]>
to contain:
 <["Something"]>
but could not find:
 <["Something"]>
 ```

nach-o-man:
#### Check List:
* Fixes #1197 fix of extracting/flatExtracting method not having proper description if set (for 2.x branch)
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
@@ -1110,7 +1110,7 @@ public abstract class AbstractIterableAssert<SELF extends AbstractIterableAssert
   @CheckReturnValue
   public <V> AbstractListAssert<?, List<? extends V>, V, ObjectAssert<V>> extracting(Extractor<? super ELEMENT, V> extractor) {
     List<V> values = FieldsOrPropertiesExtractor.extract(actual, extractor);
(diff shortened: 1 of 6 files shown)


joel-costigliola:
Integrated thanks @nach-o-man!

PR #1201, opened 2018-03-11
#1197 fix of new extracting/flatExtracting methods not having proper description if set (for master)

### LINKED ISSUE #1197: As doesn't seem to work correctly when using extracting with a method reference

nach-o-man:
#### Check List:
* Fixes #1197 fix of new extracting/flatExtracting methods not having proper description if set (for master)
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractObjectArrayAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractObjectArrayAssert.java
@@ -2060,7 +2060,7 @@ public abstract class AbstractObjectArrayAssert<SELF extends AbstractObjectArray
   @CheckReturnValue
   public <V, EXCEPTION extends Exception> AbstractListAssert<?, List<? extends V>, V, ObjectAssert<V>> extracting(ThrowingExtractor<? super ELEMENT, V, EXCEPTION> extractor) {
     List<V> values = FieldsOrPropertiesExtractor.extract(newArrayList(actual), extractor);
-    return newListAssertInstance(values);
+    return newListAssertInstance(values).as(info.description());
   }
 
   /**
--- a/src/test/java/org/assertj/core/api/iterable/IterableAssert_extracting_Test.java
+++ b/src/test/java/org/assertj/core/api/iterable/IterableAssert_extracting_Test.java
@@ -64,6 +64,14 @@ public class IterableAssert_extracting_Test {
       return input.getAge();
     }
   };
+  
+  private static final ThrowingExtractor<Employee, Object, Exception> throwingExtractor = new ThrowingExtractor<Employee, Object, Exception>() {
+    @Override
+    public Object extractThrows(Employee employee) throws Exception {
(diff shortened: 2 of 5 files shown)


src/main/java/org/assertj/core/api/AbstractObjectArrayAssert.java
```diff
@@ -2572,7 +2572,7 @@ public SELF inBinary() {
     checkNotNull(filterOperator);
     Filters<? extends ELEMENT> filter = filter(actual).with(propertyOrFieldName);
     filterOperator.applyOn(filter);
-    return newListAssertInstance(newArrayList(filter.get())).as(info.description());
+    return newListAssertInstance(newArrayList(filter.get()));
```
joel-costigliola: why not propagating the description ?

src/test/java/org/assertj/core/api/iterable/IterableAssert_extracting_Test.java
```diff
@@ -65,6 +65,14 @@ public Integer extract(Employee input) {
     }
   };
 
+  private static final ThrowingExtractor<Employee, Object, Exception> throwingExtractor = new ThrowingExtractor<Employee, Object, Exception>() {
```
joel-costigliola: use a lambda since it's for java 8

nach-o-man: ok

joel-costigliola: @nach-o-man I'm currently integrating it, don't worry, thanks!

src/main/java/org/assertj/core/api/AbstractObjectArrayAssert.java
nach-o-man: It was brought by mistake in previous pull request and isn't covered by units.
also It prevents compilation on JDK9.
Do you think I should bring it back?

PR #1206, opened 2018-03-19
Proposed solution for #1205

### LINKED ISSUE #1205: AbstractListAssert allow chainable assertions at specific index
#### Summary
I'd like to propose a new assertion for `AbstractListAssert` that would allow chainable assertions on different elements of collection. This would match various `satisfy` assertions that are already present in other assertion classes.

#### Example

```java
List<String> actual = newArrayList("Leia", "Luke", "Yoda")

assertThat(actal)
    .satisfiesAt(str -> assertThat(str).isEqualTo("Luke"), atIndex(1))
    .satisfiesAt(
        str -> {
            assertThat(str)
                .isNotEmpty()
                .endsWith("da")
        }
        , atIndex(2))
```

airborn:
#### Check List:
* Fixes #1205 
* Unit tests : YES 

This is a proposed solution for #1205 
In case the idea is accepted I'll add a necessary documentation.


src/test/java/org/assertj/core/internal/lists/Lists_satisfiesAt_Test.java
```diff
@@ -0,0 +1,58 @@
...
+public class Lists_satisfiesAt_Test extends ListsBaseTest {
+
+  private final AssertionInfo info = someInfo();
+  private final Consumer<String> requirements = str -> Assertions.assertThat(str).isEqualTo("Luke");
+  private final Index index = atIndex(1);
+  private final List<String> actual = newArrayList("Leia", "Luke", "Yoda");
+
+  @Test
+  public void should_pass_if_actual_match_requirements() {
+    lists.satisfiesAt(info, actual, requirements, index);
```
GaspardPO: I don't get it: what is "lists" and what is "actual"?
It seems to me that it should look like:

```java
assertThat(actual).satisfiesAt(info, requirements, index);
```

joel-costigliola: that's a test for `Lists` which is the class that actually implements the assertion business logic

GaspardPO: ok, thanks for the explanation.

joel-costigliola:
@airborn integrated thanks with minor code changes, thanks!

airborn:
thanks @joel-costigliola 

Out of curiosity, what are those minor code changes? I have noticed you removed documentation from not public API and changed assertion error message in case of `null requirements` is there any thing else? I'm asking in case of any subsequent PRs.

joel-costigliola:
there was a typo in the javadoc code examples : `List<String> ringBearers = newArrayList(frodo, elrond, gandalf);`.
frodo, elrond, gandalf are not String but `TolkienCharacter` ;-).

I have slightly improved some test case names (to be closer to an English sentence expressing the expected behavior)

I removed the `<? super T>` in the error message as at runtime we can't tell what `T` is (but that's really a minor change).
I only care about the user facing API, I have found over time that internal classes (even public) API is not really useful and hard to maintain, as side note I value comments when it clarifies the code (for example I think every regex should be commented as they are usually not straightforward). 

Overall it was a good PR, I did not take too much time integrating it, good job!

airborn:
Thanks for the update :smile:

PR #1212, opened 2018-03-30
Rewriting README.md to make it more technical and clear

HugoHo0212:
-You can ask questions [**stackoverflow (assertj tag)**] -> You can ask questions in [**stackoverflow (assertj tag)**]
-The idea is that, when writing unit tests, we should have at our disposal assertions specific to the type of the objects we are checking -> The idea is that disposal assertions should be specific to the type of the objects we are checking when writing unit tests
-Use Map-specific assertions, which make it easy to check the contents of the map -> Use Map-specific assertions to easily check the contents of the map
-AssertJ's assertions read very close to plain English -> AssertJ's assertions are very close to plain English
-you might want some that are specific to your own types -> you might want some assertions specific to your own types
-This is possible with AssertJ because it is easily extensible so it's simple to [write assertions for your custom types] -> It is simple to [write assertions for your custom types] with AssertJ because it is easily extensible


joel-costigliola:
Awesome, thanks for the improvements!

PR #1218, opened 2018-04-04
Fix #1217: Close stream after materialisation in ListAssert

### LINKED ISSUE #1217: Consider closing the Stream in ListAssert
#### Summary

The current implementation of converting a stream to a list uses the following logic:

```java
// org.assertj.core.api.ListAssert.ListFromStream
private List<ELEMENT> initList() {
  if (list == null) list = Lists.newArrayList(stream.iterator());
  return list;
}
```

According to my own research and https://stackoverflow.com/a/40957355/345057, this does not close the underlying stream. Most of the time this is not a problem since usually streams do not hold a resource. However, it seems way too easy to create such a situation as the example shows.

#### Example

```java
final File myFile = new File("...");

// Best case: use of AssertJ API - proper resource handling
// org.assertj.core.api.Assertions#linesOf(java.io.File)
assertThat(linesOf(myFile)).contains("foo");

// Unlucky: when using java.nio.file.Files from the standard library (the "more obvious / common" solution),
// the underlying I/O stream is not closed
// Subsequent operations (such as removing the file in test cleanup routines) may fail depending on the OS.
assertThat(Files.lines(myFile.toPath())).contains("bar");
```

#### Conclusion

Closing the stream may provide us with increased safety of use. I am currently not aware of any downsides of closing the stream right after materialization into a list. 
If you agree this is a problem (and a proper fix), I would be happy to file a PR.

f4lco:
Streams may carry resources which require closing such as
java.nio.file.Files#lines(java.nio.file.Path).

#### Check List:
* Fixes #1217 
* Unit tests : NO

### DIFF
--- a/src/main/java/org/assertj/core/api/ListAssert.java
+++ b/src/main/java/org/assertj/core/api/ListAssert.java
@@ -226,7 +226,10 @@ public class ListAssert<ELEMENT> extends
     }
 
     private List<ELEMENT> initList() {
-      if (list == null) list = Lists.newArrayList(stream.iterator());
+      if (list == null) {
+        list = Lists.newArrayList(stream.iterator());
+        stream.close();
+      }
       return list;
     }
 


joel-costigliola:
Integrated thanks @f4lco !

PR #1221, opened 2018-04-07
Adding assertj-core as Automatic-Module-Name in Manifest file.

### LINKED ISSUE #1220: AssertJ Core Does Not Define an Automatic Module Name
#### Summary
Currently when compiling a project that uses assertj-core as a dependency in Java 9+ I receive the following warning from maven-compiler-plugin.
> Required filename-based automodules detected. Please don't publish this project to a public artifact repository!

The following should be added to the assertj-core pom.xml to generate an automatic module name in the Manifest file for maven to use.

```xml
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-jar-plugin</artifactId>
  <configuration>
    <archive>
      <manifestEntries>
        <Automatic-Module-Name>assertj-core</Automatic-Module-Name>
      </manifestEntries>
    </archive>
  </configuration>
</plugin>
```

andrewauclair:
#### Check List:
* Fixes #1220 
* Unit tests : NA

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,6 +96,17 @@
           <encoding>${project.build.sourceEncoding}</encoding>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifestEntries>
+              <Automatic-Module-Name>assertj-core</Automatic-Module-Name>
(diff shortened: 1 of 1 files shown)


joel-costigliola:
I agree with @filiphr, see also https://github.com/mockito/mockito/issues/1189.

joel-costigliola:
Integrated thanks @andrewauclair!

PR #1222, opened 2018-04-09
Make mockito to be a test-only dependency

testn:


joel-costigliola:
Good catch ! :man_facepalming:

PascalSchumacher:
It was already a test-only dependency, because version and scope were inherited from dependency management of the parent pom: https://github.com/joel-costigliola/assertj-maven-parent-pom/blob/master/pom.xml#L70

joel-costigliola:
Reverted, thanks @PascalSchumacher for pointing this out. 

@testn to update the version, we need to change pom.xml in https://github.com/joel-costigliola/assertj-maven-parent-pom.

testn:
oh thanks @PascalSchumacher  I didn't look there.

PR #1226, opened 2018-04-11
Offset JavaDoc: fix fail/pass condition pseudo code

valery1707:
#### Check List:
* Fixes: NA
* Unit tests: NA
* Javadoc with a code example (API only): NA


valery1707:
JDK9 fail not correlated with changes

joel-costigliola:
Good catch thanks @valery1707 !

PR #1230, opened 2018-04-16
Fix class loading strategy to no longer rely on sun.misc.Unsafe and several performance improvements

raphw:
- Updates to Byte Buddy 1.8.4.
- Adds support for class loading under Java 11 after changing the default class loading strategy in Byte Buddy 1.8.4.
- Make interception stateless to allow for better reuse of classes.

There are still a few test failures which seem related to filtering stack traces what might have changed after my recent alterations.

### DIFF
--- a/src/main/java/org/assertj/core/api/AssertJProxySetup.java
+++ b/src/main/java/org/assertj/core/api/AssertJProxySetup.java
@@ -1,0 +1,18 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2018 the original author or authors.
+ */
+package org.assertj.core.api;
+
+public interface AssertJProxySetup {
+
+  void assertj$setup(ProxifyMethodChangingTheObjectUnderTest proxifyMethodChangingTheObjectUnderTest, ErrorCollector errorCollector);
+}
--- a/src/main/java/org/assertj/core/api/Assumptions.java
+++ b/src/main/java/org/assertj/core/api/Assumptions.java
@@ -17,8 +17,10 @@ import static org.assertj.core.util.Arrays.array;
 
 import java.io.File;
 import java.io.InputStream;
+import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.net.URI;
@@ -64,6 +66,9 @@ import java.util.stream.IntStream;
 import java.util.stream.LongStream;
 import java.util.stream.Stream;
 
+import net.bytebuddy.dynamic.loading.ClassInjector;
+import net.bytebuddy.dynamic.loading.ClassLoadingStrategy;
+import net.bytebuddy.implementation.auxiliary.AuxiliaryType;
 import org.assertj.core.api.ThrowableAssert.ThrowingCallable;
 import org.assertj.core.util.CheckReturnValue;
 
@@ -84,14 +89,29 @@ import net.bytebuddy.matcher.ElementMatchers;
  */
 public class Assumptions {
 
-  private static final TypeCache<SimpleKey> CACHE = new TypeCache.WithInlineExpunction<>(Sort.SOFT);
-  private static ByteBuddy byteBuddy = new ByteBuddy().with(TypeValidation.DISABLED);
+  private static ByteBuddy BYTE_BUDDY = new ByteBuddy()
+    .with(TypeValidation.DISABLED)
+    .with(new AuxiliaryType.NamingStrategy.SuffixingRandom("Assertj$Assumptions"));
+
+  private static final MethodHandles.Lookup LOOKUP = MethodHandles.lookup();
+  private static final Method PRIVATE_LOOKUP_IN;
+
+  private static final TypeCache<TypeCache.SimpleKey> CACHE = new TypeCache.WithInlineExpunction<>(Sort.SOFT);
+
+  static {
+    Method privateLookupIn;
+    try {
+      privateLookupIn = MethodHandles.class.getMethod("privateLookupIn", Class.class, MethodHandles.Lookup.class);
+    } catch (Exception e) {
+      privateLookupIn = null;
+    }
+    PRIVATE_LOOKUP_IN = privateLookupIn;
+  }
 
   private static final class AssumptionMethodInterceptor {
 
     @RuntimeType
-    public Object intercept(@This AbstractAssert<?, ?> assertion,
-                            @SuperCall Callable<Object> proxy) throws Exception {
+    public static Object intercept(@This AbstractAssert<?, ?> assertion, @SuperCall Callable<Object> proxy) throws Exception {
       try {
         Object result = proxy.call();
         if (result != assertion && result instanceof AbstractAssert) {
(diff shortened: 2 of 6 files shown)


src/main/java/org/assertj/core/api/AssertJProxySetup.java
```diff
@@ -0,0 +1,18 @@
...
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2018 the original author or authors.
+ */
+package org.assertj.core.api;
+
+public interface AssertJProxySetup {
+
+  void assertj$setup(ProxifyMethodChangingTheObjectUnderTest proxifyMethodChangingTheObjectUnderTest, ErrorCollector errorCollector);
```
joel-costigliola: can we clarify what is the responsibility of this method ? and why there is no implementation of it ? (is that generated by byte buddy?)

src/main/java/org/assertj/core/api/Assumptions.java
```diff
@@ -84,14 +89,29 @@
  */
 public class Assumptions {
 
-  private static final TypeCache<SimpleKey> CACHE = new TypeCache.WithInlineExpunction<>(Sort.SOFT);
-  private static ByteBuddy byteBuddy = new ByteBuddy().with(TypeValidation.DISABLED);
+  private static ByteBuddy BYTE_BUDDY = new ByteBuddy()
+    .with(TypeValidation.DISABLED)
+    .with(new AuxiliaryType.NamingStrategy.SuffixingRandom("Assertj$Assumptions"));
```
joel-costigliola: why do we need a suffix here ?

src/main/java/org/assertj/core/api/Assumptions.java
```diff
@@ -1135,30 +1155,42 @@ public static OptionalLongAssert assumeThat(OptionalLong actual) {
...
-                    .intercept(MethodDelegation.to(new AssumptionMethodInterceptor()))
+                    .intercept(MethodDelegation.to(AssumptionMethodInterceptor.class))
                     .make()
-                    .load(Assumptions.class.getClassLoader())
-                    .getLoaded();
+                    .load(Assumptions.class.getClassLoader(), classLoadingStrategy(assertionType))
+                    .getLoaded());
+  }
+
+  private static ClassLoadingStrategy<ClassLoader> classLoadingStrategy(Class<?> assertClass) {
```
joel-costigliola: looks like this can be moved in a `ClassLoadingStrategyFactory` class as the same code is used in `SoftProxies`, can you confirm there is no negative effects of sharing the same `PRIVATE_LOOKUP_IN` ?

src/main/java/org/assertj/core/api/ErrorCollector.java
```diff
@@ -36,24 +39,28 @@
...
 
   @RuntimeType
-  public Object intercept(@This Object obj, @SuperCall Callable<Object> proxy,
-                          @SuperMethod(nullIfImpossible = true) Method method) throws Exception {
+  public static Object intercept(
+    @FieldValue(FIELD_NAME) ErrorCollector errorCollector,
+    @This Object obj,
+    @SuperCall Callable<?> proxy,
+    @SuperMethod(nullIfImpossible = true) Method method,
+    @StubValue Object stub) throws Exception {
```
joel-costigliola: Can you clarify why stub is needed here ? it's not clear to me (the previous code was not to clear neither to be fair)

src/main/java/org/assertj/core/api/ErrorCollector.java
```diff
@@ -36,24 +39,28 @@
...
-      lastResult.setSuccess(false);
-      errors.add(e);
+      errorCollector.lastResult.setSuccess(false);
+      errorCollector.errors.add(e);
     }
     if (method != null && !method.getReturnType().isInstance(obj)) {
       // In case the object is not an instance of the return type, just return null to avoid ClassCastException
       // with ByteBuddy
-      return null;
+      return stub;
```
joel-costigliola: I'll remove the comment above which does not make sense

src/main/java/org/assertj/core/api/SoftProxies.java
```diff
@@ -83,58 +106,78 @@ void collectError(Throwable error) {
...
     try {
-      ClassLoader classLoader = getClass().getClassLoader();
-      SimpleKey key = new SimpleKey(assertClass);
-      Class<V> proxyClass = (Class<V>) cache.findOrInsert(classLoader, key, () -> createProxy(assertClass, collector));
-
+      Class<? extends V> proxyClass = createProxy(assertClass);
       Constructor<? extends V> constructor = proxyClass.getConstructor(actualClass);
-      return constructor.newInstance(actual);
+      V instance = constructor.newInstance(actual);
+      ((AssertJProxySetup) instance).assertj$setup(new ProxifyMethodChangingTheObjectUnderTest(this), collector);
```
joel-costigliola: I don't get this line, how come `instance ` is a `AssertJProxySetup` ? I guess the created proxy class actually implements `AssertJProxySetup` but it feels like (black) magic to me.

joel-costigliola: ok this is explained by line 131 `.implement(AssertJProxySetup.class)` later on.

src/main/java/org/assertj/core/api/SoftProxies.java
```diff
@@ -83,58 +106,78 @@ void collectError(Throwable error) {
...
-                    .intercept(MethodDelegation.to(new ProxifyMethodChangingTheObjectUnderTest(this)))
+                    .intercept(MethodDelegation.to(ProxifyMethodChangingTheObjectUnderTest.class))
+                    .defineField(ErrorCollector.FIELD_NAME, ErrorCollector.class, Visibility.PRIVATE)
                     .method(any().and(not(METHODS_CHANGING_THE_OBJECT_UNDER_TEST))
                                  .and(not(METHODS_NOT_TO_PROXY)))
-                    .intercept(MethodDelegation.to(collector))
+                    .intercept(MethodDelegation.to(ErrorCollector.class))
+                    .implement(AssertJProxySetup.class)
+                    .intercept(FieldAccessor.ofField(ProxifyMethodChangingTheObjectUnderTest.FIELD_NAME)
+                      .setsArgumentAt(0)
```
joel-costigliola: what does this do exactly ? it looks like it sets a ProxifyMethodChangingTheObjectUnderTest field but where ? in the created proxy ?

src/main/java/org/assertj/core/api/AssertJProxySetup.java
raphw: This is an interface that is implemented by all proxies and serves as a type safe setter. One could also define such setters and access them by reflection but with reflection, there is a notable performance overhead if the methods need to be located for every call.

src/main/java/org/assertj/core/api/Assumptions.java
raphw: The strategy takes the original class's name and adds a suffix to distinguish it. The default is ByteBuddy but for debugging purposes, it makes sense to add AssertJ as a name.

src/main/java/org/assertj/core/api/Assumptions.java
raphw: Yes it can. The essential part is that this lookup is done in the right module.

src/main/java/org/assertj/core/api/ErrorCollector.java
raphw: Byte Buddy provides an object for each argument whenever an intercepted method is invoked.

This: The instance of the method, the `this` reference.
SuperCall: A proxy to invoke the original method.
SuperMethod: A reference to the original method.
StubValue: A default value for the return type. null for reference type and 0 for the corresponding primitive types.

src/main/java/org/assertj/core/api/ErrorCollector.java
raphw: This is to cover for primitive return values.

src/main/java/org/assertj/core/api/SoftProxies.java
raphw: This can be done due to all proxies implementing the interface.

src/main/java/org/assertj/core/api/SoftProxies.java
raphw: Yes, to keep the proxy classes stateless, we inject the state into instance fields. This defines a setter that defines the first argument of the `AssertJProxySetup` method to one field and the second to another one.

joel-costigliola:
@raphw No significant performance changes after upgrading from 1.8.5 to 1.8.8 (in my tests at least).

joel-costigliola:
Integrated thanks @raphw!

raphw:
That is strange, @testn said he had executed the performance tests and noted a difference. I will keep on looking into how to improve performane of proxy creation in the future for such use cases. Thanks for helping out!

testn:
I tested `SoftAssertionsPerfTest` scenario and the time dropped  significantly. Which jdk do you use?

joel-costigliola:
`Java version: 1.8.0_161, vendor: Oracle Corporation`

joel-costigliola:
No significant changes after integrating https://github.com/joel-costigliola/assertj-core/pull/1233.
For the sake of testing, I ran SoftAssertionsPerfTest on 3.9.0 (in branch https://github.com/joel-costigliola/assertj-core/tree/perf-soft-proxy-3.9.0), it ran actually in ~2600ms.

testn:
Hmm interesting, I use 1.8.0_151 and it first started with around 9 seconds in 3.9.1 but down to 3-4 seconds in latest code for me.

testn:
I have a few more patches on byte-buddy side to reduce it down to < 4000 ms.

joel-costigliola:
@testn thanks for all the time you spent on improving assertj and byte buddy the performance.

PR #1240, opened 2018-05-16
Use streams instead of loops

PascalSchumacher:


joel-costigliola:
I agree as long as the resulting code is easier to read (which should be the case 90% of the time).

PascalSchumacher:
Thanks! Merged.

PR #1252, opened 2018-06-03
Improve error message for Iterables.allSatisfy

martin-tarjanyi:
* Earlier only the failure of the first item was reported in the error message
* This could lead to multiple iterations of rerunning/fixing when a failure was found
* Now the first failure for each item is reported in the error message


joel-costigliola:
Integrated thanks @martin-tarjanyi !

PR #1260, opened 2018-06-18
Add JUnit 5 Extensions for SoftAssertions

### LINKED ISSUE #1229: Add Junit5 extension for soft assertions
#### Summary
Currently there is a JUnit `@Rule` for soft assertions. In order for JUnit5.

We can use the `@RegisterWith` extensions. See more [here](https://junit.org/junit5/docs/current/user-guide/#extensions-registration-programmatic)

#### Example

```java
@RegisterWith
JUnitJupiterSoftAssertions softly = new JUnitJupiterSoftAssertions();
```

rweisleder:
This adds `JUnitJupiterSoftAssertions` and `JUnitJupiterBDDSoftAssertions` which are simple copies of corresponding JUnit 4 Rules.

#### Check List:
* Fixes #1229 
* Unit tests : YES


joel-costigliola:
Oh nice, we wanted to add support for JUnit5!
I'll have look shortly.

pom.xml
```diff
@@ -39,6 +39,27 @@
...
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <version>5.2.0</version>
+      <scope>provided</scope>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId>
+      <version>5.2.0</version>
```
joel-costigliola: should junit-jupiter-api and junit-jupiter-engine be in sync ? if so let's extract the version to a property

src/test/java/org/assertj/core/api/JUnitJupiterBDDSoftAssertionsFailureTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+@RunWith(JUnitPlatform.class)
+public class JUnitJupiterBDDSoftAssertionsFailureTest {
+
+  // we cannot register the extension here, because we need to test the failure without this test failing!
+  // @RegisterExtension
+  JUnitJupiterBDDSoftAssertions softly = new JUnitJupiterBDDSoftAssertions();
+
+  @Test
+  public void should_report_all_errors() throws Throwable {
```
joel-costigliola: keen to have a test with GIVEN WHEN THEN sections and `catchThrowable` as it makes it clearer what we are testing (the code you have rightfully duplicated was written pre java 8), somthing like:

```java
   // given some preconditions

   // when
   Throwable thrown = catchThrowable(() -> { throw new Exception("boom!"); });

   // then
   assertThat(thrown).isInstanceOf(Exception.class)
                     .hasMessageContaining("boom");
```

src/test/java/org/assertj/core/api/JUnitJupiterSoftAssertionsFailureTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+@RunWith(JUnitPlatform.class)
+public class JUnitJupiterSoftAssertionsFailureTest {
+
+  // we cannot register the extension here, because we need to test the failure without this test failing!
+  // @RegisterExtension
+  JUnitJupiterSoftAssertions softly = new JUnitJupiterSoftAssertions();
+
+  @Test
+  public void should_report_all_errors() throws Throwable {
```
joel-costigliola: same comment as in JUnitJupiterBDDSoftAssertionsFailureTest

pom.xml
```diff
@@ -39,6 +39,27 @@
...
+      <scope>test</scope>
```
joel-costigliola: I believe this should be option and provided as in `junit-jupiter-api`

pom.xml
rweisleder: Yes, they should be in sync, also the junit-platform-runner. Either we extract that as property or we use the [BOM](https://junit.org/junit5/docs/current/user-guide/#dependency-metadata-junit-bom) `org.junit:junit-bom` which is available since 5.2.0.

Should this be part of assertj-core or of assertj- parent-pom?

pom.xml
rweisleder: Yes, other projects should also have junit-jupiter-engine as dependency. But the only dependency, projects should depend on, is junit-jupiter-api. The junit-jupiter-engine is only necessary on test-runtime.

An alternative would be to declare junit-jupiter-engine as dependency of maven-surefire-plugin.
```
<plugin>
	<artifactId>maven-surefire-plugin</artifactId>
	<dependencies>
		<dependency>
			<groupId>org.junit.jupiter</groupId>
			<artifactId>junit-jupiter-engine</artifactId>
		</dependency>
	</dependencies>
</plugin>
```

See [Running Tests with Maven](https://junit.org/junit5/docs/current/user-guide/#running-tests-build-maven)

pom.xml
joel-costigliola: Let's use the BOM (good idea) and keep it in assertj-core at the moment.

pom.xml
joel-costigliola: I don't know much about plugin dependencies, what we want here is not to drag junit 5 dependencies for users that don't use it yet. would your suggestion achieve that goal ?

rweisleder: For users of assertj-core the current solution would be fine. They already have a dependency to junit-jupiter-api in test-compile scope and junit-jupiter-engine in test-runtime scope.

For those who want to build the code of assertj-core both solutions are fine, either junit-jupiter-engine as test-compile dependency (`<scope>test</scope>`) or as test-runtime dependency (the suggested dependency of maven-surefire-plugin). But I think there was an issue when running tests within an older IDE that doesn't already support JUnit 5, I think they need junit-jupiter-engine in test scope.

So I would keep the current solution with a test-compile dependency.

rweisleder:
I've pushed the changes with given-when-then and the junit-bom.

PR #1261, opened 2018-06-23
Fix if-else logic in PredicateAssert

larshvile:
Both accepts() & rejects() invoked the predicate
twice if given a single value because of broken
if-else logic.

Not a super big deal, but kind of annoying as it breaks chained `Mockito.when(..).thenReturn(..).thenReturn(..)` stuff in tests

#### Check List:
* Unit tests : YES


joel-costigliola:
Fair enough, thanks @larshvile!

PR #1265, opened 2018-07-03
Remove usages of the expected element of Test annotations.

PascalSchumacher:
First step of the JUnit 5 migration (expected element was removed from `@Test`).

Migration was done mostly by regex search and replace.


joel-costigliola:
Yes ! :+1: I never liked expected element.

PR #1267, opened 2018-07-03
Remove various ExpectedException methods

PascalSchumacher:
Another small step of the JUnit 5 migration (Rules were removed).


joel-costigliola:
I did a randomized statistical review which looked good, thanks @PascalSchumacher!

PR #1273, opened 2018-07-09
Remove ExpectedException#expectNullPointerException

PascalSchumacher:


PascalSchumacher:
Only 144 usages of `ExpectedException` to go.

PR #1285, opened 2018-07-29
Extend EnumerableAssert with new size assertion methods to improve error message

### LINKED ISSUE #1253: Add Iterable content to Iterable size assertion error messages
#### Summary
When I switch to size assertion from an iterable assertion via the `AbstractIterableSizeAssert.size()` method the error message does not include the elements of the iterable. If it included it would help to fix failed tests faster.

#### Example

```java
List<String> strings = Arrays.asList("hello", "bye", "goodbye");

assertThat(strings).size().isGreaterThanOrEqualTo(4);
```
**Actual error message:**
_Expecting:
 <3>
to be greater than or equal to:
 <4>_ 

**Expected error message:**
_Iterable does not match size requirements:
["hello", "bye", "goodbye"]
Expecting:
 <3>
to be greater than or equal to:
 <4>_

I took a look at the code but it doesn't seem to be trivial to fix since `IterableSizeAssert` extends `AbstractIntegerAssert`. With the current implementation the only way I can imagine is overriding every assertion method in `IterableSizeAssert` and decorate with it a try-catch but that would be bad for a lot of reasons. Could you maybe advise on how I (or anyone) could proceed with this?

martin-tarjanyi:
#### Check List:
* Fixes #1253
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractBooleanArrayAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractBooleanArrayAssert.java
@@ -12,11 +12,11 @@
  */
 package org.assertj.core.api;
 
-import java.util.Comparator;
-
 import org.assertj.core.data.Index;
 import org.assertj.core.internal.BooleanArrays;
 import org.assertj.core.util.VisibleForTesting;
+
+import java.util.Comparator;
 
 public abstract class AbstractBooleanArrayAssert<SELF extends AbstractBooleanArrayAssert<SELF>>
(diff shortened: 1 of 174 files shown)


joel-costigliola:
Finally integrated thanks @martin-tarjanyi !

martin-tarjanyi:
Awesome. Thank you @joel-costigliola ! Can't wait to use them.

(next time I will avoid creating such a huge PR :) )

PR #1288, opened 2018-08-03
containsAnyElementsOf should use bounded wildcard

drumonii:
`containsAnyElementsOf` takes an `Iterable<ELEMENT>` but should be a `Iterable<? extends ELEMENT>`

If I have this simple example:

```
Map<String, String> map = Map.of("some_key", "some_value");

List<String> values = List.of("some_value", "some_other_value");

assertThat(map).extracting("some_key").containsAnyElementsOf(values);
```

The `assertThat` statement will not compile with:

```
java: no suitable method found for containsAnyElementsOf(java.util.List<java.lang.String>)
method org.assertj.core.api.ObjectEnumerableAssert.containsAnyElementsOf(java.lang.Iterable<java.lang.Object>) is not applicable
  (argument mismatch; java.util.List<java.lang.String> cannot be converted to java.lang.Iterable<java.lang.Object>)
method org.assertj.core.api.AbstractIterableAssert.containsAnyElementsOf(java.lang.Iterable<java.lang.Object>) is not applicable
  (argument mismatch; java.util.List<java.lang.String> cannot be converted to java.lang.Iterable<java.lang.Object>)
```

`containsAnyElementsOf` will expect a `java.lang.Iterable<java.lang.Object>` instead of `java.lang.Iterable<?>` which will compile.

#### Check List:
* Unit tests : NA

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
@@ -745,7 +745,7 @@ public abstract class AbstractIterableAssert<SELF extends AbstractIterableAssert
    * @since 2.9.0 / 3.9.0
    */
   @Override
-  public SELF containsAnyElementsOf(Iterable<ELEMENT> iterable) {
+  public SELF containsAnyElementsOf(Iterable<? extends ELEMENT> iterable) {
     return containsAnyOf(toArray(iterable));
   }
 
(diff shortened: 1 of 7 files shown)


joel-costigliola:
Hey @drumonii, thanks for the PR!

Can you add a test with bounded wildcard to show that the PR changes fixes it ?

drumonii:
@joel-costigliola absolutely! Although I'm not entirely sure where to add. 

I'm looking at `MapAssert_extracting_Test` specifically adding to `should_allow_assertions_on_values_extracted_from_given_map_keys`.  

Or should I be creating a new test class entirely `MapAssert_extracting_containsAnyElementsOf_Test`? I don't see any other tests similar to `extracting_containsAnyElementsOf`

joel-costigliola:
You can add one test in each of these classes:
- `IterableAssert_containsAnyElementsOf_Test`
- `ObjectArrayAssert_containsAnyElementsOf_Test` 
- `AtomicReferenceArrayAssert_containsAnyElementsOf_Test`

joel-costigliola:
Integrated thanks @drumonii !

PR #1291, opened 2018-08-07
Add test packages as optional imports to OSGi manifest

### LINKED ISSUE #1290: Missing OSGi Import-Package for dynamically loaded classes - Assumptions failing
#### Summary

#1250 fixed the issue of most of the missing imports when used in an OSGi environment. However, there are a number of other packages that are dynamically accessed by AssertJ at run time using `Class.forName()`. Bnd is not able to automatically determine these references and so they must be explicitly specified in the Import-Package Bnd instruction.

The specific example I ran into was with Assumptions - which sequentially tries to load the assumption classes of JUnit, OpenTest4J and TestNG (in that order). In an OSGi environment, these load attempts will fail unless they are (optionally) imported by the bundle. This issue will only trigger if an assumption fails (as the assumptions only lazily try to load the assumption exception class).

The fix is fairly straightforward, and I have already tested this in my local directory. I hope to submit a PR with the fix shortly.

This and a few other recent OSGi-related bugs that have come up recently (eg, #1250, #1160 and #1181) also perhaps highlights that it would be nice for some kind of integration test in the continuous build process that exercises the OSGi aspects of the implementation.

kriegfrj:
#### Check List:
* Fixes #1290
* Unit tests : NO

Added org.junit, org.opentest4j and org.testng as optional imports to the OSGi Import-Package manifest entry. This allows `org.assertj.core.api.Assumptions.getAssumptionClass()` to resolve these classes if their bundles are installed in the current environment - without these import statements, they won't load and the assumption will throw an `IllegalStateException`.

Proper regression testing would be nice, but requires a full-blown OSGi integration test which does not exist yet.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -209,8 +209,19 @@
             <configuration>
               <instructions>
                 <Export-Package>org.assertj.core.*</Export-Package>
-                <!-- Don't import what assertj-core exports but import everything else -->
-                <Import-Package>!org.assertj.core.*, *</Import-Package>
+                <!--
+                  Don't import what assertj-core exports but import everything else.
+                  Also need to explicitly import packages that are dynamically loaded using
+                  Class.forName() (org.junit, org.opentest4j and org.testng are needed
+                  by the assumptions engine).
+                -->
+                <Import-Package>
+                  !org.assertj.core.*,
+                  org.junit;resolution:="optional",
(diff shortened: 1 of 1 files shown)


joel-costigliola:
Thanks @kriegfrj !

PR #1294, opened 2018-08-14
fail(): return a value

t-8ch:
This allows `fail()` to be used where the java compiler expects a value like
`Optional.orElseGet()`.

If requested I can provide more tests (and javadoc).

#### Check List:
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -1329,8 +1329,8 @@ public class Assertions {
    * @param failureMessage error message.
    * @throws AssertionError with the given message.
    */
-  public static void fail(String failureMessage) {
-    Fail.fail(failureMessage);
+  public static <T> T fail(String failureMessage) {
+    return Fail.fail(failureMessage);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


joel-costigliola:
Can you give a real example of what that would be needed ? 
It's not that I'm against the idea but I would prefer this to come from a real use case.

t-8ch:
@joel-costigliola Sure.

When writing unit tests I created a method that would extract a value of the current test state:

```java
// not the real code
private Optional<String> currentFoo() {
    return Optional.ofNullable(this.state.get("foo");
}
```

The current "foo" is needed at some point, if there is no "foo" it would mean the state was in an invalid condition and failed.

```java
// This is the function I need to call with "foo"
private doSomething(String parameter);
```

To plug my "foo" into `doSomething()` I see the following possibilities:

```java
// this will only throw a generic NoSuchElementException
doSomething(currentFoo().get();
```

```java
// two additional lines
Optional<String> foo = currentFoo();
assertThat(foo).isPresent();
doSomething(currentFoo().get());
```

```java
// Only works to replace a `fail(String)`
doSomething(currentFoo().orElseThrow(new AssertionError(""));
```

```java
// two additional lines, unnecessary line, more syntax noise
doSomething(currentFoo().orElse(() -> {
	fail("sometext);
	return null;
});
```

This PR allows the following logic:

```java
doSomething(currentFoo().orElse(() -> fail("sometext")));
```

joel-costigliola:
fair enough!

joel-costigliola:
Integrated thanks @t-8ch !

t-8ch:
Thanks!

joel-costigliola:
Just released 3.11.0 with your changes :tada:

PR #1295, opened 2018-08-14
Issue1293

### LINKED ISSUE #1293: StandardRepresentation.unambiguousToStringOf() should use identity hash codes
#### Summary

AssertJ version: 3.10.0.

Currently, `StandardRepresentation.unambiguousToStringOf()` formats the object using its own hash code:

```
  @Override
  public String unambiguousToStringOf(Object obj) {
    return obj == null ? null
        : String.format("%s (%s@%s)", toStringOf(obj), obj.getClass().getSimpleName(), toHexString(obj.hashCode()));
  }
```

But it is possible that two objects have the same hash code but are not considered equal to each other. Depending on the object's `toString()` implementation, `StandardRepresentation` could produce ambiguous messages.

By using `java.lang.System.identityHashCode()` instead this could be avoided.

#### Example

```java
class Ambiguous {

  int x;
  int y;

  Ambiguous(int x, int y) {
    this.x = x;
    this.y = y;
  }

  @Override
    if (this == o) return true;
    Ambiguous that = (Ambiguous) o;
    return this.x == that.x && this.y == that.y;
  }

  @Override
  public int hashCode() {
    return x;
  }

  @Override
  public String toString() {
    return String.format("Ambiguous(%d)", x);
  }
}

@Test
public void should_be_equal() {
  assertThat(new Ambiguous(0, 1)).isEqualTo(new Ambiguous(0, 2));
}
````

Output:

``` 
java.lang.AssertionError: 
Expecting:
 <"Ambiguous(0) (Ambiguous@0)">
to be equal to:
 <"Ambiguous(0) (Ambiguous@0)">
but was not.
```

### PR DESCRIPTION
**Pull request by adutra**
#### Check List:
* Fixes #1293 
* Unit tests : YES 

### CODE REVIEW

File: src/test/java/org/assertj/core/error/ShouldBeEqualByComparingFieldByFieldRecursively_create_Test.java [hunk 1]
```diff
@@ -66,8 +66,8 @@ public void should_use_unambiguous_fields_description_when_standard_description_
     List<Difference> differences = determineDifferences(withHashSet, withSortedSet, null, null);
     // WHEN
     // @format:off
-    String message = shouldBeEqualByComparingFieldByFieldRecursive(withSortedSet,
-                                                                   withHashSet,
+    String message = shouldBeEqualByComparingFieldByFieldRecursive(withHashSet,
```
adutra: I believe this test and the one below were giving false positives: actual and expected are inverted, but it would still pass since they contain the same elements.

src/test/java/org/assertj/core/error/ShouldBeEqualByComparingFieldByFieldRecursively_create_Test.java
```diff
@@ -152,6 +154,7 @@ public void should_precise_missing_fields_when_actual_does_not_declare_all_expec
                                          personHash, personDAOHash));
   }
 
+  @Test
```
adutra: Missing annotation.

src/test/java/org/assertj/core/presentation/StandardRepresentation_unambiguousToStringOf_Test.java
```diff
@@ -0,0 +1,431 @@
...
+import org.assertj.core.util.OtherStringTestComparatorWithAt;
+import org.assertj.core.util.StringTestComparator;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Tests for {@link StandardRepresentation#unambiguousToStringOf(Object)}.
+ *
+ * @author Alexandre Dutra
+ */
+public class StandardRepresentation_unambiguousToStringOf_Test extends AbstractBaseRepresentationTest {
```
adutra: Inspired by `StandardRepresentation_toStringOf_Test`.

src/test/java/org/assertj/core/presentation/StandardRepresentation_unambiguousToStringOf_Test.java
```diff
@@ -0,0 +1,431 @@
...
+
+  @Test
+  public void should_return_null_if_object_is_null() {
+    assertThat(unambiguousToStringOf(null)).isNull();
+  }
+
+  @Test
+  public void should_quote_String() {
+    String obj = "Hello";
+    assertThat(unambiguousToStringOf(obj)).isEqualTo(format("\"Hello\" (String@%s)",
```
adutra: I'm using IntelliJ with the Eclipse formatter profile as indicated in the contribution guidelines, and this is how the code gets formatted. Not sure if this is 100% accurate.

src/test/java/org/assertj/core/presentation/StandardRepresentation_unambiguousToStringOf_Test.java
```diff
@@ -0,0 +1,465 @@
...
+  @Test
+  public void should_return_unambiguousToStringOf_method() throws NoSuchMethodException {
+    Method method = GenericClass.class.getDeclaredMethod("someGenericMethod", Person.class, List.class, Object.class);
+    assertThat(unambiguousToStringOf(method)).isEqualTo(format("%s (Method@%s)",
+                                                               method.toGenericString(),
+                                                               toHexString(System.identityHashCode(method))));
+  }
+
+  @Test
+  public void should_disambiguate_non_equal_objects_with_same_hash_code_and_toString_representations() {
```
adutra: This is the actual test that validates that #1293 has been fixed.

src/test/java/org/assertj/core/presentation/StandardRepresentation_unambiguousToStringOf_Test.java
joel-costigliola: don't worry about formatting too much, I usually end up fixing it anyway.

PR #1299, opened 2018-08-16
Allow return value of fail* methods to be ignored

### LINKED ISSUE #1298: Assertions.fail is causing problems due to @CheckReturnValue
#### Summary
It looks like `Assertions.fail` is no longer `void` starting from AssertJ 3.11.0 (#1294). Since `Assertions` is annotated with `@CheckReturnValue` quite some projects out there might run into troubles.

#### Example

```java
import org.junit.Test;

public class AssertionsFailTest {

    @Test
    public void test() {
        Assertions.fail("failing this test on purpose");
    }
}

```

Using this example and errorprone every project which is using `Assertions.fail` will fail now with something like:

```
error: [CheckReturnValue] Ignored return value of method that is annotated with @CheckReturnValue
        Assertions.fail("failing this test on purpose");
```

gsheasby:
#### Check List:
* Fixes #1298
* Unit tests : NA

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -112,6 +112,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>com.google.errorprone</groupId>
+      <artifactId>error_prone_annotations</artifactId>
+      <version>2.3.1</version>
+    </dependency>
+    <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
       <version>25.0-jre</version>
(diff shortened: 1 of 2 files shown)


filiphr:
@gsheasby you don't have to add the error prone dependency. You can just add `@CanIgnoreReturnValue` annotation next to `@CheckReturnValue` from AssertJ

epeee:
Yes, indeed. I also would go for a new `@CanIgnoreReturnValue` annotation next to `org.assertj.core.util.CheckReturnValue`.

gsheasby:
@filiphr @epeee sure, that makes sense.

I don't have time to contribute this today, but maybe next week.

joel-costigliola:
I agree with @filiphr and @epeee, assertj has a policy of having no dependencies so we can't use `CanIgnoreReturnValue` from an external library

epeee:
@gsheasby pls let me know if you are too busy right now to finish this pr. I can continue working on this one if so.

joel-costigliola:
@epeee thanks!

joel-costigliola:
fixed by #1310

PR #1309, opened 2018-08-26
Enabled fluent String assertions for non-String objects with asString

### LINKED ISSUE #1307: Fluent assertions on the toString of an Object
#### Summary

Similar to fluent assertions on the `size()` of an `Iterable` (#301), it would be handy to have the same for the `toString()` of an `Object`.

#### Example

```java
assertThat(object).hasToString().contains("text");
```

Currently there's only `hasToString(String)`, which asserts an exact match.

Of note, this example perhaps doesn't read as nice as say, `assertThat(object).asString().contains(...)`.  As currently defined, [asString()](http://joel-costigliola.github.io/assertj/core-8/api/org/assertj/core/api/AbstractAssert.html#asString--) fails if `object` is not an instance of `String`, making that approach a breaking change.

As a workaround I'm following the recommendation in https://github.com/joel-costigliola/assertj-core/issues/357#issuecomment-89548008, and doing:

```java
assertThat(object.toString()).contains("text");
```

...however, this results in a `NullPointerException` instead of a nice `AssertionError` in the event `object` is null.

ianbrandt:
#### Check List:
* Fixes #1307
* Unit tests : YES

This is a breaking change for those that depended on `asString()` to
assert that `actual` is an instance of `String`.  It now only asserts that
`actual` is not null.  Those needing that functionality can use
`assertThat(actual).isInstanceOf(String.class).asString()`.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -456,8 +456,8 @@ public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, 
   @Override
   @CheckReturnValue
   public AbstractStringAssert<?> asString() {
-    objects.assertIsInstanceOf(info, actual, String.class);
-    return Assertions.assertThat((String) actual);
+    objects.assertNotNull(info, actual);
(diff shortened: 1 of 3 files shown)


joel-costigliola:

PR #1310, opened 2018-08-27
Fixes problems caused by @CheckReturnValue 

### LINKED ISSUE #1298: Assertions.fail is causing problems due to @CheckReturnValue

epeee:
... by using `CanIgnoreReturnValue` for `Assertions.fail*`

#### Check List:
* Fixes #1298
* Unit tests : NA

This pr is based on #1299

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -112,11 +112,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>com.google.errorprone</groupId>
-      <artifactId>error_prone_annotations</artifactId>
-      <version>2.3.1</version>
-    </dependency>
-    <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
       <version>25.0-jre</version>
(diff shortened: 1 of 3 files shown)


joel-costigliola:
Thanks @epeee and @gsheasby !

gsheasby:
Thanks @epeee - I've been on vacation, this was a pleasant thing to come back to :)

PR #1321, opened 2018-09-09
Add new assertion to check equals() and hashCode() contracts

stachenov:
Implementing equals() and hashCode() for value objects is usually very easy, assuming they are generated by the IDE. However, sometimes they can get out of date, for example, when a new field is added. Therefore, it can be useful to have a quick way to check that a given object satisfies the contracts of equals() and hashCode(). This is exactly where the proposed assertion comes into play.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
@@ -736,4 +736,51 @@ public abstract class AbstractObjectAssert<SELF extends AbstractObjectAssert<SEL
     return myself;
   }
 
+  /**
+   * Specifies objects to use for testing equals() and hashCode() contracts.
+   * <p>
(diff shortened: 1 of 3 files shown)


stachenov:
Cool. I've googled a lot before writing my own assertion (presented in this PR), but somehow missed this. However, I chose a completely different approach: instead of relying on reflection and byte code manipulation, I rely on the user to provide equal/unequal values to compare to. I think both approaches are valid, and each has its own pros and cons, so we should have both. I'm obviously biased here, but I think I'd still code my own assertion even if I knew about EqualsVerifier, simply because it's hard to figure out what it's doing under the hood, and that seems like an overkill for such a simple job.

joel-costigliola:
I need some time to think about this.

joel-costigliola:
I believe http://jqno.nl/equalsverifier/ is a better approach, it is specialized on this use case so I see little value adding these assertions to AssertJ.

PR #1332, opened 2018-09-26
Throw AssertionFailedError instead in non-strict string assertions

### LINKED ISSUE #1331: Throw AssertionFailedError instead of AssertionError in non-strict string assertions
#### Summary
Throwing AssertionFailedError from opentest4j facilitates test failure analysis in IDE providing handy "diff" view.
AssertJ already uses AssertionFailedError (if available on the classpath) but not in non-strict string assertions (from AbstractCharSequenceAssert), specifically:
* isEqualToIgnoringNewLines
* isEqualToIgnoringWhitespace
* isEqualToIgnoringNewLines

Given that two of them mention new lines it is rather common to use them to compare long multi-line strings, and IDE's "diff" feature would be very helpful to nitpick some single-letter difference in big texts (possibly coming from files).

#### Example

```java
String generatedText = sut.generateText();
assertThat(generatedText).isEqualToNormalizingNewlines(loadText("expected.txt"));
```
image

sunlaud:
#### Check List:
* Fixes #1331
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/internal/Strings.java
+++ b/src/main/java/org/assertj/core/internal/Strings.java
@@ -550,7 +550,7 @@ public class Strings {
    * @throws AssertionError if the given {@code CharSequence}s are not equal.
    */
   public void assertEqualsIgnoringCase(AssertionInfo info, CharSequence actual, CharSequence expected) {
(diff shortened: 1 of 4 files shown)


joel-costigliola:
Integrated thanks @sunlaud !

PR #1334, opened 2018-09-30
Skip JavaDoc on Java 12, because it currently fails with "javadoc: er…

PascalSchumacher:
…ror - The code being documented uses modules but the packages defined in https://docs.oracle.com/javase/8/docs/api/ are in the unnamed module."

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -346,5 +346,15 @@
         <sonar.skip>true</sonar.skip>
       </properties>
     </profile>
+    <profile>
+      <id>java12+</id>
+      <activation>
+        <jdk>[12,)</jdk>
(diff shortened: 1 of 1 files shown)


PascalSchumacher:
I'm not sure, I did not find anything yet.

The error is not AssertJ specific, because it also occurs in e.g. commons-text (see: https://travis-ci.org/apache/commons-text/jobs/435191089)

I guess it is caused by a recent change in the Java 12 builds, because everything worked a few days ago.

joel-costigliola:
ok, I'm integrating the PR, let's give a try to javadoc generation in a few weeks

PR #1346, opened 2018-10-22
Annotate Assertions and Assumptions classes with @CheckReturnValue an…

### LINKED ISSUE #1345: Assertions#assertThatThrownBy` without chained assertions should not cause static code analysis error
```java
Assertions.assertThatThrownBy(() -> new Exception());
```

causes the following error prone error:

`[CheckReturnValue] Ignored return value of method that is annotated with @CheckReturnValue`

but this is a valid use-case because:

```java
Assertions.assertThatThrownBy(() -> {});
```

fails with:

`java.lang.AssertionError: Expecting code to raise a throwable.`

PascalSchumacher:
…d explicitly annotate methods to exclude form checking with @CanIgnoreReturnValue.

This means static analysis tools can now catch more errors.

Also fixes #1345 "Assertions#assertThatThrownBy without chained assertions should not cause static code analysis error".

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractStandardSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractStandardSoftAssertions.java
@@ -38,6 +38,7 @@ import java.util.stream.Stream;
 
 import org.assertj.core.util.CheckReturnValue;
 
+@CheckReturnValue
 public abstract class AbstractStandardSoftAssertions extends Java6AbstractStandardSoftAssertions {
 
   /**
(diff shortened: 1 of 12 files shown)


joel-costigliola:
Thanks Pascal!

JoroRoss:
This change is not working for me in tools like FindBugs and IntelliJ.  Perhaps the retention should be RUNTIME?

PascalSchumacher:
@JoroRoss Sorry for breaking FindBugs. I believe FindBugs successor SpotBugs respects class-level `@CheckReturnValue` annotations.

PR #1349, opened 2018-10-28
Deprecate extractor

### LINKED ISSUE #1348: Replace Extractor with Function
#### Summary
Now that the base supported version of AssertJ is Java 8. Won't it make sense to deprecate `Extractor` (make it extend `Function`) and use `Function` everywhere instead of `Extractor`.

I found one inconsistence (in my opinion) that could be resolved with this.

In `AbstractIterableAssert`

```
public AbstractListAssert<?, List<? extends Tuple>, Tuple, ObjectAssert<Tuple>> extracting(@SuppressWarnings("unchecked") Function<ELEMENT, ?>... extractors)
```

```
public <V> AbstractListAssert<?, List<? extends V>, V, ObjectAssert<V>> extracting(Extractor<? super ELEMENT, V> extractor)
```

The first one returns `Tuple` (which is fine as it is a varargs) and the second one return a normal object assert of the extractor result. On top of that one can't use the functions with wildcards in actual.

If you are interested I can try to provide a PR that cleans this up

filiphr:
#### Check List:
* Fixes #1348 
* Unit tests : YES 

I was not sure whether you want to completely remove all places where `Extractor` is being used. For example from `Extractors`, make `ThrowingExtractor` extend `Function` and whether you want to keep tests with `Extractor`. I have no problem in completely removing all places that use that.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
@@ -1600,7 +1600,7 @@ public abstract class AbstractIterableAssert<SELF extends AbstractIterableAssert
    * @return a new assertion object whose object under test is the list of Tuples containing the extracted values.
    */
   @CheckReturnValue
-  public AbstractListAssert<?, List<? extends Tuple>, Tuple, ObjectAssert<Tuple>> extracting(@SuppressWarnings("unchecked") Function<ELEMENT, ?>... extractors) {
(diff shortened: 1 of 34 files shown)


joel-costigliola:
I'm fine with replacing `Extractor` by `Function` when possible.
Thanks for this refactoring!

filiphr:
I've replaced the `Extractor` in all places from the production code. I've left some usages in the tests so we can make sure that it works in a backwards compatible way.

joel-costigliola:
Integrated thanks @filiphr!

PR #1352, opened 2018-10-31
Do not proxy 'useComparator' method

lukas-krecan:
I am working on JsonUnit AssertJ integration. I'd like to add support for soft assertions. My class JsonAssert calls 'useComparator' in the constructor. Since the method is proxied, intercept is called before errorCollector setup and it ends-up in NPE. I can change the behavior or JsonAssert, but I assume that this is better solution.


joel-costigliola:
We missed ignoring this one. Good catch.
And thanks for JsonUnit, I have used a few times in projects I love it.

PR #1354, opened 2018-11-08
Fixes: #1324 Assertion error message sequencing

### LINKED ISSUE #1324: Assertion error message sequencing
I love AssertJ, but I find the sequencing of many of the error messages to be counterintuitive.

Take for example:

> Expecting: `<actual>` to be equal to: `<expected>` but was not.

[Per Wikipedia](https://en.wikipedia.org/wiki/Colon_(punctuation)):

> The most common use of the colon is to inform the reader that what follows the colon proves, explains, defines, describes, or lists elements of what preceded it.

Accordingly, when I begin reading that assertion error message my mind is, well, _expecting_ to see "Expecting: `<expected>`", not "Expecting: `<actual>`".  It's only when I read the closing, "but was not" that I realize I've interpreted the whole error message backwards.

I believe something like the following would be more intuitive and concise:

> Expected: `<expected>`, but was: `<actual>`.

I realize this may be a matter of personal preference, or of locale, so I'm filing this issue mainly to see if others find the current message style counterintuitive as well.

If enough do, maybe the solution would be to just change it, or perhaps some sort of localization or message style facility could be introduced to make it configurable.

7nolikov:
#### Check List:
* Fixes #1324 
* Unit tests :  Yes

Minor fix, which replaces actual and expected parameter to be more rational in error message

### DIFF
--- a/src/main/java/org/assertj/core/error/ShouldBeEqual.java
+++ b/src/main/java/org/assertj/core/error/ShouldBeEqual.java
@@ -187,8 +187,8 @@ public class ShouldBeEqual implements AssertionErrorFactory {
       Object o = constructorInvoker.newInstance("org.opentest4j.AssertionFailedError",
                                                 MSG_ARG_TYPES_FOR_ASSERTION_FAILED_ERROR,
                                                 message,
-                                                expected,
-                                                actual);
+                                                actual,
+                                                expected);
       if (o instanceof AssertionError) {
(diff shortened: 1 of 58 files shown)


7nolikov:
@joel-costigliola, i understood my mistake and replace actual/expected value only in error messageFormatter. Also i updated corresponding unit tests.

Issue #1002, opened 2017-05-29
Recursive comparison api improvements


joel-costigliola:
## Recursive comparison api

Here's an example to give users an idea of what it looks like:

```java
assertThat(person).usingRecursiveComparison()
                  .ignoringNullFields()
                  .ignoringFields("name", "age")                                                 
                  .forcingRecursiveComparisonFor(Child.class, Sister.class)
                  .isEqualTo(expectedPerson);
```

This will be in Beta as it might change according to people's feedback.

### Recursive comparison Fluent API configuration

Implements an easy to use/discover API as in:

```java
assertThat(person).usingRecursiveComparison()
                  .ignoringNullFields()
                  .ignoringFields("name", "age")                                                 
                  .isEqualTo(expectedPerson);
```

- [x] expose ignoring actual null fields  
- [x] expose ignoring fields  
- [x] expose ignoring overridden equals  
- [x] expose setting a comparison strategy per fields/types
- [x] expose enabling strict type checking
- [x] documentation

### Ignoring fields in comparison

Users can specify the comparison to ignore all null fields in the object under test. 

Users can specify to ignore specific fields in the object under test:
- giving their full path, ex: `person.sister.name`
- by regex, ex: `.*name` means `name`, `person.name` and `person.sister.name` are ignored.

Note if you ignore specific field, all sub fields are ignored too, for example if `person` is ignored then `person.name`, `person.address`, `person.address.number` are ignored too.

Nice to have: report all the fields ignored in the comparison 

#### status

- [x] ignore all null fields 
- [x] ignore fields specified with a full path 
- [ ] fail if an ignored fields specified with a full path does not match any field
- [x] ignore fields specified with a regex
- [x] report the fields configured to be ignored in the comparison  
- [ ] report all the fields actually ignored in the comparison (to show which fields the regex matched)
- [ ] documentation

### Allow to use overridden `equals` method or not

By default the recursive comparison will use `equals` if it had been overridden but users should be able to specify to ignore overridden `equals` methods and compare objects field by field.

This can be done:
- for all types matching given regexes
- a specific given list of types with `forcingRecursiveComparisonFor(Child.class, Sister.class)`
- a specific given list of fields with `forcingRecursiveComparisonForFields("child", "sister")` Nice to have

Example:

```java
assertThat(person).usingRecursiveComparison()
                  .forcingRecursiveComparisonFor(Child.class, Sister.class)
                  .forcingRecursiveComparisonForFields("name", "city")
[...]

PR #1360, opened 2018-11-13
Add support for bare-named property methods

stevenschlansker:
[Immutables](https://immutables.github.io/) is a wonderful code generator for value types.  It has support for `Optional` types, but creates the wrapper object from a nullable field under the hood.

Immutables also encourages bare-named property methods as opposed to bean-named.  (i.e. `property()` rather than `getProperty()`)

`assertj` provides great support for extracting values from types.

In the case of mixing `immutables`, `assertj`, and optional types, some confusing behavior emerges!  For a property `OptionalInt property()`, the implementation will store `Integer property;` and implement `property() { return OptionalInt.ofNullable(property); }`.  `assertj` does not consider bare-named property methods for extraction, and instead cracks open the field.  But the field is a different type than the programmer expected!  This leads to confusing assertion mismatches, "expected OptionalInt(42) got 42 instead" and the like, and it's not at all immediately clear why.

The simplest fix is to add property support for methods that match the property name exactly, without the `get` or `is` prefix.

I authored this code and release it under Apache 2.0 to the project.

### DIFF
--- a/src/test/java/org/assertj/core/extractor/ByNameSingleExtractorTest.java
+++ b/src/test/java/org/assertj/core/extractor/ByNameSingleExtractorTest.java
@@ -199,5 +199,10 @@ public class ByNameSingleExtractorTest {
     public OptionalInt value() {
       return OptionalInt.of(value);
     }
+
+    // ensure setter-like methods don't distract us
+    public BareOptionalIntHolder value(int value) {
+      throw new AssertionError("unreached");
+    }
   }
 }
--- a/src/test/java/org/assertj/core/extractor/ByNameSingleExtractorTest.java
+++ b/src/test/java/org/assertj/core/extractor/ByNameSingleExtractorTest.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.OptionalInt;
 
+import org.assertj.core.api.Assertions;
 import org.assertj.core.test.Employee;
 import org.assertj.core.test.Name;
 import org.assertj.core.util.introspection.IntrospectionError;
@@ -131,6 +132,15 @@ public class ByNameSingleExtractorTest {
     BareOptionalIntHolder holder = new BareOptionalIntHolder(42);
     assertThat(holder).extracting("value")
       .containsExactly(OptionalInt.of(42));
+  }
+
+  @Test
+  public void should_ignore_property_with_barename_method() {
+    BareOptionalIntHolder holder = new BareOptionalIntHolder(42);
+    Assertions.setExtractBareNamePropertyMethods(false);
+    assertThat(holder).extracting("value")
+      .containsExactly(42);
+    Assertions.setExtractBareNamePropertyMethods(true);
   }
 
   public static class EmployeeWithBrokenName extends Employee {
--- a/src/test/java/org/assertj/core/extractor/ByNameSingleExtractorTest.java
+++ b/src/test/java/org/assertj/core/extractor/ByNameSingleExtractorTest.java
@@ -21,6 +21,7 @@ import static org.assertj.core.extractor.Extractors.byName;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.OptionalInt;
 
 import org.assertj.core.test.Employee;
 import org.assertj.core.test.Name;
@@ -125,6 +126,13 @@ public class ByNameSingleExtractorTest {
     assertThat(extracted).isEqualTo("Young Padawan");
   }
 
+  @Test
+  public void should_extract_property_with_barename_method() {
+    BareOptionalIntHolder holder = new BareOptionalIntHolder(42);
+    assertThat(holder).extracting("value")
+      .containsExactly(OptionalInt.of(42));
+  }
+
   public static class EmployeeWithBrokenName extends Employee {
 
     public EmployeeWithBrokenName(String name) {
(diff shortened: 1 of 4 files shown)


joel-costigliola:
I'm sympathetic to the idea, I think I would prefer though to make it optional, that is users should explicitly enable bare-named property introspection. Doing that would ensure backward compatibility for users not needing this.

Enabling or disabling bare-named property introspection should be provided by the `Assertions` class which would set the flag in `Introspection` (the important part here is it is settable from the `Assertions` class  so that users can find it without too much trouble).

stevenschlansker:
I appreciate where you're coming from, but I do wonder if this is an overly conservative suggestion that really reduces the usability.  Currently we use `assertj` as a "zero configuration static library" i.e. each test case simply calls `Assertions.assertThat`.  Requiring us to configure it will mean we have to treat `assertj` itself as configurable and arrange a test preamble to configure it before every test (which currently does not exist in our build at all - we use junit `@Rule`s to decompose our test cases and not have a common superclass).

I understand the hesitation to change the rules -- but is there really a case where someone relies on the field property being read and a a similarly named method ignored, such that this change breaks it?  I'm struggling to come up with any reasonable use case for the existing behavior.  And making it configurable certainly increases the complexity of the library, requires documentation / users to be aware of the feature...

If you are sure you want it configurable, I will certainly make it so.  But I'm curious to hear if you had any specific cases or patterns of code you thought might regress.  I haven't been able to come up with any myself, which led me to think, "maybe I should just nudge to have it be default one more time" 😸

joel-costigliola:
I don't have any specific case in mind, I know it's a pain to configure libraries for tests since there is no support for a global hook to do that.
For this specific case I think I can live with bare-named property introspection being enabled by default, that will make you happy and unhappy users can disable it (hopefully there won't be too many).

src/test/java/org/assertj/core/extractor/ByNameSingleExtractorTest.java
```diff
@@ -176,4 +184,20 @@ public boolean isAdult() {
...
+          value = null;
+      }
+
+      BareOptionalIntHolder(int value) {
+          this.value = value;
+      }
+
+      public OptionalInt value() {
+          return OptionalInt.of(value);
+      }
```
joel-costigliola: what happens if you add  `public void value(int whatever)` ? I think the code to find the accessor method does not verify the method returns something and does not take any parameters.

filiphr:
Just curious about 

> I know it's a pain to configure libraries for tests since there is no support for a global hook to do that.

Isn't there the `ConfigurationProvider`? In theory assertj can expose more things thro an SPI and people would be able to configure it globally

joel-costigliola:
@filiphr I have just created https://github.com/joel-costigliola/assertj-core/issues/1361

src/test/java/org/assertj/core/extractor/ByNameSingleExtractorTest.java
stevenschlansker: I'll extend the test to verify this case.  I don't think it's a problem - `Introspection.findMethod` uses `getMethod` / `getDeclaredMethod`.  It's not obvious from a quick look, but both these are varargs invocations with a zero-length parameter list.  So it will look for nullary methods only.

joel-costigliola:
Integrated thanks @stevenschlansker!

stevenschlansker:
Any chance we could get a release soon? We'd love to integrate it.  Thanks and happy new year!

joel-costigliola:
Likely at the end of the month, depending on my progress on #1002.

wilkinsona:
> For this specific case I think I can live with bare-named property introspection being enabled by default, that will make you happy and unhappy users can disable it (hopefully there won't be too many).

I wouldn't say I'm unhappy, but I've just been bitten by this. Could the breaking change please be mentioned in the release notes?

stevenschlansker:
Just to make sure we're clear, you had a field and method with the same name, and relied on the property extractor finding the field value and *not* the method's return value?

That's unfortunate this bit you, I'm sorry.  I hope you agree with the change though, I think the new behavior is much less surprising personally.

(I'm not a maintainer so I can't add to the release notes, but I agree that it should be there.)

wilkinsona:
Yes, that's right. The class in question is [Spring Framework's `CacheControl`](https://github.com/spring-projects/spring-framework/blob/f087fd5a93dbd43700e96f81088d1372263087c2/spring-web/src/main/java/org/springframework/http/CacheControl.java). Looking more closely, I wonder if AssertJ could be a little smatter when looking for bare-named property methods. I think the problem I encountered would have been avoided if it ignored `static` methods and if it checked that the method's return type was the same as, or at least compatible with, the field's type.

stevenschlansker:
Excluding `static` methods would make sense to me.  They are never "properties" in a way that you'd assert over.

joel-costigliola:
The release notes are now part of https://github.com/assertj/doc which accepts PR ;-) but I'll update them to clarify the breaking change (I missed that one, sorry about that).

Excluding static methods make sense.

In retrospective, we shouldn't have changed the default behavior but allow to easily change it globally as you can do with representation (see https://assertj.github.io/doc/#assertj-core-custom-representation-default-global-scope).

wilkinsona:
I'm more than happy to contribute a PR for the release notes change, or anything else that you think may be appropriate here. We thoroughly enjoy using AssertJ to test Spring Boot so a contribution or two is the least we can do.

joel-costigliola:
@wilkinsona thanks for the kind words, I have just created https://github.com/joel-costigliola/assertj-core/issues/1458 to exclude static methods, let me know if you are happy to tackle it.

I'll take care of the release since it's a quick one.

joel-costigliola:
Release notes updated: https://assertj.github.io/doc/#assertj-core-3.12.0-breaking-changes

wilkinsona:
@joel-costigliola I'm happy to tackle #1458. I'll need to get approval to contribute from our legal team, but that should just be a matter of time.

PR #1363, opened 2018-11-20
Enable JavaDoc creation on Java 12+.

PascalSchumacher:
Just to check if this works.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -339,15 +339,5 @@
         <sonar.skip>true</sonar.skip>
       </properties>
     </profile>
-    <profile>
-      <id>java12+</id>
-      <activation>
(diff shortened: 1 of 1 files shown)


PascalSchumacher:
Still fails with

>An error has occurred in Javadoc report generation: 
>[ERROR] Exit code: 1 - javadoc: error - The code being documented uses modules but the packages defined in https://docs.oracle.com/javase/8/docs/api/ are in the unnamed module.

PascalSchumacher:
Javadoc still fails on Java 12.

Open JDK issue was rejected (see: https://bugs.openjdk.java.net/browse/JDK-8212233)

Looks like we have to wait for a fix in `maven-javadoc-plugin`.

PR #1382, opened 2019-01-02
Issue1377

### LINKED ISSUE #1377: Add anysatisfy  for Map
#### Summary

Allow to verify a bunch of assertions for any key-value pairs of a Map. This would serve the symmetry of the map assert. See #1030 (allSatisfy for Maps) and #889 (anySatisfy for Iterables).

#### Example

```java

class Person {
  public String firstName;
  public String lastName;

  public Person(String firstName, String lastName) {
    this.firstName = firstName;
    this.lastName = lastName;
  }
}

Person max = new Person("Max","Mustermann");
Person min = new Person("Min","Mustermann");
Map<Person, Person> knows = new HashMap<Person, Person>();
knows.put(max, min);
knows.put(max, max);
assertThat(knows).anySatisfy((key, value) -> {
  assertThat(key.firstName).isEqualTo("Max");
  assertThat(key.lastName).isEqualTo("Mustermann");
  assertThat(value.firstName).isEqualTo("Min");
  assertThat(value.lastName).isEqualTo("Mustermann");
});```

### PR DESCRIPTION
**Pull request by almondtools**
#### Check List:
* Fixes #1377
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractMapAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractMapAssert.java
@@ -100,6 +100,39 @@ public abstract class AbstractMapAssert<SELF extends AbstractMapAssert<SELF, ACT
   }
 
   /**
+   * Verifies that at least one map entry satisfies the given {@code entryRequirements} .
+   * <p>
+   * Example:
+   * <pre><code class='java'> Map&lt;TolkienCharacter, Ring&gt; elvesRingBearers = new HashMap&lt;&gt;();
+   * elvesRingBearers.put(galadriel, nenya);
+   * elvesRingBearers.put(gandalf, narya);
+   * elvesRingBearers.put(elrond, vilya);
+   *
+   * // this assertion succeeds:
+   * assertThat(elvesRingBearers).anySatisfy((character, ring) -&gt; {
+   *   assertThat(ring).isIn(vilya);
+   * });
+   *
+   * // this assertion fails as Gandalf the maia does not bear the ring vilya and the ring bearer of vilya is not a maia:
+   * assertThat(elvesRingBearers).anySatisfy((character, ring) -&gt; {
+   *   assertThat(character.getRace()).isIn(MAIA);
+   *   assertThat(ring).isIn(vilya);
+   * });</code></pre>
+   * <p>
+   *
+   * @param entryRequirements the given requirements that each entry must satisfy.
+   * @return {@code this} assertion object.
+   * @throws NullPointerException if the given entryRequirements {@link BiConsumer} is {@code null}.
+   * @throws AssertionError if the actual map is {@code null}.
+   * @throws AssertionError if all entries don't satisfy the given requirements.
+   * @since ?
+   */
+  public SELF anySatisfy(BiConsumer<? super K, ? super V> entryRequirements) {
+    maps.assertAnySatisfy(info, actual, entryRequirements);
+    return myself;
+  }
+
+  /**
    * Verifies that the {@link Map} is {@code null} or empty.
    * <p>
    * Example:
--- a/src/main/java/org/assertj/core/api/AbstractMapAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractMapAssert.java
@@ -108,23 +108,22 @@ public abstract class AbstractMapAssert<SELF extends AbstractMapAssert<SELF, ACT
    * elvesRingBearers.put(gandalf, narya);
    * elvesRingBearers.put(elrond, vilya);
    *
-   * // this assertion succeeds:
+   * // this assertion succeeds (one of the elven ring bearers bears vilya):
    * assertThat(elvesRingBearers).anySatisfy((character, ring) -&gt; {
-   *   assertThat(ring).isIn(vilya);
+   *   assertThat(ring).isEqualTo(vilya);
    * });
    *
-   * // this assertion fails as Gandalf the maia does not bear the ring vilya and the ring bearer of vilya is not a maia:
+   * // this assertion fails (one of the elven ring bearers is a maia, but none bears the One Ring):
    * assertThat(elvesRingBearers).anySatisfy((character, ring) -&gt; {
    *   assertThat(character.getRace()).isIn(MAIA);
-   *   assertThat(ring).isIn(vilya);
+   *   assertThat(ring).isEqualTo(oneRing);
    * });</code></pre>
-   * <p>
    *
-   * @param entryRequirements the given requirements that each entry must satisfy.
+   * @param entryRequirements the given requirements that at least one entry must satisfy.
    * @return {@code this} assertion object.
    * @throws NullPointerException if the given entryRequirements {@link BiConsumer} is {@code null}.
    * @throws AssertionError if the actual map is {@code null}.
-   * @throws AssertionError if all entries don't satisfy the given requirements.
+   * @throws AssertionError if no entries satisfy the given requirements.
    * @since 3.12
    */
   public SELF anySatisfy(BiConsumer<? super K, ? super V> entryRequirements) {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by joel-costigliola** [comment]
Thanks @almondtools, I will review your PR shortly !

File: src/main/java/org/assertj/core/api/AbstractMapAssert.java [hunk 1]
```diff
@@ -99,6 +99,39 @@ public SELF allSatisfy(BiConsumer<? super K, ? super V> entryRequirements) {
...
+   * <p>
+   * Example:
+   * <pre><code class='java'> Map&lt;TolkienCharacter, Ring&gt; elvesRingBearers = new HashMap&lt;&gt;();
+   * elvesRingBearers.put(galadriel, nenya);
+   * elvesRingBearers.put(gandalf, narya);
+   * elvesRingBearers.put(elrond, vilya);
+   *
+   * // this assertion succeeds:
+   * assertThat(elvesRingBearers).anySatisfy((character, ring) -&gt; {
+   *   assertThat(ring).isIn(vilya);
```
joel-costigliola: add `oneRing` and `nibelungRing` to `isIn` arguments otherwise it looks a bit weird

src/main/java/org/assertj/core/api/AbstractMapAssert.java
```diff
@@ -99,6 +99,39 @@ public SELF allSatisfy(BiConsumer<? super K, ? super V> entryRequirements) {
...
+   * });
+   *
+   * // this assertion fails as Gandalf the maia does not bear the ring vilya and the ring bearer of vilya is not a maia:
+   * assertThat(elvesRingBearers).anySatisfy((character, ring) -&gt; {
+   *   assertThat(character.getRace()).isIn(MAIA);
+   *   assertThat(ring).isIn(vilya);
```
joel-costigliola: same comment here

src/main/java/org/assertj/core/api/AbstractMapAssert.java
```diff
@@ -99,6 +99,39 @@ public SELF allSatisfy(BiConsumer<? super K, ? super V> entryRequirements) {
...
+   * });</code></pre>
+   * <p>
```
joel-costigliola: uneeded

src/main/java/org/assertj/core/api/AbstractMapAssert.java
```diff
@@ -99,6 +99,39 @@ public SELF allSatisfy(BiConsumer<? super K, ? super V> entryRequirements) {
...
+   *
+   * @param entryRequirements the given requirements that each entry must satisfy.
```
joel-costigliola: `at least one entry`

src/main/java/org/assertj/core/api/AbstractMapAssert.java
```diff
@@ -99,6 +99,39 @@ public SELF allSatisfy(BiConsumer<? super K, ? super V> entryRequirements) {
...
+   * @return {@code this} assertion object.
+   * @throws NullPointerException if the given entryRequirements {@link BiConsumer} is {@code null}.
+   * @throws AssertionError if the actual map is {@code null}.
+   * @throws AssertionError if all entries don't satisfy the given requirements.
```
joel-costigliola: `no entries satisfy the given requirements.`

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -96,6 +98,24 @@ public static Maps instance() {
...
+    boolean anyMatch = actual.entrySet().stream().anyMatch(e -> {
+      try {
+        entryRequirements.accept(e.getKey(), e.getValue());
+      } catch (AssertionError ex) {
+        return false;
+      }
+      return true;
+    });
+
+    if (!anyMatch) {
```
joel-costigliola: remove block as in:
```
if (!anyMatch) throw failures.failure(info, elementsShouldSatisfyAny(actual));
```

src/test/java/org/assertj/core/internal/maps/Maps_assertAnySatisfyingConsumer_Test.java
```diff
@@ -0,0 +1,89 @@
...
+      assertThat(team).isIn("Lakers");
+      assertThat(player.getPointsPerGame()).isGreaterThan(18);
+    });
+  }
+
+  @Test
+  public void should_fail_if_the_map_under_test_is_empty_whatever_the_assertions_requirements_are() {
+    actual.clear();
+    try {
+      maps.assertAnySatisfy(someInfo(), actual, ($1,$2) -> assertThat(true).isTrue());
```
joel-costigliola: refactor the test with `AssertionUtil.expectAssertionError` as in https://github.com/joel-costigliola/assertj-core/blob/6e006e39ab86b8ff6322c880627b354ab10b7785/src/test/java/org/assertj/core/internal/objectarrays/ObjectArrays_assertHasOnlyElementsOfTypes_Test.java#L58.

Could you also do that for other `should_fail...` tests ?

src/test/java/org/assertj/core/internal/maps/Maps_assertAnySatisfyingConsumer_Test.java
```diff
@@ -0,0 +1,89 @@
...
+    } catch (AssertionError e) {
+      verify(failures).failure(info, elementsShouldSatisfyAny(actual));
+      return;
+    }
+    failBecauseExpectedAssertionErrorWasNotThrown();
+  }
+
+  @Test
+  public void should_fail_if_actual_is_null() {
+    assertThatExceptionOfType(AssertionError.class).isThrownBy(() -> maps.assertAnySatisfy(someInfo(), null, (team, player) -> {
```
joel-costigliola: either use `expectAssertionError` or static import `AssertionsUtil.assertThatAssertionErrorIsThrownBy`which is a shortcut to `assertThatExceptionOfType(AssertionError.class).isThrownBy`

joel-costigliola:
@almondtools PR reviewed!

src/main/java/org/assertj/core/api/AbstractMapAssert.java
almondtools: I agree, yet I would resolve it another way - checking a map one any of vilya, oneRing and nibelungRing is not quite a meaningful test.

What about `isEqualTo`?

src/main/java/org/assertj/core/api/AbstractMapAssert.java
almondtools: Consequently (with the upper proposal) I would add `isEqualTo(oneRing)`. While the tuple (maia,vilya) cannot be satisfied in the tolkien universe, the tuple (maia, oneRing) can - but not by the elven ring bearers.

almondtools:
Please let me know if the update of the pull request is ok (I added some corrections and rebased it to your master). I am not really a git expert so there may have been some hidden features that would have made it easier for both of us.

joel-costigliola:
Thanks for the feedback on tests, I think what is needed is to provide a contributor guide that will show the way tests should be written.

Ideally I would have the time to migrate all the tests to the new style but we have  ~10k tests so I have a pragmatic approach: new tests must follow the new style and also change the tests for the part that are being updated.

filiphr:
@joel-costigliola just my 2 cents if you write a guide for the new tests approach maybe the community can jump in and help out. We could take package names or regex patterns and migrate them one by one. It would be much easier for you to like that. You could also create an issue summarising this and mention that it is good for contributions

almondtools:
Yet in my experience it is very hard to document code style in a way that does not create misunderstandings. Examples work almost as good (unless you are an apprentice). I found the existing sources quite instructive, I only did not know whether better to keep similar things similar (accepting old style) or new things new (accepting inconsistency between similar solutions).

So I think a task to rework all tests to the new style would be far more effective than writing documentation. Of course this would afford the community to jump in as @filiphr proposed, and also the work packages should be small such that conflicts with concurrent pull requests are small.

Yet for the first, I will follow the current strategy.

almondtools:
As I mentioned I am not common with the github reviewing process. One of the open questions is who should resolve the conversations:

- you, to accept the works on this topic)
- me, to signal that the topic of the conversation has been worked on

Since no comment occured since 2 days I assume the latter. Anyway the review could continue.

joel-costigliola:
@almondtools I don't really use conversations, I usually wait for the contributor to tell me when all the comments were addressed, I then do another review and it there are only minor things I'll fix them myself.

joel-costigliola:
Integrated, I found the time :)
Thanks @almondtools!

PR #1384, opened 2019-01-02
issue1369

### LINKED ISSUE #1369: isEqualToComparingFieldByFieldRecursively is not symmetrical when reflection is involved
#### Summary
`isEqualToComparingFieldByFieldRecursively` is not symmetrical when a field of subclass type is populated with an object with a superclass type using reflection. 

(this may sound as an obscure use case, but such 'mistakes' tend to happen with mapping/serialization libraries, which use reflection heavily. It just happened to me with Orika, for one)

#### Example

```java

import java.lang.reflect.Field;
import org.junit.Test;

public class AssertJEqualityTest {

  @Test
  public void isEqualToComparingFieldByFieldRecursively_shouldBeSymmetrical()
      throws NoSuchFieldException, IllegalAccessException {
    //given
    final ReferencingClass actual = new ReferencingClass();
    actual.setReferenced(new ChildClass());
    final ReferencingClass expected = new ReferencingClass();
    Field referencedField = ReferencingBaseClass.class.getDeclaredField("referenced");
    referencedField.setAccessible(true);
    referencedField.set(expected, new BaseClass());
    //then
    assertThat(expected).isEqualToComparingFieldByFieldRecursively(actual);
    assertThat(actual).isEqualToComparingFieldByFieldRecursively(expected);
  }
}

class BaseClass {
  private String fieldA = "fieldA";
  private String fieldB = "fieldB";

  public String getFieldA() {
    return fieldA;
  }

  public void setFieldA(String fieldA) {
    this.fieldA = fieldA;
  }

  public String getFieldB() {
    return fieldB;
  }

  public void setFieldB(String fieldB) {
    this.fieldB = fieldB;
  }
}

class ChildClass extends BaseClass {
  private String fieldC = "fieldC";

  public String getFieldC() {
    return fieldC;
  }

  public void setFieldC(String fieldC) {
    this.fieldC = fieldC;
  }
}

class ReferencingBaseClass<T extends BaseClass> {
  private T referenced;

  public T getReferenced() {
    return referenced;
  }

  public void setReferenced(T referenced) {
    this.referenced = referenced;
  }
}

class ReferencingClass extends ReferencingBaseClass<ChildClass> {
}

```

In the example above, only the second assertion fails. It is a bit similar to [this issue](https://github.com/joel-costigliola/assertj-core/issues/1054). The test case above shows, though, that one side of the comparison is favoured when AssertJ checks whether the two objects have the same fields.

dev-priporov:
#### Check List:
* Fixes #1369 
* Unit tests : NO / NA
* Javadoc with a code example (API only) : NO / NA

added symmetrical case for isEqualToComparingFieldByFieldRecursively

### DIFF
--- a/src/main/java/org/assertj/core/internal/DeepDifference.java
+++ b/src/main/java/org/assertj/core/internal/DeepDifference.java
@@ -314,14 +314,11 @@ public class DeepDifference {
       Set<String> key1FieldsNames = getFieldsNames(getDeclaredFieldsIncludingInherited(key1.getClass()));
       Set<String> key2FieldsNames = getFieldsNames(getDeclaredFieldsIncludingInherited(key2.getClass()));
       if (!key2FieldsNames.containsAll(key1FieldsNames)) {
-        Set<String> key1FieldsNamesNotInKey2 = newHashSet(key1FieldsNames);
-        key1FieldsNamesNotInKey2.removeAll(key2FieldsNames);
-        String missingFields = key1FieldsNamesNotInKey2.toString();
-        String key2ClassName = key2.getClass().getName();
-        String key1ClassName = key1.getClass().getName();
-        String missingFieldsDescription = format(MISSING_FIELDS, key1ClassName, key2ClassName, key2.getClass().getSimpleName(),
-                                                 key1.getClass().getSimpleName(), missingFields);
-        differences.add(new Difference(currentPath, key1, key2, missingFieldsDescription));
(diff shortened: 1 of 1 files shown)


joel-costigliola:
@dev-priporov we have yet decided to address #1369 so this PR is being put on hold.
One comment though, the PR does not contain any tests, if the PR solves an issue it should contain tests to prove the issue was solved (but again don't spend time on it until we decide to address the issue).

Anyway thanks for taking the time to submit a PR.

joel-costigliola:
superseded by the recursive comparison which is not symetrical by design

PR #1391, opened 2019-01-09
Issue-1390 Double and Float comparator support Infinity

champialex:
Now floating point numbers with the same infinite values will be
considered equals by default.

#### Check List:
* Fixes: https://github.com/joel-costigliola/assertj-core/issues/1390
* Unit tests : YES


joel-costigliola:
Awesome, thanks @champialex !

PR #1392, opened 2019-01-09
Fix ElementsShouldZipSatisfy that could not take objects whose string…

AHeise:
… representation contained %

#### Check List:
* Fixes https://github.com/joel-costigliola/assertj-core/issues/1387
* Unit tests : YES


joel-costigliola:
Integrated thanks @AHeise!

PR #1404, opened 2019-01-20
Add hasBinaryContent for InputSteams

### LINKED ISSUE #1395: Add hasBinaryContent for InputStream
I would like to assert that a given input string returns a given binary content.
Since `InputStream` contains bytes, I'd suggest to provide the method `hasBinaryContent` that is also available for `File`. I'm currently using `hasSameContentAs` as a workaround, but that's not optimal.

```java
byte[] expected = TEST_CONTENT;

InputStream inputStream = storage.getDownloadStream(TEST_URI);

// currently not supported:
assertThat(inputStream).hasBinaryContent(expected);
// workaround:
assertThat(inputStream).hasSameContentAs(new ByteArrayInputStream(expected));
```

ralfstx:
Add a method `hasBinaryContent` for input streams, following the
implementation of the corresponding method in `AbstractFileAssert`.

#### Check List:
* Fixes #1395
* Unit tests : YES


ralfstx:
I've created a new PR: #1413

PR #1406, opened 2019-01-22
Add missing @since

epeee:
I thought it might be good to add missing since for assertj 3.12.0.
I can link the corresponding pr's if that makes it easier to review (but maybe git annotate is easier to use).


joel-costigliola:
Cool, thanks @epeee.

Issue #1371, opened 2018-12-14
MapAssert.containsOnlyKeys() with Iterable overload


ebundy:
#### Summary
`MapAssert.containsOnlyKeys()` accepts a var-args but doesn't have an overload with `Iterable<>`, contrary to `IterableAssert`.

#### Example

```java 

void findEmployeesByCountry(Company company) {
        List<Country> countries = ...;
	Map<Country, List<Employee>> employeesByCountry = company.findEmployeesByCountry(countries);
	Assertions.assertThat(employeesByCountry)
			 .containsOnlyKeys(countries.stream().toArray(Country[]::new));         

	Assertions.assertThat(employeesByCountry.get())
}
```

That 

```java 
.containsOnlyKeys(countries.stream().toArray(Country[]::new));         
```
or  

```java 
.containsOnlyKeys(countries.toArray(new Country[countries.length]));         
```

should be just :

```java 
.containsOnlyKeys(countries);  
```

joel-costigliola:
Sure.

bastikempken:
I am working on this issue, and will provide a pull request soon.

bastikempken:
@joel-costigliola  I've implemented a possible solution on https://github.com/joel-costigliola/assertj-core/pull/1373
@ebundy It's my first participation, too and and i already worked on it. but i would really appreciate your feedback

ebundy:
@bastikempken The next time, out of courtesy for the person that asked the question, please wait for that he/she tells whether he/she would implement that.  
About the implementation : not bad but it could be much dry.  You created new methods in each layer to perform the same thing. You don't need because the single variation is `K[]` becomes `Iterable<K>`. Passing from the one to the other is straight simple. 
After I think that that is the kind of thing that deserves some discussions with the API owner and its collaborators.  
I would do things in this way  or in the other way : 
```` JAVA
public <K, V> void assertContainsOnlyKeys(AssertionInfo info, Map<K, V> actual,
                                            @SuppressWarnings("unchecked") K... keys) {
    assertContainsOnlyKeys(info, actual, java.util.Arrays.asList(keys));
}


public <K, V> void assertContainsOnlyKeys(AssertionInfo info, Map<K, V> actual,
                                             Iterable<K> keys) {
    assertNotNull(info, actual);
    failIfNull(keys);
    int sizeOfKeys = sizeOf(keys);
    if (actual.isEmpty() && sizeOfKeys == 0) {
      return;
    }
[...]

Issue #1304, opened 2018-08-23
Add satisfiesAny to object assertions


joel-costigliola:
#### Summary

This is a way to check that an object satisfies one of the given assertions requirements expressed as `Consumers`

#### Example

```java
assertThat("foo").satisfiesAny(actual -> assertThat(actual).contains("foo"),
                               actual -> assertThat(actual).isEmpty());
```

joel-costigliola:
We might want to provide an equivalent assertion for iterable/array assertions like `hasAtLeastOneElementSatisfyingAny` and `hasAllElementsSatisfyingAny`?
The challenge being not to confuse users with the existing `anySatisfy` and `allSatisfy`.

eximius313:
This option is very important to me as well.
I just have the case, when I must check if Object is nullOrEmpty *but* if it is not, it must have specific propertyWithValue.

PR #1412, opened 2019-01-28
Add missing @since 3.12.0

epeee:
let's add some more @since 3.12.0.


epeee:
(#1371 and #1304)

joel-costigliola:
Thanks for looking into it, this kind of stuff is easily missed!

PR #1419, opened 2019-02-03
noneSatisfy should report all failing elements

### LINKED ISSUE #1415: noneSatisfy should report all failing elements
#### Summary
Currently, `noneSatisfy` does only report the first element which did not satisfy the given restrictions. 
What about reporting all elements which did not satisfy the given restrictions.

#### Example

A very simple example, e.g. let's assume that we have a list `[1, 2, 3]`. If we use noneSatisfy and let's say the 2nd and the 3rd element does not meet the given restrictions we only report the 2nd one.

```
java.lang.AssertionError: 
Expecting no elements of:
  <[1, 2, 3]>
to satisfy the given assertions requirements but this one did:
  <2>
```

epeee:
#### Check List:
* Fixes #1415 
* Unit tests : YES

The example from #1415 now results in:

```
java.lang.AssertionError: 
Expecting no elements of:
  <[1, 2, 3]>
to satisfy the given assertions requirements but these elements did:
  <[2, 3]>
```

### DIFF
--- a/src/test/java/org/assertj/core/error/NoElementsShouldSatisfy_create_Test.java
+++ b/src/test/java/org/assertj/core/error/NoElementsShouldSatisfy_create_Test.java
@@ -16,7 +16,7 @@ import static java.lang.String.format;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.error.NoElementsShouldSatisfy.noElementsShouldSatisfy;
 import static org.assertj.core.presentation.StandardRepresentation.STANDARD_REPRESENTATION;
-import static org.assertj.core.util.Lists.newArrayList;
+import static org.assertj.core.util.Lists.list;
 
 import org.assertj.core.description.TextDescription;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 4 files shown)


src/test/java/org/assertj/core/internal/iterables/Iterables_assertNoneSatisfy_Test.java
```diff
@@ -67,7 +67,18 @@ public void should_fail_when_one_elements_satisfy_the_given_restrictions() {
...
+    verify(failures).failure(info, noElementsShouldSatisfy(actual, list("Yoda")));
+    assertThat(assertionError).isNotNull();
+  }
+
+  @Test
+  public void should_fail_when_two_elements_satisfy_the_given_restrictions() {
+    // GIVEN
+    Consumer<String> restrictions = name -> assertThat(name).startsWith("L");
+    // WHEN
+    Throwable assertionError = catchThrowable(() -> iterables.assertNoneSatisfy(someInfo(), actual, restrictions));
```
joel-costigliola: can be replaced by `AssertionsUtil.expectAssertionError` that checks `assertThat(assertionError).isNotNull();` internally.

joel-costigliola:
Looks good @epeee, I just had a few minor comments

src/test/java/org/assertj/core/internal/iterables/Iterables_assertNoneSatisfy_Test.java
epeee: :+1:

src/test/java/org/assertj/core/error/NoElementsShouldSatisfy_create_Test.java
```diff
@@ -16,24 +16,24 @@
...
 import org.junit.jupiter.api.Test;
 
 public class NoElementsShouldSatisfy_create_Test {
 
   @Test
-  public void should_create_error_message_any() {
+  public void should_create_error_message() {
     // GIVEN
-    ErrorMessageFactory factory = noElementsShouldSatisfy(newArrayList("Luke", "Yoda"), "Vader");
+    ErrorMessageFactory factory = noElementsShouldSatisfy(list("Luke", "Leia", "Yoda"), list("Luke", "Leia"));
```
joel-costigliola: not sure why but I had comment here to put a `%` to check that reserved `String.format` flags did not break the error message construction

joel-costigliola:
thanks one more comment and we good!

src/test/java/org/assertj/core/error/NoElementsShouldSatisfy_create_Test.java
epeee: I've added a testcase using a `%` and it looks like we are fine.

joel-costigliola:
Integrated thanks @epeee!

FYI I slightly improved the test I was commenting on this way:

```
  @Test
  public void should_fail_when_two_elements_satisfy_the_given_restrictions() {
    // GIVEN
    Consumer<String> restrictions = name -> assertThat(name).startsWith("L");
    // WHEN
    expectAssertionError(() -> iterables.assertNoneSatisfy(someInfo(), actual, restrictions));
    // THEN
    verify(failures).failure(info, noElementsShouldSatisfy(actual, list("Luke", "Leia")));
  }
```
No need to check the exception since this is done inside `expectAssertionError`.

PR #1438, opened 2019-02-15
Fix javadoc and reenable Javadoc doclint verification

PascalSchumacher:
@joel-costigliola It took me a while to figure it out, but I think the nested list error is not a bug in Javadoc. The usage of list tags was actually "incorrect" (browsers do not care, but Javadoc is more strict).

With the change below the error disappeared when using Javadoc on Java 11.0.1.


joel-costigliola:
Thanks Pascal, that's great!

PR #1439, opened 2019-02-15
Initial @Nullable annotations and using them in org.assertj.core.api.exception

### LINKED ISSUE #1414: jsr305 support for kotlin
#### Summary

probably pretty low priority, but it would be nice if assertj was annotated with jsr305 annotations so that kotlin could use them to tell where things are(n't) nullable.

https://kotlinlang.org/docs/reference/java-interop.html#jsr-305-support

checketts:
#### Check List:
* Fixes #1414 
* Unit tests : NA

So far this PR adds the various @NonNullApi annotations a `package-info.java` that default each package to assume all parameters and returns are non-null. Then I am going in and inspecting the code for null checks, javadocs comments about nullability, or where nulls are returned and are naturally appearing in the code.

Packages completed so far:

- [X] org.assertj.core.annotations
- [X] org.assertj.core.api.exception
- [ ] org.assertj.core.api.filter
- [ ] org.assertj.core.api.iterable
- [ ] org.assertj.core.api.recursive.comparison
- [ ] org.assertj.core.api (that one is huge!)
- [ ] org.assertj.core.condition
- [ ] org.assertj.core.configuration
- [ ] org.assertj.core.data
- [ ] org.assertj.core.description
- [ ] org.assertj.core.error.future
- [ ] org.assertj.core.error.uri
- [ ] org.assertj.core.error (another beast!)
- [ ] org.assertj.core.extractor 
- [ ] org.assertj.core.groups 
- [ ] org.assertj.core.internal 
- [ ] org.assertj.core.matcher 
- [ ] org.assertj.core.presentation
- [ ] org.assertj.core.util.diff.myers
- [ ] org.assertj.core.util.diff
- [ ] org.assertj.core.util.introspection
- [ ] org.assertj.core.util.xml
- [ ] org.assertj.core.util 

This is going to be a HUGE PR. What would be the best way for me to break this up so you can review it in a sane way?  Odds are I will find a potential NPE that we might want to actually handle.

### DIFF
--- a/src/main/java/org/assertj/core/annotations/NonNull.java
+++ b/src/main/java/org/assertj/core/annotations/NonNull.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2019 the original author or authors.
+ */
+package org.assertj.core.annotations;
+
+import javax.annotation.Nonnull;
+import javax.annotation.meta.TypeQualifierNickname;
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * A common annotation to declare that annotated elements cannot be {@code null}.
+ * Leverages JSR 305 meta-annotations to indicate nullability in Java to common tools with
+ * JSR 305 support and used by Kotlin to infer nullability of the API.
+ * <p>
+ * <p>Should be used at parameter, return value, and field level. Method overrides should
+ * repeat parent {@code @NonNull} annotations unless they behave differently.
+ * <p>
+ * <p>Use {@code @NonNullApi} (scope = parameters + return values) and/or {@code @NonNullFields}
+ * (scope = fields) to set the default behavior to non-nullable in order to avoid annotating
+ * your whole codebase with {@code @NonNull}.
+ *
+ * @see NonNullApi
+ * @see NonNullFields
+ * @see Nullable
+ */
+@Target({ElementType.METHOD, ElementType.PARAMETER, ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Nonnull
+@TypeQualifierNickname
+public @interface NonNull {
+}
--- a/src/main/java/org/assertj/core/api/exception/package-info.java
+++ b/src/main/java/org/assertj/core/api/exception/package-info.java
@@ -10,14 +10,9 @@
  *
  * Copyright 2012-2019 the original author or authors.
  */
+@NonNullApi
+@NonNullFields
 package org.assertj.core.api.exception;
 
-import org.assertj.core.annotations.Nullable;
-
-public final class PathsException extends RuntimeException {
-  private static final long serialVersionUID = 1L;
-
-  public PathsException(final String message, @Nullable final Throwable cause) {
-    super(message, cause);
-  }
-}
+import org.assertj.core.annotations.NonNullApi;
+import org.assertj.core.annotations.NonNullFields;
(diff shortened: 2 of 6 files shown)


src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
```diff
@@ -689,7 +690,8 @@ private static boolean expectedTypeIsNotSubtypeOfActualType(DualValue dualField)
     return expectedTypeIsNotSubtypeOfActualType(dualField.actual, dualField.expected);
   }
 
-  private static boolean expectedTypeIsNotSubtypeOfActualType(Object actual, Object expected) {
+  private static boolean expectedTypeIsNotSubtypeOfActualType(@Nullable Object actual, @Nullable Object expected) {
+    //noinspection ConstantConditions
     return !actual.getClass().isAssignableFrom(expected.getClass());
```
checketts: This is a potential NPE I found. Code that calls this method has nullable values. I added the `//noinspection`, but a better solution would be to handle the nulls

src/main/java/org/assertj/core/api/exception/package-info.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2019 the original author or authors.
+ */
+@NonNullApi
+@NonNullFields
+package org.assertj.core.api.exception;
```
checketts: To explain how this works:
I create some meta-annotations (@Nullable, @NonNullApi, etc) The meta annotations are annotated with the JSR 305 annotation, so user won't have to worry about them and the compiler can infer their use.

Then by annotating each `package-info.java` we set the API default to be `NonNull`, that way we only have to add `@Nullable` as it is appropriate.

joel-costigliola:
Thanks for the PR, I think we should start by introducing the needed annotations and apply them on a **small subset** of the code (like `org.assertj.core.api.exception`) so that I can review it without my brain exploding ;-).

checketts:
I hear you. @joel-costigliola 

OK, So I'll remove the annotations on all packages except `org.assertj.core.api.exception`, then after reviewing it, we'll merge that into master and I'll create a new PR based off that for another package and repeat.  Sound good?

joel-costigliola:
Yep!

checketts:
I've removed the other package's changes from this PR. Let me know if you have any questions.

src/main/java/org/assertj/core/annotations/NonNull.java
```diff
@@ -0,0 +1,45 @@
...
+ * @see NonNullApi
+ * @see NonNullFields
+ * @see Nullable
+ */
+@Target({ElementType.METHOD, ElementType.PARAMETER, ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Nonnull
+@TypeQualifierNickname
+public @interface NonNull {
```
joel-costigliola: should it be `Nonnull` for consitency with `javax.annotation.Nonnull` ?

src/main/java/org/assertj/core/annotations/NonNull.java
```diff
@@ -0,0 +1,45 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * A common annotation to declare that annotated elements cannot be {@code null}.
+ * Leverages JSR 305 meta-annotations to indicate nullability in Java to common tools with
+ * JSR 305 support and used by Kotlin to infer nullability of the API.
+ * <p>
+ * <p>Should be used at parameter, return value, and field level. Method overrides should
+ * repeat parent {@code @NonNull} annotations unless they behave differently.
```
joel-costigliola: Just to confirm this is not a typo, we need to annotate overridden method even if the parent was annotated ?

src/main/java/org/assertj/core/annotations/NonNull.java
checketts: I mimicked the Spring annotations. I also prefer `@NonNull`. But I'm happy to change it.

src/main/java/org/assertj/core/annotations/NonNull.java
checketts: I believe that supports what I saw in IntelliJ. It actually gave a specific warning along those lines.

src/main/java/org/assertj/core/annotations/NonNull.java
joel-costigliola: Alright, let's go with `@NonNull`

joel-costigliola:
Integrated thanks @checketts! 
The work can continue ;-)

PR #1446, opened 2019-02-25
anySatisfy: do not continue evaluating elements once we found a match

### LINKED ISSUE #1445: anySatisfy: do not continue evaluating elements once we found a match
#### Summary
In #1400 we added support for reporting all failing elements.
While doing so we changed the implementation in a case that we always evaluate all elements.

Let's optimize this and only evaluate as many elements as needed. E.g. once a element matches, we can stop evaluating all remaining elements.

I can provide a pr for this one soon.

epeee:
#### Check List:
* Fixes #1445 
* Unit tests : YES


joel-costigliola:
Integrated thanks @epeee !

Issue #1440, opened 2019-02-18
`NullPointerException` in `isEqualTo` assertion


Turbo87:
#### Summary

see https://github.com/Turbo87/intellij-emberjs/pull/246

```
java.lang.NullPointerException
        at java.util.regex.Matcher.getTextLength(Matcher.java:1283)
        at java.util.regex.Matcher.reset(Matcher.java:309)
        at java.util.regex.Matcher.<init>(Matcher.java:229)
        at java.util.regex.Pattern.matcher(Pattern.java:1093)
        at java.util.Formatter.parse(Formatter.java:2547)
        at java.util.Formatter.format(Formatter.java:2501)
        at java.util.Formatter.format(Formatter.java:2455)
        at java.lang.String.format(String.java:2940)
        at org.assertj.core.api.AbstractStringAssert.isEqualTo(AbstractStringAssert.java:310)
        at org.assertj.core.api.StringAssert$ByteBuddy$gUBfusQU.isEqualTo$accessor$e5ipmL7L(Unknown Source)
        at org.assertj.core.api.StringAssert$ByteBuddy$gUBfusQU$AssertJ$SoftProxies$xRrT7HuB.call(Unknown Source)
        at org.assertj.core.api.ErrorCollector.intercept(ErrorCollector.java:58)
        at org.assertj.core.api.StringAssert$ByteBuddy$gUBfusQU.isEqualTo(Unknown Source)
        at org.assertj.core.api.StringAssert$ByteBuddy$gUBfusQU.isEqualTo(Unknown Source)
```

joel-costigliola:
thanks for reporting this, could you provide a test case for it ?

Turbo87:
https://github.com/Turbo87/intellij-emberjs/blob/d526a6ca4974bace1d3ca396a3be57961bb0108e/src/test/kotlin/com/emberjs/resolver/EmberNameTest.kt#L107-L109 is the code that seems to trigger this. I'm not sure yet what specifically is causing it though 🤔

Turbo87:
looking though the code, could it be that `.isEqualTo(null)` is the problem?

joel-costigliola:
I won't be able to look at the code until next week, could you print actual
and  expected to get more information?

epeee:
I just had a look at this one and it looks like this one only happens if assertj is used from Kotlin.
A simple example which reproduces this one:
```kotlin
    @Test
    fun `should pass`() {
[...]

PR #1450, opened 2019-02-26
Fail with error message if the given expectedStringTemplate is null

epeee:
let's fail with a proper error message if the given 'expectedStringTemplate' is null.
note: this one is not the same as in #1440


joel-costigliola:
Fair enough

PR #1454, opened 2019-03-04
Update the maven-javadoc-plugin to version 3.1.0.

PascalSchumacher:
Let's see if it fixes javadoc on Java 11+.


PascalSchumacher:
As expected, updating the maven-javadoc-plugin is not enough:

[ERROR] Exit code: 1 - javadoc: error - The code being documented uses modules but the packages defined in https://docs.oracle.com/javase/8/docs/api/ are in the unnamed module.

joel-costigliola:
what a waste of time ... :(

PR #1456, opened 2019-03-05
Refactor macOS Conversion Script

nwinkler:
I tried to make a couple of changes to one of the macOS conversion scripts with the goal of simplifying the script and getting to a point where the script works for both GNU sed and BSD sed (macOS). Here's what I've done:

* Updated the Editorconfig configuration to work on `.sh` files as well.
* Fixed a couple of [Shellcheck](https://github.com/koalaman/shellcheck) warnings.
* Refactored the script to remove the repetition: The loop over the files is now coded only once in a function, removing the need for the copy/past code - DRY!
* Added `sed` options that should work for both GNU sed (Linux) and BSD sed (macOS), without the need for creating temp files on macOS. This is more reliable and should be a bit faster as well.

In general, this makes the script more readable, removes the repetition, makes it easier to maintain, and (most importantly) should allow to use the same script on both Linux and macOS.

Please let me know what you think about this - if this is OK, I can take a stab at refactoring the other script (JUnit5) as well, and then potentially remove the need for duplicate scripts for Linux and macOS. This last item would require updates to the documentation as well, that's why I worked on this script first to gauge whether there was interest in these changes.

Is there a test suite for these scripts that could be used to verify the results? I've run the script before/after the changes on one of my larger code bases and verified that the results were identical...


joel-costigliola:
That's awesome @nwinkler!

All your suggestions make sense, having the same script for mac and linux would be great and refactoring JUnit 5 one(s).

There is no test suite for the scripts, the fact that it worked on your code base is enough as the scripts were provided as a best effort migration tool. Having said that, creating a few files with different assertions to migrate and the expected result would be a nice thing to have.

nwinkler:
Thanks for the encouraging response! I'll work on consolidating the other scripts, so that in the end, there should only really be two scripts, one for JUnit and one for JUnit5 - I'll try to share code between them as much as possible as well. Maybe it will even be possible to merge them into one common script.

The documentation I was referring to is this one: http://joel-costigliola.github.io/assertj/assertj-core-converting-junit-assertions-to-assertj.html#automatic-conversion

I guess this is in the `gh-pages` branch, I'll try to create a second PR for that once I'm done with the scripts. I'll let you know when this is ready from my end.

In the past, I've worked with the [Bats](https://github.com/bats-core/bats-core) testing framework for Bash scripts. If I have time, I might add this at a later stage...

joel-costigliola:
The (doc) plan is to migrate http://joel-costigliola.github.io/assertj/assertj-core-converting-junit-assertions-to-assertj.html#automatic-conversion to the new site https://assertj.github.io/doc/ which written is asciidoc and much easier to maintain. I'm happy to do it unless you are keen to (I'll update the old site migration section to point to the new one too).

nwinkler:
I have now updated all three scripts (JUnit, JUnit5 and TestNG) to use the same loop/sed logic - they should all work with both GNU and BSD sed.

There's still a lot of repetition between the scripts, the only real difference is the sed expressions for search/replace. The common parts could be externalized into a separate library script, but then you'd have to deal with loading that script as a dependency, directory handling, etc.

If you don't mind, I'll leave it as-is for now - it's a lot cleaner than before works across different versions of sed. One thing that could be improved is the statement that checks for the OS, it currently only checks for `Darwin` (macOS), it does not detect any other BSD flavours. The `case` statement could be extended, or it would need to be changed to detect which options are supported by `sed`, instead of checking for the OS. I don't have a reliable way of doing that at the moment - this could be improved later.

Otherwise, this is ready from my perspective. Documentation and unit testing could be added later as separate tasks.

Please let me know what you think.

joel-costigliola:
Thanks @nwinkler, I will review this in the next few days.

joel-costigliola:
I finally had the time to give it a try at work, worked nicely, thanks for this awesome work.

nwinkler:
You're welcome - my pleasure!

PR #1460, opened 2019-03-07
Add method to use SoftAssertions with any custom Assertions

qoomon:
Add static `assertSoftly` method to be able to use  any common and custom assertion side by side within on assertSoftly call/scope e.g.
```java
        SoftAssertions.assertSoftly(
                () -> Assertions.assertThat("test1").isEqualTo("bar"),
                () -> GuavaAssertions.assertThat(...).contains(...),
                () -> MyAssertion.assertThat(...).contains(...)
        );
```

#### Check List:
* Unit tests : NO

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
@@ -36,13 +36,11 @@ public class AbstractSoftAssertions {
    *
    * @param assertion an assertion call.
    */
-  public void check(ThrowingRunnable assertion) {
+  public void check(ThrowingRunnable assertion) throws Exception {
     try {
       assertion.run();
     } catch (AssertionError error) {
       proxies.collectError(error);
-    } catch (Exception e) {
-      throw new RuntimeException(e);
     }
   }
 
--- a/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
@@ -41,6 +41,14 @@ public class AbstractSoftAssertions {
     AssertionError error = Failures.instance().failure(failureMessage);
     proxies.collectError(error);
   }
+  
+  public void assertThat(Runnable scope) {
+      try {
+        scope.run();
+      } catch (AssertionError error) {
+        proxies.collectError(error);
+      }
+  }
 
   /**
    * Fails with the given message built like {@link String#format(String, Object...)}.
--- a/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
@@ -42,9 +42,9 @@ public class AbstractSoftAssertions {
     proxies.collectError(error);
   }
   
-  public void check(ThrowingCallable assertion) {
+  public void check(ThrowingRunnable assertion) {
       try {
-        assertion.call();
+        assertion.run();
       } catch (AssertionError error) {
         proxies.collectError(error);
       }
@@ -196,7 +196,7 @@ public class AbstractSoftAssertions {
     return className.contains("$ByteBuddy$");
   }
   
-  public interface ThrowingCallable {
-    void call() throws Throwable;
+  public interface ThrowingRunnable {
+    void run() throws Exception;
   }
 }
--- a/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
@@ -44,7 +44,7 @@ public class AbstractSoftAssertions {
   
   public void check(ThrowingCallable assertion) {
       try {
-        assertion.run();
+        assertion.call();
       } catch (AssertionError error) {
         proxies.collectError(error);
       }
--- a/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
@@ -42,7 +42,7 @@ public class AbstractSoftAssertions {
     proxies.collectError(error);
   }
   
-  public void check(Runnable assertion) {
+  public void check(ThrowingCallable assertion) {
       try {
         assertion.run();
       } catch (AssertionError error) {
@@ -194,5 +194,9 @@ public class AbstractSoftAssertions {
 
   private boolean isProxiedAssertionClass(String className) {
     return className.contains("$ByteBuddy$");
+  }
+  
+  public interface ThrowingCallable {
+    void call() throws Throwable;
   }
 }
--- a/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
@@ -32,7 +32,16 @@ public class AbstractSoftAssertions {
   }
 
   /**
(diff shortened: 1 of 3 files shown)


joel-costigliola:
I will review this PR shortly, a quick look shows that it lacks tests but I need to look first at the problem you are trying to solve.

qoomon:
I will add tests and inline docuu if you approve this idea.

qoomon:
I've updated my initial description as well as simplify my implementation

qoomon:
Just had an idea for a better more generic approach
```java
        SoftAssertions softly = new SoftAssertions();
        softly.assertThat( () -> MyAssertion.assertThat(...).contains(...) );
```

joel-costigliola:
@qoomon sorry for the long delay, I'm sympathetic to the idea.

I'm not too sure about using `$` because it won't be clear what it does when discovering with code completion. I haven't a clear way of naming that I was thinking of `run`, `execute` or `check`, I'm putting all my suggestions below to get a more visual idea of it looks like:

```java
SoftAssertions softly = new SoftAssertions();
softly.check( () -> MyAssertions.assertThat(...).contains(...) );
softly.run( () -> MyAssertions.assertThat(...).contains(...) );
softly.execute( () -> MyAssertions.assertThat(...).contains(...) );
```
`check` has my preference at the moment.

qoomon:
just brain storming...
```java
SoftAssertions softly = new SoftAssertions();
softly.perform( () -> assertThat(...).contains(...) );
softly.handle( () -> assertThat(...).contains(...) );
softly.run( () -> assertThat(...).contains(...) );
softly.assertAll();
``` 
```java
SoftAssertions assertionGroup = new SoftAssertions();
assertionGroup.add( () -> assertThat(...).contains(...) );
assertionGroup.assertAll();
```

src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
```diff
@@ -42,7 +42,7 @@ public void fail(String failureMessage) {
     proxies.collectError(error);
   }
   
-  public void $(Runnable assertion) {
+  public void check(Runnable assertion) {
```
joel-costigliola: :+1:

qoomon:
Maybe we should use an own interface instead of runnable, that allows throwing exceptions on run method call.

joel-costigliola:
like this one?  https://github.com/joel-costigliola/assertj-core/blob/b82b86b2c54bb33c09103744ea0d7af4b37b4dee/src/main/java/org/assertj/core/api/ThrowableAssert.java#L32

qoomon:
exactly

joel-costigliola:
@qoomon sorry for the long delay, I'm happy with the idea and adding a `check` method.Can you add tests for it ?

qoomon:
sure I will do it during the week.

qoomon:
I've just had an other idea for the method name, what about?

`softly.defer( () -> assertThat(...).contains(...) );`

qoomon:
test is added aw well

joel-costigliola:
Sweet, I'll have a look at it shortly.

`defer` feels to me like something that is going to be computed at some point in the future but you don't know when, I still prefer `check` that is more obvious to me.

qoomon:
I agree, last naming idea :-) what about just `softly.test(...)` because it's like a sub test, and I think it's easier to understand what `test` does in an testing class than a new wording like `check` .

joel-costigliola:
IMO it is likely that what you put in `softly.test(...)` is only the assertion part of a test (the THEN part of a GIVEN WHEN THEN test) that's why I think it's more a check (or verify) than a test.

qoomon:
Okay You convinced me, check is probably the best wording

qoomon:
I've changed method name to `test`

qoomon:
One more method name idea :-D
```
softly.validate(() -> assertThat(...));
softly.validate(() -> assertThat(...));
```

qoomon:
What about `softly.declare(() -> assertThat(...).contains(...))`, actually I don't know why but I don't like `check` as method name, maybe because `check` is just a synonym for test

qoomon:
or `softly.assertion(() -> assertThat(...).contains(...))`

qoomon:
or `softly.handle(() -> assertThat(...).contains(...))`

joel-costigliola:
- `declare` : well it's not a declaration, so nope 
- `assertion`: not elegant as `assert` appears twice (`assertion` and `assertThat`)
- `handle` : not intention revealing, too generic

`check` is more specific than `test` IMO, it still has me preference.

I'm waiting for your next suggestions if you haven't exhausted all of them yet ;-)

qoomon:
Okay let's stick to test 😁

qoomon:
Sure

src/test/java/org/assertj/core/api/SoftAssertionsTest.java
```diff
@@ -151,6 +151,13 @@ public void setup() {
     iterableMap.put("rank", ranks);
   }
 
+  @Test
+  public void should_collect_soft_check_assertion() {
+    softly.check(() -> assertThat(true).isFalse());
+    softly.check(() -> assertThat(true).isTrue());
+    assertThat(softly.errorsCollected()).hasSize(1);
+  }
+
```
joel-costigliola: Add a test when the lambda throws an exception and check it is propagated as a runtime exception.

src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
```diff
@@ -31,6 +31,21 @@ public AbstractSoftAssertions() {
...
+   *
+   * @param assertion an assertion call.
+   */
+  public void check(ThrowingRunnable assertion) {
+    try {
+      assertion.run();
+    } catch (AssertionError error) {
+      proxies.collectError(error);
+    } catch (Exception e) {
+      throw new RuntimeException(e);
```
joel-costigliola: it would be nice to catch runtime and rethrow them as is instead wrapping in them in another runtime.

src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
```diff
@@ -31,6 +31,21 @@ public AbstractSoftAssertions() {
     return proxies.createSoftAssertionProxy(assertClass, actualClass, actual);
   }
 
+  /**
+   * Catch and collect assertion errors of any assertion call.
```
joel-costigliola: add an example why someone would use this method

joel-costigliola:
Thanks @qoomon, I have finished commenting your PR.

qoomon:
Rethrow exceptions and usage example added

src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
```diff
@@ -31,6 +31,28 @@ public AbstractSoftAssertions() {
...
+   *     softly.assertAll();
+   * }
+   * </pre>
+   *
+   * @param assertion an assertion call.
+   */
+  public void check(ThrowingRunnable assertion) throws Exception {
+    try {
+      assertion.run();
+    } catch (AssertionError error) {
```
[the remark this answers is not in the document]
qoomon: I am on vacation so would be great if you can take over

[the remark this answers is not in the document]
joel-costigliola: Will do!

joel-costigliola:
Integrated thanks @qoomon !

PR #1463, opened 2019-03-10
anySatisfy (Maps): do not continue evaluating elements once we found …

epeee:
…a match

pretty much the same as we did in ##1446 for iterables.


joel-costigliola:
Thanks @epeee.

PR #1465, opened 2019-03-10
Fixes #1441 use raw string when given no arguments

### LINKED ISSUE #1441: UnknownFormatConversionException when asserting String with `%` sign
#### Summary
I have a test written in Kotlin. It throws `UnknownFormatConversionException` when I try to assert a String containing `%` character. The same test in Java works properly. 

  > A hint: Java test calls method `org.assertj.core.api.AbstractAssert#isEqualTo`, while Kotlin test calls `org.assertj.core.api.AbstractStringAssert#isEqualTo` (which is buggie when there is no args). 

#### Example

1. Test in Java - works:

```java
import org.junit.Test;

public class AssertJTest {
  @Test
  public void assertJ12() {
    assertThat("foo %").isEqualTo("foo %");
  }
}
```

2. The same test in Kotlin - fails:

```kotlin
import org.assertj.core.api.Assertions.assertThat
import org.junit.Test

class AssertJTest {
  @Test
  fun assertJ12() {
    assertThat("foo %").isEqualTo("foo %")
  }
}
```

Failure message is:

```java
java.util.UnknownFormatConversionException: Conversion = '%'
	at java.base/java.util.Formatter.checkText(Formatter.java:2732)
	at java.base/java.util.Formatter.parse(Formatter.java:2718)
	at java.base/java.util.Formatter.format(Formatter.java:2655)
	at java.base/java.util.Formatter.format(Formatter.java:2609)
	at java.base/java.lang.String.format(String.java:2897)
	at org.assertj.core.api.AbstractStringAssert.isEqualTo(AbstractStringAssert.java:310)
	at controllers.api.AssertJTest.assertJ12(AssertJTest.kt:9)
```

asolntsev:
#### Check List:
* Fixes #1441 
* Unit tests : NO 

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractStringAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractStringAssert.java
@@ -312,4 +312,16 @@ public class AbstractStringAssert<SELF extends AbstractStringAssert<SELF>> exten
     return super.isEqualTo(format(expectedStringTemplate, args));
   }
 
+  /**
+   * Verifies that the actual value is equal to expected.
+   *
+   * This method needs to be overridden because otherwise another method is called from tests in Kotlin:
+   * org.assertj.core.api.AbstractStringAssert#isEqualTo(java.lang.String, java.lang.Object...)
+   *
+   * @see Assert#isEqualTo(java.lang.Object)
+   * @since 3.12.2
+   */
+  public SELF isEqualTo(String s) {
+    return super.isEqualTo(s);
+  }
 }


joel-costigliola:
Thanks @asolntsev, can you add a test for the overridden `isEqualTo` that checks it delegates to its parent method, it should be possible to do so with mockito with callRealMethod or spy

asolntsev:
@joel-costigliola As I know, it's not possible to verify that super.something() was called. 
And even if it was possible, would be such test really useful? 

A much better alternative is to add a test in Kotlin which would actually fail with current AssertJ implementation. But do you want to introduce kotlin to AssertJ project?..

joel-costigliola:
just want to make sure that the method is called in tests (which probably is but let's verify that).

Although I like Kotlin, I don't want to have a dependency on it.

joel-costigliola:
Integrated thanks @asolntsev!

PR #1468, opened 2019-03-13
Ignore static methods when finding property accessors

### LINKED ISSUE #1458: Ignore static methods when introspecting properties
#### Summary

See the discussion starting here: https://github.com/joel-costigliola/assertj-core/pull/1360#issuecomment-470200904

wilkinsona:
Previously, when looking for property accessors, Introspection would consider a static method with a name in the required form as being an accessor for a property. Where such a static method existed, this was likely to result in the wrong value being considered as the static method is unlikely to return the correct value for an instance's property.

This PR updates Introspection so that static methods are ignored when looking for property accessors.

#### Check List:
* Fixes #1458 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/util/introspection/Introspection.java
+++ b/src/main/java/org/assertj/core/util/introspection/Introspection.java
@@ -77,14 +77,15 @@ public final class Introspection {
     String capitalized = propertyName.substring(0, 1).toUpperCase(ENGLISH) + propertyName.substring(1);
     // try to find getProperty
     Method getter = findMethod("get" + capitalized, target);
-    if (getter != null) return getter;
+    if (getter != null && !Modifier.isStatic(getter.getModifiers())) return getter;
     if (bareNamePropertyMethods) {
       // try to find bare name property
       getter = findMethod(propertyName, target);
-      if (getter != null) return getter;
+      if (getter != null && !Modifier.isStatic(getter.getModifiers())) return getter;
     }
     // try to find isProperty for boolean properties
-    return findMethod("is" + capitalized, target);
+    Method isAccessor = findMethod("is" + capitalized, target);
+    return (isAccessor != null && !Modifier.isStatic(isAccessor.getModifiers())) ? isAccessor: null;
   }
 
   private static Method findMethod(String name, Object target) {
--- a/src/main/java/org/assertj/core/util/introspection/Introspection.java
+++ b/src/main/java/org/assertj/core/util/introspection/Introspection.java
@@ -34,7 +34,7 @@ public final class Introspection {
 
   /**
    * Returns the getter {@link Method} for a property matching the given name in the given object.
-   * 
+   *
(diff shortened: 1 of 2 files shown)


joel-costigliola:
Thanks @wilkinsona, I will review the PR shortly and likely cut another service pack release after it

src/main/java/org/assertj/core/util/introspection/Introspection.java
```diff
@@ -77,14 +77,15 @@ private static Method findGetter(String propertyName, Object target) {
     String capitalized = propertyName.substring(0, 1).toUpperCase(ENGLISH) + propertyName.substring(1);
     // try to find getProperty
     Method getter = findMethod("get" + capitalized, target);
-    if (getter != null) return getter;
+    if (getter != null && !Modifier.isStatic(getter.getModifiers())) return getter;
```
joel-costigliola: let' s extract a static `isValidGetter(Method method)` method to capture the `getter != null && !Modifier.isStatic(getter.getModifiers())` repeated 3 times, will be more readable too.

src/test/java/org/assertj/core/util/introspection/PropertyOrFieldSupport_getValueOf_Test.java
```diff
@@ -151,6 +151,27 @@ public void should_extract_single_value_from_maps_by_key() {
...
+    Object value = propertyOrFieldSupport.getValueOf("city", new StaticPropertyEmployee());
+
+    assertThat(value).isEqualTo("New York");
+  }
+
+  @Test
+  public void should_extract_field_value_if_only_static_is_method_matches_name() {
+    Object value = propertyOrFieldSupport.getValueOf("tall", new StaticBooleanPropertyEmployee());
+
+    assertThat(value).isEqualTo(false);
```
joel-costigliola: use `isFalse()`

joel-costigliola:
@wilkinsona I have finished commenting the PR.

src/main/java/org/assertj/core/util/introspection/Introspection.java
wilkinsona: Thanks for the suggestions. I've pushed a commit to do this.

src/test/java/org/assertj/core/util/introspection/PropertyOrFieldSupport_getValueOf_Test.java
wilkinsona: I don't think I can as `value` is an `Object` and there's no `isFalse()` method available on `ObjectAssert`.

joel-costigliola: yes, that's true

joel-costigliola:
Integrated thanks @wilkinsona!

PR #1469, opened 2019-03-14
Single parameter optimizations for extracting methods

scordio:
Several times I had the use case where a single private field was extracted and assertions were done on the resulting value, especially using `isInstanceOfSatisfying()`.

Previously, a call to `extracting("field").first()` was needed to obtain the corresponding `AbstractObjectAssert`.

This PR allows `extracting()` calls with only one string parameter to return an `AbstractObjectAssert` directly, in case of both field/property and map value extraction. The idea is similar to the existing `extracting` method which takes a single extractor `Function`.

Miror changes are also added in this PR (typos, warning suppressions and deletion of redundant `extends`).

#### Check List:
* Fixes NA
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractMapAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractMapAssert.java
@@ -575,7 +575,8 @@ public abstract class AbstractMapAssert<SELF extends AbstractMapAssert<SELF, ACT
    * @throws AssertionError if the actual map does not contain the given entries.
    */
   public SELF containsAllEntriesOf(Map<? extends K, ? extends V> other) {
-    Map.Entry<? extends K, ? extends V>[] entries = other.entrySet().toArray(new Map.Entry[other.size()]);
+    @SuppressWarnings("unchecked")
+    Map.Entry<? extends K, ? extends V>[] entries = other.entrySet().toArray(new Map.Entry[0]);
     maps.assertContains(info, actual, entries);
     return myself;
   }
@@ -618,7 +619,8 @@ public abstract class AbstractMapAssert<SELF extends AbstractMapAssert<SELF, ACT
    * @since 3.12.0
    */
   public SELF containsExactlyEntriesOf(Map<? extends K, ? extends V> map) {
-    Map.Entry<? extends K, ? extends V>[] entries = map.entrySet().toArray(new Map.Entry[map.size()]);
+    @SuppressWarnings("unchecked")
+    Map.Entry<? extends K, ? extends V>[] entries = map.entrySet().toArray(new Map.Entry[0]);
     return containsExactly(entries);
   }
 
@@ -989,7 +991,6 @@ public abstract class AbstractMapAssert<SELF extends AbstractMapAssert<SELF, ACT
    * @throws AssertionError if the actual map does not contain the given key.
    * @throws IllegalArgumentException if the given argument is an empty array.
    */
-
   public SELF containsKeys(@SuppressWarnings("unchecked") K... keys) {
     maps.assertContainsKeys(info, actual, keys);
     return myself;
@@ -1155,7 +1156,6 @@ public abstract class AbstractMapAssert<SELF extends AbstractMapAssert<SELF, ACT
    * @throws AssertionError if the actual map is {@code null}.
    * @throws AssertionError if the actual map does not contain the given values.
    */
-
   public SELF containsValues(@SuppressWarnings("unchecked") V... values) {
     maps.assertContainsValues(info, actual, values);
     return myself;
@@ -1499,12 +1499,14 @@ public abstract class AbstractMapAssert<SELF extends AbstractMapAssert<SELF, ACT
    * If a given key is not present in the map under test, a null value is extracted.
    * <p>
    * Example:
-   * <pre><code class='java'> Map&lt;String, Object&gt; map = new HashMap&lt;&gt;();
+   * <pre><code class='java'>
+   * Map&lt;String, Object&gt; map = new HashMap&lt;&gt;();
    * map.put("name", "kawhi");
    * map.put("age", 25);
    *
    * assertThat(map).extracting("name", "age")
-   *                .contains("kawhi", 25);</code></pre>
+   *                .contains("kawhi", 25);
+   * </code></pre>
    * <p>
    * Note that the order of extracted keys value is consistent with the iteration order of the array under test.
    * <p>
@@ -1515,12 +1517,43 @@ public abstract class AbstractMapAssert<SELF extends AbstractMapAssert<SELF, ACT
    * @return a new assertion object whose object under test is the array containing the extracted map values
    */
   @CheckReturnValue
-  public AbstractListAssert<?, List<? extends Object>, Object, ObjectAssert<Object>> extracting(Object... keys) {
+  public AbstractListAssert<?, List<?>, Object, ObjectAssert<Object>> extracting(Object... keys) {
     isNotNull();
     List<Object> extractedValues = Stream.of(keys).map(actual::get).collect(Collectors.toList());
     String extractedPropertiesOrFieldsDescription = extractedDescriptionOf(keys);
     String description = mostRelevantDescription(info.description(), extractedPropertiesOrFieldsDescription);
     return newListAssertInstance(extractedValues).as(description);
+  }
+
+  /**
+   * Extract the value of given key from the map under test, the extracted value becoming the new object under test.
+   * <p>
+   * For example, if you specify "id" key, then the object under test will be the map value for this key.
+   * <p>
+   * If a given key is not present in the map under test, a null value is extracted.
+   * <p>
+   * Example:
+   * <pre><code class='java'>
+   * Map&lt;String, Object&gt; map = new HashMap&lt;&gt;();
+   * map.put("name", "kawhi");
+   *
+   * assertThat(map).extracting("name")
+   *                .isEqualTo("kawhi");
+   * </code></pre>
+   * <p>
+   * Nested keys are not yet supported, passing "name.first" won't get a value for "name" and then try to extract
+   * "first" from the previously extracted value, instead it will simply look for a value under "name.first" key.
+   *
+   * @param key the key used to get value from the map under test
+   * @return a new {@link ObjectAssert} instance whose object under test is the extracted map value
+   */
+  @CheckReturnValue
+  public AbstractObjectAssert<?, ?> extracting(Object key) {
+    isNotNull();
+    Object extractedValue = actual.get(key);
+    String extractedPropertyOrFieldDescription = extractedDescriptionOf(key);
+    String description = mostRelevantDescription(info.description(), extractedPropertyOrFieldDescription);
+    return newObjectAssert(extractedValue).as(description);
   }
 
   /**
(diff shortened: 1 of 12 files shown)


scordio:
@joel-costigliola ping :slightly_smiling_face:

src/main/java/org/assertj/core/api/AbstractMapAssert.java
```diff
@@ -1499,12 +1499,14 @@ public SELF withThreadDumpOnError() {
    * If a given key is not present in the map under test, a null value is extracted.
    * <p>
    * Example:
-   * <pre><code class='java'> Map&lt;String, Object&gt; map = new HashMap&lt;&gt;();
+   * <pre><code class='java'>
+   * Map&lt;String, Object&gt; map = new HashMap&lt;&gt;();
```
joel-costigliola: please keep the original formatting

src/main/java/org/assertj/core/api/AbstractMapAssert.java
```diff
@@ -1499,12 +1499,14 @@ public SELF withThreadDumpOnError() {
...
    * map.put("name", "kawhi");
    * map.put("age", 25);
    *
    * assertThat(map).extracting("name", "age")
-   *                .contains("kawhi", 25);</code></pre>
+   *                .contains("kawhi", 25);
+   * </code></pre>
```
joel-costigliola: same here

src/main/java/org/assertj/core/api/AbstractMapAssert.java
```diff
@@ -1515,28 +1517,59 @@ public SELF withThreadDumpOnError() {
...
+  /**
+   * Extract the value of given key from the map under test, the extracted value becoming the new object under test.
+   * <p>
+   * For example, if you specify "id" key, then the object under test will be the map value for this key.
+   * <p>
+   * If a given key is not present in the map under test, a null value is extracted.
+   * <p>
+   * Example:
+   * <pre><code class='java'>
+   * Map&lt;String, Object&gt; map = new HashMap&lt;&gt;();
```
joel-costigliola: join lines and give add space after `>`

src/main/java/org/assertj/core/api/AbstractMapAssert.java
```diff
@@ -1515,28 +1517,59 @@ public SELF withThreadDumpOnError() {
...
+   * map.put("name", "kawhi");
+   *
+   * assertThat(map).extracting("name")
+   *                .isEqualTo("kawhi");
+   * </code></pre>
```
joel-costigliola: join lines

src/test/java/org/assertj/core/api/assumptions/Assumptions_assumeThat_with_extracting_Test.java
```diff
@@ -63,7 +63,7 @@ public void should_run_test_when_assumption_using_extracting_on_list_passes() {
 
   @Test
   public void should_run_test_when_assumption_using_extracting_on_object_passes() {
-    assertThatCode(() -> assumeThat(yoda).extracting("name").containsExactly("Yoda")).doesNotThrowAnyException();
+    assertThatCode(() -> assumeThat(yoda).extracting("name").isEqualTo("Yoda")).doesNotThrowAnyException();
   }
```
joel-costigliola: add a new test checking `extracting` with multiple arguments since it is now missing

src/test/java/org/assertj/core/api/iterable/IterableAssert_containsAnyElementsOf_Test.java
```diff
@@ -49,8 +49,9 @@ protected void verify_internal_effects() {
   public void should_compile_as_containsAnyElementsOf_declares_bounded_wildcard_parameter() {
     // GIVEN
     Map<String, String> map = newHashMap("some_key", "some_value");
+    Iterable<Map<String, String>> iterable = list(map);
     // THEN
-    assertThat(map).extracting("some_key").containsAnyElementsOf(list("some_value", "some_other_value"));
+    assertThat(iterable).extracting("some_key").containsAnyElementsOf(list("some_value", "some_other_value"));
```
joel-costigliola: add a new test checking `extracting` with multiple arguments since it is now missing

src/test/java/org/assertj/core/api/object/ObjectAssert_extracting_Test.java
```diff
@@ -70,10 +70,10 @@ public void should_use_property_field_names_as_description_when_extracting_tuple
...
     Employee luke = new Employee(2L, new Name("Luke", "Skywalker"), 26);
 
-    assertThatExceptionOfType(AssertionError.class).isThrownBy(() -> assertThat(luke).as("check luke first name")
-                                                                                     .extracting("name.first")
+    assertThatExceptionOfType(AssertionError.class).isThrownBy(() -> assertThat(luke).as("check luke first and last name")
+                                                                                     .extracting("name.first", "name.last")
                                                                                      .isEmpty())
-                                                   .withMessageContaining("[check luke first name]");
+                                                   .withMessageContaining("[check luke first and last name]");
   }
```
joel-costigliola: Let's add test specific to the new `extracting` method (as you did in `MapAssert_extracting_Test`)

joel-costigliola:
@scordio thanks, I have commented the PR.

src/main/java/org/assertj/core/api/AbstractMapAssert.java
scordio: Thanks for the feedback! I gave a newline to the code snippet because the rendering on IntelliJ aligns the first line wrongly:

Screenshot 2019-04-07 at 10 33 12

But it looks like [it doesn't hurt on html javadoc](http://joel-costigliola.github.io/assertj/core-8/api/org/assertj/core/api/AbstractMapAssert.html#extracting-java.lang.Object...-).

Shall I revert those lines back then? I guess if you want to address this topic, it's probably best to do it in a separate PR to have a consistent style everywhere.

src/test/java/org/assertj/core/api/assumptions/Assumptions_assumeThat_with_extracting_Test.java
scordio: I was not sure how to update this test because it is targeting an `Iterable` and from my point of view the `Map` was not really relevant here (also because the `extracting` of `AbstractIterableAssert` is never reached in case of maps). Since I was changing the `extracting` in `AbstractMapAssert` this test was also affected, therefore my quick fix was to wrap the `Map`.

Instead of adding another test here, I would suggest to remove the `Map` instance and to use any `Iterable` instead. What do you think?

src/test/java/org/assertj/core/api/iterable/IterableAssert_containsAnyElementsOf_Test.java
scordio: I was not sure how to update this test because it is targeting an `Iterable` and from my point of view the `Map` was not really relevant here (also because the `extracting` of `AbstractIterableAssert` is never reached in case of maps). Since I was changing the `extracting` in `AbstractMapAssert` this test was also affected, therefore my quick fix was to wrap the `Map`.

Instead of adding another test here, I would suggest to remove the `Map` instance and to use any `Iterable` instead. What do you think?

src/main/java/org/assertj/core/api/AbstractMapAssert.java
joel-costigliola: alright then, I remember doing this it used to add new lines but that might have changed

src/test/java/org/assertj/core/api/iterable/IterableAssert_containsAnyElementsOf_Test.java
joel-costigliola: I agree, this test actually was to make sure that `containsAnyElementsOf` could accept more than a `List<Object>`

src/test/java/org/assertj/core/api/object/ObjectAssert_extracting_Test.java
scordio: Side question: should the test `should_allow_assertions_on_array_of_properties_extracted_from_given_object_with_lambdas` stay in `ObjectAssert_extracting_with_function_Test` instead of `ObjectAssert_extracting_Test`?

scordio: I moved the test to the other class and this triggered an additional change regarding wildcards in `extracting(Function[])` methods, please double check it!

src/test/java/org/assertj/core/api/object/ObjectAssert_extracting_Test.java
```diff
@@ -67,17 +75,37 @@ public void should_use_property_field_names_as_description_when_extracting_tuple
...
+    Employee luke = new Employee(2L, new Name("Luke", "Skywalker"), 26);
+
+    assertThatExceptionOfType(AssertionError.class).isThrownBy(() -> assertThat(luke).as("check luke first and last name")
+                                                                                     .extracting("name.first", "name.last")
+                                                                                     .isEmpty())
+                                                   .withMessageContaining("[check luke first and last name]");
+  }
+
+  @Ignore
+  public void should_allow_to_specify_type_comparator_after_using_extracting_with_single_parameter_on_object() {
```
scordio: I created this test to have the same use case as the existing one with multiple parameters, but from my understanding there is no way to set a comparator for an comparison between objects. Am I missing something?
It could probably work with some more intrusive changes, maybe in `org.assertj.core.internal.Objects`.

scordio: After some investigation, I think the main "issue" is that the `extracting(String)`/`extracting(String[])` methods are returning an `ObjectAssert` instead of the specific type assertion. I have the feeling some improvement could be done in `PropertySupport` and `FieldSupport`, but I would prefer to do it in a separate pull request.

Would that be OK for you? Shall I leave this test as ignored or do you prefer to remove it?

joel-costigliola: `usingComparatorForType` is only used in the context of field by field comparison.
To use a specific comparator with `isEqualTo`, one had to call `public SELF usingComparator(Comparator<? super ACTUAL> customComparator)` but in our case ACTUAL = Object and thus it is not possible to use a `BigDecimal` comparator.

To allow this use case, we would need to allow users to specify the type of the extracted value + changing the signature of extracting to:
```
public <T> AbstractObjectAssert<?, T> extracting(String propertyOrField, Class<T> extractedType)
```
and a bunch of other changes to cascade the introduction of the generic parameter T.

That is I think valuable but definitely in another pull request.
```
T value = byName(propertyOrField).apply(actual);

Comment (joel-costigliola) [reply src/test/java/org/assertj/core/api/object/ObjectAssert_extracting_Test.java]: I "fixed" the test by using an Object comparator (hacky but possible):
```
    Comparator<Object> heightComparator = (o1, o2) -> {
      if (o1 instanceof BigDecimal) return BIG_DECIMAL_COMPARATOR.compare((BigDecimal) o1, (BigDecimal) o2);
      throw new IllegalStateException("only supported for BigDecimal");
    };
    assertThat(obiwan).extracting(Person::getHeight)
                      .usingComparator(heightComparator)
                      .isEqualTo(new BigDecimal("1.82"));
```

Comment (joel-costigliola) [reply src/test/java/org/assertj/core/api/object/ObjectAssert_extracting_Test.java]: oh and to ignore a test in JUnit 5, you must use [`@Disabled`](https://junit.org/junit5/docs/5.0.0-M2/api/org/junit/jupiter/api/Disabled.html), `@Ignore` is for JUnit 4 ;-)

Comment (scordio) [reply src/test/java/org/assertj/core/api/object/ObjectAssert_extracting_Test.java]: Nice hack :-) code updated!

**Comment by joel-costigliola** [comment]
Integrated thanks @scordio!

**Comment by PascalSchumacher** [comment]
I think we should add this as a breaking change to the release notes. This code:

```java
Map<String, Object> miRootVars = new HashMap<>();
assertThat(miRootVars).extracting("nrOfActiveInstances").containsOnly(1);
```

does not compile anymore after updating from `3.12.2` to `3.13.0`.

What do you think?

joel-costigliola:
@PascalSchumacher I fail to reproduce it, I have just tried this test which compiles and runs fine:
```java
  @Test
  public void map_extracting_example() {
    Map<String, Object> basketballPlayer = new HashMap<>();
    basketballPlayer.put("name", "kawhi");
    basketballPlayer.put("age", 25);

    // single value
    assertThat(basketballPlayer).extracting("name")
                                .isEqualTo("kawhi");
    // multiple values
    assertThat(basketballPlayer).extracting("name", "age")
                                .contains("kawhi", 25);
  }
```

joel-costigliola:
I'll update the breaking changes release notes! good catch!

joel-costigliola:
done https://assertj.github.io/doc/#assertj-core-3.13.0-breaking-changes

PascalSchumacher:
Thank you very much! 👍

PR #1476, opened 2019-03-23
adjust javadoc badge

epeee:
@joel-costigliola it looks like we have to use `org.assertj` instead of `joel-costigliola` for both, the badge and the link, isn't it?


joel-costigliola:
I was suspecting that but I'm not sure how this can work since the there is no `assertj-core` project under `org.assertj`.

PR #1477, opened 2019-03-24
Fix JavaDoc regarding `AnyOf` and `AllOf`

Riggs333:
* Fixes some mixing up between the two types `AnyOf` and `AllOf`.
* Unit tests : NO


Riggs333:
Thanks for the hint, @epeee! :+1:

joel-costigliola:
Integrated thanks @Riggs333!

PR #1478, opened 2019-03-25
Fix usingComparator() for Temporal assertions #1452

### LINKED ISSUE #1452: ZonedDateTime assertions are not consistent and don't expose isBetween
#### Summary

My test sometimes fails when the asserted value matches the (supposed to be inclusive) upper boundary. The example code fails with the exception:

java.lang.AssertionError: 
Expecting:
 <2019-03-04T10:22:45.622Z>
to be between:
 [2019-03-04T10:22:45.622Z[UTC], 2019-03-04T10:22:45.622Z[UTC]]

#### Example
``` java
ZonedDateTime boundary = ZonedDateTime.now();
assertThat(boundary).isBetween(boundary, boundary);    // passes

ZonedDateTime actual = boundary.withFixedOffsetZone();
assertThat(actual.isEqual(boundary)).isTrue();            // passes
assertThat(actual).isBetween(boundary, boundary);         //fails
```

puddlejumpr:
#### Check List:
* Fixes #1452 
* Unit tests : YES


src/main/java/org/assertj/core/api/AbstractLocalDateTimeAssert.java
```diff
@@ -229,6 +222,15 @@ public SELF isAfter(String localDateTimeAsString) {
     return isAfter(parse(localDateTimeAsString));
   }
 
+  public SELF isEqualTo(LocalDateTime expected) {
```
joel-costigliola: add javadoc and tests in `LocalDateTimeAssert_isEqualTo_Test` showing the new behavior (based on ChronoLocalDateTimeComparator).

joel-costigliola: same comment for `isNotEqualTo`

joel-costigliola: these comments also applies to `AbstractOffsetDateTimeAssert` and `AbstractZonedDateTimeAssert`.

src/main/java/org/assertj/core/api/AbstractLocalDateTimeAssert.java
```diff
@@ -68,9 +69,7 @@ protected AbstractLocalDateTimeAssert(LocalDateTime actual, Class<?> selfType) {
   public SELF isBefore(LocalDateTime other) {
```
joel-costigliola: reminder: describe the new behavior in the javadoc

joel-costigliola: Mentioning and explaining the breaking change should be done for all changed assertions, I know that's a bit of work but I think it is necessary.

src/main/java/org/assertj/core/internal/ChronoLocalDateTimeComparator.java
```diff
@@ -0,0 +1,37 @@
...
+  private ChronoLocalDateTimeComparator() {
+  }
+
+  @Override
+  public String description() {
+    return ChronoLocalDateTimeComparator.class.getSimpleName();
+  }
+
+  @Override
+  public int compare(ChronoLocalDateTime<?> o1, ChronoLocalDateTime<?> o2) {
```
joel-costigliola: Minor: rename o1 and o2 to date1 and date2

src/main/java/org/assertj/core/internal/Comparables.java
```diff
@@ -192,10 +190,50 @@ public String toString() {
    *           value is equal to or greater than the other value.
    */
   public <T extends Comparable<? super T>> void assertLessThan(AssertionInfo info, T actual, T other) {
+    this.assertLessThan(info, actual, other, ShouldBeLess::shouldBeLess);
```
joel-costigliola: minor: remove `this.`

src/main/java/org/assertj/core/internal/Comparables.java
```diff
@@ -192,10 +190,50 @@ public String toString() {
...
+   * @param info contains information about the assertion.
+   * @param actual the actual value.
+   * @param other the value to compare the actual value to.
+   * @param factory the desired error message factory to generate the suitable error message
+   * @throws AssertionError if the actual value is {@code null}.
+   * @throws AssertionError if the actual value is not before the other one: this assertion will fail if the actual
+   *           value is equal to or after the other value.
+   */
+  private <T extends Comparable<? super T>> void assertLessThan(AssertionInfo info, T actual, T other,
+                                                                TriFunction<T, T, ComparisonStrategy, ErrorMessageFactory> factory) {
```
joel-costigliola: minor: rename `factory` to `errorMessageFactory`

joel-costigliola:
The branch has conflict after we fix a typo in `isBeforeOrEqualsTo` (see https://github.com/joel-costigliola/assertj-core/issues/1506).

Hopefully, the conflicts should not be too difficult to resolve.

joel-costigliola:
A general comment on tests like `LocalDateTimeAssert_isBefore_Test`, since the implementation now delegates to `comparables`, the only test we need is to verify the delegation happens. 

You can take this test for reference:

```
public class StringAssert_isLessThan_Test extends StringAssertBaseTest {

  @Override
  protected StringAssert invoke_api_method() {
    return assertions.isLessThan("bar");
  }

  @Override
  protected void verify_internal_effects() {
    verify(comparables).assertLessThan(getInfo(assertions), getActual(assertions), "bar");
  }
}
```

src/main/java/org/assertj/core/api/AbstractLocalDateTimeAssert.java
```diff
@@ -112,9 +111,7 @@ public SELF isBefore(String localDateTimeAsString) {
   public SELF isBeforeOrEqualTo(LocalDateTime other) {
     Objects.instance().assertNotNull(info, actual);
```
joel-costigliola: this not necessary anymore as `comparables.assertIsBeforeOrEqualTo` does this check

joel-costigliola:
@puddlejumpr I have finish the first round of review on your code. Thanks for the good work!

I have mainly reviewed the changes for LocalDateTime assertions, most of my comments also apply to `AbstractOffsetDateTimeAssert` and `AbstractZonedDateTimeAssert`.

I'm planning to release 3.13.0 in a few weeks, it would be great to integrate this PR to it but it's ok it we ship it in 3.14 if you are too busy (no pressure, take the time you need).

puddlejumpr:
Hi Joel,
Sorry for the late response, I kept checking the bug thread for any replies. :(
Thank you for the review, I am going to look into your comments soon (hopefully on this week).

joel-costigliola:
@puddlejumpr no problem, take your time!

src/main/java/org/assertj/core/api/AbstractLocalDateTimeAssert.java
puddlejumpr: Could you please clarify what kind of tests you are missing exactly?

In one of your later comments you state that it is enough to test that the delegation happens to `comparables` (and `objects`, I assume). There are already tests which verify that the default comparator is `ChronoLocalDateTimeComparator` (`LocalDateTimeAssert_defaultComparator_Test`), and though the comparator is used in this assertion, it is not possible to use objects with type other than `LocalDateTime` as parameter, so the comparator is not utilized to its full extent.

What is your opinion about accepting `Object` as parameter (so API does not break) and delegating to `comparables` only if the expected value is instance of `ChronoLocalDateTime` (or `LocalDateTime`) and to `objects` otherwise?
Ofc, using `ChronoLocalDateTime` would break symmetry since there is no `ChronoLocalDateTimeAssert` currently, i.e.
```
LocalDateTime now = LocalDateTime.now();
ChronoLocalDateTime<JapaneseDate> inJapanese = JapaneseChronology.INSTANCE.localDateTime(now);
assertThat(now).isEqualTo(inJapanese); // succeed
assertThat(inJapanese).isEqualTo(now); // fails
```

puddlejumpr:
@joel-costigliola 
I have implemented most of the changes based on your comments and have them on another branch (puddlejumpr:issue-1452-updated) which is rebased on current master. I have refrained from force pushing to this pull request because i am in doubt if github can handle it without losing the review comments.

The only exception is the comments related to AbstractLocalDateTimeAssert. Can you elaborate on that?

src/main/java/org/assertj/core/api/AbstractLocalDateTimeAssert.java
joel-costigliola: :eyes:

puddlejumpr:
Closing due to new PR #1578.

PR #1480, opened 2019-03-31
Get rid of deprecated Double and Float constructor usage

epeee:
see e.g. [Double java doc](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Double.html#%3Cinit%3E(double)) for details.


joel-costigliola:
Thanks @epeee, you can directly pushed these kind of changes ;-)

PR #1486, opened 2019-04-07
Adds ByteBuffer assertions

### LINKED ISSUE #1470: Add assertion for ByteBuffer
#### Summary

Code that interacts with `ByteBuffer` objects could benefit a lot from an improved and dedicated `ByteBuffer` assertion. Typical assertions are:

* hasLength
* isEmpty
* contains/startsWith/endsWith/equals
** `String` (with `CharSet` defaulting), `byte[]`, `ByteBuffer` variants

The assertion could also benefit from a dedicated formatter that represents the buffer in a human-readable format. Currently, we get something like:

```
expected value: java.nio.HeapByteBuffer[pos=0 lim=7 cap=7]; actual value: java.nio.HeapByteBuffer[pos=0 lim=7 cap=7]
```

And this allows no diagnosis (except when using a debugger) what happened.

#### Example

```java
ByteBuffer buf = …
assertThat(buf).equalsTo("some-string");
assertThat(buf).equalsTo("some-string", Charset.defaultCharset());

assertThat(buf).contains("some-string");
assertThat(buf).contains("some-string", Charset.defaultCharset());

assertThat(buf).starts/endsWith("some-string");
```

jadeleeuw:
#### Check List:
* Fixes #1470
* Unit tests : YES

This PR adds the ByteBuffer assertions requested in #1470, along with a `Buffer.isFlipped()`, `Buffer.isNotEmpty()` and a `Buffer.hasRemainingLength` method.

Due to the amount of different parameters that these methods needed to support this has become quite a large PR. Fortunately, most of the code is therefore also quite repetitive, mostly transforming certain data types into different data types, and should therefore not be very complex.

There are several things to note however:

#### Buffers and ByteBuffers
The implementation has been split between `AbstractBufferAssert` and `AbstractByteBufferAssert`, to distinguish between assertions that are `Buffer` specific and assertions that are `ByteBuffer` specific. This allows people with custom `Buffer` implementations to still use certain assertions, and allows room in assertJ for other potential future `Buffer` assertions, such as e.g. `CharBuffer` without having to refactor `AbstractByteBufferAssert`. 

#### Buffer Flips
Due to the limitations of the buffer API there is, short of looping over the buffer and checking the contents of the buffer one by one, no way to check if the buffer is actually flipped. It is impossible to make a distinction between a buffer that is filled to its capacity and a buffer that simply has not been flipped (or manual changes in the position of the buffer). Due to these limitations, this method only checks if the position of the buffer equals 0.
The current implementations forces that the buffer under test has been flipped before the assertions happen, because that is when a buffer normally is finished writing and the reading (and therefore asserting) can proceed. If this requirement is unacceptable, please let me know.

#### Java6
I noticed that some assertion APIs such as e.g. `AbstractBDDSoftAssertions` have a "normal" variant and a "Java 6" variant. I placed the `Buffer` and `ByteBuffer` code in the "normal" variants, but I was not 100% sure if this is correct.

#### Buffer.hasRemainingLength
I defined the remaining length is the amount of space the buffer still has left, or in other words the space in the buffer that is not filled. It is defined as the capacity of the buffer minus the limit of the buffer.
I added this method because I thought it might come in handy for some people, but this is just a guess. Please let me know if this functionality needs to be removed.


joel-costigliola:
Thanks, sorry I'm been busy but I will review your PR for 3.13

jadeleeuw:
Not a problem. Take your time!

src/main/java/org/assertj/core/api/AbstractBufferAssert.java
```diff
@@ -0,0 +1,189 @@
...
+   * Example:
+   *
+   * <pre><code class='java'>
+   * Buffer buffer = ByteBuffer.allocate(10);
+   *
+   * // this assertion succeeds ...
+   * buffer.flip();
+   * assertThat(buffer).isFlipped();
+   *
+   * // ... but this one fails as "allocate" does not flip the buffer.
```
joel-costigliola: Let's not reuse buffer as it might be confusing as the previous `assertThat(buffer).isFlipped();` succeeded.
let's replace it with `assertThat(ByteBuffer.allocate(10)).isFlipped();`

src/main/java/org/assertj/core/api/AbstractBufferAssert.java
```diff
@@ -0,0 +1,189 @@
...
+   *
+   * <b>Important:</b>
+   * Due to the limitations of the Buffer API there is, short of looping over the buffer and checking the content one
+   * by one, no way to check if the Buffer is actually flipped. It is impossible to make a distinction between
+   * a buffer that is filled to its capacity and a buffer that simply has not been flipped (or manual repositioning
+   * of the buffer). Due to these limitations, this method only checks if the position of the buffer equals 0.
+   * <p>
+   * Example:
+   *
+   * <pre><code class='java'>
```
joel-costigliola: as a general comment for formatting consistency the code example should start after `<pre><code class='java'>` + one space and finish directly with `</code></pre>` without newline.
example:
```
   * <pre><code class='java'> Buffer buffer = ByteBuffer.allocate(10);
   *
   * // this assertion succeeds ...
   * buffer.flip();
   * assertThat(buffer).isFlipped();
   *
   * // ... but this one fails as "allocate" does not flip the buffer.
   * assertThat(buffer).isFlipped();</code></pre>
```

src/main/java/org/assertj/core/api/AbstractBufferAssert.java
```diff
@@ -0,0 +1,189 @@
...
+   * </code></pre>
+   *
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual {@code Buffer} is {@code null}.
+   * @throws AssertionError if the actual {@code Buffer} has not been flipped.
+   * @throws AssertionError if the actual {@code Buffer} is not empty.
+   */
+  public SELF isEmpty() {
+    isNotNull();
+    isFlipped();
```
joel-costigliola: Just curious why is it needed ? 
it it is it should made it clear the javadoc and have a failing example with an unflipped buffer.

src/main/java/org/assertj/core/api/AbstractBufferAssert.java
```diff
@@ -0,0 +1,189 @@
...
+   * </code></pre>
+   *
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual {@code Buffer} is {@code null}.
+   * @throws AssertionError if the actual {@code Buffer} has not been flipped.
+   * @throws AssertionError if the actual {@code Buffer} is empty.
+   */
+  public SELF isNotEmpty() {
+    isNotNull();
+    isFlipped();
```
joel-costigliola: same comment as above

src/main/java/org/assertj/core/api/AbstractBufferAssert.java
joel-costigliola: ok, after some googling I guess it is required to read data.

src/main/java/org/assertj/core/api/AbstractBufferAssert.java
```diff
@@ -0,0 +1,189 @@
...
+   * assertThat(buffer).isEmpty();
+   * </code></pre>
+   *
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual {@code Buffer} is {@code null}.
+   * @throws AssertionError if the actual {@code Buffer} has not been flipped.
+   * @throws AssertionError if the actual {@code Buffer} is not empty.
+   */
+  public SELF isEmpty() {
+    isNotNull();
```
joel-costigliola: `isNotNull()` is already called in `isFlipped()`.

src/main/java/org/assertj/core/api/AbstractBufferAssert.java
```diff
@@ -0,0 +1,189 @@
...
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual {@code Buffer} is {@code null}.
+   * @throws AssertionError if the actual {@code Buffer} has not been flipped.
+   * @throws AssertionError if the actual {@code Buffer}s remaining length is different than the given expected value.
+   */
+  public SELF hasRemainingLength(int expected) {
+    isNotNull();
+    isFlipped();
+
+    int remaining_length = actual.capacity() - actual.limit();
```
joel-costigliola: Use camelCase in production code for consistency (but can be used for test)

joel-costigliola:
I have started reviewing it but I'm not an expert in buffer so it's a bit hard. WIP !

commenting on your different sections:

- Buffers and ByteBuffers: good design !
- Buffer Flips: I'm not yet sure what to think yet due to my lack of expertise on buffers
- Java6: you did right, we are stopping doing Java 6 specific work.
- Buffer.hasRemainingLength : maybe people interested in buffer assertions can answer, otherwise I'd rather not put it and wait for someone to request it (to avoid unnecessary API bloating). @Stephan202 ?

@epeee or @PascalSchumacher can you review this PR if you have good knowledge of buffer ? (and enough spare time to review it)

joel-costigliola:
@jadeleeuw could you add some tests for soft assertions and assumptions ?

For soft assertions we want a test to make sure that in case one assertion calls another that raise an AssertionError, the error is collected but the main assertion method calls return.

Ex:
```
 public SELF contains(String expected) {
    isFlipped();

    String contentString = new String(getContent(actual));
    if (!contentString.contains(expected)) throwAssertionError(contentsShouldContain(expected, actual, Charset.defaultCharset()));
    return myself;
  }
```
we want to make sure that a non flipped buffer fails the soft assertion in `isFlipped`, the error is collected and we don't execute the remaining `contains` assertion code (hope my explanation is clear enough).

The tests should be added to `SoftAssertionsTest.should_propagate_AssertionError_from_nested_proxied_calls`.

jadeleeuw:
Thank you for the feedback! Unfortunately I have an import deadline coming up next week, so I am unsure whether I will be able to make adjustments according to the feedback before then. But if not, it will be soon after next week.

joel-costigliola:
@jadeleeuw no problem, take the time you need!

joel-costigliola:
@jadeleeuw I'm planning to release 3.13 in a couple weeks, I wonder if you want and can work on this PR to make for 3.13. If you don't that's not a problem, it will ship in 3.14.

jadeleeuw:
@joel-costigliola I think I can work on this to make it ready for 3.13. I'll incorporate your suggestions this weekend.

src/main/java/org/assertj/core/api/AbstractBufferAssert.java
jadeleeuw: Well the problem is that it is not exactly required (as in you could read data from the buffer without flipping first), but it is more of a convention to properly distinguish between writing to the buffer and reading from the buffer.

I decided to "enforce" this convention by doing the isFlipped check on all operations (which does a very basic check to determine if the buffer is flipped or not, but cannot actually guarantee it as explained in the "Buffer Flips" section of this PR), because I assume that if you are going to perform assertions on your buffer, that you have finished writing to the buffer.

I am not a heavy user of buffers however, so there might be a valid use case where you do not want to flip the buffer before performing assertions on it, proving my assumption wrong. In which case we should perhaps reconsider this isFlipped check.

jadeleeuw:
As I already noted in the comments, I am not a heavy user of buffers and therefore I agree that a review of this PR from someone with more expertise in that area would be very beneficial. Perhaps @mp911de who requested this feature could take a look or knows someone who could take a look?

**hasRemainingLength**
Regarding the `hasRemainingLength` buffer assertion, I propose to leave it in for now until either of the following has happened:
1. Someone with expertise regarding buffers has shared his opinion on the usefulness of this method.
2. It is time to merge this PR.

The reason for my proposal is that I think it is better to leave it in until it is time to merge and then remove it, rather than removing first and then potentially re-adding it before merging this PR. If you would rather have it removed right away, please let me know.

**Proxied calls**
Judging by the content in `SoftAssertionsTest.should_propagate_AssertionError_from_nested_proxied_calls` I assume that only one nested proxy call test is enough? Or would you like me to add a case for every assertion method that uses the `isFlipped` method (for both all `Buffer` and `ByteBuffer` assertions)?

I added one test case to `SoftAssertionsTest.should_propagate_AssertionError_from_nested_proxied_calls`. I hope I did it correctly but from your explanation I understand that the idea is that in that test case, two errors should trigger: the buffer both not being flipped (which is proxied) and the buffer not being empty. But the amount of errors collected only increases by one rather than by two, thereby confirming that only the proxied call is performed. Is that correct? Are other SoftAssertion test cases needed as well?

src/main/java/org/assertj/core/api/AbstractByteBufferAssert.java
```diff
@@ -0,0 +1,498 @@
...
+    isFlipped();
+
+    byteArrays.assertEndsWith(info, getContent(actual), getContent(expected));
+    return myself;
+  }
+
+  private byte[] getContent(ByteBuffer buffer) {
+    byte[] content = new byte[buffer.limit()];
+    for (int i = 0; i < content.length; i++) {
+      content[i] = buffer.get();
```
mp911de: This operation starts consuming the buffer at the current position and rewinds its position to zero afterward. There's nothing wrong with the current position, however the buffer position should be reset to the same position.

something like:

```
int position = buffer.position();

// get cotent
buffer.position(position);
```

Also, retrieval could be slightly simplified to `buffer.get(content)` instead of looping.

src/main/java/org/assertj/core/api/AbstractByteBufferAssert.java
```diff
@@ -0,0 +1,498 @@
...
+   */
+  public SELF endsWith(ByteBuffer expected) {
+    isFlipped();
+
+    byteArrays.assertEndsWith(info, getContent(actual), getContent(expected));
+    return myself;
+  }
+
+  private byte[] getContent(ByteBuffer buffer) {
+    byte[] content = new byte[buffer.limit()];
```
mp911de: That should be rather `buffer.remaining()`

src/main/java/org/assertj/core/api/AbstractBufferAssert.java
```diff
@@ -0,0 +1,175 @@
...
+   * // ... but this one fails as "buffer" has a different remaining length than the given expected value.
+   * assertThat(buffer).hasRemainingLength(10);</code></pre>
+   *
+   * @param expected integer value representing the expected length of the buffer.
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual {@code Buffer} is {@code null}.
+   * @throws AssertionError if the actual {@code Buffer} has not been flipped.
+   * @throws AssertionError if the actual {@code Buffer}s remaining length is different than the given expected value.
+   */
+  public SELF hasRemainingLength(int expected) {
```
mp911de: Remaining at `Buffer` is defined as

> the number of elements between the current position and the limit.

This means that the verification is to check how many readable bytes are left.

If you want to verify remaining capacity (that's what this method is doing), then let's rename this method to `hasRemainingCapacity()`

mp911de (changes requested):
Assertion methods regarding size/emptiness focus a lot on the buffer capacity and not so much about the actual content. I would expect that 

```java
ByteBuffer buf = ByteBuffer.allocate(100);

assertThat(buf).isEmpty();
assertThat(buf).hasSize(0);
```

succeeds because no data was written on the buffer.

To verify capacity, we would do the following:

```java
ByteBuffer buf = ByteBuffer.allocate(100);

assertThat(buf).hasCapacity(100);
assertThat(buf).hasLimit(0);
```

After writing data to a buffer:

```java
ByteBuffer buf = ByteBuffer.allocate(100);
buf.write("foo".getBytes()).flip();

assertThat(buf).isNotEmpty();
assertThat(buf).hasSize(3);
assertThat(buf).hasCapacity(100);
assertThat(buf).hasLimit(3);
```

jadeleeuw:
@mp911de First of all, thank you for your review! Secondly, I have a question I would like to ask you, because I have a feeling that I might have over-engineerd the assertions.

Each of the assertion methods requires the buffer under test to be flipped/in position 0, and will throw an assertion error otherwise. I did this because I assumed users would be done writing to the buffer before starting to perform assertions on the buffer, but I am second guessing my assumption and would like to ask you to share your opinion on this so that I can remove this requirement from the assertion methods if necessary.

src/test/java/org/assertj/core/api/BDDSoftAssertionsTest.java
```diff
@@ -741,8 +743,10 @@ public void should_propagate_AssertionError_from_nested_proxied_calls() {
     softly.then((Predicate<String>) null).accepts("a", "b", "c");
     // nested proxied call to isCompleted
     softly.then(new CompletableFuture<String>()).isCompletedWithValue("done");
+    //nested proxied call to isFlipped
+    softly.then((Buffer) ByteBuffer.allocate(10).put((byte) 10)).isEmpty();
     // it must be caught by softly.assertAll()
-    assertThat(softly.errorsCollected()).hasSize(4);
+    assertThat(softly.errorsCollected()).hasSize(5);
```
joel-costigliola: @jadeleeuw good test!

joel-costigliola:
@mp911de thanks for the feedback!

mp911de:
> Each of the assertion methods requires the buffer under test to be flipped/in position 0

I'm not sure this works for every scenario. When working with binary protocols, buffers can have a header and a body segment. Sometimes you verify the header and then you run body verifications in a subsequent block of assertions. Between these two, you'd advance the position.

Having `isNotEmpty()` basically verifies that there are readable bytes. Non-flipped buffers are a common source of bugs. I think getting rid of `isFlipped()` and embracing `isNotEmpty()` instead should be the way to go.

src/main/java/org/assertj/core/api/AbstractByteBufferAssert.java
jadeleeuw: Good suggestion! It did made me question the following however:
The current implementation asserts from the current position i.e.:
```
ByteBuffer buffer = ByteBuffer.wrap("someString".getBytes())
buffer.startsWith("some") // this assertion will succeed
```
But that also means that moving the position of the buffer will change the outcome of the assertions, i.e.:
```
ByteBuffer buffer = ByteBuffer.wrap("someString".getBytes())
buffer.position(1)
buffer.startsWith("some") // this assertion will fail
```

I thought that this behaviour would perhaps be desirable, as it allows you to use the assertion methods as you move through the buffer, allowing for combinations like:
```
ByteBuffer buffer = ByteBuffer.wrap("header:someHeader;body:someBody".getBytes())
buffer.startsWith("header")
... perform header assertions
buffer.startsWith("body")
... perform body assertions
```

But should this indeed be the desired behaviour? Or would it be more useful to make all (or perhaps some?) of the assertions check the entire buffer instead? A third option would be to have assertion methods for both ways.

mp911de: Assertions should not alter the position IMO but rather be good citizens and remember the last position.

The second example should be the following code to make it work:

```
ByteBuffer buffer = ByteBuffer.wrap("header:someHeader;body:someBody".getBytes())
buffer.startsWith("header")
// …
buffer.position(18);
buffer.startsWith("body")
... perform body assertions
```

I suggest to come up with a minimal API for now and learn from feedback.

> But should this indeed be the desired behaviour

It makes sense from how one works with buffers, so yes, I think so.

> assertions check the entire buffer instead

These approaches do not contradict. Initially, the buffer is at position `0` and that allows running assertions against the entire buffer. An extension to the API could look like `assertThat(buffer).atPosition(0).startsWith(…)`. For now, I'd keep it as simple as possible.

jadeleeuw:
My apologies for not updating for a while, I plan to have all the current feedback worked in on Monday, so that we can hopefully finish this before the 3.13 milestone.

joel-costigliola:
No problem @jadeleeuw, I prefer not to rush it. If it does not make it for 3.13 we will ship in the next release.

jadeleeuw:
There is still one large issue that I have trouble dealing with:

I am not sure how to properly implement a `isNotEmpty` method. The current implementation just checks  if the limit of the buffer and throws an assertion error if it is equal to zero. The problem with this approach is the following:

```
ByteBuffer buffer = ByteBuffer.allocate(10);
assertThat(buffer).isNotEmpty(); // this will succeed
```
Because `allocate` will set the `limit` to the same value as the `capacity` to prepare for writing, the current implementation of `isNotEmpty` will be fooled and assume that there are readable bytes, even though there are none. I could not come up with a solution for `isNotEmpty` that would work for all cases (at least not while limiting myself to the `Buffer` API, perhaps something could be done by letting `isNotEmpty` be an abstract method and let each `Buffer` subtype overwrite it and let them check if there is actual data there by trying to fetch it or something).

@mp911de I was wondering if you perhaps have any ideas on how to deal with this issue?

mp911de:
@jadeleeuw perhaps we should not try to be smart about it. An empty buffer is not empty in the sense that it can be consumed. On the flip side (pun intended!), when the buffer is not flipped, the results will be either quite strange or `isNotEmpty()` fails if the buffer was fully written.

jadeleeuw:
@mp911de I think you are right. I will update the code accordingly and clearly document the limitations/edge cases in the javadoc. When the solution proves to be insufficient by users then it can always be changed and that way we have some extra input on why the solution is not sufficient.

jadeleeuw:
As you have all noticed I have been quite busy lately and as such there has not been much progress on this PR. Unfortunately this will continue for a little while until around half October. I am unsure of how pressing this PR is, so if someone is urgently waiting for this and wants to finish it feel free, otherwise I will finish it when I come back. Apologies for my lack of available time.

joel-costigliola:
Thanks for the status, no need to apologize, we all contribute in our spare time!

jadeleeuw:
Small status update: My schedule is clearing up again, so the coming days I will pick up where I left off and then hopefully we can see it all the way through this time :smile:

jadeleeuw:
So it took a while for me to understand the comment given by @mp911de last Summer about how it should work, but now that I do it all makes sense. So I changed relatively little, the biggest difference being that the `AbstractBufferAssert` assertions are now relative to the current position of the buffer (just like the `AbstractByteBufferAssert` assertions), instead of absolute. This allows for the chaining of these assertions. So for example:
```
ByteBuffer buffer = ByteBuffer.wrap("SomeText");
assertThat(buffer).isNotEmpty() // This will succeed
... consume the content of the buffer until the end is reached
assertThat(buffer).isEmpty() // This will also succeed
```
Whereas previously the `AbstractBufferAssert` assertions only took the buffer as a whole into account. As already pointed out by @mp911de, there are also some aspects to this approach that might not be fully intuitive at first sight (but makes complete sense later on) for people new to buffers. For example, calling on a `isNotEmpty()` on a buffer that is fully written will fail if it has been fully consumed. Furthermore, forgetting to `flip()` the buffer will result in weird behaviour. I have documented this a little bit in the code, but let me know if you think more extensive documentation is needed and where the best place would be to put this information.

wsztajerowski:
Hi guys!
I was looking for a way to test ByteBuffer using AssertJ recently and found this PR. Having it in the released version would make my life much easier (I wouldn't have to re-implemented part of it by myself ;)
@jadeleeuw @joel-costigliola - is there anything I can help you with to merge it?

joel-costigliola:
@wsztajerowski if you are knowledgeable in byte buffers (seems like you do), we would be grateful if you could take a look at the PR and check if the implementation makes sense.

@jadeleeuw are you fine with that ?

wsztajerowski:
@joel-costigliola - sure - will do.

src/test/java/org/assertj/core/api/buffer/Buffer_isEmpty_Test.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Tests for <code>{@link AbstractBufferAssert#isEmpty()}</code>.
+ * @author Jean de Leeuw
+ */
+public class Buffer_isEmpty_Test {
+
+  @Test
+  public void should_pass_when_empty() {
+    ByteBuffer byteBuffer = ByteBuffer.allocate(10);
+    Buffer buffer = byteBuffer;
```
wsztajerowski, line 34: Out of curiosity, why are you using the Buffer reference here?

wsztajerowski (commented):
I've made a few comments - @jadeleeuw  @joel-costigliola - please let me know if you agree. Generally, I would keep the assertion API close to what the [Byte]Buffer class provides - by that, I mean using assertion names similar to the class method, e.g. `hasLimit()` instead of `hasLength()`.

In addition, I'd think of assertions for the buffer's position and mark - the way I see assertions for ByteBuffer is being able to test all basic properties in the fluent interface manner, like:
```
assertThat(byteBuffer)
  .hasPosition(3)
  .hasCapacity(10)
  .hasRemaining(5)
  .hasLimit(8)
  .doesNotHaveMark()
  .contains(expectedByteArray);
```

jadeleeuw:
> @jadeleeuw are you fine with that ?
Of course! 

> I've made a few comments - @jadeleeuw @joel-costigliola - please let me know if you agree.
Thank you for the comments! Unfortunately I don't have time to reply today, but let me get back to you on that this weekend.

src/main/java/org/assertj/core/api/AbstractBufferAssert.java
```diff
@@ -0,0 +1,194 @@
...
+   *
+   * @param expected integer value representing the expected length of the buffer.
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual {@code Buffer} is {@code null}.
+   * @throws AssertionError if the actual {@code Buffer} is empty.
+   * @throws AssertionError if the actual {@code Buffer} length is different than the given expected value.
+   */
+  public SELF hasLength(int expected) {
+    isNotEmpty();
+    if (actual.limit() != expected) throwAssertionError(shouldHaveLength(expected, actual.limit(), actual));
```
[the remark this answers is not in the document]
joel-costigliola: agreed, having an API close the byte buffer one will make it easier to understand

src/main/java/org/assertj/core/api/AbstractBufferAssert.java
```diff
@@ -0,0 +1,194 @@
...
+   * assertThat(buffer).hasLength(testArray.length - 1);</code></pre>
+   *
+   * @param expected integer value representing the expected length of the buffer.
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual {@code Buffer} is {@code null}.
+   * @throws AssertionError if the actual {@code Buffer} is empty.
+   * @throws AssertionError if the actual {@code Buffer} length is different than the given expected value.
+   */
+  public SELF hasLength(int expected) {
+    isNotEmpty();
```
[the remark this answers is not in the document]
joel-costigliola: agreed

src/main/java/org/assertj/core/api/AbstractBufferAssert.java
```diff
@@ -0,0 +1,194 @@
...
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual {@code Buffer} is {@code null}.
+   * @throws AssertionError if the actual {@code Buffer} is empty.
+   * @throws AssertionError if the actual {@code Buffer}s remaining number of elements is different than the given
+   *  expected value.
+   */
+  public SELF hasRemaining(int expected) {
+    isNotEmpty();
+
+    int remainingLength = actual.limit() - actual.position();
```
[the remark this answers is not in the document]
joel-costigliola: go for it

src/main/java/org/assertj/core/api/AbstractByteBufferAssert.java
```diff
@@ -0,0 +1,498 @@
...
+   * @throws AssertionError if the actual {@code ByteBuffer}s content does not end with the given expected {@code ByeBuffer}.
+   */
+  public SELF endsWith(ByteBuffer expected) {
+    isNotEmpty();
+
+    byteArrays.assertEndsWith(info, getContent(actual), getContent(expected));
+    return myself;
+  }
+
+  private byte[] getContent(ByteBuffer buffer) {
```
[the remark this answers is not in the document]
joel-costigliola: Good idea, are there different implementations/subclasses of byte buffer that would behave differently if you duplicate not as a subclass but as a ByteBuffer ? 

I'm asking because we did a similar thing with maps for some assertions and duplicating a specific implementation of map that accepted string keys ignoring their case, we ended with a duplicated hash map that did not ignore the case.

Overall, we can duplicate the byte buffer and see later if we have some corner cases because of it,  we should make it clear in the javadoc that we are duplicating the buffer.

src/main/java/org/assertj/core/presentation/StandardRepresentation.java
```diff
@@ -299,6 +303,21 @@ protected String toStringOf(SimpleDateFormat dateFormat) {
...
+  }
+
+  protected String toStringOf(ByteBuffer b) {
+    String separator = " ";
+    String result = Hexadecimals.byteArrayToHexString(b.array(), separator);
+    int arrowLocation = b.position() * (2 + separator.length());  // 2 = byte representation length in chars.
+    if (b.position() == b.capacity()) {  // Accommodate for no separator at the end
+      arrowLocation--;
+    }
+    result = "[" + result.substring(0, arrowLocation) + "->" + result.substring(arrowLocation) + "]";
```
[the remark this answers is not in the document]
joel-costigliola: let's do that if it helps to understand better the description of the buffer.

src/main/java/org/assertj/core/api/AbstractByteBufferAssert.java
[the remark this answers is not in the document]
wsztajerowski: My point is that Buffer#duplicate() creates a shallow copy of the buffer with shared data under the hood, so it's safe to manipulate the buffer's pointers (position, limit, mark) without the need to restore it after modification.

Regarding ByteBuffer subclasses, there is only one - MappedByteBuffer - which behaves in the same manner as the rest.

wsztajerowski:
Hi @jadeleeuw!
Have you had a chance to take a look at it? :)

src/main/java/org/assertj/core/api/AbstractBufferAssert.java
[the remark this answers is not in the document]
jadeleeuw: Same 👍

src/main/java/org/assertj/core/api/AbstractBufferAssert.java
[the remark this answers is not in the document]
jadeleeuw: 👍

src/main/java/org/assertj/core/api/AbstractByteBufferAssert.java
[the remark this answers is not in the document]
jadeleeuw: I am fine with this change, but I am not sure I understand the benefit. Ultimately, it seems both approaches leave the buffer in the same state as it originally was, so what is the benefit of doing the shallow copy?

src/main/java/org/assertj/core/presentation/StandardRepresentation.java
```diff
@@ -299,6 +303,21 @@ protected String toStringOf(SimpleDateFormat dateFormat) {
     return dateFormat.toPattern();
   }
 
+  protected String toStringOf(Buffer b) {
+    return b.toString();
+  }
+
+  protected String toStringOf(ByteBuffer b) {
+    String separator = " ";
+    String result = Hexadecimals.byteArrayToHexString(b.array(), separator);
```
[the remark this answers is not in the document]
jadeleeuw: Good catch!

src/main/java/org/assertj/core/presentation/StandardRepresentation.java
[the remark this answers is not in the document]
jadeleeuw: Sounds good!

jadeleeuw:
> Hi @jadeleeuw!
> Have you had a chance to take a look at it? :)
Your suggestions/comment look good! Would you like to make the changes or would you like me to make them? (Although arguably the biggest effort/challenge for this PR to be in a good state against seems to be the rebase required for ~5 years of changes 😅 )

jadeleeuw:
>  (Although arguably the biggest effort/challenge for this PR to be in a good state against seems to be the rebase required for ~5 years of changes 😅 )

Actually cherry-picking would make much more sense 🤦

joel-costigliola:
Don't hesitate to squash all your commits before cherry picking, at the end when integrate this work, it's going to be one commit anyway.

src/main/java/org/assertj/core/api/AbstractByteBufferAssert.java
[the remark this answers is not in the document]
wsztajerowski: IMHO, it is a good approach to work on a copy of the original buffer in all places by default - it's extremely easy to change the buffer's pointers, which would fail following assertions.
You have the right in this case - the result would be the same with both implementations.

wsztajerowski:
> > Hi @jadeleeuw!
> > Have you had a chance to take a look at it? :)
> 
> Your suggestions/comment look good! Would you like to make the changes or would you like me to make them? (Although arguably the biggest effort/challenge for this PR to be in a good state against seems to be the rebase required for ~5 years of changes 😅 )

Hey @jadeleeuw, I'll leave it to you if you have time to make these changes.

jadeleeuw:
> Hey @jadeleeuw, I'll leave it to you if you have time to make these changes.

I'll try to make the changes soon. I just rebased so it aligns again with upstream, but I still have to double check if I didn't miss anything/everything still makes sense :smile:

assertj-core/src/main/java/org/assertj/core/api/AbstractByteBufferAssert.java
```diff
@@ -0,0 +1,644 @@
...
+   */
+  public SELF isMarked() {
+    isNotNull();
+    try {
+      actual.duplicate().reset();
+    } catch (InvalidMarkException e) {
+      throwAssertionError(shouldBeMarked(actual));
+    }
+    return myself;
+  }
```
jadeleeuw, line 572: Unfortunately I had to place the `mark` related assertions on the `ByteBuffer` assertions even though the method is defined on `Buffer`. The reason for this is that the only way I found on how to check if something is marked is by calling `reset()`, which throws the `InvalidMarkException` when the buffer is not marked. But `reset()` modifies the state of the buffer, so I needed a copy of the buffer in order to do so. And while `Buffer` has a `duplicate()` method in more recent versions of Java, in Java 8 (for which AssertJ also has to work if I am not mistaken) the `duplicate()` method was only defined on `ByteBuffer`, and not on `Buffer`.

assertj-core/src/main/java/org/assertj/core/api/AbstractByteBufferAssert.java
```diff
@@ -0,0 +1,644 @@
...
+  public SELF isMarkedAt(int expected) {
+    isNotNull();
+    try {
+      int markedPosition = actual.duplicate().reset().position();
+      if (markedPosition != expected) throwAssertionError(shouldBeMarkedAt(expected, markedPosition, actual));
+    } catch (InvalidMarkException e) {
+      throwAssertionError(shouldBeMarked(actual));
+    }
+    return myself;
+  }
```
jadeleeuw, line 605: I took the liberty of creating a `isMarkedAt` assertion as well. I figured that asserting the position of the mark could also be interesting, but let me know if you think that brings no real value.

src/main/java/org/assertj/core/api/AbstractBufferAssert.java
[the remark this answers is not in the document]
jadeleeuw: I ended up removing both the `hasLength` and the `hasRemainingLength` assertions.

assertj-core/src/main/java/org/assertj/core/error/ShouldBeMarked.java
```diff
@@ -23,11 +24,35 @@ public static ErrorMessageFactory shouldBeMarked(AtomicMarkableReference<?> actu
     return new ShouldBeMarked(actual, true);
   }
 
+  /**
+   * Creates a new <code>{@link ShouldBeMarked}</code>.
+   *
+   * @param actual the actual buffer in the failed assertion.
+   * @return the created {@code ErrorMessageFactory}.
+   */
+  public static ErrorMessageFactory shouldBeMarked(Buffer actual) {
```
wsztajerowski, line 33: I don't know whether mixing Errors from AtomicMarkableReference and ByteBuffer is a good idea - they are separate abstractions even if they both have a "marker".
@joel-costigliola - is there any strict convention for re-using ErrorMessageFactories inside AsssertJ?

src/main/java/org/assertj/core/api/AbstractByteBufferAssert.java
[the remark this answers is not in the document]
jadeleeuw: I agree that it is quite easy to make changes to the buffer's pointers, so I am making the suggest change. 
I noticed however that the method you suggested above uses `copy.remaining()` instead of `copy.limit()` for the size of the `content` array.

I think in this case, it should be `copy.limit()` to prevent errors in cases where the buffer's position has already moved, as otherwise the entire content of the buffer will not fit in the `content` array.

But it did get me thinking: all the assertions I defined here are under the assumption the want to assert on the contents of the _entire_ buffer. So **not** relative to the current position. So for example, assuming I have a buffer with the String `test` in it, and my position is already on the `s`. Then doing `assertThat(buffer).endsWith("est")` would pass. Whereas if I did assertions relative to the position then `assertThat(buffer).endsWith("est")` would fail. But `assertThat(buffer).endsWith("st")` would succeed for both. 

Personally I think doing assertions on the content of the entire buffer makes sense. But I don't use buffers that often, so I don't want make any assumptions as to what others might find most useful.

dbarenholz:
Leaving this comment here for those (like me) who are less experience in Java: if you only want to check if any `Buffer` (be it `ByteBuffer` or `CharBuffer` or whatever) is `null`, then you can cast to `Object` and test on that:

```java
void test() {
  Buffer buffer = null;
  assertThat((Object) buffer).isNotNull(); // this fails, since it is null
}
```

In the same vein, `hasToString` just works as well (at least in my simple usecases).

```java
void test() {
  CharBuffer buffer = CharBuffer.wrap("hello");
  assertThat((Object) buffer).as("buffer is not null").isNotNull();
  assertThat((Object) buffer).hasToString("hello");
}
```

Of course it would be great if we could do this without casts, but at least this is a feasible workaround for me, and so hopefully also for someone else!

PR #1488, opened 2019-04-15
Add support for combined millisecond and timezone parsing (#1487)

### LINKED ISSUE #1487: yyyy-MM-dd'T'HH:mm:ss.SSSX is not supported in time parsing.
#### Summary

Related to #1133. Time parsing supports either milliseconds OR timezone, it does not work when both are specified.

#### Example
 ```java
 Calendar cal = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
 cal.clear();
 cal.set(2017, Calendar.NOVEMBER, 27, 20, 13, 24);
 Date date = cal.getTime();
 
 assertThat(date)
     .isEqualTo("2017-11-27T20:13:24.000+00:00");
 ```

Above test fails if target machine's default tiemzone is any other than UTC.

matejdro:
#### Check List:
* Fixes #1487 
* Unit tests : YES 

I had to add new parser to the first place, otherwise regular parser with ms (but without timezone) captures it.

One thing to note is that I could not get all unit tests to pass on my local machine because different regional settings on my PC cause test messages to be different than hardcoded ones. Not sure what I should do in this case.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractDateAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractDateAssert.java
@@ -63,7 +63,8 @@ public abstract class AbstractDateAssert<SELF extends AbstractDateAssert<SELF>> 
   /**
    * the default DateFormat used to parse any String date representation.
    */
-  private static final List<DateFormat> DEFAULT_DATE_FORMATS = newArrayList(newIsoDateTimeWithMsFormat(),
+  private static final List<DateFormat> DEFAULT_DATE_FORMATS = newArrayList(newIsoDateTimeWithMsAndUtcTimeZoneFormat(),
+                                                                            newIsoDateTimeWithMsFormat(),
                                                                             newTimestampDateFormat(),
                                                                             newIsoDateTimeWithUtcTimeZoneFormat(),
                                                                             newIsoDateTimeFormat(),
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -2431,7 +2431,8 @@ public class Assertions {
    *   assertThat(date).isEqualTo("2003/04/26");
    * } catch (AssertionError e) {
    *   assertThat(e).hasMessage("Failed to parse 2003/04/26 with any of these date formats: " +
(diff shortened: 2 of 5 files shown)


matejdro:
Should I just fix my code against travis messages? My local PC gives out different messages due to different locale. It is slightly annoying for tests to be locale-dependant.

joel-costigliola:
I had a quick look and from what I see in Travis, the error comes from the order of date formats in the error message.
```
[ERROR]   DateAssert_with_string_based_date_representation_Test.should_fail_if_given_date_string_representation_cant_be_parsed_with_any_custom_date_formats:134 
Expecting message:
 <"Failed to parse 2003 04 26 with any of these date formats:
   [yyyy-MM-dd'T'HH:mm:ss.SSSX,
    yyyy/MM/dd'T'HH:mm:ss,
    yyyy/MM/dd,
    yyyy-MM-dd'T'HH:mm:ss.SSS,
    yyyy-MM-dd HH:mm:ss.SSS,
    yyyy-MM-dd'T'HH:mm:ssX,
    yyyy-MM-dd'T'HH:mm:ss,
    yyyy-MM-dd]">
but was:
 <"Failed to parse 2003 04 26 with any of these date formats:
   [yyyy/MM/dd'T'HH:mm:ss,
    yyyy/MM/dd,
    yyyy-MM-dd'T'HH:mm:ss.SSSX,
    yyyy-MM-dd'T'HH:mm:ss.SSS,
    yyyy-MM-dd HH:mm:ss.SSS,
    yyyy-MM-dd'T'HH:mm:ssX,
    yyyy-MM-dd'T'HH:mm:ss,
    yyyy-MM-dd]">
```
As user custom date formats always come first, I believe the correct order is:
```
   [yyyy/MM/dd'T'HH:mm:ss,
    yyyy/MM/dd,
    yyyy-MM-dd'T'HH:mm:ss.SSSX,
    yyyy-MM-dd'T'HH:mm:ss.SSS,
    yyyy-MM-dd HH:mm:ss.SSS,
    yyyy-MM-dd'T'HH:mm:ssX,
    yyyy-MM-dd'T'HH:mm:ss,
    yyyy-MM-dd]">
```
which is what your code is doing so in that case the test assertion is wrong.

matejdro:
Thanks for the comments, I have commited fixed version.

joel-costigliola:
Integrated thanks @matejdro!

PR #1497, opened 2019-05-01
Add missing @since for single parameter 'extracting'

scordio:
Trivial: adding missing `@since` for the methods introduced in #1469.


joel-costigliola:
Good catch, thanks @scordio!

PR #1498, opened 2019-05-02
Add 'asInstanceOf' to Assert

### LINKED ISSUE #1020: Idea: complete AbstractAssert.as...() methods
Hi,
just an idea, similar to AbstractAssert.asList(), AbstractAssert could have asOptional, asMap, ... returning the respective specialized Assertions instance.

I think this is particularly useful in combination with .extracting()

### LINKED ISSUE #1109: isInstanceOf() should narrow the returned assertion type
#### Summary

This is related to the concept behind #1108.

At the moment, 
It would be handy if `AbstractAssert.isInstanceOf()` assertion could narrow the return type to return the most specific assertion class available.

#### Example

Current example (slightly contrived, but illustrates the point):
```java
// Code under test
Object obj = "This is a test";

// Test code
assertThat(obj).isInstanceOf(String.class);

String str = (String)obj;
assertThat(str).hasLength(14);
```

At the moment, AbstractAssert.isInstanceOf() returns the assertion's SELF type. So because the first `assertThat()` is called on a reference variable of type `Object`, the compiler will expect `isInstanceOf()` to return an `ObjectAssert`. Of course, this class doesn't have the String-specific assertions that I would like to do, so having now established that it is in fact a String I would like to do String-specific assertions on it. Hence the need to cast. I find that this is a pattern that occurs frequently in my testing (particularly using legacy interfaces that don't use genericised return values to help get around this problem at compile time).

A simple workaround would be to define a method as follows (which I've done in my own code), very similar to `catchThrowableOfType()`:

```java
public static <T> T assertAndCast(Object actual, Class<T> type) {
  assertThat(actual).isInstanceOf(type);
  return (T)actual;
}

// Test code
String str = assertAndCast(obj, String.class);
assertThat(str).hasLength(14);
```

However, a syntactically nicer solution (though much more difficult to implement) would be the following:

```java
// Test code
assertThat(obj)
  .isInstanceOf(String.class)
  .hasLength(14);
```
In order for this to work, `AbstractAssert.isInstanceOf()` must return a StringAssert instead of the self type when called with a Class<String> parameter.

This could be done by adding corresponding `isInstanceOf()` methods to `Assert` and `Abstract` assert for every known assertion type (similar to all the `assertThat()` methods), eg:

```java
public StringAssert isInstanceOf(Class<? extends String> type) {
  isInstanceOf((Class<Object>)type);
  return new StringAssert(actual);
}
public ListAssert isInstanceOf(Class<? extends List> type) {
  isInstanceOf((Class<Object>)type);
  return new ListAssert(actual);
}
```

The same principle could also apply to methods like `isSameAs()`:
```java
public StringAssert isSameAs(String expected) {
  isSameAs((Object)expected);
  return new StringAssert(actual);
}
public ListAssert isSameAs(List expected) {
  isSameAs((Object)expected);
  return new ListAssert(actual);
}
```

Of course, it would become a bit unwieldy trying to maintain all of these methods; it would make sense to have automatic code generator do this at compile time (it could also automatically generate all of the assertThat() methods in Assertions/AssertionsForClassTypes/AssertionsForInterfaceTypes). The disadvantage of generating them at compile time is that you wouldn't be able to add your custom assertion types into the hierarchy. I can't see an easy way around this other than having a custom classloader that can add the methods at load time, or maybe by having your custom deployment define its own versions of `Assert.java` and `AbstractAssert.java` and having them earlier in the class path. Even then, you'd need to convince your IDE to load the classes through your custom classloader in order for it to work, and I'm not sure how simple that would be.

#### Java 8 specific ?

I'm not sure if any attempt to implement this might fall foul of the same issues that required the split out into `AssertionsForClassTypes` and `AssertionsForInterfaceTypes` in Java 8.

### LINKED ISSUE #1490: Incorrect Assertion type while using extraction by function (BUG)
#### Summary
`assertThat` method is contextual:
```
  IterableAssert<Object> objectIterableAssert = assertThat(new HashSet<>());
  OptionalAssert<Object> objectOptionalAssert = assertThat(Optional.empty());
```
But `extracting` method, used with function, even though the method returns HashSet, still uses `AbstractObjectAssert` instead of contextual `IterableAssert`

#### Example

```java
public class User {
  pivate Set<String> messages; 
  public Set<String> getMessages() {
    return messages;
  }
}

//Correct:
IterableAssert<Object> objectIterableAssert = assertThat(new User().getMessages());

//Incorrect:
AbstractObjectAssert<?, ?> extracting = assertThat(user).extracting(User::getMessages);
```

scordio:
Following https://github.com/joel-costigliola/assertj-core/issues/1109#issuecomment-487390269, this PR adds an `asInstanceOf` method to `Assert`, which checks the value type and returns a new `Assert` with narrowed type.

Briefly, the client code would look like:

```java
assertThat((Object) 0).asInstanceOf(INTEGER).isZero();
assertThat((Object) "string").asInstanceOf(STRING).startsWith("str");
assertThat((Object) Function.class).asInstanceOf(CLASS).hasAnnotations(FunctionalInterface.class);
```

Some minor refactoring and cleanup is also done in this PR.

#### Check List:
* Fixes #1020, fixes #1109 and fixes #1490
* Unit tests : YES

#### To Do:
- [x] Additional static factories in `InstanceOfAssertFactories` for all the types supported by `Assertions#assertThat`
- [x] Additional typed factories for types supporting generics (e.g., `List`, `Predicate`, etc.)
- [x] Support for soft assertions and assumptions
- [x] Javadoc for `InstanceOfAssertFactories` static fields and methods


src/main/java/org/assertj/core/api/AssertFactory.java
```diff
@@ -12,28 +12,21 @@
...
-
-  /**
-   * Creates the custom Assert object for the given element value.
-   *
-   * Typically this will just invoke <code>assertThat(t)</code>
-   * @param t the type to convert to an Assert object
-   * @return returns <code>assertThat(t)</code>
-   */
-  ASSERT createAssert(T t);
+public interface AssertFactory<T, ASSERT extends Assert<?, ?>> extends Function<T, ASSERT> {
```
filiphr: My 2 cents about this change. Doing this change is a backwards incompatible change:

* Removal of the `createAssert` method
* Changing the generic `ASSERT` to extend `Assert`

IMHO changes to this interface is not really needed.

In theory the `AssertFactory` itself could be annotated with `@FunctionalInterface`

scordio: IMHO without those changes the semantic of this interface was wrong, there was nothing ensuring that the produced object was an instance an `Assert`. The interface was basically only a copy of `Function` with different method name.

With these changes the assert concept is enforced and `AssertFactory` gets any existing default method in `Function`. (BTW, it was already a `@FunctionalInterface`)

Assuming the implementing classes were respecting the contract (i.e., returning an `Assert` instance), the only breaking change is the method name changing from `createAssert` to `get`, which should not be a big deal. For example, the change on [`ObjectAssertFactory`](https://github.com/joel-costigliola/assertj-core/pull/1498/files/660461e1aa04f61e974554c8ba54ceb09ef50be8#diff-d14732ccd64a699c94b14a7b26ae9976) is just the method name.

filiphr: >IMHO without those changes the semantic of this interface was wrong, there was nothing ensuring that the produced object was an instance of `Assert`

On the contrary, different libraries might choose not to implement `org.assertj.core.api.Assert` and there is no need for an `AssertFactory` to enforce that. `AssertProvider` is something similar and doesn't enforce `Assert` as well.

> the only breaking change is the method name changing from `createAssert` to `get`, which should not be a big deal

Maybe it is not a big deal in an end user application project. However, there might be other libraries out there that are building on top of AssertJ. This would mean that if a user is using AssertJ and some other library the user would not be able to only upgrade AssertJ, but will need to wait for the other library to update to the latest version of AssertJ. Therefore, if I put on the hat as another library developer this is quite a big binary incompatible change.

P.S. I am not a member of the AssertJ team, only a contributor and a fellow open source developer. If the AssertJ team decides that this change is acceptable I am fine with it. I only wanted to voice my opinion

scordio: No worries and I definitely appreciate your feedback :-) your point is absolutely reasonable. I challenged this because I introduced another breaking change in #1469 and it was considered fine.

filiphr: Just checked that one as well, that is an end user breaking change (it doesn't really affect other libraries building on top of AssertJ).

I don't know the scope of using `AssertFactory` though. Maybe it is like you are saying as well.

joel-costigliola: I haven't had the time to look into this PR into details but I would rather avoid breaking changes unless it is not possible otherwise.

@scordio is it possible to keep `AssertFactory` as it is for that feature ? If it is not possible I'd rather prefer another interface to be defined independently of  `AssertFactory`.

scordio: @joel-costigliola it is possible, but before doing it I'd like to stress this point a bit more.

There are two main changes here:
1. Restricting the base output type to `Assert` instead of `Object`
2. Inheriting from `Function`

About point 1.:
* PROS: right semantic (why an `AssertFactory` should return something that is not an `Assert` at all?)
* CONS: breaking change only if an implementing class was returning a type which was not a child of `Assert` (which didn't happen with any assertj internal classes, so internally the semantic was already implicitly right)

About point 2.:
* PROS: avoiding repeating what `Function` already provides and gaining [`Function#andThen`](https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html#andThen-java.util.function.Function-), [`Function#compose`](https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html#compose-java.util.function.Function-) and [`Function#identity`](https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html#identity--)
* CONS: bigger breaking change compared to 1., because the functional interface method name changes from `createAssert` to `get`. Who would be affected by this change is any class which implemented the functional interface explicitely instead of using lambdas or method reference. Who is not affected is any existing lambda and method reference already in place. Since Assertj 3 is Java 8 based, I see a quite reduced risk here.

If you are still not convinced, I'm totally fine to revert 1., 2. or both!

joel-costigliola: @scordio I'm convinced for point 1 but not for point 2, I prefer keeping the `createAssert` method (the name is more expressive of the assertions domain) and I don't see a huge benefit for the inherited methods of `Function`.

scordio: @joel-costigliola Done! You may want to do the same for `AssertProvider` (not needed for the feature in this PR).

joel-costigliola:
I don't find `instanceOf` to be fluent enough when I was reviewing some of the examples in `InstanceOfAssertFactoriesTest`, I think `asAnInstanceOf` is a bit better in term or readability, ex:

```java
Object value = completedFuture("done");

assertThat(value).instanceOf(COMPLETABLE_FUTURE).isDone();
assertThat(value).asAnInstanceOf(COMPLETABLE_FUTURE).isDone();
```

WDYT ?

src/main/java/org/assertj/core/api/Assert.java
```diff
@@ -272,6 +272,32 @@
...
+  /**
+   * Uses an {@link InstanceOfAssertFactory} to verify that the actual value is an instance of a given type and to produce
+   * a new {@link Assert} narrowed to that type.
+   * <p>
+   * {@link InstanceOfAssertFactories} provides static factories for all the types supported by {@link Assertions#assertThat}.
+   * Additional factories can be created with custom {@code InstanceOfAssertFactory} instances.
+   * <p>
+   * Example:
+   * <pre><code class='java'> // assertions will pass
+   * assertThat((Object) &quot;abc&quot;).instanceOf(InstanceOfAssertFactories.STRING).startsWith(&quot;ab&quot;);
```
joel-costigliola: let's extract the actual value as an `Object` local variable for readability and to emphasize that are able to chain specific type assertions from an object

src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.function.IntPredicate;
+import java.util.function.LongPredicate;
+import java.util.function.Predicate;
+
+/**
+ * Static {@link InstanceOfAssertFactory InstanceOfAssertFactories} for {@link Assert#instanceOf(InstanceOfAssertFactory)}.
+ *
+ * @since 3.13.0
+ */
+public final class InstanceOfAssertFactories {
```
joel-costigliola: In order to make it easy for users to discover the factories they should be available in the entry point classes like `Assertions`, `BDDAssertions` and `WithAssertions`. 

To avoid the hassle of keeping the entry point classes in sync, we can simply make this class an interface and make the entry point classes to implement them.

src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
```diff
@@ -0,0 +1,70 @@
...
+
+  public static final InstanceOfAssertFactory<Predicate, PredicateAssert<Object>> PREDICATE = new InstanceOfAssertFactory<>(Predicate.class,
```
joel-costigliola: there is warning `Predicate is a raw type. References to generic type Predicate<T> should be parameterized` but I'm not sure this is easily solvable, if not, suppress the warning and add a comment explaining why this is ok.

src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
```diff
@@ -0,0 +1,70 @@
...
+                                                                                                                            PredicateAssert::new);
+
+  @SuppressWarnings("unchecked")
+  public static <T> InstanceOfAssertFactory<Predicate, PredicateAssert<T>> predicate(Class<T> type) {
+    return new InstanceOfAssertFactory<>(Predicate.class, predicate -> new PredicateAssert<>(predicate));
```
joel-costigliola: use `PredicateAssert::new`

joel-costigliola: same comment whenever we can use method reference

src/test/java/org/assertj/core/api/InstanceOfAssertFactoriesTest.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.concurrent.CompletionStage;
+import java.util.function.DoublePredicate;
+import java.util.function.IntPredicate;
+import java.util.function.LongPredicate;
+import java.util.function.Predicate;
+
+import org.assertj.core.util.Strings;
+import org.junit.jupiter.api.Test;
+
+public class InstanceOfAssertFactoriesTest {
```
joel-costigliola: a few general comments for most of the tests:
- use GIVEN WHEN THEN to better separate the data from the action, here we can skip WHEN and simply use WHEN THEN
- declare the actual value as `Object` to avoid the cast when practical (so not for test where the value is lambda)

example for `completable_future_factory_should_allow_completable_future_assertions`:
```java
// GIVEN
Object value = completedFuture("done");
// THEN
assertThat(value).instanceOf(COMPLETABLE_FUTURE).isDone();
```

src/test/java/org/assertj/core/api/InstanceOfAssertFactoriesTest.java
```diff
@@ -0,0 +1,105 @@
...
+  @Test
+  public void list_factory_should_allow_list_assertions() {
+    List<Object> list = asList("Homer", "Marge", "Bart", "Lisa", "Maggie");
+    assertThat((Object) list).instanceOf(LIST).contains("Bart", "Lisa");
+  }
+
+  @Test
+  public void typed_list_factory_should_allow_typed_list_assertions() {
+    List<String> list = asList("Homer", "Marge", "Bart", "Lisa", "Maggie");
+    assertThat((Object) list).instanceOf(list(String.class)).contains("Bart", "Lisa");
```
joel-costigliola: replace the assertion by this one that better shows the benefit of `instanceOf`:
```java
assertThat(list).first().instanceOf(STRING).startsWith("Ho");
```

src/test/java/org/assertj/core/api/abstract_/AbstractAssert_instanceOf_Test.java
```diff
@@ -0,0 +1,73 @@
...
+
+import org.assertj.core.api.*;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mock;
+
+/**
+ * Tests for <code>{@link AbstractAssert#instanceOf(InstanceOfAssertFactory)}</code>.
+ */
+public class AbstractAssert_instanceOf_Test extends AbstractAssertBaseTest {
```
joel-costigliola: don't extend `AbstractAssertBaseTest` as it is not relevant for this particular test

src/test/java/org/assertj/core/api/abstract_/AbstractAssert_instanceOf_Test.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @Mock
+  private InstanceOfAssertFactory<?, ?> mockFactory;
+
+  @Mock
+  private Assert<?, ?> mockAssert;
+
+  @BeforeEach
+  public void initMockito() {
+    initMocks(this);
```
joel-costigliola: let's use the JUnit 5 mockito extension: https://stackoverflow.com/questions/40961057/how-to-use-mockito-with-junit5

scordio: Thanks! I'm still a legacy Junit 4 user :-) I saw that page but I wasn't sure you wanted to add a new dependency for it. Will do then!

scordio:
@joel-costigliola `asAnInstanceOf` is definitely better, I was thinking to use an overloaded `as` because of fluency but it would have been overlapped the description api. What about the naming of the related classes? As a first feeling, I'd keep `InstanceOfAssertFactory` and related.

src/test/java/org/assertj/core/api/InstanceOfAssertFactoriesTest.java
scordio: The purpose of this test case is to show the typed List cast, the `String` one is already covered in `string_factory_should_allow_string_assertions()`.

Actually, this test case could be more expressive with:

`assertThat((Object) list).instanceOf(list(String.class)).first().startsWith("Ho");`

but I just noticed `first()` is returning an `ObjectAssert`. I'll investigate a bit if this can easily be changed.

src/test/java/org/assertj/core/api/abstract_/AbstractAssert_instanceOf_Test.java
scordio: I was trying to keep the style of the neighbor tests, then I'll get rid of it!

src/test/java/org/assertj/core/api/InstanceOfAssertFactoriesTest.java
scordio: That's also my preferred style but I noticed that was not widely used in the codebase. I'll adapt my tests to it.

src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
scordio: For some reason the compiler was not able to infer the right return type in case of method reference, this is why I kept the lambda. I'll play a bit more with this, if I don't find a solution I'll put a suppress warning and a comment on these cases.

scordio:
@joel-costigliola regarding `asAnInstanceOf`, what do you think if we take away the _An_, resulting in`asInstanceOf`? It's closer to the existing `isInstanceOf` methods

src/test/java/org/assertj/core/api/InstanceOfAssertFactoriesTest.java
```diff
@@ -0,0 +1,105 @@
...
+  @Test
+  public void predicate_factory_should_allow_predicate_assertions() {
+    Predicate<Object> predicate = Objects::isNull;
+    assertThat((Object) predicate).instanceOf(PREDICATE).accepts((Object) null);
+  }
+
+  @Test
+  public void typed_predicate_factory_should_allow_typed_predicate_assertions() {
+    Predicate<String> predicate = Strings::isNullOrEmpty;
+    assertThat((Object) predicate).instanceOf(predicate(String.class)).accepts("");
```
scordio: I'm not completely happy on how the typed factories should be used. Specifically, the current implementation has a type parameter which is actually not used for the factory creation but just for type inference:
```java
public static <T> InstanceOfAssertFactory<Predicate, PredicateAssert<T>> predicate(Class<T> type) {
    return new InstanceOfAssertFactory<>(Predicate.class, PredicateAssert<T>::new);
}
```

If I drop the unused parameter, then the caller must specify the type parameter explicitly:

```java
// With unused parameter
assertThat((Object) predicate).instanceOf(predicate(String.class)).accepts("");
// Without parameter
assertThat((Object) predicate).instanceOf(InstanceOfAssertFactories.<String>predicate()).accepts("");
```

I tend to favor the unused parameter to have better readability.
@joel-costigliola what do you think?

src/test/java/org/assertj/core/api/WithAssertions_delegation_Test.java
```diff
@@ -385,15 +380,15 @@ public void withAssertions_assertThat_iterable_assert_class_Test() {
...
+    assertThat((Iterable<TestItem>) Arrays.asList(ITEMS), ObjectAssert::new).first().isEqualTo(ITEMS[0]);
   }
 
   /**
    * Test that the delegate method is called.
    */
   @Test
   public void withAssertions_assertThat_iterator_Test() {
-    assertThat(list(ITEMS).iterator()).hasNext();
+    assertThat(Lists.list(ITEMS).iterator()).hasNext();
```
joel-costigliola: let's keep the static import for (slightly) better readability

joel-costigliola:
I'm happy with `asInstanceOf` :)

src/test/java/org/assertj/core/api/InstanceOfAssertFactoriesTest.java
joel-costigliola: I agree, `assertThat((Object) predicate).instanceOf(predicate(String.class)).accepts("");` is way better than  `assertThat((Object) predicate).instanceOf(InstanceOfAssertFactories.<String>predicate()).accepts("");` which is not elegant at all.

src/test/java/org/assertj/core/api/InstanceOfAssertFactoriesTest.java
joel-costigliola: I'm happy with the refactored tests you have pushed, as for `first()` let's discuss that in another issue/PR, this one is big enough ;-)

src/main/java/org/assertj/core/api/Assert.java
```diff
@@ -272,6 +272,35 @@
...
+   * <p>
+   * Example:
+   * <pre><code class='java'> // assertions will pass
+   * Object string = &quot;abc&quot;;
+   * assertThat(string).asInstanceOf(InstanceOfAssertFactories.STRING).startsWith(&quot;ab&quot;);
+   *
+   * Object integer = 1;
+   * assertThat(integer).asInstanceOf(InstanceOfAssertFactories.INTEGER).isNotZero();
+   *
+   * // assertions will fail
```
joel-costigliola: `assertion` (we show only one)

src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -136,7 +136,7 @@
  * @author William Delanoue
  */
 @CheckReturnValue
-public class Assertions {
+public class Assertions implements InstanceOfAssertFactories {
```
joel-costigliola: let's also make `BDDAssertions` implement `InstanceOfAssertFactories`

src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.function.Predicate;
+
+/**
+ * Static {@link InstanceOfAssertFactory InstanceOfAssertFactories} for {@link Assert#asInstanceOf(InstanceOfAssertFactory)}.
+ *
+ * @since 3.13.0
+ */
+public interface InstanceOfAssertFactories {
+
+  InstanceOfAssertFactory<Predicate, PredicateAssert<Object>> PREDICATE = new InstanceOfAssertFactory<>(Predicate.class,
```
joel-costigliola: Fix the warning with `@SuppressWarnings("rawtypes")`

src/test/java/org/assertj/core/api/InstanceOfAssertFactoriesTest.java
```diff
@@ -0,0 +1,161 @@
...
+    // WHEN
+    ListAssert<Object> result = assertThat(value).asInstanceOf(LIST);
+    // THEN
+    result.contains("Bart", "Lisa");
+  }
+
+  @Test
+  void typed_list_factory_should_allow_typed_list_assertions() {
+    // GIVEN
+    Object value = asList("Homer", "Marge", "Bart", "Lisa", "Maggie");
```
joel-costigliola: Minor: let's use `list` instead

src/test/java/org/assertj/core/api/InstanceOfAssertFactoriesTest.java
```diff
@@ -0,0 +1,161 @@
...
+    IntegerAssert result = assertThat(value).asInstanceOf(INTEGER);
+    // THEN
+    result.isZero();
+  }
+
+  @Test
+  void class_factory_should_allow_class_assertions() {
+    // GIVEN
+    @SuppressWarnings("deprecation")
+    Object value = Certificate.class;
```
joel-costigliola: replace by `Certificate` by `Function` to avoid using a deprecated class ;-)

joel-costigliola:
@scordio I had only a few minor comments, and a bigger one: support for soft assertions.

I'm happy to do it if you are busy, what we need to do is adding:
- `"asInstanceOf"` in `SoftProxies.METHODS_CHANGING_THE_OBJECT_UNDER_TEST`
- tests in `SoftAssertionsTest` and `BDDSoftAssertionsTest` to make sure proxying works correctly
- tests `Assumptions_assumeThat_with_asInstanceOf_Test` for assumptions

E.g.:

```
  @Test
  public void soft_assertions_should_work_with_asInstanceOf() {
    // GIVEN
    Object s = "abc";
    // WHEN
    softly.assertThatObject(s)
          .as("startsWith")
          .asInstanceOf(Assertions.STRING)
          .startsWith("b");
    // THEN
    List<Throwable> errorsCollected = softly.errorsCollected();
    assertThat(errorsCollected).hasSize(1);
    assertThat(errorsCollected.get(0)).hasMessageContaining("[startsWith]");
  }
```

src/main/java/org/assertj/core/api/Assertions.java
scordio: `BDDAssertions` extends `Assertions` so it's already getting the factories, do you want the `implements` to be explicit also here?

src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
scordio: I wanted to play a bit more to see if I can avoid the suppression, if not I'll use it

src/test/java/org/assertj/core/api/InstanceOfAssertFactoriesTest.java
scordio: `Lists.list()` clashes with `InstanceOfAssertFactories.list()` so the static import for the former would not be an option, should I use Lists.list() without the static import?

src/test/java/org/assertj/core/api/WithAssertions_delegation_Test.java
scordio: This test class implements `WithAssertions` which extends `InstanceOfAssertFactories`, so now in the scope we have `InstanceOfAssertFactories.list()` which is clashing with `Lists.list()`

scordio:
I'd love to cover also soft assertions and assumptions, it's a great chance to learn more about assertj internals :-)
About my time, I think I won't be able to make some significant progress before the end of this week. If it's not an issue for you, I'd be happy to continue.
To consider this PR complete, what is also missing is more factories aligned with `assertThat` methods in `Assertions`

src/test/java/org/assertj/core/api/InstanceOfAssertFactoriesTest.java
scordio: I investigated a bit and the main issue for `first()` is the `ObjectAssertFactory` in the `ListAssert` constructor, some refactoring would be needed there. I'll open a separate issue about it.

scordio:
BTW, before addressing `first()`, I think a nice extension of this PR for `AbstractObjectAssert` could be:
```java
<ASSERT extends Assert<?, ?>> ASSERT extracting(String, InstanceOfAssertFactory<?, ASSERT>);
<ASSERT extends Assert<?, ?>> ASSERT extracting(Function<? super ACTUAL, ?>, InstanceOfAssertFactory<?, ASSERT>);
```
and also for `AbstractMapAssert`:
```java
<ASSERT extends Assert<?, ?>> ASSERT extracting(Object, InstanceOfAssertFactory<?, ASSERT>);
```

They are basically shortcuts for chaining `.extracting(something).asInstanceOf(somethingElse)`.
Maybe some changes there could help the `first()` issue.

joel-costigliola:
@scordio take your time, just let me know if you can't make it for next week.

Let's discuss/make the `extracting` improvement in another issue so that we can get this one done!

scordio:
@joel-costigliola sure, I'll keep you updated. Having a separate issue/PR for `extracting` was implied :-)

scordio:
@joel-costigliola almost done, please check last commits. Some other factories are still missing but it's just a time consuming activities. You could merge the current stuff and I'll create another PR for the missing ones if you want to speed up.

I was thinking to replace the inner implementation of `asList` using `asInstanceOf`, but then I noticed in `asInstanceOf` I don't propagate the assert description to the new instance. I don't think it's an easy change because of generics parameters, I'd prefer to address it in a separate PR.

Summarizing, new issues/PRs would be needed for:
- missing factories aligned with `assertThat` methods
- `asInstanceOf` should keep the existing description:
```java
// not working
public <ASSERT extends Assert<?, ?>> ASSERT asInstanceOf(InstanceOfAssertFactory<?, ASSERT> instanceOfAssertFactory) {
  return instanceOfAssertFactory.createAssert(actual).as(info.description());
}
```
- `extracting` extension for `AbstractObjectAssert`:
```java
<ASSERT extends Assert<?, ?>> ASSERT extracting(String, InstanceOfAssertFactory<?, ASSERT>);
<ASSERT extends Assert<?, ?>> ASSERT extracting(Function<? super ACTUAL, ?>, InstanceOfAssertFactory<?, ASSERT>);
```
- `extracting` extension for `AbstractMapAssert`:
```java
<ASSERT extends Assert<?, ?>> ASSERT extracting(Object, InstanceOfAssertFactory<?, ASSERT>);
```
- `first()` returning `ObjectAssert`

scordio:
@joel-costigliola Javadoc is done and I wouldn't add/change anything else to the PR unless you think it's required

joel-costigliola:
Sweet, I'll have a look at it after I'm done with https://github.com/joel-costigliola/assertj-core/pull/1453

joel-costigliola:
@scordio I have started reviewing the PR, and trying it as a user would. It works fine but (in eclipse at least) I'm not getting code completion for any `InstanceOfAssertFactory` of `InstanceOfAssertFactories`. 

That's a bit unfortunate as it makes it harder to find what to put in `asInstanceOf()`. I managed to get the expected code completion after adding the `InstanceOfAssertFactories` fields directly in `Assertions` but obviously this would need to be done on all other entry point class (WithAssertions, BDDAssertions ...) - not great because of the duplication and the involved maintenance when new types are added.

I was wondering if that is eclipse specific and whether Idea is able to provide better code completion (if you happen to use Idea) ?

scordio:
Yes I use IntelliJ and I think the code completion worked fine but let me do some tests again from user perspective. I'll be able to come back to you in half day approximately.

scordio:
To have a 100% independent test, I ran `mvn install` against my branch to have the snapshot jar in my local maven repo and then I created a new app with it as dependency.

* Basic completion:

Screenshot 2019-05-28 at 22 42 14

* Smart completion:

Screenshot 2019-05-28 at 22 42 27

It seems ok so far! Do you have issues with both static fields and methods or just with fields?

joel-costigliola:
Looks good, I think eclipse code completion is pretty basic but having `InstanceOfAssertFactories` as a favorite static import does the trick.

BTW, I use [assertj-examples](https://github.com/joel-costigliola/assertj-examples) when I want to try assertions as a user.

joel-costigliola:
Integrated, thanks @scordio for the great work!

scordio:
Thanks @joel-costigliola! You may want to follow up/close some of #1020, #1109 and #1490 which were related to this PR.

joel-costigliola:
@scordio I'm starting to document the `asInstanceOf`  starting with some examples in [assertions-examples](https://github.com/joel-costigliola/assertj-examples/tree/master/assertions-examples/src/test/java/org/assertj/examples).

I'm wondering if we should make all lowercase methods in `InstanceOfAssertFactories` uppercase for consistency.

The example below shows the inconsistency:
```java
Object value = "Once upon a time in the west";
assertThat(value).asInstanceOf(STRING).startsWith("Once");

value = list(vilya, nenya, narya);
assertThat(value).asInstanceOf(list()).contains(nenya);
assertThat(value).asInstanceOf(list(Ring.class)).contains(nenya);
```

scordio:
Having uppercase methods would be a against Java naming convention, it could seem strange to the developer.

What if we get rid of the methods without parameters? they can be easily converted to a static field. The concept would be like with the static field you get the "base" factory, but if you want more control you use the method and the Javadoc of the field could point to the corresponding method version.

Something like:
```java
Object value = "Once upon a time in the west";
assertThat(value).asInstanceOf(STRING).startsWith("Once");

value = list(vilya, nenya, narya);
assertThat(value).asInstanceOf(LIST).contains(nenya);
assertThat(value).asInstanceOf(list(Ring.class)).contains(nenya);
```

Or, crazy idea, everything is a method, but it seems to me overkilling.

scordio:
@joel-costigliola BTW, let me know if you would like me to help somehow for the documentation!

joel-costigliola:
Having `LIST` is good idea, the only disadvantage is that it makes a bit harder to discover `list(Class)` but I guess this can be mitigated with Javadoc and the documentation.

Not keen for moving everything to methods, I prefer to avoid `()` when possible for readability ;-)

Thanks for the documentation help, I have already started so I'm going to finish it but I'm keen to get your feedback when I have a first version.

scordio:
Ok, then I'll take care of the method conversions in a PR!

scordio:
@joel-costigliola Following the last refactoring, I feel like the sync test should be split in two parts:
1. One checking that for each assertion there is a corresponding field factory
2. Another checking that for each assertion having a type parameter there is a corresponding method factory

For point 1, I see two corner cases:
* Field factory missing for `ObjectAssert` --> easily fixable with an `OBJECT` factory, I just see this factory as useless (why would you use it? To "downgrade" to an `ObjectFactory` with `Object` as value type?)
* Field factory missing for `AbstractComparableAssert` --> in this case I don't see how to have a field factory (it doesn't make much sense) so probaby this factory should be filtered out from the input assertions in case of the field test

WDYT? I guess it would make sense to spend some time on it.

joel-costigliola:
Good idea for 1. and 2., I understand the first corner case and as you do I don't want to have an `OBJECT` factory, so the way to go is to ignore the Object assertion in the first place.

I also agree on the second corner case.

Issue #1502, opened 2019-05-10
Support formatting overloads in more AbstractThrowableAssert methods


antoniomacri:
#### Summary
In particular, I'd use `hasMessageContaining(String message, Object... parameters)`.

#### Example
Instead of:
```java
assertThat(thrown).hasMessageContaining(String.format("My %s format string", "beautiful"));
```
write
```java
assertThat(thrown).hasMessageContaining("My %s format string", "beautiful");
```
just like:
```java
assertThat(thrown).hasMessage("My %s format string", "beautiful");
```

joel-costigliola:
Interestingly #1501 goes in a different way unfortunately incompatible with your suggestion.

I totally get why you ask for it but I think allowing multiple string to be checked is more useful than supporting `String.format` syntax. 

@antoniomacri have you got a compelling argument for your suggestion that I have missed ?

antoniomacri:
No, it's just convenience.

But instead of adding varargs to `withMessageContaining` a new `withMessageContainingAll` could be defined?

What do you think?

joel-costigliola:
Good idea, we'll make the change to `withMessageContainingAll` / `hasMessageContainingAll`.

@antoniomacri are you keen to contribute this improvement ? If so I suggest to wait for #1501 to be resolved to avoid merge conflicts.

joel-costigliola:
Actually for consistency with collection assertions `contains`, I suggest we change the existing `with/hasMessageContaining(String)` to  `with/hasMessageContainingFormattedMessage(String, Object...)` .

It will make it obvious what the difference is and with #1501, we are backward compatible so all good.

comments ?

antoniomacri:
Since we are checking exception messages, the consistency with collections is probably not a big problem.

`hasMessageContainingFormattedMessage` seems ugly to me (maybe better `hasMessageContainingFormattedString`) and is longer than `String.format`, therefore I'd probably keep using the last. 😄 ...no ok maybe not, just to avoid the extra parenthesis.

`containsFormattedMessage` (as suggested by Phil [here](https://github.com/joel-costigliola/assertj-core/pull/1503#issuecomment-491558079)) has a subtle difference in its name, since it refers not to the exception itself but actually to its `getMessage` result, that's why `hasMessage...`/`has...` methods are clearer to me.

joel-costigliola:
Good point for `hasMessageContainingFormattedMessage`, having `message` twice is not great.

[...]

PR #1503, opened 2019-05-10
Support var-arg throwable message assertions (#1501)

### LINKED ISSUE #1501: Use var-args with ThrowableAssertAlternative.withMessageContaining
#### Summary
I currently often find myself calling `withMessageContaining` multiple times on an exception assertions. It would be more convenient if I could roll these up into a single line, similar to the way that String assertions support `contains(CharSequence... values)`.

#### Example
The following current code:

```java
assertThatExceptionOfType(BeanCreationException.class).isThrownBy(lbf::preInstantiateSingletons)
	.withMessageContaining("Circular")
	.withMessageContaining("'tb3'")
	.withMessageContaining("'tb1'");
```

could become:

```java
assertThatExceptionOfType(BeanCreationException.class).isThrownBy(lbf::preInstantiateSingletons)
	.withMessageContaining("Circular", "'tb3'", "'tb1'");
```

philwebb:
Update AbstractThrowableAssert and ThrowableAssertAlternative so
that the hasMessageContaining and hasMessageNotContaining now accepts
a var-arg String array.

This unifies the API with the existing String assertions and allows
calls such as this:

	assertThatExceptionOfType(MyException.class).isThrownBy(service::call)
		.withMessageContaining("very")
		.withMessageContaining("bad")
		.withMessageContaining("stuff");

to be simplified to:

	assertThatExceptionOfType(MyException.class).isThrownBy(service::call)
		.withMessageContaining("very", "'bad'", "'stuff'");

#### Check List:
* Fixes #1501
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
@@ -195,6 +195,30 @@ public abstract class AbstractThrowableAssert<SELF extends AbstractThrowableAsse
   }
 
   /**
+   * Verifies that the message of the actual {@code Throwable} contains all the given values.
+   * <p>
+   * Examples:
+   * <pre><code class='java'> Throwable throwableWithMessage = new IllegalArgumentException("wrong amount 123");
+   * Throwable throwableWithoutMessage = new IllegalArgumentException();
+   *
+   * // assertion will pass:
+   * assertThat(throwableWithMessage).hasMessageContaining("amount", "123");
+   *
+   * // assertions will fail:
+   * assertThat(throwableWithoutMessage).hasMessageContaining("123");
+   * assertThat(throwableWithMessage).hasMessageContaining("234"); </code></pre>
+   *
+   * @param values the Strings expected to be contained in the actual {@code Throwable}'s message.
+   * @return this assertion object.
+   * @throws AssertionError if the actual {@code Throwable} is {@code null}.
+   * @throws AssertionError if the message of the actual {@code Throwable} does not contain the given description.
+   */
+  public SELF hasMessageContainingAll(CharSequence... values) {
+    throwables.assertHasMessageContainingAll(info, actual, values);
+    return myself;
+  }
+
+  /**
    * Verifies that the message of the actual {@code Throwable} does not contain the given content or is {@code null}.
    * <p>
    * Examples:
@@ -216,6 +240,31 @@ public abstract class AbstractThrowableAssert<SELF extends AbstractThrowableAsse
    */
   public SELF hasMessageNotContaining(String content) {
     throwables.assertHasMessageNotContaining(info, actual, content);
+    return myself;
+  }
+
+  /**
+   * Verifies that the message of the actual {@code Throwable} does not contain any of the given values or is {@code null}.
+   * <p>
+   * Examples:
+   * <pre><code class='java'> Throwable throwableWithMessage = new IllegalArgumentException("wrong amount 123");
+   * Throwable throwableWithoutMessage = new IllegalArgumentException();
+   *
+   * // assertions will pass:
+   * assertThat(throwableWithMessage).hasMessageNotContaining("234");
+   * assertThat(throwableWithoutMessage).hasMessageNotContaining("foo");
+   *
+   * // assertion will fail:
+   * assertThat(throwableWithMessage).hasMessageNotContaining("foo", "amount");</code></pre>
+   *
+   * @param values the contents expected to not be contained in the actual {@code Throwables}'s message.
+   * @return this assertion object.
+   * @throws AssertionError if the actual {@code Throwable} is {@code null}.
+   * @throws AssertionError if the message of the actual {@code Throwable} contains the given content.
+   * @since 3.12.0
+   */
+  public SELF hasMessageNotContainingAny(CharSequence... values) {
+    throwables.assertHasMessageNotContainingAny(info, actual, values);
     return myself;
   }
 
--- a/src/test/java/org/assertj/core/api/Assertions_assertThat_with_Throwable_Test.java
+++ b/src/test/java/org/assertj/core/api/Assertions_assertThat_with_Throwable_Test.java
@@ -83,9 +83,9 @@ public class Assertions_assertThat_with_Throwable_Test {
     // WHEN
     AssertionError assertionError = expectAssertionError(() -> catchThrowableOfType(codeThrowingException, IOException.class));
     // THEN
-    assertThat(assertionError).hasMessageContaining(IOException.class.getName())
-                              .hasMessageContaining(Exception.class.getName())
-                              .hasMessageContaining(getStackTrace(exception));
+    assertThat(assertionError).hasMessageContainingAll(IOException.class.getName(),
+                                                    Exception.class.getName(),
+                                                    getStackTrace(exception));
   }
 
   @Test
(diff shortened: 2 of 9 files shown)


src/test/java/org/assertj/core/internal/throwables/Throwables_assertHasMessageContaining_Test.java
```diff
@@ -46,11 +53,24 @@ public void should_fail_if_actual_is_null() {
...
   public void should_fail_if_actual_has_message_not_containing_with_expected_description() {
     AssertionInfo info = someInfo();
+    String content = "expected description part";
+    try {
+      throwables.assertHasMessageContaining(info, actual, content);
+      fail("AssertionError expected");
+    } catch (AssertionError err) {
+      verify(failures).failure(info, shouldContain(actual.getMessage(), content), actual.getMessage(), content);
+    }
+  }
```
joel-costigliola: let's rewrite these tests with java 8 love like this:
```java
import static org.assertj.core.util.AssertionsUtil.expectAssertionError;

    // GIVEN
    AssertionInfo info = someInfo();
    String content = "expected description part";
    // WHEN
    expectAssertionError(() -> throwables.assertHasMessageContaining(info, actual, content));
    // THEN
    verify(failures).failure(info, shouldContain(actual.getMessage(), content"));
```
I think this is much more readable and elegant than `try catch` (due to the tests being written for Java 7 at the time).

src/test/java/org/assertj/core/api/Assertions_assertThat_with_Throwable_Test.java
```diff
@@ -111,6 +111,15 @@ public void should_fail_with_good_message_when_assertion_is_failing() {
...
                                                    .withMessageContaining("<\"boom\">");
   }
+  
+  @Test
+  public void should_fail_with_good_message_when_vararg_has_message_containing_assertion_is_failing() {
+    assertThatExceptionOfType(AssertionError.class).isThrownBy(() -> assertThatThrownBy(raisingException("boom")).hasMessageContaining("bam", "boom"))
+                                                   .withMessageContaining("Expecting:")
+                                                   .withMessageContaining("<\"boom\">")
+                                                   .withMessageContaining("<[\"bam\", \"boom\"]>").withMessageContaining("but could not find:")
+                                                   .withMessageContaining("<[\"bam\"]>");
```
joel-costigliola: let's avoid repeating call to `withMessageContaining` ;-)
for readability put each expected string on a new line (expect for the first).

joel-costigliola: and for consistency, let's update the other tests in this file the same way.

src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
```diff
@@ -171,31 +171,31 @@ public SELF hasMessageStartingWith(String description) {
...
    *
-   * @param description the description expected to be contained in the actual {@code Throwable}'s message.
+   * @param values the Strings expected to be contained in the actual {@code Throwable}'s message.
    * @return this assertion object.
    * @throws AssertionError if the actual {@code Throwable} is {@code null}.
    * @throws AssertionError if the message of the actual {@code Throwable} does not contain the given description.
    */
-  public SELF hasMessageContaining(String description) {
-    throwables.assertHasMessageContaining(info, actual, description);
+  public SELF hasMessageContaining(String... values) {
```
joel-costigliola: I think we should check that `values` is not empty  which can happen if given no arguments, i.e. `hasMessageContaining()`, at the moment the assertion would pass but I'm not sure how to interpret it `hasMessageContaining()`, is it contains nothing ?  I feel it is more a user mistake than anything.

thoughts ?

philwebb: I think we should be consistent with whatever the String asserts currently do, I'll double check that behavior.

philwebb: String assertions have the following behavior:

```java
assertThat("foo").contains();
```

```
java.lang.IllegalArgumentException: The array of values to look for should not be empty
	at org.assertj.core.internal.CommonErrors.arrayOfValuesToLookForIsEmpty(CommonErrors.java:30)
	at org.assertj.core.internal.Strings.checkIsNotEmpty(Strings.java:489)
	at org.assertj.core.internal.Strings.doCommonCheckForCharSequence(Strings.java:1109)
	at org.assertj.core.internal.Strings.assertContains(Strings.java:457)
	at org.assertj.core.api.AbstractCharSequenceAssert.contains(AbstractCharSequenceAssert.java:656)
```

philwebb:
Thanks for the suggestions! I've forced pushed an update with the comments addressed.

joel-costigliola:
Sweet :)

One more thing and we are good to go, I was discussing #1502  and I think it would be better to rename these assertions to `withMessageContainingAll` / `hasMessageContainingAll` so that we can modify `withMessageContaining` / `hasMessageContaining` to support the `String.format` syntax.

I'm happy to do these changes if you don't have the time.

joel-costigliola:
There is javadoc failure on Travis for oraclejdk8:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.0.1:javadoc (default-cli) on project assertj-core: An error has occurred in Javadoc report generation: 
[ERROR] Exit code: 1 - /home/travis/build/joel-costigliola/assertj-core/src/main/java/org/assertj/core/api/ThrowableAssertAlternative.java:204: error: reference not found
[ERROR]    * @see AbstractThrowableAssert#hasMessageContaining(String...)
[ERROR]           ^
[ERROR] /home/travis/build/joel-costigliola/assertj-core/src/main/java/org/assertj/core/api/ThrowableAssertAlternative.java:206: warning - Tag @see: can't find hasMessageContaining(String...) in org.assertj.core.api.AbstractThrowableAssert
[ERROR] /home/travis/build/joel-costigliola/assertj-core/src/main/java/org/assertj/core/api/ThrowableAssertAlternative.java:233: error: reference not found
[ERROR]    * @see AbstractThrowableAssert#hasMessageNotContaining(String...)
[ERROR]           ^
[ERROR] /home/travis/build/joel-costigliola/assertj-core/src/main/java/org/assertj/core/api/ThrowableAssertAlternative.java:235: warning - Tag @see: can't find hasMessageNotContaining(String...) in org.assertj.core.api.AbstractThrowableAssert
```
I won't block the PR as I feel it is not relevant as the javadoc is successfully built on the other jdks, do you know why by any chance ?

philwebb:
Javadoc is because I changed the `String...` to `CharSequence...` to match the String assertions. I just forced pushed an update.

philwebb:
> I think it would be better to rename these assertions to withMessageContainingAll / hasMessageContainingAll so that we can modify withMessageContaining / hasMessageContaining to support the String.format syntax.

The one thing I don't like so much about that is that it's a little inconsistent with `AbstractCharSequenceAssert` which uses `contains` rather than `containsAll`. I wonder if we should instead use a different method name for the format based version:

```java
assertThat(someString).containsFormattedString("a %s", value);
assertThat(someException).containsFormattedMessage("a %s", value);
```

I'm happy to rework it whatever you decide.

joel-costigliola:
@philwebb after a bit of reflection, my preference goes to renaming the assertion to `with/hasMessageContainingAll(String...)` and `with/hasMessageContaining(String, Object...)`.

philwebb:
Sounds good to me. I've forced pushed another update.

joel-costigliola:
Integrated thanks @philwebb !

PR #1510, opened 2019-05-21
Fix Javadoc on Java 11+.

PascalSchumacher:
Let's see if this works.


PascalSchumacher:
Just to show that it works.

Is superseded by https://github.com/joel-costigliola/assertj-maven-parent-pom/pull/37

joel-costigliola:
So we should simply release a new parent pom and update this project pom.xml to use it, right ?

PascalSchumacher:
Yes, indeed.

PR #1514, opened 2019-05-30
Add sync test between Assertions and InstanceOfAssertFactories

scordio:
Sync test for `InstanceOfAssertFactories` (follows #1498).

Adds also missing factories (spotted by the test, yay!) for `StringBuilder` and `StringBuffer` and a `type(Class)` factory to obtain an `ObjectAssert` with updated type.

### DIFF
--- a/src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
+++ b/src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
@@ -288,8 +288,8 @@ public interface InstanceOfAssertFactories {
   /**
    * {@link InstanceOfAssertFactory} for a {@link Future}.
    *
-   * @param <RESULT>    the {@code Future} result type.
-   * @param resultType  the result type instance.
+   * @param <RESULT>   the {@code Future} result type.
+   * @param resultType the result type instance.
    * @return the factory instance.
    */
   @SuppressWarnings({ "rawtypes", "unused" }) // rawtypes: using Class instance, unused: parameter needed for type inference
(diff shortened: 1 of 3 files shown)


scordio:
@joel-costigliola I see Travis is using OpenJDK version 1.8.0_141 and Oracle JDK version 1.8.0_151.

I don't have the same versions locally and I'm not able to reproduce the issue with any of Oracle JDK 1.8.0_112, OpenJDK 1.8.0_181 and OpenJDK 1.8.0_212. Any suggestions?

joel-costigliola:
I have found the cause of the error after playing a bit with this test, I could only reproduce with maven on the command line, printing the values before calling `extractTypeParameters` showed the culprit:

```
field public static final org.assertj.core.api.InstanceOfAssertFactory org.assertj.core.api.InstanceOfAssertFactories.INT_STREAM
org.assertj.core.api.InstanceOfAssertFactory<java.util.stream.IntStream, org.assertj.core.api.ListAssert<java.lang.Integer>>

field public static final org.assertj.core.api.InstanceOfAssertFactory org.assertj.core.api.InstanceOfAssertFactories.PATH
org.assertj.core.api.InstanceOfAssertFactory<java.nio.file.Path, org.assertj.core.api.AbstractPathAssert<?>>

field public static final boolean[] org.assertj.core.api.InstanceOfAssertFactories.$jacocoData
class [Z
[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.74 s <<< FAILURE! - in org.assertj.core.api.Assertions_sync_with_InstanceOfAssertFactories_Test
[ERROR] each_standard_assertion_should_have_an_instance_of_assert_factory  Time elapsed: 0.548 s  <<< FAILURE!
```

`$jacocoData` is added by JaCoCo for code coverage, it is a synthetic fields and should be ignored (we already had a similar issue in https://github.com/joel-costigliola/assertj-core/issues/590 and https://github.com/jacoco/jacoco/issues/168 is also a good pointer).

 I fixed the problem by adding `.filter(not(Field::isSynthetic))` in:

```
  private Map<Type, Type> findFieldFactoryTypes() {
    return Stream.of(InstanceOfAssertFactories.class.getFields())
                 .filter(not(Field::isSynthetic))
```

Can you apply this fix and see how it goes ?

scordio:
Thanks @joel-costigliola, it looks good now!

PR #1515, opened 2019-06-02
RecursiveIterableComparisonAssert first implementation with isEqualTo and contains methods

### LINKED ISSUE #1238: Add usingRecursiveComparison for Iterables
#### Summary
Currently there is a feature to exclude fields using isEqualToIgnoringGivenFields while asserting a bean
http://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html 

We are exactly looking for similar feature while asserting list (or Iterable Objects)

#### Example
List<TolkienCharacter> frodos = Arrays.asList(
                                   new TolkienCharacter("Frodo1", 33, HOBBIT),
                                   new TolkienCharacter("Frodo2", 34, MAN));
List<TolkienCharacter> sams= Arrays.asList(
                                   new TolkienCharacter("Sam1", 38, HOBBIT);
                                   new TolkienCharacter("Sam2", 39, MAN);
// frodo and sam are equals when ignoring name and age as the only remaining field is race
assertThat(frodos).isEqualToIgnoringGivenFields(sams, "name", "age"); // OK both are HOBBIT

// ... but they are not equals if only age is ignored as their names differ.
assertThat(frodos).isEqualToIgnoringGivenFields(sams, "age"); // FAIL

haruntuncay:
#### Check List:
* Fixes #1238 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
@@ -193,7 +193,8 @@ public abstract class AbstractIterableAssert<SELF extends AbstractIterableAssert
    * <p>
    * The default recursive iterable comparison behavior has the same default configuration as {@link RecursiveComparisonAssert}, that is:
    * <ul>
-   * <li>elements of actual and expected iterables and their fields were compared field by field recursively even if they were not of the same type, this allows for example to compare a Person to a Doctor (call {@link RecursiveIterableComparisonAssert#withStrictTypeChecking() withStrictTypeChecking()} to change that behavior). </li>
+   * <li>elements of actual and expected iterables and their fields were compared field by field recursively even if they were not of the same type,
+   * this allows for example to compare a Person to a Doctor (call {@link RecursiveIterableComparisonAssert#withStrictTypeChecking() withStrictTypeChecking()} to change that behavior). </li>
    * <li>overridden equals methods were used in the comparison (unless stated otherwise)</li>
    * <li>these types were compared with the following comparators:
    *   <ul>
@@ -207,11 +208,15 @@ public abstract class AbstractIterableAssert<SELF extends AbstractIterableAssert
    * <p>
    * In addition to default behavior, {@link RecursiveIterableComparisonAssert RecursiveIterableComparisonAssert} has few additional points that can be configured:
    * <ul>
-   * <li>different types of iterables can be compared by default, this allows to compare for example an ArrayList<Person> and a LinkedHashSet<Person>. This behavior can be turned off by calling {@link RecursiveIterableComparisonAssert#withStrictTypeCheckingOnActualIterable() withStrictTypeCheckingOnActualIterable}.</li>
+   * <li>different types of iterables can be compared by default, this allows to compare for example an ArrayList<Person> and a LinkedHashSet<Person>.
+   * This behavior can be turned off by calling {@link RecursiveIterableComparisonAssert#withStrictTypeChecking()} () withStrictTypeChecking}.
+   * Note that calling {@link RecursiveIterableComparisonAssert#withStrictTypeChecking}, in addition to enforcing type compatibility of Iterables themselves,
+   * it also enforces type compatibility of the elements contained in those Iterables. For an example, see {@link RecursiveIterableComparisonAssert#isEqualTo(Iterable)}.</li>
    * <li>
    * although different type of iterables can be compared, order does matter. Comparing a List<Person> to a Set<Person> will almost always fail,
-   * but the order of comparison can be configured to be ignored by calling {@link RecursiveIterableComparisonAssert#ignoringActualIterableOrder() igroringActualIterableOrder}.
+   * but the order of comparison can be configured to be ignored by calling {@link RecursiveIterableComparisonAssert#ignoringCollectionOrder()} () ignoringCollectionOrder}.
    * This allows the comparison of ordered/unordered iterable pairs, such as comparing a a List<Person> to a Set<Person>.
+   * Note that calling {@link RecursiveIterableComparisonAssert#ignoringCollectionOrder} also causes the order of any fields/properties that are iterables/collections to be ignored.
    * </li>
    * </ul>
    * </p>
@@ -230,7 +235,7 @@ public abstract class AbstractIterableAssert<SELF extends AbstractIterableAssert
    * @return a new {@link RecursiveIterableComparisonAssert} instance
    */
   @Beta
-  public RecursiveIterableComparisonAssert<?> usingRecursiveComparison() {
+  public RecursiveIterableComparisonAssert<?, ACTUAL> usingRecursiveComparison() {
     return usingRecursiveComparison(new RecursiveComparisonConfiguration());
   }
 
@@ -241,7 +246,7 @@ public abstract class AbstractIterableAssert<SELF extends AbstractIterableAssert
    * @return a new {@link RecursiveIterableComparisonAssert} instance built with the given {@link RecursiveComparisonConfiguration}.
    */
   @Beta
-  public RecursiveIterableComparisonAssert<?> usingRecursiveComparison(RecursiveComparisonConfiguration recursiveComparisonConfiguration) {
+  public RecursiveIterableComparisonAssert<?, ACTUAL> usingRecursiveComparison(RecursiveComparisonConfiguration recursiveComparisonConfiguration) {
     SELF myselfCopy = usingElementComparator(new RecursiveComparisonDifferenceComparator(recursiveComparisonConfiguration));
     return new RecursiveIterableComparisonAssert<>(actual, recursiveComparisonConfiguration, myselfCopy)
       .withAssertionState(myselfCopy)
--- a/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
@@ -48,9 +48,12 @@ import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import org.assertj.core.annotations.Beta;
 import org.assertj.core.api.filter.FilterOperator;
 import org.assertj.core.api.filter.Filters;
 import org.assertj.core.api.iterable.ThrowingExtractor;
+import org.assertj.core.api.recursive.comparison.RecursiveComparisonConfiguration;
+import org.assertj.core.api.recursive.comparison.RecursiveComparisonDifferenceComparator;
 import org.assertj.core.condition.Not;
 import org.assertj.core.description.Description;
 import org.assertj.core.groups.FieldsOrPropertiesExtractor;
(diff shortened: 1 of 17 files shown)


src/main/java/org/assertj/core/api/RecursiveIterableComparisonAssert.java
```diff
@@ -0,0 +1,285 @@
...
+   *                       .isEqualTo(peopleSet);
+   *
+   * // assertion succeeds because both iterables contain exact same items and comparison order is ignored.
+   * assertThat(peopleList).usingRecursiveComparison()
+   *                       .ignoringActualIterableOrder()
+   *                       .isEqualTo(peopleArray);</code></pre>
+   *
+   * @return this {@link RecursiveIterableComparisonAssert} to chain other methods.
+   */
+  public SELF ignoringActualIterableOrder() {
```
joel-costigliola: I would rename it to `ignoringIterableOrder`. `ignoringElementsOrder` or `ignoringOrder` as we ignore as much actual order as we ignore expected order.

Also I expect this to also ignore order for any element fields that are iterable too (the implementation might already do that, I haven't checked) which should be documented in the javadoc.

src/main/java/org/assertj/core/api/AbstractIterableAssert.java
```diff
@@ -143,6 +146,108 @@ public SELF isNotEmpty() {
...
+   * Set<Doctor> doctorsSet = new HashSet<>(doctors);
+   *
+   * // assertion succeeds as both lists contains the same items in same orders.
+   * assertThat(doctors).usingRecursiveComparison().isEqualTo(people);
+   *
+   * // assertion fails because the order of items are different. (HashSet doesn't maintain insertion order)
+   * assertThat(doctors).usingRecursiveComparison().isEqualTo(doctorsSet);
+   *
+   * // assertion succeeds as doctors list does contain sheldon and raj.
+   * assertThat(doctors).usingRecursiveComparison().contains(sheldon, raj);</code></pre>
```
joel-costigliola: great examples!

src/main/java/org/assertj/core/api/RecursiveIterableComparisonAssert.java
```diff
@@ -0,0 +1,285 @@
...
+   *                       .isEqualTo(peopleSet);
+   *
+   * // assertion fails because iterables types are restricted to be compatible, which in this case they are not.
+   * assertThat(peopleList).usingRecursiveComparison()
+   *                       .withStrictTypeCheckingOnActualIterable()
+   *                       .isEqualTo(peopleSet);</code></pre>
+   *
+   * @return this {@link RecursiveIterableComparisonAssert} to chain other methods.
+   */
+  public SELF withStrictTypeCheckingOnActualIterable() {
```
joel-costigliola: For the same reason as the previous comment, I would rename it to `withStrictTypeCheckingOnIterable` or `withStrictIterableTypeChecking`.

src/main/java/org/assertj/core/api/RecursiveIterableComparisonAssert.java
```diff
@@ -0,0 +1,285 @@
...
+   *                       .isEqualTo(peopleArray);
+   *
+   * // assertion succeeds because array type is explicitly made allowed.
+   * assertThat(peopleList).usingRecursiveComparison()
+   *                       .allowingArrayTypeForExpected()
+   *                       .isEqualTo(peopleArray);</code></pre>
+   *
+   * @return this {@link RecursiveIterableComparisonAssert} to chain other methods.
+   */
+  public SELF allowingArrayTypeForExpected() {
```
joel-costigliola: I would rename it to `allowingComparingIterableWithArray`

src/main/java/org/assertj/core/api/RecursiveIterableComparisonAssert.java
```diff
@@ -0,0 +1,285 @@
...
+   *                       .isEqualTo(peopleLinkedSet);</code></pre>
+   *
+   * @param expected the object to compare {@code actual} to.
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual and the given objects are not deeply equal property/field by property/field.
+   * @throws IntrospectionError if one property/field to compare can not be found.
+   */
+  @SuppressWarnings("unchecked")
+  public SELF isEqualTo(Iterable<?> expected) {
+    if((actual == null && expected == null) || (actual == expected))
```
joel-costigliola: I believe `(actual == null && expected == null)` can be removed as we check `actual == null || expected == null` afterward

src/main/java/org/assertj/core/api/RecursiveIterableComparisonAssert.java
joel-costigliola: The RecursiveComparisonAssert API already has `ignoringCollectionOrder` methods, what use case `ignoringActualIterableOrder` covers that these methods don't.

src/main/java/org/assertj/core/api/RecursiveIterableComparisonAssert.java
joel-costigliola: The implementation should be like the start of `RecursiveComparisonAssert.isEqualTo`:
```
    // deals with both actual and expected being null
    if (actual == expected) return myself;
    if (expected == null) {
      // for the assertion to pass, actual must be null but this is not the case since actual != expected
      // => we fail expecting actual to be null
      objects.assertNull(info, actual);
    }
    // at this point expected is not null, which means actual must not be null for the assertion to pass
```
Extract these common check to a protected method to avoid duplication.

src/main/java/org/assertj/core/api/RecursiveIterableComparisonAssert.java
haruntuncay: From my understanding, `RecursiveComparisonAssert#ignoringCollectionOrder` works when the **field** of an element is an iterable, not the element itself, which is the case for `RecursiveIterableComparisonAssert`. For example:

```
Person p1 = new Person("harun", 1);
Person p2 = new Person("harun", 2);
Person p3 = new Person("harun", 3);
List<Person> list1 = Arrays.asList(p1, p2, p3);
List<Person> list2 = Arrays.asList(p2, p3, p1);
assertThat(list1).usingRecursiveComparison().ignoringCollectionOrder().isEqualTo(list2);
```

This assertion fails despite the usage of `ignoringCollectionOrder`, since `ignoringCollectionOrder `only applies to the **fields** under comparison and not the actual collection as a whole. Based on my understanding, I thought that I should provide a method that would actually ignore the order of actual/expected iterables.

src/main/java/org/assertj/core/api/RecursiveIterableComparisonAssert.java
```diff
@@ -0,0 +1,285 @@
...
+        return myself;
+
+    if(actual == null || expected == null)
+      throw new AssertionError("Iterables to compare can not be null.");
+
+    if(recursiveComparisonConfiguration.isInStrictTypeCheckingOnActualMode() &&  !actual.getClass().isAssignableFrom(expected.getClass()))
+      throw new AssertionError("types are not compatible.");
```
joel-costigliola: The error message must more informative and show the types, a bit like the one in `RecursiveComparisonAssert`:
```
"the fields are considered different since the comparison enforces strict type check and %s is not a subtype of %s"
```

src/main/java/org/assertj/core/api/RecursiveIterableComparisonAssert.java
```diff
@@ -0,0 +1,285 @@
...
+
+    if(sizeOf((Iterable<?>) actual) != sizeOf(expected))
+      throw new AssertionError("Sizes are not the same.Actual is of size "
```
joel-costigliola: reuse `RecursiveComparisonAssert.DIFFERENT_SIZE_ERROR`

src/main/java/org/assertj/core/api/RecursiveIterableComparisonAssert.java
```diff
@@ -0,0 +1,285 @@
...
+    if((actual == null && expected == null) || (actual == expected))
+        return myself;
+
+    if(actual == null || expected == null)
+      throw new AssertionError("Iterables to compare can not be null.");
+
+    if(recursiveComparisonConfiguration.isInStrictTypeCheckingOnActualMode() &&  !actual.getClass().isAssignableFrom(expected.getClass()))
+      throw new AssertionError("types are not compatible.");
+
+    if(sizeOf((Iterable<?>) actual) != sizeOf(expected))
```
joel-costigliola: We should try to make `actual` an Iterable, this is possible by adding a generic type to `RecursiveComparisonAssert` as in `AbstractAssert`.

Generics being tricky, if we find some incompatibilities I wouldn't mind keeping actual as an object.

src/main/java/org/assertj/core/api/RecursiveIterableComparisonAssert.java
```diff
@@ -0,0 +1,285 @@
...
+      throw new AssertionError("Sizes are not the same.Actual is of size "
+                                          + sizeOf((Iterable<?>) actual)
+                                          + ", Expected is of size " + sizeOf(expected));
+
+    boolean ignoreActualIterableOrder = recursiveComparisonConfiguration.getIgnoreActualIterableOrder();
+    // order can not be ignored
+    if(!ignoreActualIterableOrder)
+      iterableAssert.isEqualTo(expected);
```
joel-costigliola: To be clear, the reference order is the one specified by expected.

src/main/java/org/assertj/core/api/RecursiveIterableComparisonAssert.java
```diff
@@ -0,0 +1,285 @@
...
+   *                       .isEqualTo(peopleArray);</code></pre>
+   *
+   * @param values the objects that must be contained in {@code actual}.
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual and the given objects are not deeply equal property/field by property/field.
+   * @throws IntrospectionError if one property/field to compare can not be found.
+   */
+  public SELF isEqualTo(Object... values) {
+    if(!recursiveComparisonConfiguration.getAllowArrayTypeForExpected())
+      throw new AssertionError("Type array is not acceptable when comparing iterables, try calling #allowingArrayTypeForExpected configuration.");
```
joel-costigliola: Reword suggestion:
`By default array can't be compared to Iterable, call allowingArrayTypeForExpected to allow it.`

src/main/java/org/assertj/core/api/RecursiveIterableComparisonAssert.java
```diff
@@ -0,0 +1,285 @@
...
+    return isEqualTo(Arrays.asList(values));
```
joel-costigliola: use AssertJ the more elegant `Lists.list` and static import it for optimal readability.

src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration.java
```diff
@@ -57,6 +57,12 @@
   private Set<FieldLocation> ignoredCollectionOrderInFields = new LinkedHashSet<>();
   private List<Pattern> ignoredCollectionOrderInFieldsMatchingRegexes = new ArrayList<>();
 
+  // RecursiveIterableComparisonAssert configuration section
+  private boolean ignoreActualIterableOrder = false; // ignore the order of iterables when comparing their elements
+  private boolean strictTypeCheckingOnActualIterable = false; // don't allow different type of iterables to be compared (i.e. List vs Set)
+  private boolean allowArrayTypeForExpected = false;
```
joel-costigliola: We need to describe this properties in `multiLineDescription` but that will impact `RecursiveComparisonAssert` assertion. 

Thinking out loud, I believe we only need `RecursiveComparisonConfiguration` for both `RecursiveComparisonAssert` and `RecursiveIterableComparisonAssert` as `RecursiveComparisonAssert` must be able to compare iterable fields.

joel-costigliola: It means the field must not named only for `RecursiveIterableComparisonAssert`, hopefully me renaming suggestions in the method exposing these fields applies to both recursive assert classes.

src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceComparator.java
```diff
@@ -0,0 +1,27 @@
+package org.assertj.core.api.recursive.comparison;
+
+import org.assertj.core.internal.FieldByFieldComparator;
+
+public class RecursiveComparisonDifferenceComparator extends FieldByFieldComparator {
```
joel-costigliola: add javadoc

src/test/java/org/assertj/core/api/recursive/comparison/iterables/RecursiveIterableComparisonAssert_contains_Test.java
```diff
@@ -0,0 +1,60 @@
...
+import org.assertj.core.internal.objects.data.PersonDto;
+import org.junit.jupiter.api.Test;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.util.AssertionsUtil.expectAssertionError;
+
+public class RecursiveIterableComparisonAssert_contains_Test {
```
joel-costigliola: I love tests following the GIVEN WHEN THEN pattern, let's do that for all tests ;-)

joel-costigliola:
@haruntuncay I have finished the first round of review, I have looked at the tests into too much details though, I'll do that after the other comments are addressed.

src/main/java/org/assertj/core/api/RecursiveIterableComparisonAssert.java
haruntuncay: It seems adding a new generic type to `RecursiveComparisonAssert` will require minor modifications in a few other files. Just to confirm, should I continue with adding the new generic type ?

src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration.java
haruntuncay: How about I remove `ignoreActualIterableOrder` and `strictTypeCheckingOnActualIterable` and use `ignoreCollectionOrder` and `strictTypeChecking` respectively for both purposes as you suggested earlier ?

src/main/java/org/assertj/core/api/RecursiveIterableComparisonAssert.java
joel-costigliola: The best way forward is to make `ignoringCollectionOrder` effective on actual iterables, the rationale being that what applies for fields should also apply for top level objects.

src/main/java/org/assertj/core/api/RecursiveIterableComparisonAssert.java
joel-costigliola: Yes!

src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration.java
joel-costigliola: That the idea, as I said in a previous comment what applies for field should also work for top level objects.

haruntuncay:
Hello @joel-costigliola,
I have gone through the issues and applied the changes you have requested. I only wasn't able to fix this one:
> reuse `RecursiveComparisonAssert.DIFFERENT_SIZE_ERROR`.
> 
> In addition, we should display the two iterables to help users figure out the difference. (I was planning to do that in `RecursiveComparisonAssert`).

Because `DIFFERENT_SIZE_ERROR` was private in `RecursiveDifferenceComparisonCalculator` and I wasn't sure whether it was okay to make it public for one use case, so for the time being, I just used the `AssertionError` with the message similar to `DIFFERENT_SIZE_ERROR`.

By the way, thank you for the time you've spent reviewing and thanks for the opportunity given to help this project. I am really learning a lot about api design just by studying this project 😄 .

haruntuncay:
By the way, I realized that, because we use `IterableAssert` with `new IterableElementComparisonStrategy(new RecursiveComparisonDifferenceComparator(new RecursiveComparisonConfiguration()))`, and since IterableAssert has it's own error report mechanism, differences found by `RecursiveComparisonDifferenceCalculator `gets swallowed.

```
  // Since  RecursiveIterableComparisonAssert#isEqualTo calls IterableAssert#isEqualTo
  // and this is what IterableAssert#isEqualTo boils down to:

  public void assertEqual(AssertionInfo info, Object actual, Object expected) {
    if (!areEqual(actual, expected))
      throw failures.failure(info, shouldBeEqual(actual, expected, comparisonStrategy, info.representation()));
  }

  // differences found by `RecursiveComparisonDifferenceCalculator` are completely ignored.
```
What can we do about it so that actual error report could point out the differences found by `RecursiveComparisonDifferenceCalculator` ? Delegating to IterableAssert seems unfulfilling here.

haruntuncay:
Hello @joel-costigliola 
Hope you are doing well. When you have the time, could you please check out the updates and let me know what you think ? Thanks in advance 😊

joel-costigliola:
Let's not worry too much about DIFFERENT_SIZE_ERROR, I think it would actually be better to use `"actual and expected are iterables of different sizes` (I used the word `values` in the DIFFERENT_SIZE_ERROR to express that they were field values not top level objects as in our case - I might change that to be clearer).

For `isEqualTo` we need to think of a way to report the differences found.
When the collection order is enforced, since we compare elements at the same index, we would ideally report the differences found per index, concretely if elements at index 2 and 5 differ, we would report something like:

```
elements at index 2 differ:
- actual element was: ...
- expected element was: ...
Differences found: ... 

elements at index 5 differ:
- actual element was: ...
- expected element was: ...
Differences found: ... 
```
You get the idea.

When the comparison is not ordered, it is way more complicated to show the differences since we have to compare each actual elements with all expected ones! 
For example if actual element 2 was not found in expected, should we report the differences between it and each expected elements:

```
elements at index 2 was not found in expected:
- actual element was: ...

Differences found with expected element 1 : ... 
Differences found with expected element 2 : ... 
Differences found with expected element 3 : ... 
...
Differences found with expected element n : ... 
```
It's going to be super long and might not be so useful2

joel-costigliola:
I'll continue the review tomorrow (if I have the time), in the meantime let's think about the best way to report the differences.

src/main/java/org/assertj/core/api/AbstractIterableAssert.java
```diff
@@ -143,6 +146,113 @@ public SELF isNotEmpty() {
...
+   * @param recursiveComparisonConfiguration the {@link RecursiveComparisonConfiguration} used in the chained {@link RecursiveIterableComparisonAssert} assertions.
+   *
+   * @return a new {@link RecursiveIterableComparisonAssert} instance built with the given {@link RecursiveComparisonConfiguration}.
+   */
+  @Beta
+  public RecursiveIterableComparisonAssert<?, ACTUAL> usingRecursiveComparison(RecursiveComparisonConfiguration recursiveComparisonConfiguration) {
+    SELF myselfCopy = usingElementComparator(new RecursiveComparisonDifferenceComparator(recursiveComparisonConfiguration));
+    return new RecursiveIterableComparisonAssert<>(actual, recursiveComparisonConfiguration, myselfCopy)
+      .withAssertionState(myselfCopy)
+      .withTypeComparators(comparatorsByType);
```
joel-costigliola: remove the call to `withTypeComparators` as the `RecursiveComparisonConfiguration` provides the type comparators (and much more).

src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceComparator.java
```diff
@@ -0,0 +1,31 @@
+package org.assertj.core.api.recursive.comparison;
+
+import org.assertj.core.internal.FieldByFieldComparator;
+
+/**
+ * Compares two objects field/property by field/property recursively using the new {@link RecursiveComparisonDifferenceCalculator} api,
+ * configured by the given {@link RecursiveComparisonConfiguration}.
+ */
+public class RecursiveComparisonDifferenceComparator extends FieldByFieldComparator {
```
joel-costigliola: Add a `toString` method to show the `RecursiveComparisonConfiguration` used otherwise we get the `FieldByFieldComparator` one which is misleading:
```
when comparing elements using field/property by field/property comparator on all fields/properties
Comparators used:
- for elements fields (by type): {Double -> DoubleComparator[precision=1.0E-15], Float -> FloatComparator[precision=1.0E-6]}
```
^^ this is not correct as the `RecursiveComparisonConfiguration` comparators were used.

haruntuncay:
Hello @joel-costigliola,
Hope you are well. I have had the time to implement a proof of concept solution for error reporting and wanted to ask for your opinion. In short, I found that using already existing `Iterables` class for custom error report is not very easy since it handles iteration implicitly, where I think that we would benefit more handling iteration explicitly, instead.

haruntuncay:
Hello @joel-costigliola,
Hope you are well. I have had the time to implement a proof of concept solution for error reporting and wanted to ask for your opinion. In short, I found that using already existing Iterables class for custom error report is not very easy since it handles iteration implicitly, where I think that we would benefit more handling iteration explicitly, instead.

haruntuncay:
Hello @joel-costigliola,

First of all, I apologize that I have accidentally hit "close and comment" button instead of comment.

Second, I hope you are well. I have had the time to implement a proof of concept solution for error reporting and wanted to ask for your opinion.

In short, I found that using already existing Iterables class for custom error report is not very easy since it handles iteration implicitly, where I think that we would benefit more handling iteration explicitly, instead. I have opened another branch, seperate from this one, so that my demo commits wouldn't go to PR. When you have the time, could you please check it out from [here](https://github.com/haruntuncay/assertj-core/commit/f1f867bab4a82e60a4db19d7dd38ea0d41cb3541) ?

To give a clearer idea of how it currently works:

```
    Person person1 = new Person("harun", 1, new Field("field1", "val1"));
    Person person2 = new Person("harun2", 2, new Field("field2", "val2"));

    List<Person> list1 = Arrays.asList(person1, person2, person1);
    List<Person> list2 = Arrays.asList(person2, person1, person1);

    assertThat(list1)
      .usingRecursiveComparison()
      .isEqualTo(list2);
  }
```

This code currently yields this error message:

```
Exception in thread "main" java.lang.AssertionError: 
Expecting iterable:
  <[[name=harun, age=1], [name=harun2, age=2], [name=harun, age=1]]>
to be equal to:
  <[[name=harun2, age=2], [name=harun, age=1], [name=harun, age=1]]>
when recursively comparing iterables field by field, but found the following 2 differences:

elements at index 0 differ:
- actual was: [name=harun, age=1]
- expected was: [name=harun2, age=2]
Differences found:
field/property 'age' differ:
- actual value   : 1
- expected value : 2
field/property 'field.name' differ:
- actual value   : "field1"
- expected value : "field2"
field/property 'field.val' differ:
- actual value   : "val1"
- expected value : "val2"
field/property 'name' differ:
- actual value   : "harun"
- expected value : "harun2"


elements at index 1 differ:
- actual was: [name=harun2, age=2]
- expected was: [name=harun, age=1]
Differences found:
field/property 'age' differ:
- actual value   : 2
- expected value : 1
field/property 'field.name' differ:
- actual value   : "field2"
- expected value : "field1"
field/property 'field.val' differ:
- actual value   : "val2"
- expected value : "val1"
field/property 'name' differ:
- actual value   : "harun2"
- expected value : "harun"


The recursive comparison was performed with this configuration:
- overridden equals methods were used in the comparison
- these types were compared with the following comparators:
  - java.lang.Double -> DoubleComparator[precision=1.0E-15]
  - java.lang.Float -> FloatComparator[precision=1.0E-6]
- actual and expected objects and their fields were compared field by field recursively even if they were not of the same type, this allows for example to compare a Person to a PersonDto (call strictTypeChecking(true) to change that behavior).
- comparing iterables with arrays was not allowed.

	at app.Main.main(Main.java:22)

```

joel-costigliola:
I ended with adding `usingRecursiveFieldByFieldElementComparator(configuration)` to iterable assertions - https://assertj.github.io/doc/#assertj-core-recursive-comparison-for-iterable

PR #1513, opened 2019-05-30
Add containsExactlyInAnyOrderEntriesOf() to AbstractMapAssert

scordio:
Adds missing `containsOnlyEntriesOf()` to the `containsEntriesOf` family.

#### Check List:
* Fixes N/A
* Unit tests : YES
* Javadoc with a code example (API only) : YES

PR #1521, opened 2019-06-08
Convert InstanceOfAssertFactories no param methods to static fields

scordio:
Follows https://github.com/joel-costigliola/assertj-core/pull/1498#issuecomment-500118663.

In case you have time to take a look at #1513, it could help to clean a bit the `THEN` of  `Assertions_sync_with_InstanceOfAssertFactories_Test#each_standard_assertion_should_have_an_instance_of_assert_factory()` :-)

#### Check List:
* Fixes N/A
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
+++ b/src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
@@ -68,13 +68,11 @@ public interface InstanceOfAssertFactories {
 
   /**
    * {@link InstanceOfAssertFactory} for a {@link Predicate}, assuming {@code Object} as input type.
-   *
-   * @return the factory instance.
+   * 
+   * @see #predicate(Class)
    */
   @SuppressWarnings("rawtypes") // rawtypes: using Class instance
-  static InstanceOfAssertFactory<Predicate, PredicateAssert<Object>> predicate() {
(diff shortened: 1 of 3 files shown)


joel-costigliola:
Integrated thanks @scordio!

scordio:
@joel-costigliola I noticed you refactored the `THEN` of `each_standard_assertion_should_have_an_instance_of_assert_factory()` with:
```java
then(factories).hasSameSizeAs(assertThatMethods)
               .containsAllEntriesOf(assertThatMethods);
```
This way is harder to understand what is missing, e.g. you get:
```
java.lang.AssertionError: 
Actual and expected should have same size but actual size is:
 <68>
while expected is:
 <70>
Actual was:
<...>
```
instead of:
```
java.lang.AssertionError: 
Expecting:
<...>
to contain only:
<...>
  <[MapEntry[key=java.lang.Comparable, value=org.assertj.core.api.AbstractComparableAssert],
    MapEntry[key=java.lang.Object, value=org.assertj.core.api.ObjectAssert]]>
```
I'd probably put `containsAllEntriesOf()` before `hasSameSizeAs()`.

joel-costigliola:
Good point, I'll change that. 

Another improvement would be to display the collections in the size error message.

PR #1523, opened 2019-06-10
Split InstanceOfAssertFactories sync test into fields and methods tests

scordio:
Follows https://github.com/joel-costigliola/assertj-core/pull/1498#issuecomment-500384530.

#### Check List:
* Fixes NA
* Unit tests : NA


joel-costigliola:
Integrated thanks @scordio!

Issue #1504, opened 2019-05-14
Java module descriptor for `org.assertj.core`


sormuras:
### Old caption

`[ByteBuddy] NoClassDefFoundError: java.lang.instrument.Instrumentation`

#### Summary

Running tests included in JUnit 5's [documentation](https://github.com/junit-team/junit5/tree/master/documentation/src/test/java) project (backing the User Guide) on the `class-path` works fine. Converting those tests and demos into a Java module and running it on the `module-path` fails with several errors. One of the error is related to AssertJ/ByteBuddy.

Found this issue while working on https://github.com/junit-team/junit5/issues/1877

/cc @raphw 

```
  JUnit Jupiter:EngineTestKitAllEventsDemo:verifyAllJupiterEvents()
    MethodSource [className = 'example.testkit.EngineTestKitAllEventsDemo', methodName = 'verifyAllJupiterEvents', methodParameterTypes = '']
    => java.lang.IllegalArgumentException: Could not create type
       org.assertj.core@3.12.2/org.assertj.core.internal.bytebuddy.TypeCache.findOrInsert(TypeCache.java:154)
       org.assertj.core@3.12.2/org.assertj.core.internal.bytebuddy.TypeCache$WithInlineExpunction.findOrInsert(TypeCache.java:365)
       org.assertj.core@3.12.2/org.assertj.core.api.SoftProxies.createSoftAssertionProxyClass(SoftProxies.java:118)
       org.assertj.core@3.12.2/org.assertj.core.api.SoftProxies.createSoftAssertionProxy(SoftProxies.java:104)
       org.assertj.core@3.12.2/org.assertj.core.api.AbstractSoftAssertions.proxy(AbstractSoftAssertions.java:31)
       [...]
     Caused by: java.lang.NoClassDefFoundError: java/lang/instrument/Instrumentation
       org.assertj.core@3.12.2/org.assertj.core.internal.bytebuddy.utility.JavaModule$Dispatcher$CreationAction.run(JavaModule.java:270)
       org.assertj.core@3.12.2/org.assertj.core.internal.bytebuddy.utility.JavaModule$Dispatcher$CreationAction.run(JavaModule.java:251)
       java.base/java.security.AccessController.doPrivileged(Native Method)
       org.assertj.core@3.12.2/org.assertj.core.internal.bytebuddy.utility.JavaModule.<clinit>(JavaModule.java:45)
[...]

PR #1525, opened 2019-06-12
Introduce explicit module descriptor

sormuras:
Replaces the `<Automatic-Module-Name>` entry from the manifest.

#1504 draft/scaffold/proof-of-concept PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,6 @@ language: java
 install: true
 
 addons:
-  apt:
-    packages:
-      - oracle-java8-installer
   sonarcloud:
     organization: "default"
     token:
@@ -18,8 +15,6 @@ cache:
     - '$HOME/.sonar/cache'
 
 jdk:
-  - openjdk8
-  - oraclejdk8
   - openjdk11
   - openjdk12
   - openjdk-ea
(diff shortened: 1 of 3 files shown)


PascalSchumacher:
I am not a fan of dropping the java 8 build on travis.

If we drop the build we should at least add something like `animal sniffer` to ensure that we use only the Java 8 API.

sormuras:
`javac --release 8` takes care of this.

sormuras:
Copied from https://docs.oracle.com/en/java/javase/11/tools/javac.html

`--release release`
_Compiles against the public, supported and documented API for a specific VM version. Supported release targets are 6, 7, 8, 9, 10, and 11._

The JUnit team uses this option/feature since version 5.0 (2017) without any issues.

PascalSchumacher:
Thanks. I missed that. Sorry.

sormuras:
No problem, you're welcome.

For completeness, here are the lines configuring the `maven-compiler-plugin`:

https://github.com/joel-costigliola/assertj-core/blob/4a93d9b7bc76d452833a6688b3d58d6ad4ccb9d5/pom.xml#L133-L137

PR #1527, opened 2019-06-13
Retain assertion state in asInstanceOf()

### LINKED ISSUE #1517: asInstanceOf() should retain description
Currently `asInstanceOf()` is not keeping the existing assertion description.

To achieve it, the return statement of:

https://github.com/joel-costigliola/assertj-core/blob/aee33e05bdba3e39376abd9083941c5fc632a905/src/main/java/org/assertj/core/api/AbstractAssert.java#L371-L373

should be changed to:

```java
return instanceOfAssertFactory.createAssert(actual).as(info.description());
```

but this cannot work with the current method declaration because of the type parameter `ASSERT` extending `Assert<?, ?>` instead of `Assert<ASSERT, ?>`.

To fix it, some changes would be needed which cause the replacement of all the `assertThat()` abstract return types with the corresponding concrete ones.

My main question is: why are some `assertThat()` returning an abstract class instead of the concrete one, while others (e.g., `assertThat(Class<?>`) are already returning concrete types? 
Is there a design reason behind it which I'm not understanding?

scordio:
#### Check List:
* Fixes #1517
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -16,6 +16,7 @@ import static java.util.Arrays.stream;
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.toList;
 import static org.assertj.core.error.ShouldMatch.shouldMatch;
+import static org.assertj.core.error.ShouldNotBeNull.shouldNotBeNull;
 import static org.assertj.core.util.Preconditions.checkArgument;
 import static org.assertj.core.util.Strings.formatIfArgs;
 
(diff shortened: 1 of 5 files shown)


scordio:
Hi @joel-costigliola, did you have a chance to take a look at this PR? I'd like to propose another PR related to `asInstanceOf()` but I prefer to consolidate this one first.

joel-costigliola:
@scordio I was about to do it but I got stuck on the java modules issue, will have a look this we!

scordio:
@joel-costigliola thanks, take your time!

joel-costigliola:
Good work @scordio  (as usual !), this is integrated.

PR #1531, opened 2019-06-19
Travis CI: Add Java 13 build and restore Java EA build

PascalSchumacher:
I do not know if we should keep https://github.com/joel-costigliola/assertj-core/pull/1531/commits/eea3712e4aa0f0237a0c3dab1d1757d6707feec2

No idea if this is bug in the first Java 14 EA build or a deliberate change in Java 14 (that may require changes in mockito).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ cache:
 jdk:
   - openjdk11
   - openjdk12
+  - openjdk13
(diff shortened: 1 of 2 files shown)


joel-costigliola:
I'm integrating the PR and see how it goes with Java 14

PR #1532, opened 2019-06-19
Add missing `uses` directives to module descriptor

sormuras:
See #1504


joel-costigliola:
Integrated thanks @sormuras !

PR #1534, opened 2019-06-20
Allow return values of WithAssertions#fail methods to be ignored

### LINKED ISSUE #1533: Allow return values of WithAssertions#fail* methods to be ignored
#### Summary
`WithAssertions#fail` methods don't return void (from #1294). This causes violations in static analysis tools because the class `WithAssertions` is annotated with `@CheckReturnValue`. 

`WithAssertions#fail` methods  should be annotated with `@CanIgnoreReturnValue`.
The similar fix has been made for `Assertions#fail` in #1298.

jiri-pejchal:
#### Check List:
* Fixes #1533
* Unit tests : NA


joel-costigliola:
Good catch @jiri-pejchal and thanks for the PR!

PR #1537, opened 2019-06-26
Issues/1418 jupiter soft assertions

### LINKED ISSUE #1418: SoftAssertions for JUnit Jupiter should reset tracked exceptions between tests
## Overview

The following code looks up tracked exceptions in an instance field.

https://github.com/joel-costigliola/assertj-core/blob/9692f7dc339765e9d7ef2725c89b3241c613268f/src/main/java/org/assertj/core/api/JUnitJupiterSoftAssertions.java#L45

The problem with doing that is that the tracked exceptions do not get reset between test method invocations if the test class is configured with `@TestInstance(PER_CLASS)` lifecycle semantics, since the test instance is cached along with the instance of `JUnitJupiterSoftAssertions` stored in an instance field.

I became aware of this due to the following comment in JUnit's issue tracker: https://github.com/junit-team/junit5/issues/1500#issuecomment-459720295

## Affected Extensions

- `JUnitJupiterSoftAssertions`
- `JUnitJupiterBDDSoftAssertions`
- any other extensions implemented in a similar manner

## Proposals

Modify `JUnitJupiterSoftAssertions` so that it implements JUnit Jupiter's `BeforeEachCallback` extension API and clears the collected errors in `SoftProxies` before each test method.

Another option would be to store the collected errors in Jupiter's `ExtensionContext.Store` for the currently executing scope.

## Relate Issues

- #1229

joel-costigliola:
Fixes #1418

### DIFF
--- a/src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
+++ b/src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
@@ -1,0 +1,139 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2019 the original author or authors.
+ */
+
+package org.assertj.core.api.junit.jupiter;
+
+import static org.junit.platform.commons.support.AnnotationSupport.isAnnotated;
+
+import java.lang.reflect.Executable;
+import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
+import java.util.Optional;
+import java.util.function.Consumer;
+
+import org.assertj.core.api.BDDSoftAssertions;
+import org.assertj.core.api.SoftAssertions;
+import org.junit.jupiter.api.extension.AfterTestExecutionCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
+import org.junit.jupiter.api.extension.ExtensionContext.Store;
+import org.junit.jupiter.api.extension.ParameterContext;
+import org.junit.jupiter.api.extension.ParameterResolutionException;
+import org.junit.jupiter.api.extension.ParameterResolver;
+import org.junit.platform.commons.annotation.Testable;
+
+/**
+ * Extension for JUnit Jupiter that provides support for injecting an instance
+ * of {@link SoftAssertions} or {@link BDDSoftAssertions} into test methods.
+ *
+ * <h3>Applicability</h3>
+ *
+ * <p>In this context, the term "test method" refers to any method annotated with
+ * {@code @Test}, {@code @RepeatedTest}, {@code @ParameterizedTest},
+ * {@code @TestFactory}, or {@code @TestTemplate}. This extension does not inject
+ * {@code SoftAssertions} or {@code BDDSoftAssertions} arguments into test
+ * constructors or lifecycle methods.
+ *
+ * <h3>Scope</h3>
+ *
+ * <p>The scope of the {@code SoftAssertions} or {@code BDDSoftAssertions} instance
+ * managed by this extension begins when a parameter of type {@code SoftAssertions}
+ * or {@code BDDSoftAssertions} is resolved for a test method. The scope of the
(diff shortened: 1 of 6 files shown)


src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
```diff
@@ -0,0 +1,119 @@
...
+		boolean isBddSoftAssertionsParameter = parameterContext.getParameter().getType() == BDDSoftAssertions.class;
+		return (isTestLevel && (isSoftAssertionsParameter || isBddSoftAssertionsParameter));
+	}
+
+	@Override
+	public Object resolveParameter(ParameterContext parameterContext, ExtensionContext extensionContext) {
+		if (parameterContext.getParameter().getType() == SoftAssertions.class) {
+			return getStore(extensionContext).getOrComputeIfAbsent(SoftAssertions.class);
+		}
+		return getStore(extensionContext).getOrComputeIfAbsent(BDDSoftAssertions.class);
```
joel-costigliola: Could we do this ?
```java
return getStore(extensionContext).getOrComputeIfAbsent(parameterContext.getParameter().getType());
```

src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	@Override
+	public void afterEach(ExtensionContext extensionContext) throws Exception {
+		SoftAssertions softAssertions = getStore(extensionContext).remove(SoftAssertions.class, SoftAssertions.class);
+		if (softAssertions != null) {
```
joel-costigliola: More a matter of style but I'd rather use an `Optional` instead of `null` check, something like:
```
Optional.of(getStore(extensionContext).remove(SoftAssertions.class, SoftAssertions.class))
        .ifPresent(softAssertions::assertAll)
```
I think it should work but I haven't tried it

src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
```diff
@@ -0,0 +1,119 @@
...
+
+		BDDSoftAssertions bddSoftAssertions = getStore(extensionContext).remove(BDDSoftAssertions.class,
+			BDDSoftAssertions.class);
+		if (bddSoftAssertions != null) {
+			bddSoftAssertions.assertAll();
+		}
+	}
+
+	private Store getStore(ExtensionContext extensionContext) {
+		return extensionContext.getStore(Namespace.create(getClass(), extensionContext.getUniqueId()));
```
joel-costigliola: Would `extensionContext.getStore(Namespace.create(getClass()));` be enough ? 

`getStore` being an `ExtensionContext` method it has access to the `getUniqueId()`, why do we need to pass it ?

src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
sbrannen: Yes, I suppose we could do that. It looks a bit like voodoo, but with a comment I think it would fine. 😉

src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
sbrannen: If you slightly modify it as follows, it works.

```java
Optional.ofNullable(getStore(extensionContext).remove(SoftAssertions.class, SoftAssertions.class))
	.ifPresent(SoftAssertions::assertAll);
```

As a general rule of thumb, I would not create an `Optional` just to replace an `if` statement without using any other features of `Optional` such as mapping, filtering, etc.

But it's your code base. So feel free to change whatever I've done once/if you merge the work. 😇 

Having said that, I _may_ introduce a helper method that does that behind the scenes in order to create a mini DSL for those two blocks of code.

src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
sbrannen: I already changed that in my latest commit. 😇

src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
sbrannen: Done in latest commit: 21f56775f63cfbbe4dbaccf2d06d4cca52e3537d

src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
sbrannen: Done in latest commit: 21f5677

sbrannen:
@joel-costigliola, it appears that GitHub will not let me submit a PR from my branch since you already opened this one.

Would you mind deleting this PR so that I can submit the PR myself?

joel-costigliola:
Just closed it @sbrannen

PR #1547, opened 2019-07-04
Add asInstanceOf(Class) to AbstractObjectAssert

scordio:
I imagine that the [`type()`](https://github.com/joel-costigliola/assertj-core/blob/94307cabbfce604becf260b01dadc6e628b386f6/src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java#L342-L354) factory can be used mostly to obtain a new `ObjectAssert` for types which do not have a dedicated `Assert` class. For example, the result of an `extracting("myField")` provides an `ObjectAssert<Object>`, making it difficult to perform other operations which need to have the real type (e.g., a chained `extracting(Function)` with a class getter).

This PR is meant to be a shortcut for these use cases where, starting from an `ObjectAssert`, `asInstanceOf(type(MyClass.class))` would be used. E.g., (using String just for demonstration):
```java
Object value = "string";

// Bad
assertThat(value) // ObjectAssert<Object>
    .extracting(String::length); // not compiling

// Good 
assertThat(value) // ObjectAssert<Object>
     .asInstanceOf(type(String.class)) // ObjectAssert<String>
     .extracting(String::length); // compiling

// Better 
assertThat(value) // ObjectAssert<Object>
     .asInstanceOf(String.class) // ObjectAssert<String>
     .extracting(String::length); // compiling
```

I decided to put the new method under `AbstractObjectAssert` instead of `Assert`/`AbstractAssert` because I think there is no realistic use case which starts from assertions not inheriting from `AbstractObjectAssert`. E.g.:
```java
List list = Arrays.asList(1,2,3);
assertThat(list) // ListAssert<Integer>
     .asInstanceOf(List.class); // ObjectAssert<List> - What?
```

#### Questions
* Opinions about the overall idea?
* As this method delegates to `asInstanceOf(InstanceOfAssertFactories.type(type))`, is any change needed for `SoftAssertionsTest` and `BDDSoftAssertionsTest`?
* Similar to the previous point, are additional `Assumptions_assumeThat_with_instanceOf_*` tests needed?

#### Check List:
* Fixes NA
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
@@ -15,6 +15,7 @@ package org.assertj.core.api;
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.toList;
 import static org.assertj.core.description.Description.mostRelevantDescription;
+import static org.assertj.core.error.ShouldNotBeNull.shouldNotBeNull;
 import static org.assertj.core.extractor.Extractors.byName;
 import static org.assertj.core.extractor.Extractors.extractedDescriptionOf;
 import static org.assertj.core.internal.TypeComparators.defaultTypeComparators;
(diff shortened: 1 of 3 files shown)


joel-costigliola:
@scordio that's a good idea :+1:, I agree there is no need for additional soft assertions / assumption tests.

I'll integrate this for 3.13.0 which is going to be released soon, I just have created a first draft of the release notes here https://github.com/assertj/doc/pull/13, it would be great if you could have a look at the section about `asIstanceOf`.

joel-costigliola:
I gave it more thoughts and now I'm not sure we should add this one, nothing wrong with the idea but my main concern is users can be confused when choosing between 
```java
assertThat(value).asInstanceOf(STRING)... // InstanceOfAssertFactories.STRING
```
and
```java
assertThat(value).asInstanceOf(String.class)...
```
I believe most users will expect `asInstanceOf(String.class)` to give them `String` assertions (that the API I would have liked for narrowing type assertions if it was technically possible) so to avoid confusion I prefer not to add this capability.

WDYT ?

scordio:
You're definitely right about the possible confusion and the bold text I put in the javadoc [should have been a red flag for me](https://twitter.com/francesc/status/508066994491387904) :-)

What about using a complete different name for the API?

In this case, a proper shortcut implementation could be provided in all the assert classes which have a corresponding parameterized factory. E.g.:
* `AbstractObjectAssert#withValueType(String.class)` --> `asInstanceOf(type(String.class))`
* `AbstractObjectArrayAssert#withElementType(String.class)` --> `asInstanceOf(array(String.class))`
* `ListAssert#withElementType(String.class)` --> `asInstanceOf(list(String.class))`

My only doubt is what to do with children of `AbstractObjectAssert` which are setting the `ACTUAL` type parameter to a specific value and therefore don't have a corresponding parameterized factory. They could maybe throw an `UnsupportedOperationException` but I don't like it 100%.

What do you think?

scordio:
Slightly related to this topic, I found a possible improvement with `extracting(Function)`  that I think would be nice for 3.13.

PR coming soon :-)

joel-costigliola:
@scordio is this idea still relevant with the [extracting improvement](https://assertj.github.io/doc/#assertj-core-3.13.0-extracting-type-propagation) ?

scordio:
@joel-costigliola I think the extracting improvement partially covers one of the use cases mentioned here and arrays and collections are still uncovered.

However, I still have the feeling this would be an api that can confuse the end user and right now I don't see a clear benefit. I will try first to enhance `extracting(String)` and `first()` methods to support better assertion types via `InstanceOfAssertFactory`.

I'm going to close this PR to avoid confusion.

joel-costigliola:
I had the same feeling too, I think that enhancing the other `extracting` and navigational methods (`first`, `last` ... ) should likely be enough.

PR #1553, opened 2019-07-10
add convenient assertSoftly for arbitrary commands  

qoomon:
#### Check List:
* Unit tests : YES

simplified approach of PR #1460 (Add method to use SoftAssertions with any custom Assertions)

### DIFF
--- a/src/test/java/org/assertj/core/api/BDDSoftAssertionsTest.java
+++ b/src/test/java/org/assertj/core/api/BDDSoftAssertionsTest.java
@@ -874,6 +874,17 @@ public class BDDSoftAssertionsTest extends BaseAssertionsTest {
   }
 
   @Test
+  public void should_assert_using_assertSoftlyExecutables() {
+    assertThatThrownBy(() -> assertSoftly(
+      () -> assertThat(true).isFalse(),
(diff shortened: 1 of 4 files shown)


joel-costigliola:
Thanks, we will look at it after releasing 3.13.0.

qoomon:
Nice, basically it provides the same functionality as `org.junit.jupiter.api.Assertions.assertAll(...)` from jUnit5

qoomon:
Maybe it make sense to add the functionality to pass multiple conditions(varargs) to all `satisfy` methods as well. Implementation is the done by `assertAll(conditions)`

src/test/java/org/assertj/core/api/BDDSoftAssertionsTest.java
```diff
@@ -873,6 +873,17 @@ public void should_assert_using_assertSoftly() {
        .hasMessageContaining("blue");
   }
 
+  @Test
+  public void should_assert_using_assertSoftlyExecutables() {
```
joel-costigliola: is there a difference with the test you have added in `SoftAssertionsTest` ? it there is not remove this one in favor of the other.

qoomon:
Sorry for the long delay

PR #1557, opened 2019-07-15
Disable example test cases for SoftAssertionsExtension

sbrannen:
Prior to this commit, the example test cases used to test the `SoftAssertionsExtension` were properly excluded from the Maven build by intentionally not following the "*Test" naming convention; however, those test cases were executed within IDEs, resulting in failures.

This commit addresses this inconvenience by annotating all example test cases with `@Disabled` and deactivating the `DisabledCondition` when actually executing the test cases via the `EngineTestKit`.

See #1418

### DIFF
--- a/src/test/java/org/assertj/core/api/junit/jupiter/AbstractSoftAssertionsExtensionIntegrationTests.java
+++ b/src/test/java/org/assertj/core/api/junit/jupiter/AbstractSoftAssertionsExtensionIntegrationTests.java
@@ -68,10 +68,11 @@ abstract class AbstractSoftAssertionsExtensionIntegrationTests {
   private void assertExecutionResults(Class<?> testClass, boolean nested) {
     EngineTestKit.engine("junit-jupiter")
                  .selectors(selectClass(testClass))
+                 .configurationParameter("junit.jupiter.conditions.deactivate", "*")
                  .execute().tests()
                  .assertStatistics(stats -> stats.started(nested ? 8 : 4).succeeded(nested ? 4 : 2).failed(nested ? 4 : 2))
                  .failed()
-                 // @format:off
+                 // @formatter:off
                  .assertThatEvents().haveExactly(nested ? 2 : 1,
                                                  event(test("multipleFailures"),
                                                        finishedWithFailure(instanceOf(AssertJMultipleFailuresError.class),
@@ -80,7 +81,7 @@ abstract class AbstractSoftAssertionsExtensionIntegrationTests {
                                                  event(test("parameterizedTest"),
                                                        finishedWithFailure(instanceOf(AssertJMultipleFailuresError.class), 
                                                                            message(msg -> msg.contains("Multiple Failures (1 failure)")))));
-    // @format:on
+                 // @formatter:on
   }
 
 }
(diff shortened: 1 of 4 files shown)


src/test/java/org/assertj/core/api/junit/jupiter/AbstractSoftAssertionsExtensionIntegrationTests.java
```diff
@@ -68,10 +68,11 @@ final void test_instance_per_class_with_nested_tests() {
   private void assertExecutionResults(Class<?> testClass, boolean nested) {
     EngineTestKit.engine("junit-jupiter")
                  .selectors(selectClass(testClass))
+                 .configurationParameter("junit.jupiter.conditions.deactivate", "org.junit.jupiter.engine.extension.DisabledCondition")
```
joel-costigliola: It would have been better to use `DisabledCondition.class.getName()` to avoid relying on a String value but `DisabledCondition` is not public, is making it public something that can be considered ?

This is not a blocker for this PR ;-)

src/test/java/org/assertj/core/api/junit/jupiter/BDDSoftAssertionsExtensionIntegrationTest.java
```diff
@@ -94,29 +95,31 @@ void parameterizedTest(int a, int b, int sum, BDDSoftAssertions softly) {
 	}
 
 	@TestInstance(PER_METHOD)
-	// Uses "Example" suffix to ensure that this class is not executed as part of the Maven build.
+	@Disabled("Executed via the JUnit Platform Test Kit")
```
joel-costigliola: Very informative disabled description 👍

src/test/java/org/assertj/core/api/junit/jupiter/BDDSoftAssertionsExtensionIntegrationTest.java
```diff
@@ -40,7 +41,7 @@
  * @since 3.13
```
scordio: Few rows above there is probably a `@link` typo: `SoftAssertionsExtensionTest` instead of `SoftAssertionsExtensionIntegrationTest`?

src/test/java/org/assertj/core/api/junit/jupiter/AbstractSoftAssertionsExtensionIntegrationTests.java
sbrannen: No, we have no intention to make that class public: it's an internal implementation detail of the Jupiter test engine.

But... good that you brought it up, since there is in fact no need to be that specific. What we are really trying to do here is disable the tests unless we execute them via the Test Kit. So I'll simplify that by deactivating all conditions with `"*"`. 😉

src/test/java/org/assertj/core/api/junit/jupiter/BDDSoftAssertionsExtensionIntegrationTest.java
sbrannen: Yep, good catch! I'll fix that.

sbrannen:
Changes have been force-pushed.

joel-costigliola:
Integrated thanks @sbrannen !

sbrannen:
Thanks for the quick turnaround, and... you're welcome!

PR #1562, opened 2019-07-24
Enforce assertj-core exclusion in maven-enforcer-plugin

scordio:
Improvement for #1552.

I think the `maven-enforcer-plugin` version should come from the parent POM, I plan to update it soon.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -141,6 +141,31 @@
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>enforce-banned-dependencies</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <bannedDependencies>
+                  <includes>
+                    <include>net.bytebuddy:byte-buddy</include>
+                  </includes>
+                  <excludes>
+                    <exclude>org.assertj:assertj-core</exclude>
+                    <exclude>*:*:*:jar:compile</exclude>
+                  </excludes>
+                </bannedDependencies>
+              </rules>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
           <source>1.8</source>


pom.xml
```diff
@@ -139,6 +139,28 @@
...
+          <execution>
+            <id>enforce-assertj-exclusion</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <bannedDependencies>
+                  <excludes>
+                    <exclude>org.assertj:assertj-core</exclude>
```
joel-costigliola: the policy in assertj-core is no dependency at all for the main code which I think  could be translated with (TBC)
```
*:*:*:jar:compile:*
```

scordio: `bytebuddy` is currently a compile dependency but I guess is fine, so I would add it as allowed. I would still exclude `assertj-core` in case of test scope.

joel-costigliola: Indeed, but that's the only one and it is shaded, I think we can just include it and exclude everything else.

joel-costigliola: @scordio juts added you as a collaborator on assertj parent pom, just commit directly the maven-enforcer-plugin declaration in there

scordio: Thanks!

Not sure I got you correctly, summarizing I'd put just the plugin version in the parent pom and the inclusions/exclusions should stay here in the core project, correct?

BTW I didn't understand why the build is failing and how it's related to my changes, locally that test seems fine both under maven and in the IDE.

joel-costigliola: I was able to reproduce it locally, the issue comes from the `ConfigurationForTests` loaded when the first Assert instance is created (this create a `WritableAssertionInfo` which calls `CONFIGURATION_PROVIDER.representation()` which loads `ConfigurationForTests`).

`ConfigurationForTests` disable filtering AssertJ elements from the stack trace when the failing test itself wanted to enable it.

Why it fails just now, I have no clue as I have added it in f258fa9604540d9587a41faf8bd8fb492d5fa920 a few month ago.

Anyway this is actually a bug since the intention of registering a configuration was that it was applied before any tests was run, it turns out that it is loaded when the first assertion is built which is after the first test started.

I think I'll disable the Configuration auto discovery for the moment.

It's actually lucky as I was about to release 3.13.0 :sweat_smile:

scordio: @joel-costigliola plugin config updated and conflict resolved, I think this PR should be ready now

scordio:
Integrated in ad5fb35519abd3e9aac8af1de9b1adbd1a60d4b2.

PR #1564, opened 2019-07-28
Make dependency on OpenTest4J optional again (fixes #1563).

### LINKED ISSUE #1563: AssertJ 3.13.0 requires OpenTest4J (java.lang.NoClassDefFoundError: org/opentest4j/MultipleFailuresError)
If the classpath does not contain OpenTest4J (e.g. you use JUnit 4 instead of JUnit 5) AssertJ `3.13.0` fails with:

`java.lang.NoClassDefFoundError: org/opentest4j/MultipleFailuresError`

A sample project to reproduce this issue: https://github.com/PascalSchumacher/assertj-3.13.0-requires-junit5-bug

PascalSchumacher:

### DIFF
--- a/src/main/java/org/assertj/core/error/AssertionErrorCreator.java
+++ b/src/main/java/org/assertj/core/error/AssertionErrorCreator.java
@@ -111,7 +111,10 @@ public class AssertionErrorCreator {
       if (multipleFailuresError instanceof AssertionError) { // means that we were able to build a MultipleFailuresError
         List<Throwable> failures = extractFailuresOf(multipleFailuresError);
         // we switch to AssertJMultipleFailuresError in order to control the formatting of the error message.
-        AssertionError assertionError = new AssertJMultipleFailuresError(heading, failures);
+        // we use reflection to avoid making opentest4j a required dependency
+        AssertionError assertionError = (AssertionError) constructorInvoker.newInstance("org.assertj.core.error.AssertJMultipleFailuresError",
+                                                                                        MULTIPLE_FAILURES_ERROR_ARGUMENT_TYPES,
+                                                                                        array(heading, failures));
         Failures.instance().removeAssertJRelatedElementsFromStackTraceIfNeeded(assertionError);
         return Optional.of(assertionError);
       }


filiphr:
@PascalSchumacher how sure are you that this fixes it? Looking at the code on line 108 means that OpenTest4J is present otherwise there would be an exception on that line instead.

PascalSchumacher:
I build AssertJ from source with this change and checked that it fixes the error in my sample project: https://github.com/PascalSchumacher/assertj-3.13.0-requires-junit5-bug

My knowledge of java class loading is very limited, but it seems that `AssertJMultipleFailuresError` is loaded because it is used by `AssertionErrorCreator`. `org.opentest4j.MultipleFailuresError` is loaded because  `AssertJMultipleFailuresError` extends it.

Use this simple test:

`assertThat("a").isNotNull();`

and the JVM argument `-verbose:class` shows that it is loaded (when added to the classpath):

>...
[Loaded org.assertj.core.error.AssertionErrorCreator from file:/.m2/repository/org/assertj/assertj-core/3.13.0/assertj-core-3.13.0.jar]
[Loaded org.opentest4j.MultipleFailuresError from file:/.m2/repository/org/opentest4j/opentest4j/1.1.1/opentest4j-1.1.1.jar]
[Loaded org.assertj.core.error.AssertJMultipleFailuresError from file:/.m2/repository/org/assertj/assertj-core/3.13.0/assertj-core-3.13.0.jar]
...

With my pull request is is not loaded anymore:

>...
[Loaded org.assertj.core.error.AssertionErrorCreator from file:/.m2/repository/org/assertj/assertj-core/3.13.1-SNAPSHOT/assertj-core-3.13.1-SNAPSHOT.jar]
[Loaded org.assertj.core.util.Arrays from file:/.m2/repository/org/assertj/assertj-core/3.13.1-SNAPSHOT/assertj-core-3.13.1-SNAPSHOT.jar]
...

filiphr:
That's weird. I thought that classes are usually loaded when they are first used and the `AssertJMultipleFailuresError` is first used when it is instantiated. It is not used in the functions as well.

Anyways, I mostly commented because I was curious about the error, as it should not happen :). Thanks for the fast fix

PascalSchumacher:
>classes are usually loaded when they are first used

That was what I thought too, so I was also suprised by this behavior.

joel-costigliola:
Thanks @PascalSchumacher  for the quick fix.

PR #1565, opened 2019-07-28
Make ConstructorInvoker#newInstance throw ReflectiveOperationExceptio…

PascalSchumacher:
…n (instead of Exception) and catch it (instead of Throwable) in code using this method.


PascalSchumacher:
ConstructorInvoker can also throw a `ClassNotFoundException`, so this in not valid. Sorry!

PR #1568, opened 2019-07-29
Add TYPE suffix to URI and URL factories to prevent shadowing

### LINKED ISSUE #1567: import static org.assertj.core.api.Assertions.* clashes with URI (java.net.URI)
#### Summary

Statically importing AssertJ assertions imports fields from `InstanceOfAssertFactories`. The `URI` field shadows `java.net.URI` and requires fully qualified naming when using `java.net.URI`. 

There's a similar issue for `URL` as these two types match exactly constant names from `InstanceOfAssertFactories`.

#### Example

```java
import static org.assertj.core.api.Assertions.*;

URI.create("http://foo")
```

Compilation failure:

```

PagedResourcesAssemblerUnitTests.java:[272,79] cannot find symbol
   symbol:   method create(java.lang.String)
   location: variable URI of type org.assertj.core.api.InstanceOfAssertFactory<java.net.URI,org.assertj.core.api.AbstractUriAssert<?>>
```

scordio:
Fixes #1567.

### DIFF
--- a/src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
+++ b/src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
@@ -204,15 +204,25 @@ public interface InstanceOfAssertFactories {
 
   /**
    * {@link InstanceOfAssertFactory} for a {@link URI}.
+   * <p>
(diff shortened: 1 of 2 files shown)


joel-costigliola:
Cool, thanks.

PR #1578, opened 2019-08-17
Instant-based comparators for Temporal assertions

### LINKED ISSUE #1452: ZonedDateTime assertions are not consistent and don't expose isBetween

puddlejumpr:
Compares OffsetDateTime, ZonedDateTime and LocalDateTime
objects using their timeLineOrder() comparator as default.

#### Check List:
* Fixes #1452 
* Unit tests : YES


joel-costigliola:
:eyes:

src/test/java/org/assertj/core/api/localdatetime/LocalDateTimeAssert_isEqualTo_Test.java
```diff
@@ -12,43 +12,62 @@
...
-    assertThatIllegalArgumentException().isThrownBy(() -> assertThat(LocalDateTime.now()).isEqualTo((String) null))
-                                        .withMessage("The String representing the LocalDateTime to compare actual with should not be null");
+  public void should_fail_if_given_string_parameter_cant_be_parsed() {
+    assertThatThrownBy(() -> assertions.isEqualTo("not a LocalDateTime")).isInstanceOf(DateTimeParseException.class);
   }
 
+  @Test
+  public void should_fail_if_given_would_be_equal_to_actual_with_default_comparator() {
+    ChronoLocalDateTime<JapaneseDate> inJapan = JapaneseChronology.INSTANCE.localDateTime(now);
+    assertThatThrownBy(() -> assertThat(now).isEqualTo(inJapan)).isInstanceOf(AssertionFailedError.class);
```
joel-costigliola: Since `inJapan` is not a `LocalDateTime`, `AbstractLocalDateTimeAssert.isEqualTo(LocalDateTime other)` is not called but `AbstractAssert.isEqualTo(Object other)` is which makes the assertion to fail.

But I think the assertion should pass as we want `isEqualTo` assertion to be consistent with  `LocalDateTime.isEqual`.

`AbstractLocalDateTimeAssert.isEqualTo` should actually override `AbstractAssert.isEqualTo` and take an `Object`.

So at the end the test assertions could be:
```java 
    assertThat(now.equals(inJapan)).isFalse();
    assertThat(now.isEqual(inJapan)).isTrue();
    assertThat(now).isEqualTo(inJapan);
    Object inJapanObject = inJapan;
    assertThat(now).isEqualTo(inJapanObject);
```

joel-costigliola:
@puddlejumpr I tool the liberty to push a commit with the changes I was suggesting to do but this is not final and I still want to hear your thoughts.

joel-costigliola:
Integrated thanks @puddlejumpr!

PR #1582, opened 2019-08-22
Decorate with JetBrains' @Language("regexp") annotations for regex arguments

xkr47:
This adds `@Language("regexp")` annotations to all public method arguments accepting regexps as `CharSequence`/`String`. This enables IntelliJ IDEA editor to

1. perform regexp syntax hilighting inside string arguments
2. allow unquoted editing of regexps in a separate editor (see https://confluence.jetbrains.com/display/CONTEST/IntelliLang)

BLOCKER: @Language not currently supported for `CharSequence` arguments; created issues https://youtrack.jetbrains.net/issue/IDEA-221169 and https://github.com/JetBrains/java-annotations/issues/16 for this to get resolved.

#### Check List:
* Unit tests : NA

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -136,6 +136,11 @@
       <version>3.9</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jetbrains</groupId>
+      <artifactId>annotations</artifactId>
+      <version>17.0.0</version>
+    </dependency>
   </dependencies>
   <build>
     <plugins>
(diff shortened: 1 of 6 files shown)


scordio:
To keep AssertJ free of IDE-specific dependencies, what about creating an `IntelliLang.xml` for AssertJ that can be imported into the IDE?

We could even try to ask JetBrains to put it as built-in configuration like the rules for Spring and Hibernate :slightly_smiling_face:

xkr47:
Cool ideas! Makes sense.

xkr47:
I'll see if I can make a new PR for this.

PR #1583, opened 2019-08-23
Add IntelliJ IDEA metadata conf for regexp arguments of AssertJ methods

xkr47:
By importing this configuration, AssertJ method regexp arguments will
get the benefits of the "Language Injection" feature which gives e.g.
regexp syntax hilighting, unquoted editing etc.

#### Check List:
* Replaces #1582
* Unit tests : NA

### DIFF
--- a/src/ide-support/IntelliLang.xml
+++ b/src/ide-support/IntelliLang.xml
@@ -1,0 +1,31 @@
+<application>
+  <component name="LanguageInjectionConfiguration">
+    <injection language="RegExp" injector-id="java">
+      <display-name>AssertJ.AbstractCharSequenceAssert (org.assertj.core.api)</display-name>
+      <place><![CDATA[psiParameter().ofMethod(0, psiMethod().withName("containsPattern").withParameters("java.lang.CharSequence").definedInClass("org.assertj.core.api.AbstractCharSequenceAssert"))]]></place>
+      <place><![CDATA[psiParameter().ofMethod(0, psiMethod().withName("doesNotMatch").withParameters("java.lang.CharSequence").definedInClass("org.assertj.core.api.AbstractCharSequenceAssert"))]]></place>
+      <place><![CDATA[psiParameter().ofMethod(0, psiMethod().withName("matches").withParameters("java.lang.CharSequence").definedInClass("org.assertj.core.api.AbstractCharSequenceAssert"))]]></place>
+    </injection>
+    <injection language="RegExp" injector-id="java">
+      <display-name>AssertJ.AbstractThrowableAssert (org.assertj.core.api)</display-name>
+      <place><![CDATA[psiParameter().ofMethod(0, psiMethod().withName("hasMessageFindingMatch").withParameters("java.lang.String").definedInClass("org.assertj.core.api.AbstractThrowableAssert"))]]></place>
+      <place><![CDATA[psiParameter().ofMethod(0, psiMethod().withName("hasMessageMatching").withParameters("java.lang.String").definedInClass("org.assertj.core.api.AbstractThrowableAssert"))]]></place>
+    </injection>
+    <injection language="RegExp" injector-id="java">
+      <display-name>AssertJ.RecursiveComparisonAssert (org.assertj.core.api)</display-name>
+      <place><![CDATA[psiParameter().ofMethod(0, psiMethod().withName("ignoringCollectionOrderInFieldsMatchingRegexes").withParameters("java.lang.String...").definedInClass("org.assertj.core.api.RecursiveComparisonAssert"))]]></place>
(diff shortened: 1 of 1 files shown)


xkr47:
@scordio wrote in #1582:
> We could even try to ask JetBrains to put it as built-in configuration like the rules for Spring and Hibernate slightly_smiling_face

Yeah, or could we even ask them to support IntelliLang.xml in root of .jar files so users would always get the correct version of the configuration? :thinking:

xkr47:
Nevertheless, I put the file in `src/main/resources/` so it would get included in the .jar and thus (hopefully) be easier for users to import into IDEA.

xkr47:
I hand-edited the `<display-name>` parts slightly so they would group & render nicely in File ➜ Settings / Editor ➜ Language Inspections..

xkr47:
If you want I could create a symlink from `src/ide-support/` to the file so it would be visible there as well.

Also, should write some documentation for https://assertj.github.io/doc/#ide-configuration

filiphr:
Is it possible for this to be done with an IntelliJ plugin?

xkr47:
No idea, never tried that..

scordio:
> Yeah, or could we even ask them to support IntelliLang.xml in root of .jar files so users would always get the correct version of the configuration? 🤔 

@xkr47 working this way could become a bit unpredictable, including a jar in an application could have side effects in the IDE that you don't immediately realize.

> If you want I could create a symlink from src/ide-support/ to the file so it would be visible there as well.

@xkr47 I'm not sure it's worth to package it in the jar. I'd start conservative putting it under `src/ide-support/`, following the same pattern like the assertion template. 

> Is it possible for this to be done with an IntelliJ plugin?

@filiphr `IntelliLang` should be the plugin which already provides the functionality. Is maybe a bit of overkilling to extend the configuration of a plugin with another plugin? I would try first to reach JetBrains.

filiphr:
>`IntelliLang` should be the plugin which already provides the functionality. Is maybe a bit of overkilling to extend the configuration of a plugin with another plugin? I would try first to reach JetBrains.

What I meant is the registration for the methods to be done by some AssertJ plugin not to add the functionality for the regex there.

xkr47:
Hmm.. IntelliJ does not like to import xml files, but it wants them inside zip files. While a bit clumsier to keep up-to-date, should I just include the zip file instead with the config inside? An option would be to use maven to create the zip but that takes some work..

scordio:
> What I meant is the registration for the methods to be done by some AssertJ plugin not to add the functionality for the regex there. 

Got it :+1: I still have the feeling the plugin could be the plan B if JetBrains doesn't accept it as built-in configuration.

xkr47:
@scordio will you formulate an issue for IntelliJ? https://youtrack.jetbrains.net/newIssue?draftId=25-2473393 and then for "Subsystem" parameter in the panel to the right, select option "Code Analysis. Language Injection" I guess..

scordio:
> @scordio will you formulate an issue for IntelliJ? https://youtrack.jetbrains.net/newIssue?draftId=25-2473393 and then for "Subsystem" parameter (in the panel to the right) select option "Code Analysis. Language Injection" I guess..

I can take care of it but let's have also a feedback from @joel-costigliola

scordio:
> Hmm.. IntelliJ does not like to import xml files, but it wants them inside zip files. While a bit clumsier to keep up-to-date, should I just include the zip file instead with the config inside? An option would be to use maven to create the zip but that takes some work.. Funnily they have a list of contributed settings that can be downloaded on their web page and those are plain xml..

@xkr47 I was able to import the XML from Settings > Editor > Language injections > Import

xkr47:
Ah, yes I tried the File -> Import Settings... Good to know.

joel-costigliola:
I'm not an Idea expert, so as long as we don't add a dependency to assertj, fine by me.
From my non expert eyes, the PR looks good, I'll let @scordio finish taking care of this PR.

scordio:
Integrated, thanks @xkr47!

joel-costigliola:
thanks @xkr47 !

scordio:
> @scordio will you formulate an issue for IntelliJ? https://youtrack.jetbrains.net/newIssue?draftId=25-2473393 and then for "Subsystem" parameter (in the panel to the right) select option "Code Analysis. Language Injection" I guess..

Here it is: [IDEA-221416](https://youtrack.jetbrains.net/issue/IDEA-221416)

scordio:
> Here it is: [IDEA-221416](https://youtrack.jetbrains.net/issue/IDEA-221416)

IDEA issue has been fixed in build 193.4225, the configuration should appear in one of the next releases!

scordio:
IDEA 2019.3 is out and contains built-in AssertJ rules!

@joel-costigliola should we remove https://github.com/joel-costigliola/assertj-core/blob/master/src/ide-support/IntelliLang.xml?
I don't think it would be used by anyone, the users will rather update the IDE. I would not even keep the file as a reference for additional rules, we can always search for this PR.

joel-costigliola:
let's remove it, it would be more confusing than anything.

Issue #1574, opened 2019-08-16
escapePercent (%) missing in org.assertj.core.internal.Failures#assertionErrorMessage


grigoni:
#### Summary
when using
        assertThatThrownBy(() -> restCal(XYZ))
                .isInstanceOf(MyServiceErrorException.class)
                .hasMessage(expectedMessage)

it throws MissingFormatArgumentException

#### Example

java.util.MissingFormatArgumentException: Format specifier '%3A'

	at java.util.Formatter.format(Formatter.java:2519)
	at java.util.Formatter.format(Formatter.java:2455)
	at java.lang.String.format(String.java:2940)
	at org.assertj.core.util.Strings.formatIfArgs(Strings.java:86)
	at org.assertj.core.error.MessageFormatter.format(MessageFormatter.java:62)
	at org.assertj.core.error.BasicErrorMessageFactory.create(BasicErrorMessageFactory.java:109)
	at org.assertj.core.internal.Failures.assertionErrorMessage(Failures.java:142)
	at org.assertj.core.internal.Failures.failure(Failures.java:132)
	at org.assertj.core.internal.Throwables.assertHasMessage(Throwables.java:87)
	at org.assertj.core.api.AbstractThrowableAssert.hasMessage(AbstractThrowableAssert.java:66)

epeee:
Thank your for reporting this one.
I will add a reproducer to our test base and provide a pr soon.

joel-costigliola:
In addition to this bug, we we should also look for all usage of assertions taking a `String value, Object...` parameters to call `String.format` to make sure we fix similar potential issues (it's not the first time we had this kind of bug, let's try to fix once for all)

epeee:
@joel-costigliola, yes, indeed. I will have a look at other usages too.

grigoni:
I am very thankful the issue is receiving a lot of care, I could only provide few lines of code and no proper test case due to time constraints, I'll gladly look at your code changes
thanks again!
G.

epeee:
Some more details, it looks like we introduced this one in 3.13.0.

A sample to reproduce this one:

```java
    assertThatThrownBy(() -> { throw new IllegalStateException("%3A"); } )
      .hasMessage("message");
```

PR #1585, opened 2019-08-25
fix possible MissingFormatArgumentException in ShouldContain

epeee:
As a follow up of #1574 I had a quick look about similar cases like the one described in the linked ticket.


joel-costigliola:
Integrated thanks @epeee !

PR #1588, opened 2019-08-30
Add extracting(Function, InstanceOfAssertFactory) to AbstractObjectAssert

### LINKED ISSUE #1490: Incorrect Assertion type while using extraction by function (BUG)

scordio:
#### Check List:
* Fixes #1490
* Unit tests : YES

#### Open issue

Instead of writing the inner implementation from scratch like in:
https://github.com/joel-costigliola/assertj-core/blob/167c63e4a901d598ea9418eef6a6e885aa7ba279/src/main/java/org/assertj/core/api/AbstractObjectAssert.java#L704-L712
I tried to delegate the actual logic to the existing `extracting` methods, but I'm having issues in `SoftAssertionsTest`. E.g., when executing `should_pass_when_using_extracting_with_object()`, I get:
```
	at org.assertj.core.api.ErrorCollector.intercept(ErrorCollector.java:59)
	at org.assertj.core.api.StringAssert$ByteBuddy$RyD7qQEf$ByteBuddy$WbC2sF8N.assertj$setup(Unknown Source)
	at org.assertj.core.api.SoftProxies.createSoftAssertionProxy(SoftProxies.java:109)
	at org.assertj.core.api.ProxifyMethodChangingTheObjectUnderTest.createAssertProxy(ProxifyMethodChangingTheObjectUnderTest.java:55)
	at org.assertj.core.api.ProxifyMethodChangingTheObjectUnderTest.intercept(ProxifyMethodChangingTheObjectUnderTest.java:45)
	at org.assertj.core.api.ProxyableObjectAssert$ByteBuddy$yrZx1PLg.extracting(Unknown Source)
	at org.assertj.core.api.SoftAssertionsTest.should_pass_when_using_extracting_with_object(SoftAssertionsTest.java:428)
```

I guess the ByteBuddy configuration should be adjusted somehow, maybe to avoid proxying `extracting(..., InstanceOfAssertFactory)` methods, but I am not very familiar with ByteBuddy.

Any suggestion?

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
@@ -659,7 +659,7 @@ public abstract class AbstractObjectAssert<SELF extends AbstractObjectAssert<SEL
     Object value = byName(propertyOrField).apply(actual);
     String extractedPropertyOrFieldDescription = extractedDescriptionOf(propertyOrField);
     String description = mostRelevantDescription(info.description(), extractedPropertyOrFieldDescription);
-    return newObjectAssert(value).as(description);
+    return newObjectAssert(value).withAssertionState(myself).as(description);
   }
 
   /**
@@ -699,16 +699,10 @@ public abstract class AbstractObjectAssert<SELF extends AbstractObjectAssert<SEL
    *
    * @since 3.14.0
    */
-  @SuppressWarnings("unchecked")
   @CheckReturnValue
   public <ASSERT extends AbstractAssert<?, ?>> ASSERT extracting(String propertyOrField,
                                                                  InstanceOfAssertFactory<?, ASSERT> assertFactory) {
-    requireNonNull(assertFactory, shouldNotBeNull("assertFactory").create());
-    Object value = byName(propertyOrField).apply(actual);
-    objects.assertIsInstanceOf(info, value, assertFactory.getType());
-    String extractedPropertyOrFieldDescription = extractedDescriptionOf(propertyOrField);
-    String description = mostRelevantDescription(info.description(), extractedPropertyOrFieldDescription);
-    return (ASSERT) assertFactory.createAssert(value).as(description);
+    return extracting(propertyOrField).asInstanceOf(assertFactory);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


joel-costigliola:
I'll have a look at it

scordio:
@joel-costigliola thanks!

I tried to add
`.or(named("extracting").and(takesArgument(1, InstanceOfAssertFactory.class)))` 
to `METHODS_NOT_TO_PROXY` but it was not enough.

joel-costigliola:
I got it working by moving the common extracting code into methods not named extracting, I saw in the debugger that your implementation led to have two erroCollector fields injected when `generateProxyClass` was called, one being null.

This piece of code is quite tricky I have to say (most of it was written by Raphael Winterhalter, byte buddy author).

joel-costigliola:
I also found a bug when extracting multiple times like:
```
    softly.assertThat(frodo)
          .overridingErrorMessage("error 7")
          .extracting(TolkienCharacter::getRace)
          .extracting(Race::getClass, InstanceOfAssertFactories.CLASS)
```
this leads `ProxifyMethodChangingTheObjectUnderTest.actualClass` to fail.

it can be fixed by adding this line to `actualClass`:
```
    if (result instanceof ClassAssert) return Class.class;
```
I think the reason is that `Class` is a generic, the bug would likely occur with any generic type, I'll look into that.

scordio:
Amazing, thank you!

scordio:
I just noticed my current implementation has a bug. At first I thought it would have been a good idea to enforce the factory type at compile time (i.e., using factories corresponding to the extractor type or to a parent of it), but it excludes the use case where the value could have been extracted with a super type, like `Object`.

I'm going to adjust the implementation and add a test for this use case, then the PR will be ready.

scordio:
@joel-costigliola PR ready!

joel-costigliola:
thanks @scordio

PR #1589, opened 2019-08-31
Upgrade parent to version 2.2.6 and Mockito to version 3.0.0

scordio:
This introduces dependency and plugin management with overridable versions, a dependency convergence rule for the enforcer and a GitHub-friendly license header.


scordio:
For some reason there are tests failing only under maven but not under the IDE, it's probably some plugin interfering. I'll investigate a bit more.

scordio:
When you realize after days that a test is failing only because of a changed line number 😅 

As mentioned in https://github.com/joel-costigliola/assertj-maven-parent-pom/pull/41#issuecomment-529232642, using `LICENSE.txt` for file headers is not so nice due to code formatting, therefore I put the smaller header in the pom but I plan to move it to the parent pom.

@joel-costigliola PR ready!

joel-costigliola:
looks good :)

PR #1608, opened 2019-09-18
Fix #1607: Improve HamcrestCondition generic type inference

### LINKED ISSUE #1607: HamcrestCondition matcher generic is too strict
#### Summary

I'm working on an ErrorProne check which produces 'suggested fixes' that can be automatically applied to migrate from junit asserts to AssertJ. `assertThat(Object, Matcher)` is more difficult than other methods, so for the time being, I've simply fallen back to using `assertThat(actual).is(new HamcrestCondition<>(ORIGINAL_HAMCREST_MATCHER)`.
Unfortunately there are many cases where this doesn't compile!

#### Example

```java
Collection<String> value = ImmutableList.of();
assertThat(value).is(new HamcrestCondition<>(not(hasItems("one", "two"))));
﻿                                          ﻿﻿^ 🌶 Cannot infer arguments 🌶 (I hope the spaces align!)
```

#### Recommendation

We should loosen the Matcher to `Matcher<? extends T>` rather than `T`.

Aforementioned ErrorProne migration is [here](https://github.com/palantir/gradle-baseline/blob/develop/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferAssertj.java) if you're interested. The [test](https://github.com/palantir/gradle-baseline/blob/develop/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/PreferAssertjTests.java) are probably the most interesting place to start.

carterkozak:
#### Check List:
* Fixes #1607
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/HamcrestCondition.java
+++ b/src/main/java/org/assertj/core/api/HamcrestCondition.java
@@ -32,14 +32,14 @@ import org.hamcrest.StringDescription;
 */
 public class HamcrestCondition<T> extends Condition<T> {
 
-  private Matcher<T> matcher;
+  private Matcher<? extends T> matcher;
 
   /**
    * Constructs a {@link Condition} using the matcher given as a parameter.
    * 
    * @param matcher the Hamcrest matcher to use as a condition
    */
- public HamcrestCondition(Matcher<T> matcher) {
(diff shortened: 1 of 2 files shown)


joel-costigliola:
Integrated thanks !

Issue #1611, opened 2019-09-21
Use assertThatAssertionErrorIsThrownBy instead of assertThatThrownBy(() -> ...).isInstanceOf(AssertionError.class)


joel-costigliola:
#### Summary

`AssertionsUtil.assertThatAssertionErrorIsThrownBy` checks the caught throwable is an `AssertionError`.
Make sure the tests are written following the GIVEN WHEN THEN pattern

#### Example

```java
CompletableFuture<String> future = new CompletableFuture<>();

assertThatThrownBy(() -> assertThat(future).hasFailed()).isInstanceOf(AssertionError.class)
                                                        .hasMessage(shouldHaveFailed(future).create());
```
should be written as
```java
// GIVEN
CompletableFuture<String> future = new CompletableFuture<>();
// WHEN
ThrowingCallable code = () -> assertThat(future).hasFailed();
// THEN
assertThatAssertionErrorIsThrownBy(code).withMessage(shouldHaveFailed(future).create());
```
Note the use of GIVEN WHEN THEN to clarify the test steps.

cgrabmann:
working on it

cgrabmann:
@joel-costigliola By now I just replaced every occurance in the CompletableFutureAssert tests, as well as replaced and refactored all LocalDateAssert and all InstantAssert tests to follow the current test guidelines. I still have to do all LocalDateTimeAssert, LocalTimeAssert and OffsetDateTimeAssert tests and I already have around 30 changed files.
Since big PRs aren't much fun, would you like the changes in batches or in one big bulk?

joel-costigliola:
Good idea, let's split the PR in per type.

fdooom663:
Gfdnvzh

quaythuoc-org:
Thanks you about this

MilkTimeplus:
wow, difficult, I couldn't understand it

chandra20sekhar:
i didn't understand it

PR #1618, opened 2019-09-25
#1611: rewrite CompletableFutureAssert test classes to use assertThatAssertionErrorIsThrownBy

cgrabmann:
#### Check List:
* relates to #1611 (multiple parts per Assertion class tests)
* Unit tests : NA


joel-costigliola:
Integrated thanks @cgrabmann !

Issue #1605, opened 2019-09-15
API improvements related to asInstanceOf


scordio:
## Summary
Some existing APIs can benefit from a chained `asInstanceOf` invocation in order to obtain more specialized assertions, for example the methods which change the current object under test.

For these cases, it would be convenient to have an overloaded method which expects an `InstanceOfAssertFactory` as additional parameter.

**Happy to receive any feedback!**

### Syntax sugar for InstanceOfAssertFactory parameter
- [x] To improve readability for the cases mentioned above, a static method `as` can be provided in ~`InstanceOfAssertFactories`~ `Assertions` (#1609).
```java
public static <T, ASSERT extends AbstractAssert<?, ?>> InstanceOfAssertFactory<T, ASSERT> as(InstanceOfAssertFactory<T, ASSERT> assertFactory) {
  return assertFactory;
}
```
The approach is similar to [`Assertions#from`](https://github.com/joel-costigliola/assertj-core/blob/7deb7821cc84991bc23375ca2b8054cea5a9200c/src/main/java/org/assertj/core/api/Assertions.java#L1961-L1977).

### AbstractObjectAssert
- [X] `extracting(String, InstanceOfAssertFactory)` (#1581):
```java
TolkienCharacter frodo = new TolkienCharacter("Frodo", 33, HOBBIT);
// Current
assertThat(value).extracting("name")
                 .asInstanceOf(STRING)
                 .startsWith("Fro");
// Improvement
assertThat(value).extracting("name", as(STRING))
                 .startsWith("Fro");
```
- [X] `extracting(Function, InstanceOfAssertFactory)` (#1588):
```java
TolkienCharacter frodo = new TolkienCharacter("Frodo", 33, HOBBIT);
// Current
assertThat(value).extracting(TolkienCharacter::getName)
                 .asInstanceOf(STRING)
                 .startsWith("Fro");
// Improvement
assertThat(value).extracting(TolkienCharacter::getName, as(STRING))
                 .startsWith("Fro");
```

### AbstractMapAssert
- [x] `extractingByKey(Object, InstanceOfAssertFactory)` (#1623, #1619):
```java
Map<String, Object> map = new HashMap<>();
map.put("name", "kawhi");
// Current
assertThat(map).extractingByKey("name")
                 .asInstanceOf(STRING)
                 .startsWith("kaw");
// Improvement
assertThat(value).extractingByKey("name", as(STRING))
                 .startsWith("kaw");
```

[...]

PR #1619, opened 2019-09-26
Add extractingByKey(KEY, InstanceOfAssertFactory) to AbstractMapAssert

scordio:
Also PR for release notes: assertj/doc#19

#### Check List:
* Fixes N/A
* Unit tests : YES

### DIFF
--- a/src/test/java/org/assertj/core/api/SoftAssertionsTest.java
+++ b/src/test/java/org/assertj/core/api/SoftAssertionsTest.java
@@ -556,6 +556,9 @@ public class SoftAssertionsTest extends BaseAssertionsTest {
           .extractingByKey("name")
           .isEqualTo("kawhi");
     softly.assertThat(map)
+          .extractingByKey("name", as(STRING))
+          .startsWith("kaw");
+    softly.assertThat(map)
           .extractingFromEntries(Map.Entry::getKey, Map.Entry::getValue)
           .contains(tuple("name", "kawhi"), tuple("age", 25));
     softly.assertThat(map)
@@ -1708,9 +1711,14 @@ public class SoftAssertionsTest extends BaseAssertionsTest {
           .overridingErrorMessage("error message")
           .extractingByKey("a")
           .isEqualTo("456");
+    softly.assertThat(map)
+          .as("extracting(\"a\") as string")
+          .overridingErrorMessage("error message")
+          .extractingByKey("a", as(STRING))
+          .startsWith("456");
     // THEN
     List<Throwable> errors = softly.errorsCollected();
-    assertThat(errors).hasSize(16);
+    assertThat(errors).hasSize(17);
     assertThat(errors.get(0)).hasMessageContaining("MapEntry[key=\"abc\", value=\"ABC\"]");
     assertThat(errors.get(1)).hasMessageContaining("empty");
     assertThat(errors.get(2)).hasMessageContaining("gh")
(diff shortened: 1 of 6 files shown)


src/test/java/org/assertj/core/api/SoftAssertionsTest.java
```diff
@@ -1687,9 +1713,15 @@ public void map_soft_assertions_should_report_errors_on_final_methods_and_method
           // convert to Object otherwise will use extracting(String) in AbstractObjectAssert
           .extracting((Object) "a")
           .isEqualTo("456");
+    softly.assertThat(map)
+          .as("extracting(\"a\") as string")
+          .overridingErrorMessage("error message")
+          // convert to Object otherwise will use extracting(String, InstanceOfAssertFactory) in AbstractObjectAssert
+          .extracting((Object) "a", as(STRING))
```
joel-costigliola: Thinking out loud, I wonder if we should rename `extracting` to `extractingByKey` to avoid the collision with `extracting(String, InstanceOfAssertFactory)` in `AbstractObjectAssert`, we would the same for `extracting(Object key)`, it would have the advantage to be obvious, still it might be confusing to see `extracting(Object )` from `AbstractObjectAssert` along with `extractingByKey`

We could even use `extractingByKey(K key)` since if you are extracting a key is should be compatible with the map key type.

If we go this way we would create another PR for renaming `extracting(Object key)` to `extractingByKey(K key)` but keeping the `extracting(Object key)` and deprecate it.

scordio: You're right, the current `extracting(Object key)` is confusing also because it has an `extracting(String)` example in its javadoc 😅 (I just realized it)
I like `extractingByKey(K key)` also because of key type safety.

Shall I pause this PR and introduce `extractingByKey(K key)` first? Afterwards I can rebase this one adding the `InstanceOfAssertFactory` variant only to the new method.

joel-costigliola: It does not matter if we do `extractingByKey(K, InstanceOfAssertFactory)` first and then `extractingByKey(K)` as long as we do both for 3.14.

Goo point for renaming `extracting(Object... keys)` too.

Let's finish this PR first, see if there is no issue using the `K` generic type instead of `Object` and then move on to `extractingByKey(K)`.

I'm planning to release 3.14 soon as soon as we have these changes done.

scordio: About `extracting(Object... keys)`, it should become `extractingByKeys(Object... keys)`, right?

joel-costigliola: should be `extractingByKeys(K... keys)`, I suspect will lead to a javac warning about heap pollution as in `extracting(Function...)`

scordio: Yes I meant with the generic 😅 ok will do!

About 3.14 release, would you like to include also the remaining in #1605 for `IterableAssert` and `OptionalAssert`or shall these go into the next one?

joel-costigliola: I think for consistency it would be better to finish #1605, how long do you think it's going to take ? (don't feel pressure, just want to have a gut feeling about it)

scordio: I would say additional 7-10 days, not because of complexity but due to the limited time I will have in the upcoming week

scordio: I tried to close this PR first but having partial renaming was a bit confusing, so I preferred to do the deprecations first in #1623.

scordio: PR ready again! #1623 should be checked and merged first, then I will probably need to rebase this one.

joel-costigliola: yup, I have just commented it

scordio:
Closed by my reference mistake 😅 reopening it

scordio:
@joel-costigliola PR ready!

joel-costigliola:
Integrated thanks @scordio!

PR #1622, opened 2019-09-26
#1611: rewrite InstantAssert test classes to use assertThatAssertionErrorIsThrownBy and be more compliant with current guidelines

cgrabmann:
#### Check List:
* relates to #1611 (multiple parts per Assertion class tests)
* Unit tests : NA


joel-costigliola:
Integrated thanks for the great refactoring @cgrabmann!

PR #1623, opened 2019-09-28
Deprecate extracting() in favor of extractingByKey()/extractingByKeys() for MapAssert

scordio:
#### Check List:
* Fixes https://github.com/joel-costigliola/assertj-core/pull/1619#discussion_r328548320
* Unit tests : YES


src/main/java/org/assertj/core/api/AbstractMapAssert.java
```diff
@@ -1566,6 +1569,40 @@ public SELF withThreadDumpOnError() {
...
+   * map.put("name", "kawhi");
+   * map.put("age", 25);
+   *
+   * assertThat(map).extractingByKeys("name", "age")
+   *                .contains("kawhi", 25);</code></pre>
+   * <p>
+   * Note that the order of extracted keys value is consistent with the iteration order of the array under test.
+   * <p>
+   * Nested keys are not yet supported, passing "name.first" won't get a value for "name" and then try to extract
+   * "first" from the previously extracted value, instead it will simply look for a value under "name.first" key.
```
joel-costigliola: I'm not so sure if this section is valuable as it only applies for String keys, thoughts ?

src/test/java/org/assertj/core/api/BDDSoftAssertionsTest.java
```diff
@@ -1522,8 +1522,7 @@ public void map_soft_assertions_should_report_errors_on_final_methods_and_method
     softly.then(map)
           .as("extracting(\"a\")")
           .overridingErrorMessage("error message")
-          // convert to Object otherwise will use extracting(String) in AbstractObjectAssert
-          .extracting((Object) "a")
+          .extractingByKey("a")
```
joel-costigliola: that looks better indeed! :)

src/main/java/org/assertj/core/api/AbstractMapAssert.java
```diff
@@ -1566,6 +1569,40 @@ public SELF withThreadDumpOnError() {
...
+   * <p>
+   * Example:
+   * <pre><code class='java'> Map&lt;String, Object&gt; map = new HashMap&lt;&gt;();
+   * map.put("name", "kawhi");
+   * map.put("age", 25);
+   *
+   * assertThat(map).extractingByKeys("name", "age")
+   *                .contains("kawhi", 25);</code></pre>
+   * <p>
+   * Note that the order of extracted keys value is consistent with the iteration order of the array under test.
```
joel-costigliola: Add another note to say that providing no keys will result in an empty list.

src/test/java/org/assertj/core/api/map/MapAssert_extractingByKeys_Test.java
```diff
@@ -0,0 +1,118 @@
...
+    then(error).hasMessage(actualIsNull());
+  }
+
+  @Test
+  void should_allow_list_assertions_on_values_extracted_from_given_map_keys() {
+    // WHEN
+    AbstractListAssert<?, List<?>, Object, ObjectAssert<Object>> result = assertThat(map).extractingByKeys(NAME, "age");
+    // THEN
+    result.containsExactly("kawhi", 25);
+  }
```
joel-costigliola: add a tests showing the behavior when providing no keys = will result in an empty list.

joel-costigliola:
Good work, just a few comments!

src/main/java/org/assertj/core/api/AbstractMapAssert.java
scordio: I agree, it doesn't make much sense now. I was copy-pasting it from the corresponding `extracting` and even there it doesn't seem 100% right. The original source of this sentence was the javadoc for `extracting(Object... keys)` which has last change in 2017, probably the original concept has changed over time.

Also, this seems to overlap the description of `extracting` in `AbstractObjectAssert`:
https://github.com/joel-costigliola/assertj-core/blob/215c93d4b0ac3c040772ab552470e956d2cabb62/src/main/java/org/assertj/core/api/AbstractObjectAssert.java#L621

I'll remove this sentence from here.

scordio:
@joel-costigliola PR ready!

joel-costigliola:
Integrated thanks @scordio!

joel-costigliola:
^^ Notice how I'm not putting a space before `!` 😆

PR #1625, opened 2019-09-30
Adds Behavior-driven development style entry point for assumptions.

### LINKED ISSUE #1613: Add 'given' method for BDD assumptions
#### Summary

Just as there is an [`org.assertj.core.api.BDDAssertions`](https://assertj.github.io/doc/#alternative-entry-points) that provides an equivalent BDD approach to `org.assertj.core.api.Assertions` (`org.assertj.core.api.BDDAssertions` supplies the method `then`, an entrypoint for a BDD assertion), A **`org.assertj.core.api.BDDAssumptions`** (equivalent to [`org.assertj.core.api.Assumptions`](https://assertj.github.io/doc/#assertj-core-assumptions)) **could be introduced to provide the method `given`, an entrypoint for a BDD assumption**.

#### Example

```java
@Test
public void when_all_assumptions_are_met_the_test_is_run_normally() {
  // since this assumption is true ...
  given(frodo.getRace()).isEqualTo(HOBBIT);
  // ... this assertion is performed
  then(fellowshipOfTheRing).doesNotContain(sauron);
}
```

gmullerb:
* Fixes #1613
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/BDDAssumptions.java
+++ b/src/main/java/org/assertj/core/api/BDDAssumptions.java
@@ -1,0 +1,2392 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2019 the original author or authors.
+ */
+package org.assertj.core.api;
+
+import java.io.File;
+import java.io.InputStream;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.net.URI;
+import java.net.URL;
+import java.nio.file.Path;
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.OffsetDateTime;
+import java.time.OffsetTime;
+import java.time.ZonedDateTime;
+import java.util.Date;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.OptionalDouble;
+import java.util.OptionalInt;
+import java.util.OptionalLong;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicIntegerArray;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicLongArray;
+import java.util.concurrent.atomic.AtomicLongFieldUpdater;
+import java.util.concurrent.atomic.AtomicMarkableReference;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.atomic.AtomicReferenceArray;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import java.util.concurrent.atomic.AtomicStampedReference;
+import java.util.function.DoublePredicate;
+import java.util.function.IntPredicate;
+import java.util.function.LongPredicate;
+import java.util.function.Predicate;
+import java.util.stream.DoubleStream;
+import java.util.stream.IntStream;
+import java.util.stream.LongStream;
+import java.util.stream.Stream;
+
+import org.assertj.core.api.ThrowableAssert.ThrowingCallable;
+import org.assertj.core.util.CheckReturnValue;
+
+/**
+ * Behavior-driven development style entry point for assumption methods for different types, which allow to skip test execution when assumptions are not met.
+ * <p>
+ * The difference with the {@link Assumptions} class is that entry point methods are named <b><code>given</code></b> instead of
+ * <code>assumeThat</code>.
+ * <p>
+ * {@link BDDAssumptions} and {@link BDDAssertions} complement each other to allow a fluent Behavior-driven development.
+ * <p>
+ * Examples:
+ * <pre>{@code
+ * String frodo = "HOBBIT";
+ * List<String> fellowshipOfTheRing = Arrays.asList("Aragorn", "Gandalf", "Frodo", "Legolas"); // and more
+ * }</pre>
+ * <pre><code class='java'>
+ * {@literal @Test}
+ *  public void given_the_assumption_is_not_met_the_test_is_skipped() {
+ *    given(frodo).isEqualTo("ORC");
+ *
+ *    // ... following code is not executed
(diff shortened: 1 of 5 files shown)


src/main/java/org/assertj/core/api/BDDAssumptions.java
```diff
@@ -0,0 +1,2392 @@
...
+ * {@link BDDAssumptions} and {@link BDDAssertions} complement each other to allow a fluent Behavior-driven development.
+ * <p>
+ * Examples:
+ * <pre>{@code
+ * String frodo = "HOBBIT";
+ * List<String> fellowshipOfTheRing = Arrays.asList("Aragorn", "Gandalf", "Frodo", "Legolas"); // and more
+ * }</pre>
+ * <pre><code class='java'>
+ * {@literal @Test}
+ *  public void given_the_assumption_is_not_met_the_test_is_skipped() {
```
joel-costigliola: code must start after adding one space to `<pre><code class='java'>` and </code></pre>must not be put on a new line, here's a reference to follow:
```
   * <pre><code class='java'> // Create frodo, setting its name, age and Race
   * TolkienCharacter frodo = new TolkienCharacter(&quot;Frodo&quot;, 33, HOBBIT);
   *
   * assertThat(frodo).extracting(TolkienCharacter::getName, as(InstanceOfAssertFactories.STRING))
   *                  .startsWith(&quot;Fro&quot;);</code></pre>
```

this comment applies for all methods in this class, sorry that's a bit of work! but thanks for having out code examples for each method that's really great! :+1:

src/main/java/org/assertj/core/api/BDDAssumptions.java
```diff
@@ -0,0 +1,2392 @@
...
+ *    given(frodo).isEqualTo("ORC");
+ *
+ *    // ... following code is not executed
+ *    then(fellowshipOfTheRing).contains("Sauron");
+ *  }
+ *
+ * {@literal @Test}
```
joel-costigliola: also `@Test` was not aligned with `public`, see
image

src/main/java/org/assertj/core/api/BDDAssumptions.java
```diff
@@ -0,0 +1,2392 @@
...
+ *    // ... following code is executed
+ *    then(fellowshipOfTheRing).doesNotContain("Sauron");
+ *  }
+ * </code></pre>
+ *
+ * @since 3.14.0
+ * @author Gonzalo Müller
+ */
+@CheckReturnValue
+public class BDDAssumptions extends Assumptions {
```
joel-costigliola: Add a test to make sure this class will always be in sync with `Assertions` so that if we add another `assertThat` method the test would fail if we forget to add the corresponding `given` method here.

The easiest is to add a  `Assertions_sync_with_BDDAssumptions_Test` test similar to  `Assertions_sync_with_Assumptions_Test` but for `BDDAssumptions`.

joel-costigliola: don't extend `Assumptions`, use delegation instead with static import.

src/test/java/org/assertj/core/api/BDDAssumptionsTest.java
```diff
@@ -0,0 +1,1529 @@
...
+
+import static org.mockito.BDDMockito.mock;
+import static org.mockito.BDDMockito.willReturn;
+
+/**
+ * Tests for <code>{@link org.assertj.core.api.BDDAssumptions}</code>.
+ *
+ * @author Gonzalo Müller
+ */
+public class BDDAssumptionsTest {
```
joel-costigliola: move to `org.assertj.core.api.assumptions` with the other assumptions tests

src/test/java/org/assertj/core/api/BDDAssumptionsTest.java
```diff
@@ -0,0 +1,1529 @@
...
+    @Test
+    public void should_run_test_when_assumption_passes() {
+      thenCode(() -> given(actual).isTrue())
+        .doesNotThrowAnyException();
+    }
+
+    @Test
+    public void should_ignore_test_when_assumption_fails() {
+      thenThrownBy(() -> given(actual).isFalse())
+        .isInstanceOf(AssumptionViolatedException.class);
```
joel-costigliola: let's extract a static util method `expectAssumptionViolatedException(ThrowingCallable shouldRaiseThrowable)` in `AssertionsUtil` to simply write every `should_ignore_test_when_...` like:
```
expectAssumptionViolatedException(() -> given(actual).isFalse());
```
we could go even further and make `should_ignore_test_when_assumption_fails()` a parameterized test taking a `ThrowingCallable` parameters

joel-costigliola: if we do that then the Nested class are not so useful anymore and should be changed to simple test like `should_run_test_when_boolean_assumption_passes`, `should_run_test_when_boolean_array_assumption_passes`, ...

src/main/java/org/assertj/core/api/BDDAssumptions.java
gmullerb: Absolutely agree, I did it because `BDDAssertions` extends `Assertions` (I thought it was some kind of pattern in the lib). Thanks!

src/main/java/org/assertj/core/api/BDDAssumptions.java
gmullerb: I doing the change (locally it was looking fine), please let me know if the issue continues. Thanks!

src/test/java/org/assertj/core/api/BDDAssumptionsTest.java
gmullerb: I'm doing partially changes, I do not understood all the idea. Did you mean remove `actual` field, and pass it as argument so nested tested are removed and may use Parameterized tests.
Something like this:
```
    @Test
    public void should_run_test_when_boolean_assumption_passes() {
      thenCode(() -> given(true).isTrue())
        .doesNotThrowAnyException();
    }

    @Test
    public void should_ignore_test_when_boolean_assumption_fails() {
      expectAssumptionViolatedException((true) -> given(actual).isFalse());
    }
```
and then parameterized:
```
Thanks!
    @ParameterizedTest
    // ... parameters
    public void should_run_test_when_assumption_passes(ThrowingCallable mayRaiseError) {
      thenCode(mayRaiseError)
        .doesNotThrowAnyException();
    }

    @ParameterizedTest
    // ... parameters
    public void should_ignore_test_when_assumption_fails(ThrowingCallable shouldRaiseError) {
      expectAssumptionViolatedException(shouldRaiseError);
    }
```

gmullerb:
Thanks for all the feedback!!

src/test/java/org/assertj/core/api/BDDAssumptionsTest.java
gmullerb: For some tests this "may be not" possible,  e.g. `BDDAssumptions_given_Comparable_Test`.

src/test/java/org/assertj/core/api/BDDAssumptionsTest.java
gmullerb: Done! Thanks!

src/main/java/org/assertj/core/api/BDDAssumptions.java
gmullerb: Done! Thanks!

scordio:
It seems you have different formatting settings. Are you using [assertj-eclipse-formatter.xml](https://github.com/joel-costigliola/assertj-core/blob/master/src/ide-support/assertj-eclipse-formatter.xml)?

joel-costigliola:
Integrated thanks @gmullerb!

gmullerb:
My Pleasure!

PR #1626, opened 2019-10-01
Support nested maps and fields within map entry values using .extracting()

svjson:
* Fixes issue with extracting() with nested maps. (No tracker issue, to my knowledge)
* Unit tests : YES

Consider the following:
```
public class Employee {
   private Map<String, Employee> relations = new HashMap<>();
}
```

```
assertThat(employee).extracting("relations.padawan.name.first").isEqualTo(..)
```
This will break, since the current implementation only considers Maps as the first element in a nested property path. This pull request allows for getting a nested map entry value and for reaching further into nested properties of a nested map entry value.

Potential blocker for this PR could be the fact that the logic used by the extractor after the first element in the dot notated chain is located in PropertyOrFieldSupport, which may or may not be intended to deal with maps.

### DIFF
--- a/src/main/java/org/assertj/core/util/introspection/PropertyOrFieldSupport.java
+++ b/src/main/java/org/assertj/core/util/introspection/PropertyOrFieldSupport.java
@@ -17,6 +17,8 @@ import static org.assertj.core.util.Preconditions.checkArgument;
 
 import org.assertj.core.util.VisibleForTesting;
 
+import java.util.Map;
+
 public class PropertyOrFieldSupport {
   private static final String SEPARATOR = ".";
   private PropertySupport propertySupport;
@@ -54,11 +56,18 @@ public class PropertyOrFieldSupport {
       // extract next sub-property/field value until reaching the last sub-property/field
       return getValueOf(nextNameFrom(propertyOrFieldName), propertyOrFieldValue);
     }
+
     return getSimpleValue(propertyOrFieldName, input);
   }
 
   public Object getSimpleValue(String propertyOrFieldName, Object input) {
-    // first try to get given property values from objects, then try fields
+    // first check if input object is a map
+    if (input instanceof Map) {
+      Map<?, ?> map = (Map<?, ?>) input;
+      return map.get(propertyOrFieldName);
+    }
+
+    // then try to get given property values from objects, then try fields
     try {
       return propertySupport.propertyValueOf(propertyOrFieldName, Object.class, input);
     } catch (IntrospectionError propertyIntrospectionError) {
--- a/src/test/java/org/assertj/core/api/object/ObjectAssert_extracting_Test.java
+++ b/src/test/java/org/assertj/core/api/object/ObjectAssert_extracting_Test.java
@@ -38,6 +38,7 @@ public class ObjectAssert_extracting_Test {
   @BeforeEach
   public void setup() {
     luke = new Employee(2L, new Name("Luke", "Skywalker"), 26);
+    luke.setAttribute("side", "light");
   }
 
   @Test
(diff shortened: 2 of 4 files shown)


joel-costigliola:
I get the idea, the thing I wonder if that is not a better use case for `extracting(Function<? super ACTUAL, T> extractor)` where you can extract whatever you want programmatically.

Passing a String to specify a property to extract was really useful for java 7 but much less with java 8 lambdas.

Having said that I'm not against having it, just need to a bit more convincing.

jmondo:
I'm trying to test graphql results which are almost always a combination of deeply nested lists and maps and this would be tremendously useful. Using a predicate would be fine if it were a one-off but we're doing this consistently and the interface is very slick.

svjson:
> I get the idea, the thing I wonder if that is not a better use case for `extracting(Function<? super ACTUAL, T> extractor)` where you can extract whatever you want programmatically.

Well, yes, in the example I gave, I agree. I may have oversimplified it in the pull request description.

The exact case that I was working on when I ran into the need for this feature was one where a list of instances of a class is generated by the code under test, and I want to do my assertions on a few select properties of that object, where some are deeply nested, some in values of a map, and some on the top level of the class. This syntax allows me to write my assertions in an almost tabular format, which I find very descriptive of the expected results once the tests are written.

So my actual case is something more akin to:
```
assertThat(myList)
 .extracting("countryCode", "id.type", "identifierMap.typeId", "some.other.deep.field")
 .containsExactly(
   tuple(CountryCode.SE, Type.PRODUCT, "44224", "bouillabaisse"),
   tuple(CountryCode.SE, Type.CATEGORY, "72", "thingamagoops"),
   tuple(CountryCode.SE, Type CATEGORY, "2", "blubber")
   ...
);
```

To me it's a lot less noisy than using lambdas/extractors, although those definitely also have their uses.

scordio:
> Passing a String to specify a property to extract was really useful for java 7 but much less with java 8 lambdas.

I still find it useful for extracting properties with no getters, this helps the user to avoid reflection.

However, I have the feeling that we don't have a clear separation of concerns about this feature in the existing code. `AbstractObjectAssert#extracting(String)` has currently the side effect that the parameter is used as a key if actual is a `Map` and the magic is currently done by `ByNameSingleExtractor`. On the other side, no explicit `extracting` with String parameters appears in `AbstractMapAssert`, but we have `extracting(Object)` / `extracting(Object...)` (together with the new `extractingByKey`/`extractingByKeys`).

Maybe `ByNameSingleExtractor` and `ByNameMultipleExtractor` have too much complexity, we could instead delegate to the extracting methods of the corresponding assertion classes.

Coming back to this PR, if PropertyOrFieldSupport is enhanced to handle maps, does it still make sense to keep the following in `ByNameSingleExtractor`?
https://github.com/joel-costigliola/assertj-core/blob/2daea5c5235237763ac5b0fe9f639cdeeff884a7/src/main/java/org/assertj/core/extractor/ByNameSingleExtractor.java#L38-L42

joel-costigliola:
The code for extracting values should be in one place which at the moment I think is `PropertyOrFieldSupport` which should deal with map

> On the other side, no explicit extracting with string parameters appears in AbstractMapAssert, but we have extracting(Object) / extracting(Object...) (together with the new extractingByKey/extractingByKeys).

I tend to think that extracting for map assertion should be by simple key, it could be confusing to support nested key for map, if I write: `assertThat(myMap).extracting("id.type")` does it mean it is a `map.get("id.type")` or `map.get("id").getType()` ? It should be `map.get("id.type")` IMO because it is the less surprising behavior (and a good majority of maps use String as keys, it should be obvious what `extracting` means for such map assertions).

joel-costigliola:
Integrated thanks @svjson! 
@scordio I have removed the code you indicated from `ByNameSingleExtractor`.

jmondo:
i saw https://github.com/joel-costigliola/assertj-core/commit/337f99eee5cdcab9270145c47b429cbc48fb0134 how exciting! will you be cutting a release?

PR #1631, opened 2019-10-07
Improve assertion message for #shouldHaveRootCause and #shouldHaveRootCauseMessage

gagoman:
Follow up for #1624 

#### Check List:
* Fixes -
* Unit tests : YES


joel-costigliola:
Integrated thanks @gagoman!

PR #1639, opened 2019-10-15
Add isEmpty and isNotEmpty assertions for File

### LINKED ISSUE #1638: Add isEmpty and isNotEmpty assertions for File
#### Summary
I would like to contribute a couple of assertions to check whether the content of a given file is empty or not. These assertion could simply check file.length(), which returns the length of a file in bytes. At present I'm having to do something like `assertThat(file.length()).isGreaterThan(0)`

#### Example
```java
assertThat(file)
.isEmpty()
.isNotEmpty()
```

steveorourke:
#### Check List:
* Fixes #1638
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/internal/Files.java
+++ b/src/main/java/org/assertj/core/internal/Files.java
@@ -301,6 +301,18 @@ public class Files {
   }
 
   /**
+   * Asserts that the given {@code File} is not empty (i.e. size is greater than zero bytes).
+   * @throws AssertionError if the given {@code File} is {@code null}.
+   * @throws AssertionError if the given {@code File} does not exist.
+   * @throws AssertionError if the given {@code File} is empty.
+   */
+  public void assertIsNotEmptyFile(AssertionInfo info, File actual) {
+    assertIsFile(info, actual);
+    if (actual.length() > 0) return;
+    throw failures.failure(info, shouldNotBeEmpty(actual));
+  }
+
+  /**
    * Asserts that the given file can be read by the application.
    * @param info contains information about the assertion.
    * @param actual the given file.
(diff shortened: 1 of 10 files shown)


src/main/java/org/assertj/core/internal/Files.java
```diff
@@ -287,6 +288,30 @@ public void assertCanWrite(AssertionInfo info, File actual) {
...
+  /**
+   * Asserts that the given {@code File} is not empty (i.e. size is greater than zero bytes).
+   * @throws AssertionError if the given {@code File} is {@code null}.
+   * @throws AssertionError if the given {@code File} does not exist.
+   * @throws AssertionError if the given {@code File} is empty.
+   */
+  public void assertIsNotEmptyFile(AssertionInfo info, File actual) {
+    assertIsFile(info, actual);
+    if (actual.length() > 0) return;
+    throw failures.failure(info, shouldNotBeEmpty());
```
joel-costigliola: introduce `shouldNotBeEmpty*(File)` with message `"%nExpecting file <%s> not to be empty"` (add the corresponding test in `ShouldNotBeEmpty_create_Test`.

src/test/java/org/assertj/core/error/ShouldBeEmpty_create_Test.java
```diff
@@ -28,18 +28,28 @@
  * @author Alex Ruiz
  * @author Yvonne Wang
  */
-public class ShouldBeEmpty_create_Test {
+@DisplayName("ShouldBeEmpty create")
+class ShouldBeEmpty_create_Test {
```
joel-costigliola: thanks for the refactoring!

steveorourke:
Thanks for the feedback @joel-costigliola. I've updated the PR with the suggested changes.

joel-costigliola:
Integrated thanks @steveorourke!

PR #1649, opened 2019-10-20
Add element(int, InstanceOfAssertFactory) to AbstractIterableAssert

scordio:
**NOTE: this PR is currently rebased onto #1647 to (hopefully) prevent merge conflicts, so the latter should be checked first!**

#### Check List:
* Fixes N/A
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
@@ -2536,9 +2536,47 @@ public abstract class AbstractIterableAssert<SELF extends AbstractIterableAssert
    * @return the assertion on the given element
    * @throws AssertionError if the given index is out of bound.
    * @since 2.5.0 / 3.5.0
+   * @see #element(int, InstanceOfAssertFactory)
    */
   @CheckReturnValue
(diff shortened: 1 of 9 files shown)


joel-costigliola:
Integrated thanks @scordio!

FYI we can now use `AssertionsUtil.expectAssumptionViolatedException` for assumptions instead of `assertThatExceptionOfType(AssumptionViolatedException.class).isThrownBy`.

I also added the following to `should_propagate_AssertionError_from_nested_proxied_calls` which was missing (but no biggie).
```
    // the nested proxied call to isNotEmpty() throw an Assertion error that must be propagated to the caller.
    softly.assertThat(emptyList()).element(0);
    // the nested proxied call to isNotEmpty() throw an Assertion error that must be propagated to the caller.
    softly.assertThat(emptyList()).element(0, as(STRING));
```

scordio:
> FYI we can now use `AssertionsUtil.expectAssumptionViolatedException` for assumptions instead of `assertThatExceptionOfType(AssumptionViolatedException.class).isThrownBy`.

Thanks! I was planning to refactor that group of classes as one of my next activities.

PR #1655, opened 2019-10-27
Fix #1654: Make conversion script work on Windows

### LINKED ISSUE #1654: Conversion script error on Windows
The conversion shell script fails when running in the bash of Git for Windows:
`./convert-junit-assertions-to-assertj.sh: line 45: replace-in-all-files: not a valid identifier`

It is sufficient to rename the function to not contain hyphen characters.

Bananeweizen:
When executing the conversion script on Windows (in the bash shell of Git for Windows) it stops with an error:
`./convert-junit-assertions-to-assertj.sh: line 45: `replace-in-all-files': not a valid identifier`

Shortening the function name to not contain hyphen characters fixes the problem.

* Fixes #1654
* Unit tests : NA


joel-costigliola:
Integrated thanks @Bananeweizen!

PR #1664, opened 2019-10-31
Make OSGi import of jdk.* packages optional

bjhargrave:
Fixes https://github.com/joel-costigliola/assertj-core/issues/1663


joel-costigliola:
Thanks @bjhargrave!

PR #1687, opened 2019-11-03
 use bnd 4.3.0 to a) use -noclassforname instruction b) generate most up to date OSGi metadata c) add verification that additional package imports never sneak in

rotty3000:


rotty3000:
@joel-costigliola as requested here https://github.com/joel-costigliola/assertj-core/issues/1663#issuecomment-548611969 @bjhargrave and I worked out an approach to handle the OSGi metadata issues. Testing surfaced a few small issues (such as two versions of hamcrest being on your compile classpath).
I'd be more than happy to answer any of your questions about this process.

rotty3000:
To be clear, this PR contains a verification check that will catch any inadvertent dependency creep on the OSGi metadata. :)

src/main/java/org/assertj/core/configuration/ConfigurationProvider.java
```diff
@@ -27,6 +30,8 @@
  * @author Filip Hrisafov
  * @since 2.9.0 / 3.9.0
  */
+@ServiceConsumer(value = Configuration.class, resolution = OPTIONAL)
+@ServiceConsumer(value = Representation.class, resolution = OPTIONAL)
```
joel-costigliola: What problem does putting the service optional solves?

I'm asking the question because we prefer having no additional dependencies.

joel-costigliola:
Thanks for the PR @rotty3000 and @bjhargrave, I only have one comment so far.

src/main/java/org/assertj/core/configuration/ConfigurationProvider.java
rotty3000: You know what? The service loader change is frivolous and takes focus ways from the verification issue so I just remove that commit. Please take another look.

pom.xml
```diff
@@ -37,6 +37,7 @@
     <!-- Dependency versions overriding -->
     <junit-jupiter.version>5.5.2</junit-jupiter.version>
     <mockito.version>3.1.0</mockito.version>
+    <bnd.version>4.3.0</bnd.version>
```
scordio: Just a minor comment here: this should go three lines up right after `byte-buddy` as it's not overriding a dependency from `assertj-parent-pom`

pom.xml
```diff
@@ -70,6 +71,16 @@
...
       <scope>provided</scope>
       <optional>true</optional>
+      <exclusions>
+        <!--
+        This has to be excluded because it's subjecting the code to the
+        incorrect version of certain hamcrest packages.
+        -->
+        <exclusion>
+          <groupId>org.hamcrest</groupId>
+          <artifactId>hamcrest-core</artifactId>
```
scordio: Nice catch! I would also put an additional `<exclude>org.hamcrest:hamcrest-core</exclude>` line into the `maven-enforcer-plugin` configuration to avoid the legacy hamcrest artifact

pom.xml
rotty3000: will do!

pom.xml
rotty3000: I can add that.

joel-costigliola:
Integrated thanks @rotty3000!

PR #1689, opened 2019-11-05
improve .hasToString() reporting

t-8ch:
* Throw full AssertionFailedErrors instead of plain AssertionErrors.
* Unify formatting:

Before `actual` was formatted as an object, without surrounding quotes,
while the expected string was formatted as string without surrounding quotes,
which is confusing.

```
java.lang.AssertionError:
Expecting actual's toString() to return:
  <"a">
but was:
  <x>
```

By formatting the failure message based on `acutal.toString()` directly the
errormessage is consistent:

```
java.lang.AssertionError:
Expecting actual's toString() to return:
  <"a">
but was:
  <"x">
```

#### Check List:
* Fixes NA
* Unit tests : YES


joel-costigliola:
Integrated thanks @t-8ch!

PR #1702, opened 2019-11-15
Implemented isPackagePrivate into AbstractClassAssert

### LINKED ISSUE #1700: isPackagePrivate() missing from AbstractClassAssert
#### Summary
Currently `AbstractClassAssert` provides `isPublic()` and `isProtected()` methods to assert access level modifiers. I would also add `isPackagePrivate()` for completeness.

#### Current solution
```java
Condition<Class<?>> packagePrivate = new Condition<>(clazz -> {
	int mod = clazz.getModifiers();
	return !(Modifier.isPublic(mod) || Modifier.isProtected(mod) || Modifier.isPrivate(mod));
}, "package-private");

assertThat(MyClass.class).is(packagePrivate);
```

#### Improvement
```java
assertThat(MyClass.class).isPackagePrivate();
```

I was also thinking if `isPrivate()` would make sense but probably there is no real use case for it as the class should be declared inside the test class itself.

HaydenMeloche:
Implemented `isPackagePrivate()` as described in #1700.

#### Check List:
* Fixes #1700 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractClassAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractClassAssert.java
@@ -273,8 +273,8 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
    * assertThat(MyClass.class).isPackagePrivate();
    *
    * // these assertions fail:
-   * assertThat(String.class).isProtected();
-   * assertThat(Math.class).isProtected();</code></pre>
+   * assertThat(String.class).isPackagePrivate();
+   * assertThat(Math.class).isPackagePrivate();</code></pre>
    *
    * @return {@code this} assertions object
    * @throws AssertionError if {@code actual} is {@code null}.
--- a/src/main/java/org/assertj/core/api/AbstractClassAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractClassAssert.java
@@ -267,7 +267,7 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
    * Verifies that the actual {@code Class} is package-private (has no modifier).
    * <p>
    * Example:
-   * <pre><code class='java'> public MyClass { }
+   * <pre><code class='java'> class MyClass { }
    *
    * // this assertion succeeds:
    * assertThat(MyClass.class).isPackagePrivate();
--- a/src/main/java/org/assertj/core/api/AbstractClassAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractClassAssert.java
@@ -278,9 +278,9 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
    *
    * @return {@code this} assertions object
    * @throws AssertionError if {@code actual} is {@code null}.
-   * @throws AssertionError if the actual {@code Class} is not protected.
+   * @throws AssertionError if the actual {@code Class} is not package private.
    *
-   * @since 2.7.0 / 3.7.0
+   * @since 3.15.0
    */
   public SELF isPackagePrivate() {
     classes.assertIsPackagePrivate(info, actual);
--- a/src/main/java/org/assertj/core/api/AbstractClassAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractClassAssert.java
@@ -278,7 +278,7 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
    *
    * @return {@code this} assertions object
    * @throws AssertionError if {@code actual} is {@code null}.
-   * @throws AssertionError if the actual {@code Class} is not package private.
+   * @throws AssertionError if the actual {@code Class} is not package-private.
    *
    * @since 3.15.0
    */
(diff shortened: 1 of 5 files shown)


src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -263,6 +263,30 @@ public SELF isProtected() {
...
+   *
+   * // these assertions fail:
+   * assertThat(String.class).isProtected();
+   * assertThat(Math.class).isProtected();</code></pre>
+   *
+   * @return {@code this} assertions object
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} is not protected.
+   *
+   * @since 2.7.0 / 3.7.0
```
HaydenMeloche: What should this be?

joel-costigliola: 3.15.0

src/main/java/org/assertj/core/internal/Classes.java
```diff
@@ -206,6 +207,23 @@ public void assertIsProtected(AssertionInfo info, Class<?> actual) {
...
+   * @param info contains information about the assertion.
+   * @param actual the "actual" {@code Class}.
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} is not protected.
+   */
+  public void assertIsPackagePrivate(AssertionInfo info, Class<?> actual) {
+    assertNotNull(info, actual);
+    if ((Modifier.isPublic(actual.getModifiers())
+      || Modifier.isProtected(actual.getModifiers())
+      || Modifier.isPrivate(actual.getModifiers()))) {
```
joel-costigliola: extract `actual.getModifiers()` to a local variable.

src/main/java/org/assertj/core/internal/Classes.java
```diff
@@ -206,6 +207,23 @@ public void assertIsProtected(AssertionInfo info, Class<?> actual) {
...
+   * Verifies that the actual {@code Class} is package-private.
+   *
+   * @param info contains information about the assertion.
+   * @param actual the "actual" {@code Class}.
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} is not protected.
+   */
+  public void assertIsPackagePrivate(AssertionInfo info, Class<?> actual) {
+    assertNotNull(info, actual);
+    if ((Modifier.isPublic(actual.getModifiers())
```
joel-costigliola: I believe we can remove one pair of `()` here

src/test/java/org/assertj/core/internal/classes/Classes_assertIsPackagePrivate_Test.java
```diff
@@ -0,0 +1,47 @@
...
+
+  @Test
+  public void should_pass_if_actual_is_a_package_private_class() {
+    classes.assertIsPackagePrivate(someInfo(), PackagePrivateClass.class);
+  }
+
+  @Test
+  public void should_fail_if_actual_is_not_a_package_private_class() {
+    assertThatExceptionOfType(AssertionError.class).isThrownBy(() -> classes.assertIsPackagePrivate(someInfo(), Object.class))
+                                                   .withMessage(shouldBePackagePrivate(Object.class).create());
```
joel-costigliola: more a style thingy but we prefer using `AssertionsUtil.expectAssertionError` as shown https://github.com/joel-costigliola/assertj-core/blob/master/CONTRIBUTING.md:
```java
  @Test
  void should_fail_if_optional_is_empty() {
    // GIVEN
    Optional<Object> actual = Optional.empty();
    // WHEN
    AssertionError assertionError = expectAssertionError(() -> assertThat(actual).containsInstanceOf(Object.class));
    // THEN
    then(assertionError).hasMessage(shouldBePresent(actual).create());
  }
```

HaydenMeloche:
Thanks for your review @joel-costigliola & @scordio. Mind checking again? 😄

src/main/java/org/assertj/core/internal/Classes.java
```diff
@@ -206,6 +207,22 @@ public void assertIsProtected(AssertionInfo info, Class<?> actual) {
     }
   }
 
+  /**
+   * Verifies that the actual {@code Class} is package-private.
+   *
+   * @param info contains information about the assertion.
+   * @param actual the "actual" {@code Class}.
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} is not protected.
```
scordio: Whoops I missed this before 🤦‍♂ it should be `package-private` instead of `protected`. We can adjust it during the merge

src/test/java/org/assertj/core/api/classes/ClassAssert_isPackagePrivate_Test.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.assertj.core.api.classes;
+
+import org.assertj.core.api.ClassAssert;
+import org.assertj.core.api.ClassAssertBaseTest;
+
+import static org.mockito.Mockito.verify;
+
+/**
+ * Tests for <code>{@link ClassAssert#isPackagePrivate()} ()} ()}</code>.
```
scordio: Probably ` ()} ()}` can be removed

joel-costigliola:
@scordio I leave this one to you, thanks!

scordio:
Integrated, thanks @HaydenMeloche!

PR #1703, opened 2019-11-17
Setup GitHub actions

filiphr:
This is an initial go of setting up GitHub actions for AssertJ.

I added https://github.com/sormuras/bach/blob/master/install-jdk.sh so we can keep using testing with OpenJDK. The GitHub actions have a setup for JDK Azul.

The EA build is currently not setup to continue on error true. The reason for that is that with GitHub actions even if something fails the job would be marked as green, only way to see that it failed is to check the logs.

I also added MacOS and Windows builds only for Java 11.

The Windows build and EA are failing. I can try to fix them both if you like the approach (I don't have Windows, but I could perhaps figure it out from the test failures 😄).

I updated the Maven version to 3.6.1 so I can use `--no-show-progress`, this removes the downloading progress from the console, we could alternatively use `-B`, with that there would be no coloring in the console.

I also configured the sonar cloud to only be done in this repo and not on others. You would probably need to add a `secrets.SONAR_TOKEN`. Having said this now it seems like this would be printed on the console so we need to resolve it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,49 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  test_linux:
+    name: Linux (OpenJDK ${{ matrix.java }})
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [11, 13, ea]
+    steps:
+      - uses: actions/checkout@master
+      # If installing the JDK is in a different step then wrong java would be used
+      - name: Install JDK and Build Project
+        run: |
+          . ./.github/scripts/install-jdk.sh --feature ${{ matrix.java }} --os linux-x64
+          ./mvnw -V --no-transfer-progress clean verify javadoc:javadoc
+
+  test_os:
+    name: OS ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macOS-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@master
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Test
+        run: ./mvnw -V --no-transfer-progress clean verify javadoc:javadoc
+
+  sonar:
+    name: Sonar Cloud
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


scordio:
About the JDK I think we could switch to Azul, or do you see something against it?

filiphr:
I have no problem with the ones from Azul. However, they don't provide EA JDK. For windows, macOS and the sonar build the Zuul JDK is used. If we want to keep testing against EA we'll need to use instal-jdk.sh.

.github/workflows/main.yml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push' && github.ref == 'master' && github.repository == 'joel-costigliola/assertj-core'
+    steps:
+      - uses: actions/checkout@master
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Test
+        run: ./mvnw -V --no-transfer-progress clean org.jacoco:jacoco-maven-plugin:prepare-agent install javadoc:javadoc sonar:sonar -Dsonar.login=${{ secrets.SONAR_TOKEN }}
```
joel-costigliola: note to myself ;-) https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets

filiphr:
I think that the Windows tests would be fixed with https://github.com/joel-costigliola/assertj-core/pull/1724 :)

joel-costigliola:
This issue is more to get noticed of future windows only problems instead of relying on our dear community to fix them for us (which we appreciate!). 

I think we will only do a simple build for windows as we have re-enabled sonar in the current travis build.

joel-costigliola:
Thanks @filiphr , I have integrated the PR for windows and macos only, the windows build fails when setting up java. will look into that later.

scordio:
> the windows build fails when setting up java.

It looks like it was just a temporary connection timeout, now it runs fine and the issue reported in #1724 appeared.

filiphr:
>Thanks @filiphr , I have integrated the PR for windows and macos only, the windows build fails when setting up java. will look into that later.

No problem, glad that I could help.

Btw why only the windows and macos? Don't you want to migrate everything to one single CI?

joel-costigliola:
We might at some point, at the moment travis is working fine for us s no need.

PR #1712, opened 2019-11-29
Fix typo in javadoc

gyumin-kim:
#### Check List:
* Removes unmeaningful whitespace in javadoc.
* Unit tests : NA


joel-costigliola:
Integrated thanks @gyumin-kim!

Issue #1715, opened 2019-12-01
Floating point comparisons consider positive zero greater than negative zero


carterkozak:
#### Summary

StandardComparisonStrategy compares unboxed floating point values in an unexpected and arguably incorrect way around positive and negative zero.

#### Example

While refactoring tests to take advantage of assertj fluent assertions for better failure messages, I found a case that fails after the migration in some cases:

```diff
- assertThat( floatA >= floatB).isTrue();
+ assertThat(floatA).isGreaterThanOrEqualTo(floatB);
```
The new code fails when floatA is `-0f` and be is `0f` due to natural ordering based on Comparable, but is unexpected in this case.

carterkozak:
I've provided a PR with four failing unit tests: #1716

joel-costigliola:
Thanks for reporting this @carterkozak!

joel-costigliola:
Indeed the behavior comes from the fact that doubles are compared according to https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Double.html#compareTo(java.lang.Double) which:

> Compares two Double objects numerically. There are two ways in which comparisons performed by this method differ from those performed by the Java language numerical comparison operators (<, <=, ==, >=, >) when applied to primitive double values:
> -   Double.NaN is considered by this method to be equal to itself and greater than all other double values (including Double.POSITIVE_INFINITY).
> -   0.0d is considered by this method to be greater than -0.0d.  

I think it is fair in that case to follow equals behavior and consider 0.0d to be equal to -0.0d,  to follow the behavior less surprising for most users.

Users who want to use the `compareTo` semantics can use `isEqualByComparingTo`.

We will check the behavior for other primitive types and make similar adjustment.

carterkozak:
Thanks @joel-costigliola! jls-5.6.2 may be relevant here as well, depending how we go about implementing the solution https://docs.oracle.com/javase/specs/jls/se11/html/jls-5.html#jls-5.6.2

joel-costigliola:
My bad, I actually was wrong about my root cause analysis, AssertJ does not use `compareTo` in this case but [`Double.equals`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Double.html#equals(java.lang.Object)) which states:

> If d1 represents +0.0 while d2 represents -0.0, or vice versa, the equal test has the value false, even though +0.0==-0.0 has the value true. 

This explains the observed behavior.

[...]

PR #1716, opened 2019-12-01
Tests to reproduce #1715

carterkozak:
Floating point comparisons consider positive zero greater than negative
zero.

Note this PR only provides failing unit tests, it does not fix the bug.

#### Check List:
* Fixes #???
* Unit tests : YES


joel-costigliola:
Fixed in 71ae7a4c1440b8b49c8cc55555fc789cf3383047

carterkozak:
Thanks!

PR #1711, opened 2019-11-27
Revise module declaration

sormuras:
Relates to #1656 -- here are the changes that made AssertJ Core compile with Java 9+, including the module descriptor.

- Added the "base" source root `/src/main/java` to the compilation set of the `java9+` profile.
- Encountered a lot of "invisible package" errors...
- ...and added a bunch of `requires ...;` statements to fix those. Left some comments in the `module-info.java` file.

Open ends:
- [ ] Delete/ignore all Java-9+ compiled classes, except for `module-info.class`
- [x] `java.instrument` is a special snow flake, it's not required directly to compile AssertJ Core, but ByteBuddy seems to need it. @raphw ... any new input here? (removed in 116798b15518f16597e8754b83c06cd3b89ebe24)
- [x] Module `java.sql` is only(?) needed due to a link reference in the javadoc of `Assertions.java` (not needed since 5c136d6268a6e0c375c6d6ae266c6a6e2b2b0c6d)
- [x] Module name `junit` is file named based -- JUnit 4.13 will fix this, though. (fixed in b0385fcf6459bb1a5ddb1c2783dc7dc6c054df30)
- [x] Module name `jsr305` is file name based -- don't see a fix in the near future, do you? (not needed since 24176f5450b8e3624a819fdc594f709a378ca09f)

TODO:
- [ ] Provide integration test that uses module `org.assertj.core` on the module path.
- [ ] Run `jdeps --check ...`, similar to https://github.com/junit-team/junit5/issues/1939

Issue #1656, opened 2019-10-27
AbstractThrowableAssert#isInstanceOf() throws IllegalAccessError


leadpony:
#### Summary

`AbstractThrowableAssert#isInstanceOf()` throws an `IllegalAccessError` exception 
when assertj-core is used as a modular jar. 

#### Example

```java
Throwable thrown = new IllegalArgumentException();
assertThat(thrown).isInstanceOf(NullPointerException.class);
```
Building and runing this example with JDK 9 or higher:
```bash
mvn package
java -p target/modules -m example/example.Example
```
outputs the following error messages:
```
Exception in thread "main" java.lang.IllegalAccessError: class org.assertj.core.util.Closeables (in module org.assertj.core) cannot access class java.util.logging.Logger (in module java.logging) because module org.assertj.core does not read module java.logging
	at org.assertj.core@3.14.0-SNAPSHOT/org.assertj.core.util.Closeables.<clinit>(Closeables.java:25)
	at org.assertj.core@3.14.0-SNAPSHOT/org.assertj.core.util.Throwables.getStackTrace(Throwables.java:172)
	at org.assertj.core@3.14.0-SNAPSHOT/org.assertj.core.error.ShouldBeInstance.<init>(ShouldBeInstance.java:67)
	at org.assertj.core@3.14.0-SNAPSHOT/org.assertj.core.error.ShouldBeInstance.shouldBeInstance(ShouldBeInstance.java:34)
	at org.assertj.core@3.14.0-SNAPSHOT/org.assertj.core.internal.Objects.assertIsInstanceOf(Objects.java:136)
	at org.assertj.core@3.14.0-SNAPSHOT/org.assertj.core.api.AbstractAssert.isInstanceOf(AbstractAssert.java:382)
	at example@1.0.0-SNAPSHOT/example.Example.main(Example.java:9)
```
The problem happens because `org.assertj.core.util.Closeables` requires `java.logging` module both at compile-time and at runtime.
Adding the following line to `src/main/java9/module-info.java` in asssertj-core repository will resolve the problem.
```java
requires java.logging;
```
Please find the attached project which reproduces the problem.
[assertj-example.zip](https://github.com/joel-costigliola/assertj-core/files/3775868/assertj-example.zip)

Thank you.

joel-costigliola:
Thanks for reporting this @leadpony!

scordio:
[...]

PR #1717, opened 2019-12-01
Add modular integration test

scordio:
Related to #1711, this introduces a skeleton for modular integration testing based on the `maven-invoker-plugin`. As an initial test case I chose the issue reported in #1656: it fails correctly when the test runs in the IDE, but it doesn't fail when executed by the invoker plugin.

Open points:
- [ ] Understand why `issue_1656()` doesn't fail under `maven-invoker-plugin`
- [ ] Allow the IDE to automatically detect the project generated by the plugin, maybe `build-helper-maven-plugin`? (for now I manually added `assertj-core/target/it/modular-tests` to the project modules)
- [ ] Replace AssertJ 3.14.0 version with `@project.version@` in the integration test POM (the IDE is currently not able to detect the AssertJ module with the parameterized version and the test build fails -- I still need to try if the changes in #1711 can help)
 
@sormuras please check it! This follows more your [Testing In The Modular World](https://sormuras.github.io/blog/2018-09-11-testing-in-the-modular-world.html) rather than the [mainrunner](https://github.com/sormuras/mainrunner)+[Bach](https://github.com/sormuras/bach) combo you suggested in #1711, I need a bit more time to understand how the latter works 😅

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -545,6 +545,31 @@
             </plugin>
           </plugins>
         </pluginManagement>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-invoker-plugin</artifactId>
+            <version>3.2.1</version>
+            <configuration>
+              <cloneProjectsTo>${project.build.directory}/it</cloneProjectsTo>
+              <settingsFile>src/it/settings.xml</settingsFile>
+              <localRepositoryPath>${project.build.directory}/local-repo</localRepositoryPath>
+              <goals>
+                <goal>clean</goal>
+                <goal>test</goal>
+              </goals>
+            </configuration>
+            <executions>
+              <execution>
+                <id>integration-test</id>
+                <goals>
+                  <goal>install</goal>
+                  <goal>run</goal>
+                </goals>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
       </build>
     </profile>
     <profile>
--- a/src/it/modular-tests/pom.xml
+++ b/src/it/modular-tests/pom.xml
@@ -1,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>it</groupId>
+  <artifactId>modular-tests</artifactId>
+  <version>0</version>
+
+  <properties>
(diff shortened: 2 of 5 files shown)


sormuras:
Looks good to me.

Including the fact, that it doesn't fail using `invoker`: 

1. your root module is `black.box`
2. it `requires org.junit.jupiter.api;`
3. which in turn`requires transitive org.junit.platform.commons;`
4. `requires java.logging;` is declared in `org.junit.platform.commons`.

Thus, `java.logging` is added to the module graph. All is fine.

sormuras:
Having said that, as `org.assertj.core` already refers to types declared in `java.logging`, it must state it via a `requires java.logging;` statement ... and not rely on 3rd-party modules to trigger its resolution.

scordio:
> Looks good to me.
> 
> Including the fact, that it doesn't fail using `invoker`:
> 
>     1. your root module is `black.box`
> 
>     2. it `requires org.junit.jupiter.api;`
> 
>     3. which in turn`requires transitive org.junit.platform.commons;`
> 
>     4. `requires java.logging;` is declared in `org.junit.platform.commons`.
> 
> 
> Thus, `java.logging` is added to the module graph. All is fine.

Then I don't understand what's wrong when I run the test in IntelliJ... need to dig a bit more.

I didn't realize the junit module could hide the `java.logging` case (still learning modules), probably these are corner cases for the modules used by junit. But then having the `issue_1656()` test seems to be pointless.

> Having said that, as `org.assertj.core` already refers to types declared in `java.logging`, it must state it via a `requires java.logging;` statement ... and not rely on 3rd-party modules to trigger its resolution.

If we apply your changes in #1711, especially `<compileSourceRoot>${project.basedir}/src/main/java</compileSourceRoot>`, we should fail during build phase, right? Probably this would be the right way to spot issues similar to #1656.

sormuras:
> Then I don't understand what's wrong when I run the test in IntelliJ... need to dig a bit more.

Your version of IJ doesn't run your tests on the module path then. With 2019.3 on, it should work.

Try to add `--show-module-resolution` to your (Maven, Gradle, IDE, ...) `java` call that is running tests. 

> If we apply your changes in #1711, especially `<compileSourceRoot>${project.basedir}/src/main/java</compileSourceRoot>`, we should fail during build phase, right? 

Aye. Yes, all core types are being compiled twice then, but that only takes some ... as can be seen here: https://travis-ci.org/joel-costigliola/assertj-core/jobs/617590898#L260 ... and it emits some expected warning regarding module name being based on their file names. Which can be muted, iirc.

sormuras:
Oh noes ... just (re)discovered this PR, after working on and pushing #1778 ... 🙈

scordio:
> Oh noes ... just (re)discovered this PR, after working on and pushing #1778 ... 🙈

No worries, the ByteBuddy issue could be the motivation to wrap this topic up, it doesn't matter in which PR 🙂

sormuras:
I'll finish this one, soon. Let's compare them later.

PR #1720, opened 2019-12-03
Improve perf of containsOnly() assertion on very large collections

### LINKED ISSUE #1718: StandardComparisonStrategy.iterableRemoves() inefficient on ArrayList
#### Summary
```
iterableRemoves:102, StandardComparisonStrategy (org.assertj.core.internal)
iterablesRemove:357, Iterables (org.assertj.core.internal)
assertContainsOnly:383, Iterables (org.assertj.core.internal)
containsOnly:354, AbstractIterableAssert (org.assertj.core.api)
containsOnly:258, ListAssert (org.assertj.core.api)
```
This stack works very slow when the list checked in `containsOnly()` has ~ 1 m elements, because of O(N^2) complexity.

leventov:
#### Check List:
* Fixes #1718 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/internal/ComparatorBasedComparisonStrategy.java
+++ b/src/main/java/org/assertj/core/internal/ComparatorBasedComparisonStrategy.java
@@ -98,11 +98,12 @@ public class ComparatorBasedComparisonStrategy extends AbstractComparisonStrateg
     // Avoid O(N^2) complexity of serial removal from an iterator of collections like ArrayList
     if (iterable instanceof Collection) {
       ((Collection<?>) iterable).removeIf(o -> comparator.compare(o, value) == 0);
-    }
-    Iterator<?> iterator = iterable.iterator();
-    while (iterator.hasNext()) {
-      if (comparator.compare(iterator.next(), value) == 0) {
-        iterator.remove();
+    } else {
+      Iterator<?> iterator = iterable.iterator();
+      while (iterator.hasNext()) {
+        if (comparator.compare(iterator.next(), value) == 0) {
+          iterator.remove();
+        }
       }
     }
   }
--- a/src/main/java/org/assertj/core/internal/ComparatorBasedComparisonStrategy.java
+++ b/src/main/java/org/assertj/core/internal/ComparatorBasedComparisonStrategy.java
@@ -15,6 +15,7 @@ package org.assertj.core.internal;
 import static org.assertj.core.configuration.ConfigurationProvider.CONFIGURATION_PROVIDER;
 import static org.assertj.core.util.IterableUtil.isNullOrEmpty;
 
+import java.util.Collection;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.Set;
@@ -94,6 +95,10 @@ public class ComparatorBasedComparisonStrategy extends AbstractComparisonStrateg
   @SuppressWarnings("unchecked")
   public void iterableRemoves(Iterable<?> iterable, Object value) {
     if (iterable == null) return;
+    // Avoid O(N^2) complexity of serial removal from an iterator of collections like ArrayList
+    if (iterable instanceof Collection) {
+      ((Collection<?>) iterable).removeIf(o -> comparator.compare(o, value) == 0);
+    }
     Iterator<?> iterator = iterable.iterator();
     while (iterator.hasNext()) {
       if (comparator.compare(iterator.next(), value) == 0) {
--- a/src/test/java/org/assertj/core/perf/ContainsOnlyPerfTest.java
+++ b/src/test/java/org/assertj/core/perf/ContainsOnlyPerfTest.java
@@ -1,0 +1,44 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2019 the original author or authors.
+ */
+package org.assertj.core.perf;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.ArrayList;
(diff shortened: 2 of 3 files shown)


joel-costigliola:
Thanks, I will review your PR shortly!

src/main/java/org/assertj/core/internal/ComparatorBasedComparisonStrategy.java
```diff
@@ -94,6 +95,10 @@ public boolean iterableContains(Iterable<?> iterable, Object value) {
   @SuppressWarnings("unchecked")
   public void iterableRemoves(Iterable<?> iterable, Object value) {
     if (iterable == null) return;
+    // Avoid O(N^2) complexity of serial removal from an iterator of collections like ArrayList
```
joel-costigliola: Unless I'm missing something the original removal is O(n) since it is just iterating over the elements. do you agree?

src/main/java/org/assertj/core/internal/ComparatorBasedComparisonStrategy.java
```diff
@@ -94,6 +95,10 @@ public boolean iterableContains(Iterable<?> iterable, Object value) {
...
+    if (iterable instanceof Collection) {
+      ((Collection<?>) iterable).removeIf(o -> comparator.compare(o, value) == 0);
```
joel-costigliola: the javadoc of `removeIf` states:
> The default implementation traverses all elements of the collection using its iterator. Each matching element is removed using Iterator.remove(). If the collection's iterator does not support removal then an UnsupportedOperationException will be thrown on the first matching element.

this is what we do here below so before accepting these changes I would like to see before and after performance measurements that would prove that performance was indeed improved (not as an automated test, just as a few perf run).

src/test/java/org/assertj/core/perf/ContainsOnlyPerfTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.ArrayList;
+import java.util.concurrent.ThreadLocalRandom;
+
+import org.junit.Test;
+
+/**
+ * See https://github.com/joel-costigliola/assertj-core/issues/1718
+ */
+public class ContainsOnlyPerfTest {
```
joel-costigliola: Unfortunately this can't be an automated unit test, I don't want this test to fail only because it was run on a slow machine.

we can keep that class though but without the test part, just a good old main method (or simply disable the tests)
if we keep the tests then I would prefer them to be written with JUnit 5.

src/main/java/org/assertj/core/internal/ComparatorBasedComparisonStrategy.java
leventov: Each removal is O(N), and there are O(N) of them

src/main/java/org/assertj/core/internal/ComparatorBasedComparisonStrategy.java
leventov: Please check out `ArrayList.removeIf()` implementation, it overrides the default impl.

You can try to comment out the changes in the Strategy and run the `ContainsOnlyPerfTest` without a time limit. I don't how much it takes to complete. I don't have enough patience, after waiting for more than a few dozen seconds. Its complexity is O(1000 billion).

joel-costigliola: I have run `ContainsOnlyPerfTest` with and without the changes, it is indeed much faster, I still fail to see where to is the O(n2), enlighten me please ;-)

src/test/java/org/assertj/core/perf/ContainsOnlyPerfTest.java
leventov: To fail on a slow machine, each operation should take ~1 microsecond = 1000 nanoseconds, whereas the "operation" is some cheap manipulation with Array, with memory which resides caches. I don't believe there is such a test agent where that would fail, perhaps apart from some ultra-micro-embedded machine, in conjunction with JVM running in interactive (no compilation) mode.

OTOH, O(N^2) complexity pushes the time of this test far beyond 1 second, on a however powerful machine. Therefore I think this is a good guard against accidental regression to an O(N^2) complexity.

I could change this timeout to 10 seconds, which would still absolutely surely fail on any machine if the complexity was O(N^2).

joel-costigliola: happy with 10s + adding a comment explaining that.

src/main/java/org/assertj/core/internal/ComparatorBasedComparisonStrategy.java
leventov: `ArrayList$Iterator.remove()`, while at the element at index 0, should shift all elements until the end (say there is 1 million of them) by one array slot:

https://github.com/openjdk/jdk/blob/36bd09dead31636e71ab9d74edff130cbd5124a4/src/java.base/share/classes/java/util/ArrayList.java#L636-L642

`System.arrayCopy()` is an O(N) operation here

joel-costigliola: ok I got it I think in `ArrayList.removeIf`:
```
        // Tolerate predicates that reentrantly access the collection for
        // read (but writers still get CME), so traverse once to find
        // elements to delete, a second pass to physically expunge.
```

joel-costigliola:
thanks @leventov, so sum up only two things to do:

- migrate the test to JUnit 5
- use 10s timeout and comment why it is a good threshlod

leventov:
I think timeout = 10 seconds may be actually flaky on some modern machines, because `System.arraycopy` may be very fast. So I stopped at 5 seconds.

src/main/java/org/assertj/core/internal/ComparatorBasedComparisonStrategy.java
```diff
@@ -94,6 +95,10 @@ public boolean iterableContains(Iterable<?> iterable, Object value) {
...
+    }
     Iterator<?> iterator = iterable.iterator();
     while (iterator.hasNext()) {
```
filiphr: One side note from me. I think that this should be in an `else` statement. Otherwise it would traverse the collection again. This is not really needed, since `removeIf` handles everything. Same for the changes in the `StandardComparisonStrategy`

leventov: Thanks! Fixed.

joel-costigliola:
Integrated thanks @leventov!

PR #1723, opened 2019-12-11
Remove duplication for Descriptable implementations using default methods

kriegfrj:
Just an idea I had. A precursor for a bigger idea.

### DIFF
--- a/src/main/java/org/assertj/core/api/Descriptable.java
+++ b/src/main/java/org/assertj/core/api/Descriptable.java
@@ -13,6 +13,7 @@
 package org.assertj.core.api;
 
 import org.assertj.core.description.Description;
+import org.assertj.core.description.TextDescription;
 
 /**
  * An object that has a description.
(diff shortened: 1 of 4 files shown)


src/main/java/org/assertj/core/api/Descriptable.java
```diff
@@ -23,7 +24,7 @@
  * @author Yvonne Wang
  * @author Mikhail Mazursky
  */
-public interface Descriptable<SELF> {
+public interface Descriptable<SELF extends Descriptable<SELF>> {
```
filiphr: Just FYI this change is not a backwards compatible change. Why is it needed to force that `SELF` extends from `Descriptable`?

kriegfrj: This is true. However, it's not really a "SELF" type if it doesn't extend itself? I think it was probably an oversight when the type was originally created - I can't think of when you'd have a use case that SELF is not an instance of `Descriptable`.

On the other hand, it is not essential to the main purpose of this PR, so I can leave this change out of this PR for further discussion. The rest of the PR should (hopefully!) be less controversial.

joel-costigliola: Let's take the safest approach and not extend from Descriptable.

kriegfrj: Latest push has removed the constraint on `SELF`.

joel-costigliola:
Fixed in 618609cc0a05414ef1a117cf616621310cd062ee, thanks @kriegfrj.

PR #1724, opened 2019-12-11
Fix tests that fail on Windows

kriegfrj:
These tests made assumptions about paths that don't apply when running on Windows. `new File("/test.txt")` produced `C:\test.txt` in the error message output, while the test was expecting `/test.txt`.

Updated it to call `getAbsolutePath()` to generate the correct expected value in a platform-inspecific way.


joel-costigliola:
Integrated thanks @kriegfrj! We now build assertj on windows (see https://github.com/joel-costigliola/assertj-core/runs/348372175) to catch this kind of mistakes.

Issue #1355, opened 2018-11-08
Describe actual Throwables when ThrowableAssert fails


jakubzytka:
### Summary
When some `Throwable`-related assertion fails the actual `Throwable` that failed the check is not described beyond what actually caused assertion failure (e.g. message mismatch). In non-trivial code when unexpected `Exception`s are thrown this makes debugging inconvenient.

Dumping more comprehensive (class, message, stack, perhaps even recursive info about causes) info about the actual `Throwable` could ease assertion failure investigation.

Relevant message factories (e.g. `ShouldHaveMessage`) already have access to actual `Throwable`.

It may be tempting to use `util.Throwables.ERROR_DESCRIPTION_EXTRACTOR` to describe throwables, but a more powerful mechanism would be appreciated. 
Providing precise information in case of assertion failure doesn't seem to be overkill (unlike as in #864)

### Error message to improve

- [x] ShouldHaveCause #2872
- [ ] ShouldHaveCauseExactlyInstance
- [ ] ShouldHaveCauseInstance
- [ ] ShouldHaveCauseReference - @weiyilei
- [ ] ShouldHaveNoCause
- [ ] ShouldHaveNoSuppressedExceptions
- [x] ShouldHaveRootCauseExactlyInstance #2910
- [x] ShouldHaveRootCauseInstance #2910
- [ ] ShouldHaveSuppressedException
- [ ] ShouldNotContainCharSequence 
- [ ] ShouldStartWith
- [ ] ShouldEndWith

If you are interested in contributing to this issue, select one error message, it will be assigned to you.

joel-costigliola:
do you mean printing the stack trace ?

jakubzytka:
Example of my ideal message:
```java
Expecting Throwable message:
<"expected exception message">
but was:
<"some other unexpected exception">

Throwable that failed the check:
Exception in thread "main" java.lang.RuntimeException: some other unexpected exception
	at Test.functionThatIsSupposedToThrowExpectedException(Test.java:119)
	at Test.main(Test.java:69)
Caused by: java.lang.RuntimeException: unexpected exception in recurseToGenerateSomeStack
	at Test.recurseToGenerateSomeStack(Test.java:133)
	at Test.recurseToGenerateSomeStack(Test.java:138)
	at Test.recurseToGenerateSomeStack(Test.java:138)
	at Test.recurseToGenerateSomeStack(Test.java:138)
	at Test.recurseToGenerateSomeStack(Test.java:138)
[...]

PR #1725, opened 2019-12-11
Add additional stacktrace information for throwable assertions

benoitdupont:
Show the stack trace for the Throwable under test when assertions for Throwable hasMessageContaining and hasMessageContainingAll fails

#### Check List:
* Fixes #1355
* Unit tests : YES 

### DIFF
--- a/src/main/java/org/assertj/core/error/ShouldContainCharSequence.java
+++ b/src/main/java/org/assertj/core/error/ShouldContainCharSequence.java
@@ -12,6 +12,9 @@
  */
 package org.assertj.core.error;
 
+import static org.assertj.core.util.Strings.escapePercent;
+import static org.assertj.core.util.Throwables.getStackTrace;
+
 import java.util.Set;
 
 import org.assertj.core.internal.ComparisonStrategy;
@@ -37,6 +40,36 @@ public class ShouldContainCharSequence extends BasicErrorMessageFactory {
   public static ErrorMessageFactory shouldContain(CharSequence actual, CharSequence sequence) {
     return new ShouldContainCharSequence("%nExpecting:%n <%s>%nto contain:%n <%s> %s", actual, sequence,
                                    StandardComparisonStrategy.instance());
+  }
+
+  public static ErrorMessageFactory shouldContain(Throwable actual, CharSequence sequence) {
+    String format = "%n" +
+                    "Expecting:%n" +
+                    "  <%s>%n" +
+                    "to contain:%n" +
+                    "  <%s>%n" +
+                    "but did not.%n" +
+                    "%n" +
+                    "Throwable that failed the check:%n" +
+                    "%n" + escapePercent(getStackTrace(actual)); // to avoid AssertJ default String formatting
+
+    return new ShouldContainCharSequence(format, actual.getMessage(), sequence, StandardComparisonStrategy.instance());
+  }
+
+  public static ErrorMessageFactory shouldContain(Throwable actual, CharSequence[] sequence,
+                                                  Set<? extends CharSequence> notFound) {
+    String format = "%n" +
+                    "Expecting:%n" +
+                    "  <%s>%n" +
+                    "to contain:%n" +
+                    "  <%s>%n" +
+                    "but could not find:%n" +
+                    "  <%s>%n" +
+                    "%n" +
+                    "Throwable that failed the check:%n" +
+                    "%n" + escapePercent(getStackTrace(actual)); // to avoid AssertJ default String formatting
+
+    return new ShouldContainCharSequence(format, actual.getMessage(), sequence, notFound, StandardComparisonStrategy.instance());
   }
 
   /**
--- a/src/main/java/org/assertj/core/error/ShouldContainCharSequence.java
+++ b/src/main/java/org/assertj/core/error/ShouldContainCharSequence.java
@@ -44,7 +44,7 @@ public class ShouldContainCharSequence extends BasicErrorMessageFactory {
 
   public static ErrorMessageFactory shouldContain(Throwable actual, CharSequence sequence) {
     String format = "%n" +
-                    "Expecting:%n" +
+                    "Expecting throwable message:%n" +
                     "  <%s>%n" +
                     "to contain:%n" +
                     "  <%s>%n" +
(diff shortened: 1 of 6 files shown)


benoitdupont:
Hello, this is my first Open Source contribution, any feedback welcome !

Here are some questions and remarks to be considered before merging (maybe I should have done a draft PR instead ?):

* I did not add specific methods for theses signatures, as they are not used in the caller code:
  * `ShouldContainCharSequence.shouldContains(Throwable actual, CharSequence sequence, ComparisonStrategy comparisonStrategy)`
  * `ShouldContainCharSequence.shouldContains(Throwable actual, CharSequence sequence, Set<? extends CharSequence> notFound, ComparisonStrategy comparisonStrategy)`

* I don't know if you want to extract the added logic for `Throwable` from `ShouldContainCharSequence` in a specific class, that delegate then to `ShouldContainCharSequence` which is String specific, or if a small mix of Throwable and String concepts is acceptable in the same class.

* I added `ShouldContainThrowable_create_Test#should_create_error_message_with_escaping_percent` but not `ShouldContainThrowable_create_Test#should_create_error_message` without the escaping percent, I don't think it's pertinent to have both, unless ?

* `assertHasStackTraceContaining` still need to be adapted, it calls `ShouldContainCharSequence.shouldContains`, but there's some investigation needed as it doesn't work as-is.

* Can I migrate `Throwables_assertHasMessageContaining_Test.java` and `Throwables_assertHasMessageContainingAll_Test.java` (because I modified them for the original issue) to junit5 by reducing visibility of class and methods from public to package private in a separate commit, or we should avoid to do multiples things in a single PR ?

joel-costigliola:
Thanks @benoitdupont , I will review your PR shortly!

src/test/java/org/assertj/core/error/ShouldContainThrowable_create_Test.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+@DisplayName("ShouldContainThrowable create")
+class ShouldContainThrowable_create_Test {
+  @Test
+  void should_create_error_message_with_escaping_percent() {
+    // GIVEN
+    RuntimeException actual = new RuntimeException("You know nothing %");
+    // WHEN
+    String errorMessage = ShouldContainCharSequence.shouldContain(actual, "You know nothing % Jon Snow")
+                                                   .create(new TestDescription("TEST"));
```
joel-costigliola: use static import om `shouldContain` to make the code a bit lighter

src/test/java/org/assertj/core/error/ShouldContainThrowable_create_Test.java
```diff
@@ -0,0 +1,75 @@
...
+    // THEN
+    assertThat(errorMessage).isEqualTo("[TEST] %n" +
```
joel-costigliola: Replace `assertThat` by `then` from `BDDAssertions`.

src/test/java/org/assertj/core/error/ShouldContainThrowable_create_Test.java
```diff
@@ -0,0 +1,75 @@
...
+                                       getStackTrace(actual));
+  }
+
+  @Test
+  void should_create_error_message_with_several_values_not_found() {
+    // GIVEN
+    RuntimeException actual = new RuntimeException("You know nothing");
+    // WHEN
+    String errorMessage = ShouldContainCharSequence.shouldContain(actual, array("You", "know", "nothing", "Jon", "Snow"),
+                                                                  newSet("Jon", "Snow"))
```
joel-costigliola: good test data :smile:

src/main/java/org/assertj/core/error/ShouldContainCharSequence.java
```diff
@@ -39,6 +42,36 @@ public static ErrorMessageFactory shouldContain(CharSequence actual, CharSequenc
                                    StandardComparisonStrategy.instance());
   }
 
+  public static ErrorMessageFactory shouldContain(Throwable actual, CharSequence sequence) {
+    String format = "%n" +
+                    "Expecting:%n" +
```
joel-costigliola: I would reword this to ` "Expecting throwable message:%n" ` as we know the following string is not any string but a throwable message

src/test/java/org/assertj/core/error/ShouldContainThrowable_create_Test.java
benoitdupont: I just discovered BDDAssertions and saw that it's encouraged in the contributing guidelines. Is it to _allow to skip test execution when assumptions are not met_ as described in https://assertj.github.io/doc/#assertj-core-3.14.0-bdd-assumptions ?

src/test/java/org/assertj/core/error/ShouldContainThrowable_create_Test.java
benoitdupont: Glad you like it ;-)

benoitdupont:
Voilà :-) All remarks has been fixed, just one question remaining in the upper discussion. And I see that I forgot to rebase the latest commit, I'm battling with git now to do that.

benoitdupont:
Ok seems good, next time I won't forget to rebase the upstream before

src/test/java/org/assertj/core/error/ShouldContainThrowable_create_Test.java
joel-costigliola: yes but I'm not sure what the use case is here, can you clarify?

joel-costigliola:
@benoitdupont  as long as you are the only one working on a branch you can do whatever git operations on it (like rewrite the history and force push).

src/test/java/org/assertj/core/error/ShouldContainThrowable_create_Test.java
benoitdupont: What I meant is: should we use BDDAssertions when we use the _given_ method to be able to skip the test execution when assumptions are not met, or use BDDAssertions by default instead of _assertThat_ as a general coding guideline ?

joel-costigliola:
Integrated thanks @benoitdupont!

PR #1732, opened 2019-12-16
Adding hasSameBinaryContentAs

### LINKED ISSUE #1059: Comparing equal binary files fails with java.nio.charset.MalformedInputException

ngeor:
#### Check List:
* Fixes #1059 
* Unit tests : YES

### DIFF
--- a/src/test/java/org/assertj/core/api/file/FileAssert_hasSameBinaryContentAs_Test.java
+++ b/src/test/java/org/assertj/core/api/file/FileAssert_hasSameBinaryContentAs_Test.java
@@ -1,0 +1,73 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2019 the original author or authors.
+ */
+package org.assertj.core.api.file;
+
+import org.assertj.core.api.FileAssert;
+import org.assertj.core.api.FileAssertBaseTest;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import java.io.File;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.BDDAssertions.then;
+import static org.assertj.core.util.AssertionsUtil.expectAssertionError;
+import static org.assertj.core.util.TempFileUtil.createTempFileWithContent;
+import static org.mockito.Mockito.verify;
+
+/**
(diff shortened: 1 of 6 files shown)


src/test/java/org/assertj/core/api/file/FileAssert_hasSameBinaryContentAs_Test.java
```diff
@@ -0,0 +1,78 @@
...
+    assertThat(actual).hasSameBinaryContentAs(expected);
+  }
+
+  @Test
+  public void should_fail_on_different_files() throws Exception {
+    File actual = createDeleteOnExitTempFileWithContent("assertJ");
+    File expected = createDeleteOnExitTempFileWithContent("assertJ++");
+    assertThatExceptionOfType(AssertionError.class).isThrownBy(() -> {
+      assertThat(actual).hasSameBinaryContentAs(expected);
+    }).withMessageContaining("does not have expected binary content at offset");
```
joel-costigliola: prefer `expectAssertionError` as per https://github.com/joel-costigliola/assertj-core/blob/master/CONTRIBUTING.md
add GIVEN WHEN THEN sections

src/test/java/org/assertj/core/api/file/FileAssert_hasSameBinaryContentAs_Test.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  private File createDeleteOnExitTempFileWithContent(String content) throws IOException {
```
joel-costigliola: rename to `createTempFileWithContent`

scordio:
Closing and reopening to trigger Travis again

ngeor:
> we need to deprecate `hasSameContentAs` for `hasSameTextualContentAs` to fix #1059 otherwise
> just some minor comments, thanks @ngeor!

Thanks, I  processed the minor code changes (I should've read the Contributing guidelines first).
Just to double check, shall I introduce this new method `hasSameTextualContentAs`?

joel-costigliola:
Yes let's add `hasSameTextualContentAs` (a copy of `hasSameContentAs`)

ngeor:
I added the new method and marked the old ones as deprecated. Some extra things:
- I added a `@since` tag to the new methods
- I added a `TempFileUtil` class because the `createTempFile` method was copy pasted around
- Fixed a typo in the docs `Turkisch -> Turkish`

joel-costigliola:
Integrated thanks @ngeor! (I have added the same assertion for path)

PR #1742, opened 2020-01-01
Deprecate Preconditions#checkNotNull(Object) and Preconditions#checkNotNull(Object, String)

PascalSchumacher:

### DIFF
--- a/src/main/java/org/assertj/core/util/Preconditions.java
+++ b/src/main/java/org/assertj/core/util/Preconditions.java
@@ -13,6 +13,7 @@
 package org.assertj.core.util;
 
 import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
 
 import org.assertj.core.api.filter.FilterOperator;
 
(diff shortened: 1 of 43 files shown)


src/main/java/org/assertj/core/util/Preconditions.java
```diff
@@ -61,24 +62,11 @@ public static CharSequence checkNotNullOrEmpty(CharSequence s, String message) {
...
 
-  /**
-   * Verifies that the given object reference is not {@code null}.
-   * 
-   * @param <T> the type of the reference to check.
-   * @param reference the given object reference.
-   * @return the non-{@code null} reference that was validated.
-   * @throws NullPointerException if the given object reference is {@code null}.
-   */
-  public static <T> T checkNotNull(T reference) {
```
PascalSchumacher: Maybe we have to keep the method around to avoid breaking binary compatiblity as the class is not part of the `internal` package?

scordio: I would keep it but marking as deprecated and then it could be dropped with the next major release

joel-costigliola: agreed, along with indicating the alternative in the deprecated javadoc tag description

PascalSchumacher:
Merged in https://github.com/joel-costigliola/assertj-core/commit/d4287dbd03ba84e941d999d135d567fc683b1b54

PR #1743, opened 2020-01-02
Add hasSuperclass(Class<?>) to AbstractClassAssert

### LINKED ISSUE #1512: Add hasSuperclass(Class<?>) to AbstractClassAssert
#### Summary
Sometimes I feel the lack of `isNotAssignableFrom()` in `AbstractClassAssert`.
Recently I needed a defensive test to enforce an exception class staying "checked" (i.e., not inheriting from `RuntimeException`). To do that, I created a custom condition:

```java
private static Condition<Class<?>> runtimeException = new Condition<>(RuntimeException.class::isAssignableFrom, RuntimeException.class.getSimpleName());
```
used in this way:
```java
assertThat(MyException.class).isNot(runtimeException);
```

#### Example
Following the behavior of the existing `isAssignableFrom(Class<?>...)`, a new api could check that `actual` is not assignable from any of the `others`:
```java
public SELF isNotAssignableFrom(Class<?>... others)
```
With it, my example would not require a custom condition:
```java
assertThat(RuntimeException.class).isNotAssignableFrom(MyException.class);
```

Would this new api make sense?

#### Additional Thoughts
Would it make sense to have some assertions for `Throwable` classes, for example to verify if the class is inheriting from `Error`/`Exception`/`RuntimeException`? It would address the ugly backwards readability of `isAssignableFrom` :-)
Maybe a new `AbstractThrowableClassAssert` inheriting from `AbstractClassAssert` could to the job.

scordio:
#### Check List:
* Fixes #1512 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractClassAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractClassAssert.java
@@ -350,6 +350,45 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
   }
 
   /**
+   * Verifies that the actual {@code Class} has the given class as direct superclass (as in {@link Class#getSuperclass()}).
+   * <p>
+   * Example:
+   * <pre><code class='java'> // this assertion succeeds as Object has no superclass:
+   * assertThat(Object.class).hasSuperclass(null);
+   *
+   * // this assertion succeeds as interfaces have no superclass:
+   * assertThat(Cloneable.class).hasSuperclass(null);
+   *
+   * // this assertion succeeds as primitive types have no superclass:
+   * assertThat(Integer.TYPE).hasSuperclass(null);
+   *
+   * // this assertion succeeds as void type has no superclass:
+   * assertThat(Void.TYPE).hasSuperclass(null);
+   *
+   * // this assertion succeeds:
+   * assertThat(Integer.class).hasSuperclass(Number.class);
+   *
+   * // this assertion succeeds as superclass for array classes is Object:
+   * assertThat(Integer[].class).hasSuperclass(Object.class);
+   *
+   * // this assertion fails as only direct superclass matches:
+   * assertThat(Integer.class).hasSuperclass(Object.class);
+   *
+   * // this assertion fails as interfaces are not superclasses:
+   * assertThat(Integer.class).hasSuperclass(Comparable.class);</code></pre>
+   *
+   * @param superclass the class which must be the direct superclass of actual.
+   * @return {@code this} assertions object
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} doesn't have the given class as direct superclass.
+   * @since 3.15.0
+   */
+  public SELF hasSuperclass(Class<?> superclass) {
+    classes.assertHasSuperclass(info, actual, superclass);
+    return myself;
+  }
+
+  /**
    * @deprecated use {@link #hasPublicFields(String...)} instead.
    * @param fields the fields who must be in the class.
    * @return {@code this} assertions object
--- a/src/main/java/org/assertj/core/error/ShouldHaveSuperclass.java
+++ b/src/main/java/org/assertj/core/error/ShouldHaveSuperclass.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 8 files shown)


src/main/java/org/assertj/core/error/ShouldHaveSuperclass.java
```diff
@@ -0,0 +1,48 @@
...
+public class ShouldHaveSuperclass extends BasicErrorMessageFactory {
+
+  private static final String SHOULD_HAVE_SUPERCLASS = new StringJoiner("%n", "%n", "")
+                                                                                       .add("Expecting")
+                                                                                       .add("  <%s>")
+                                                                                       .add("to have superclass:")
+                                                                                       .add("  <%s>")
+                                                                                       .add("but had:")
+                                                                                       .add("  <%s>")
+                                                                                       .toString();
```
scordio: I used StringJoiner instead of standard string concatenation to improve readability, but happy to revert it if it's too deviating from the standard approach.

joel-costigliola: I don't mind at all, I think we used other implementations when the code base was java < 8

src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -349,6 +349,45 @@ public SELF hasAnnotation(Class<? extends Annotation> annotation) {
     return myself;
   }
 
+  /**
+   * Verifies that the actual {@code Class} has the given class as direct superclass (as in {@link Class#getSuperclass()}).
+   * <p>
+   * Example:
+   * <pre><code class='java'> // this assertion succeeds as Object has no superclass:
+   * assertThat(Object.class).hasSuperclass(null);
```
joel-costigliola: Minor `hasNoSuperclass` would be better than `hasSuperclass(null)` but in this PR.

src/main/java/org/assertj/core/error/ShouldHaveSuperclass.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Creates an error message indicating that an assertion that verifies that a class has a given superclass failed.
+ * 
+ * @author Stefano Cordio
+ */
+public class ShouldHaveSuperclass extends BasicErrorMessageFactory {
+
+  private static final String SHOULD_HAVE_SUPERCLASS = new StringJoiner("%n", "%n", "")
+                                                                                       .add("Expecting")
```
joel-costigliola: minor: join lines

src/main/java/org/assertj/core/error/ShouldHaveSuperclass.java
joel-costigliola: I don't mind as long as it is formatted so that we can have a good idea of how the message will look like

src/main/java/org/assertj/core/internal/Classes.java
```diff
@@ -258,6 +259,23 @@ public void assertContainsAnnotations(AssertionInfo info, Class<?> actual,
...
+   * @param info       contains information about the assertion.
+   * @param actual     the "actual" {@code Class}.
+   * @param superclass the direct superclass, which can be null according to {@link Class#getSuperclass()}.
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} superclass does not have the expected superclass.
+   */
+  public void assertHasSuperclass(AssertionInfo info, Class<?> actual, Class<?> superclass) {
+    assertNotNull(info, actual);
+    Class<?> actualSuperclass = actual.getSuperclass();
+    if (actualSuperclass == null && superclass != null || actualSuperclass != null && !actualSuperclass.equals(superclass)) {
```
joel-costigliola: add `()` to group && conditions for clarity

src/test/java/org/assertj/core/internal/classes/Classes_assertHasSuperclass_Test.java
```diff
@@ -0,0 +1,94 @@
...
+    // THEN
+    then(assertionError).hasMessage(actualIsNull());
+  }
+
+  @Test
+  void should_pass_if_actual_has_given_class_as_direct_superclass() {
+    // GIVEN
+    Class<?> actual = Integer.class;
+    Class<?> superclass = Number.class;
+    // WHEN
```
joel-costigliola: // WHEN/THEN

joel-costigliola:
thanks @scordio!

src/main/java/org/assertj/core/api/AbstractClassAssert.java
scordio: Done in #1752.

PR #1752, opened 2020-01-12
Add hasNoSuperclass() to AbstractClassAssert

scordio:
Main behavior change: `hasSuperclass(Class<?>)` now fails with `null` parameter.

#### Check List:
* Fixes https://github.com/joel-costigliola/assertj-core/pull/1743#discussion_r362622859
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/error/ShouldHaveSuperclass.java
+++ b/src/main/java/org/assertj/core/error/ShouldHaveSuperclass.java
@@ -25,9 +25,14 @@ public class ShouldHaveSuperclass extends BasicErrorMessageFactory {
                                                                                        .add("  <%s>")
                                                                                        .add("to have superclass:")
                                                                                        .add("  <%s>")
-                                                                                       .add("but had:")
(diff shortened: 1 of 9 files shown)


src/main/java/org/assertj/core/error/ShouldHaveSuperclass.java
```diff
@@ -37,11 +42,19 @@
    * @return the created {@code ErrorMessageFactory}.
    */
   public static ErrorMessageFactory shouldHaveSuperclass(Class<?> actual, Class<?> superclass) {
-    return new ShouldHaveSuperclass(actual, superclass);
+    Class<?> actualSuperclass = actual.getSuperclass();
+    if (actualSuperclass == null) {
+      return new ShouldHaveSuperclass(actual, superclass);
```
joel-costigliola: is it possible? I think we are checking that `superClass` is not null in `hasSuperClass` so we should not need this one

joel-costigliola:
Nice, thanks @scordio

PR #1755, opened 2020-01-14
Add ignoreFieldsForTypes to RecursiveComparisonConfiguration

### LINKED ISSUE #1749: Recursive comparison that ignore fields by types
`.usingRecursiveComparison()` is really useful for objects that don't override equals and hashcode.

In order to assert an object with fields that has random values, like id, uuid or datetime, I'll need to ignore them using this method:
https://github.com/joel-costigliola/assertj-core/blob/master/src/main/java/org/assertj/core/api/RecursiveComparisonAssert.java#L251

```
RecursiveComparisonConfiguration ignoreIdConfig = new RecursiveComparisonConfiguration();
ignoreIdConfig.ignoreFields("idFieldName");
assertThat(actual)
        .usingRecursiveComparison(ignoreIdConfig)
        .isEqualTo(expected);
```

It works like a charm, but supplying the field name with a string is not type-safe, I wonder if it's possible to add a method that can take field type as arguments like this: 

```
ignoreIdConfig.ignoreFields(ZonedDateTime.class, UUID.class);
```

So it will match and ignore any fields with the given types.

tommysitu:
#### Check List:
* Fixes #1749 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration.java
+++ b/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration.java
@@ -43,6 +43,7 @@ public class RecursiveComparisonConfiguration {
   private boolean ignoreAllActualNullFields = false;
   private Set<FieldLocation> ignoredFields = new LinkedHashSet<>();
   private List<Pattern> ignoredFieldsRegexes = new ArrayList<>();
+  private Set<Class<?>> ignoredFieldsForTypes = new LinkedHashSet<>();
 
   // overridden equals method to ignore section
   private List<Class<?>> ignoredOverriddenEqualsForTypes = new ArrayList<>();
@@ -133,12 +134,32 @@ public class RecursiveComparisonConfiguration {
   }
 
   /**
+   * Adds the given types to the list of the object under test fields to ignore in the recursive comparison.
+   * <p>
+   * See {@link RecursiveComparisonAssert#ignoringFields(String...) RecursiveComparisonAssert#ignoringFields(String...)} for examples.
+   *
+   * @param types the types of the object under test to ignore in the comparison.
+   */
+  public void ignoreFieldsForTypes(Class<?>... types) {
+    ignoredFieldsForTypes.addAll(list(types));
+  }
+
+  /**
    * Returns the list of the object under test fields to ignore in the recursive comparison.
    *
    * @return the list of the object under test fields to ignore in the recursive comparison.
    */
   public Set<FieldLocation> getIgnoredFields() {
     return ignoredFields;
+  }
+
+  /**
+   * Returns the set of the object under test field types to ignore in the recursive comparison.
+   *
+   * @return the set of the object under test field types to ignore in the recursive comparison.
+   */
+  public Set<Class<?>> getIgnoredFieldsForTypes() {
+    return ignoredFieldsForTypes;
   }
 
   /**
@@ -324,6 +345,7 @@ public class RecursiveComparisonConfiguration {
     describeIgnoreAllActualNullFields(description);
     describeIgnoredFields(description);
     describeIgnoredFieldsRegexes(description);
+    describeIgnoredFieldsForTypes(description);
     describeOverriddenEqualsMethodsUsage(description, representation);
     describeIgnoreCollectionOrder(description);
     describeIgnoredCollectionOrderInFields(description);
@@ -339,7 +361,8 @@ public class RecursiveComparisonConfiguration {
   boolean shouldIgnore(DualValue dualKey) {
     return matchesAnIgnoredNullField(dualKey)
            || matchesAnIgnoredField(dualKey)
-           || matchesAnIgnoredFieldRegex(dualKey);
+           || matchesAnIgnoredFieldRegex(dualKey)
+           || matchesAnIgnoredFieldType(dualKey);
   }
 
   Predicate<String> shouldKeepField(String parentConcatenatedPath) {
@@ -382,6 +405,11 @@ public class RecursiveComparisonConfiguration {
   private void describeIgnoredFields(StringBuilder description) {
     if (!ignoredFields.isEmpty())
       description.append(format("- the following fields were ignored in the comparison: %s%n", describeIgnoredFields()));
+  }
+
+  private void describeIgnoredFieldsForTypes(StringBuilder description) {
+    if (!ignoredFieldsForTypes.isEmpty())
+      description.append(format("- the following types were ignored in the comparison: %s%n", describeIgnoredTypes()));
   }
 
   private void describeIgnoreAllActualNullFields(StringBuilder description) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration.java
```diff
@@ -141,6 +153,15 @@ public void ignoreFieldsMatchingRegexes(String... regexes) {
     return ignoredFields;
   }
 
+  /**
+   * Returns the set of the object under test field types to ignore in the recursive comparison.
+   *
+   * @return the set of the object under test field types to ignore in the recursive comparison.
+   */
+  public Set<Class<?>> getIgnoredFieldsForTypes() {
```
joel-costigliola: rename to `getIgnoredTypes` as this is not returning fields (or `FieldLocation`)

src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration.java
```diff
@@ -477,6 +505,10 @@ private boolean matchesAnIgnoredField(DualValue dualKey) {
     return matchesAnIgnoredField(dualKey.concatenatedPath);
   }
 
+  private boolean matchesAnIgnoredFieldType(DualValue dualKey) {
+    return ignoredFieldsForTypes.contains(dualKey.actual.getClass());
```
joel-costigliola: `dualKey.actual` can be null, in that case it's a bit tricky, we could look at `dualKey.expected` but since we can compare objects of different types (ex: `Person` vs `PersonDto`) this might not be relevant.

Options when `dualKey.actual` is null:
* return false (don't ignore as we can't evaluate the type)
* evaluate `dualKey.expected` instead assuming `actual` and `expected` (not always true)
* ?

thoughts

src/test/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration_shouldIgnoreFields_Test.java
```diff
@@ -137,10 +140,48 @@ public void should_ignore_fields(DualValue dualKey) {
...
                      arguments(dualKeyWithPath("number")),
                      arguments(dualKeyWithPath("surname")),
-                     arguments(dualKeyWithPath("first", "name")));
+                     arguments(dualKeyWithPath("first", "name")),
+                     arguments(new DualValue(randomPath(), "actual", "expected")));
 
   }
 
+  @Test
+  public void ignoring_fields_for_types_does_not_replace_previous_types() {
```
joel-costigliola: rename to `ignoring_fields_for_types_does_not_replace_previous_ignored_types`

src/test/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration_shouldIgnoreFields_Test.java
```diff
@@ -137,10 +140,48 @@ public void should_ignore_fields(DualValue dualKey) {
...
+
+  @ParameterizedTest(name = "{0} should be ignored with these ignored fields {1}")
+  @MethodSource("ignoringSpecifiedTypesSource")
+  public void should_ignore_fields_for_specified_types(DualValue dualKey, List<Class<?>> ignoredTypes) {
+    // GIVEN
+    recursiveComparisonConfiguration.ignoreFieldsForTypes(ignoredTypes.toArray(new Class<?>[0]));
+    // WHEN
+    boolean ignored = recursiveComparisonConfiguration.shouldIgnore(dualKey);
+    // THEN
+    assertThat(ignored).as("%s should be ignored with these ignored types %s", dualKey, ignoredTypes).isTrue();
```
joel-costigliola: minor: move `isTrue()` to the next line for readability

src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration.java
tommysitu: The use case for this method would be when someone needs to ignore a field with random values. 

As a user, I maybe not aware of the actual field is null, but I do want to ignore it whenever possible. 

In that sense, the second option is better, to check `dualKey.expected`'s type. And there are three scenarios: 

1. If the expected type matches, then obviously it should ignore (return true)
2. If the expected type does not match, it should not ignore (return false)
3. If both are null, then there is no need to ignore (return false)

User ignoring the types should be providing an expected value with the same type, so hopefully scenario 2 is unlikely to occur. And most of the time, the behaviour for handling null `dualKey.actual` should make sense. What do you think?

joel-costigliola: > If the expected type matches, then obviously it should ignore (return true)
> If the expected type does not match, it should not ignore (return false)

we can't know that since if actual is null, you can't get its type

> If both are null, then there is no need to ignore as they are untyped and equal (return false)

I agree this is not an issue

joel-costigliola: I think we should not ignore the field if actual is null otherwise we would ignore all null fields in actual.

tommysitu: > I think we should not ignore the field if actual is null otherwise we would ignore all null fields in actual.

So evaluting`dualKey.expected` type will avoid ignoring all null fields, as in Scenario 1, you only ignore a null actual only when the expected has a matching type. 

Here is an example: 

I have a person with a pet: 
```
public class Person {
       private final String name;
       private final Pet pet;
}
```

I'd like to ignore the Pet, so I would do this: 

```
RecursiveComparisonConfiguration ignorePetConfig = new RecursiveComparisonConfiguration();
ignorePetConfig.ignoreFieldsForTypes(Pet.class);

assertThat(actual)
        .usingRecursiveComparison(ignorePetConfig)
        .isEqualTo(new Person("john", new Pet());
```

where `actual = new Person()` with null fields, so by checking the `dualKey.expected` type, it should only ignore the `pet` field, not the `name` field.  Do you see what I meant? 

It's basically what you suggest: 
`evaluate dualKey.expected instead assuming actual and expected (not always true)`

joel-costigliola: what if you have a `PersonDto` with a `PetDto`?

tommysitu: haven't used much DTO, but I'm taking the suggestion on returning false if the type cannot be evaluated, it's simple to reason about.

joel-costigliola:
@tommysitu thanks, starting from this PR I have started to complete the work for this feature, i'm almost done!

tommysitu:
> @tommysitu thanks, based on your PR I have started to complete the work for this feature, I'm almost done!

Hey @joel-costigliola I can still wrap this up this weekend. Is it fine?

joel-costigliola:
Sure, go ahead!

tommysitu:
Hi @joel-costigliola thanks for your patience. Please review my latest commit. I believe that's everything for this PR.

joel-costigliola:
Thanks, I will review it shortly !

joel-costigliola:
Integrated thanks @tommysitu! 
I have made a few changes:

tommysitu:
> Integrated thanks @tommysitu!
> I have made a few changes:
> 
> * use expected class  when actual is null but only in `strictTypeChecking` mode where we enforce that actual and expected have the same types (this one was tricky)
> * renamed the api method to `ignoreFieldsOfTypes`
> * handle primitive types (int.class) as their wrapper counterpart (Integer.class)

Fab! Thank you for refinining it!

PR #1762, opened 2020-01-25
Replaces String by CharSequence in textual diff.

fduminy:
#### Check List:
* Extends textual diff to all CharSequence implementations
* Unit tests : NA


joel-costigliola:
@fduminy what is the problem solved by this PR?

fduminy:
@joel-costigliola It simply makes the code more generic because it's better to depend on an interface (CharSequence) than depend on a concrete class (String). 

That makes the code more reusable for other implementations of CharSequence like StringBuffer, StringBuilder, CharBuffer ...

PR #1767, opened 2020-01-29
Deprecate areEqual() and areEqualArrays(), introduce EqualsVerifier

scordio:
Following #1764, this PR deprecates and reduces a bit more the usage of `areEqual()` and `areEqualArrays()`. Also it shows how an `EqualsVerifier` test would look like.

### DIFF
--- a/src/main/java/org/assertj/core/groups/Tuple.java
+++ b/src/main/java/org/assertj/core/groups/Tuple.java
@@ -12,45 +12,39 @@
  */
 package org.assertj.core.groups;
 
-import static java.util.Collections.addAll;
 import static org.assertj.core.configuration.ConfigurationProvider.CONFIGURATION_PROVIDER;
-import static org.assertj.core.util.Lists.newArrayList;
-import static org.assertj.core.util.Objects.areEqual;
+import static org.assertj.core.util.Lists.list;
 
+import java.util.Arrays;
 import java.util.List;
 
-public class Tuple {
+public final class Tuple {
 
-  private final List<Object> datas = newArrayList();
+  private final List<Object> values;
 
   public Tuple(Object... values) {
-    addAll(datas, values);
+    this.values = list(values);
   }
 
   public Object[] toArray() {
-    return datas.toArray();
+    return values.toArray();
   }
 
   public List<Object> toList() {
-    return datas;
-  }
-
-  @Override
-  public int hashCode() {
-    final int prime = 31;
-    int result = 1;
-    result = prime * result + datas.hashCode();
-    return result;
+    return values;
   }
 
   @Override
   public boolean equals(Object obj) {
     if (this == obj) return true;
-    if (obj == null) return false;
     if (!(obj instanceof Tuple)) return false;
     Tuple other = (Tuple) obj;
-    // datas can't be null
-    return areEqual(datas.toArray(), other.datas.toArray());
+    return Arrays.deepEquals(values.toArray(), other.values.toArray());
+  }
+
+  @Override
+  public int hashCode() {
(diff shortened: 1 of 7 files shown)


src/main/java/org/assertj/core/groups/Tuple.java
```diff
@@ -12,45 +12,38 @@
...
-  public int hashCode() {
-    final int prime = 31;
-    int result = 1;
-    result = prime * result + datas.hashCode();
-    return result;
-  }
-
-  @Override
-  public boolean equals(Object obj) {
+  public final boolean equals(Object obj) {
```
scordio: The `final` on `equals` and `hashCode` is [a suggestion from EqualsVerifier](https://jqno.nl/equalsverifier/errormessages/subclass-equals-is-not-final/), but there are other strategies if we don't want to be so strict.

src/test/java/org/assertj/core/internal/objects/Objects_assertEqual_Test.java
```diff
@@ -80,10 +80,10 @@ public void should_fail_if_objects_are_not_equal_according_to_custom_comparison_
   }
 
   @Test
-  public void should_fail_with_my_exception_if_compared_with_null() {
+  public void should_fail_if_compared_with_null() {
     Throwable error = catchThrowable(() -> objects.assertEqual(someInfo(), new MyObject(), null));
 
-    assertThat(error).isInstanceOf(MyObject.NullEqualsException.class);
+    assertThat(error).isInstanceOf(AssertionError.class);
```
scordio: This changed because `java.util.Objects#deepEquals` performs a null check first and then invokes `equals()`.

src/main/java/org/assertj/core/groups/Tuple.java
joel-costigliola: I'm ok with that

src/main/java/org/assertj/core/groups/Tuple.java
```diff
@@ -12,45 +12,38 @@
...
 import java.util.List;
 
 public class Tuple {
 
-  private final List<Object> datas = newArrayList();
+  private final List<Object> values;
 
   public Tuple(Object... values) {
-    addAll(datas, values);
+    this.values = Arrays.asList(values);
```
joel-costigliola: minor point: I usually prefer a static import of `org.assertj.core.util.Lists.list` for conciseness (feel free to ignore me).

scordio: Fair point, I'll change it! Maybe when we move to java 9+ we could replace these util methods with the corresponding `of()` factories (e.g., `List.of()`)

src/main/java/org/assertj/core/groups/Tuple.java
joel-costigliola: I'm ok with that, we might even put the class as final, I don't see why people would subclass `Tuple` it's really an immutable data class.

src/main/java/org/assertj/core/groups/Tuple.java
joel-costigliola: `List.of` is better than `asList` I sill find `list` to be better but I don't mind if people use `List.of`, what I'm not too fond of is `of` because it's not always obvious what is being built

Examples:
```
this.values = List.of(values);  
```
```
this.values = list(values);  
```
```
// is that a list, an array, a set  or something else?
this.values = of(values) ; 
```

scordio: In that case not using the static import would be mandatory 🙂

PR #1769, opened 2020-01-30
Fix trailing space when class is only package-private

### LINKED ISSUE #1768: Trailing space in the modifiers list for ClassModifierShouldBe
#### Summary
When a class is only `package-private` and a `ClassModifierShouldBe` message is created, the list of modifiers in the failure message contains a trailing space right after the `package-private` value.

#### Example

Given:

```java
class PackagePrivate {
}
```

`assertThat(PackagePrivate.class).isPublic()` yields:

```
java.lang.AssertionError: 
Expecting:
  <org.example.PackagePrivate>
to be a "public" class but was "package-private ".
```

scordio:
#### Check List:
* Fixes #1768 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/error/ClassModifierShouldBe.java
+++ b/src/main/java/org/assertj/core/error/ClassModifierShouldBe.java
@@ -12,12 +12,13 @@
  */
 package org.assertj.core.error;
 
-import static java.lang.String.format;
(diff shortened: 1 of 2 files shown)


joel-costigliola:
thanks!

PR #1770, opened 2020-01-31
Add linux OS and cross-version build for jdk 13, 14 and 15

scordio:


PascalSchumacher:
Imho we should not test with Java 12 as it has be superseded by Java 13 and is no longer supported.

scordio:
> If we use Github Actions for Linux testing we should remove the travis build.

I didn't remove it yet because Sonar is still missing, I was planning to do it in a second round.

PR #1771, opened 2020-02-02
Extends SoftAssertionsExtension to injection of constructor parameters.

fduminy:
#### Check List:
* Extends SoftAssertionsExtension to injection of constructor parameters
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/ISoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/ISoftAssertions.java
@@ -1,0 +1,18 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2020 the original author or authors.
+ */
+package org.assertj.core.api;
+
+public interface ISoftAssertions {
+  void clearErrorsCollected();
+  void assertAll();
+}
--- a/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
@@ -12,13 +12,13 @@
  */
 package org.assertj.core.api;
 
-import static java.lang.String.format;
-import static java.util.stream.Collectors.toList;
+import org.assertj.core.internal.Failures;
 
 import java.lang.reflect.Constructor;
 import java.util.List;
 
-import org.assertj.core.internal.Failures;
+import static java.lang.String.format;
+import static java.util.stream.Collectors.toList;
 
 public class AbstractSoftAssertions implements InstanceOfAssertFactories {
 
@@ -126,6 +126,13 @@ public class AbstractSoftAssertions implements InstanceOfAssertFactories {
    */
   public List<Throwable> errorsCollected() {
     return decorateErrorsCollected(proxies.errorsCollected());
+  }
+
+  /**
+   * Clears the list of soft assertions collected errors.
+   */
+  public void clearErrorsCollected() {
+    proxies.clearErrorsCollected();
   }
 
   /**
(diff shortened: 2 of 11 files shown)


joel-costigliola:
@fduminy can you explain with an example the problem yo are trying to solve with this PR?

fduminy:
@joel-costigliola here is an example : 

```java
ExtendWith(SoftAssertionsExtension.class)
 class ExampleTestCase {
    private final SoftAssertions softly;

    ExampleTestCase(SoftAssertions softly) {
      this.softly = softly;
    }

    @Test
    void multipleFailures() {
       softly.assertThat(2 * 3).isEqualTo(0);
       softly.assertThat(Arrays.asList(1, 2)).containsOnly(1);
       softly.assertThat(1 + 1).isEqualTo(2);
    }
}
```

It can be particularly useful when there are many test methods using SoftAssertions in the same class.

joel-costigliola:
We already have support for JUnit 5 https://assertj.github.io/doc/#assertj-core-junit5-soft-assertions, I'm not sure to see what is missing with the current support, can clarify what is missing?

fduminy:
The link you gave doesn't show an example with injection in constructor. I don't think it's currently supported by assertj.

I think we should support different styles of junit 5 tests to give users more flexibility instead of forcing them to use only method injection style. Moreover, migration from junit 4 rule is easier with constructor injection since it avoids changing each test method.

joel-costigliola:
Ok that makes sense now

joel-costigliola:
@fduminy what about taking https://github.com/joel-costigliola/assertj-core/pull/1784 approach? WDTY?

fduminy:
I replyed in #1784

joel-costigliola:
@fduminy as in https://github.com/joel-costigliola/assertj-core/pull/1784#issuecomment-593849219 I think there is a fundamental issue: if we run tests in parallel, the soft assertions field is going to be reset while test instances are being initialized, leading to mix soft assertions errors collected.

Do you see away around that?

filiphr:
JUnit creates a new test instance for every test (unless `Lifecycle.PER_CLASS` is used). According to [Parallel Execution](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parallel-execution), even if you enable parallelization per test node (parallel for the same class) you still have to explicitly mark the class or test method with `@Execution(CONCURRENT)`.

I think that for a constructor injection, or even for #1784 it should be OK to inject the field

src/main/java/org/assertj/core/api/ISoftAssertions.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2020 the original author or authors.
+ */
+package org.assertj.core.api;
+
+public interface ISoftAssertions {
```
dstekanov: maybe it is worth to add javadoc here

joel-costigliola:
@filiphr thanks for the comment.

kriegfrj:
> @filiphr thanks for the comment.
> 
> @sbrannen do you think as @filiphr suggested that we can use a `SoftAssertions` field initialized with constructor injection? (except for classes combining `Lifecycle.PER_CLASS` with `@Execution(CONCURRENT)`)

Actually, I think that constructor injection would be a problem for `SoftAssertions` if you're using `PER_CLASS`, regardless of whether it's single threaded or concurrent. That's because in `PER_CLASS` the constructor is only called once, so you don't get a fresh `SoftAssertions` object for each test.

I don't think that this should be a reason not to support constructor injection though (more on that in https://github.com/joel-costigliola/assertj-core/issues/1824#issuecomment-608992199). But to avoid nasty surprises, we should do a sanity check - `supportsParameter()` should throw a `ParameterResolutionException` if it notices that we're trying to inject a soft assertions object into a constructor and ExtensionContext.getTestInstanceLifecycle() is `PER_CLASS`.

joel-costigliola:
@fduminy I have just pushed a [commit](https://github.com/joel-costigliola/assertj-core/commit/a4c847cd49b0537702b7756389d63f9d526993b8) with a refactored `SoftlyExtension` approach without annotating fields, I'm not sure there is a lot of value doing another one for constructor, WDYT?

src/main/java/org/assertj/core/api/ISoftAssertions.java
kriegfrj: `ISoftAssertions` is similar in concept and scope to `SoftAssertionsProvider`, which was merged in #1817. I think that at least this part of this PR is redundant.

joel-costigliola:
I'm closing this issue as we now support field injection, even though it's not constructor injection per se, I believe it solves the same issue.
Please comment if you disagree.

PR #1772, opened 2020-02-03
Add Sonar scan on push events

scordio:
This workflow requires a `SONAR_TOKEN` secret.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,3 +19,23 @@ jobs:
           java-version: 11
       - name: Test
         run: ./mvnw -V --no-transfer-progress clean verify javadoc:javadoc
+
+  sonar:
+    name: Sonar
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


filiphr:
Is it possible to also add

```
 if: github.repository == 'joel-costigliola/assertj-core'
```

for the job?

This way the job would be skipped on forks. Otherwise, contributors will also have this running and would be getting emails due to failures (the `SONAR_TOKEN` not being present)

joel-costigliola:
I have added the token with the value we used for travis, hope it is enough

.github/workflows/sonar.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Test with Sonar
+        run: >
+          ./mvnw -V --no-transfer-progress clean verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar
```
joel-costigliola: let's add javadoc building too

PR #1782, opened 2020-02-10
Allow assertions on the (root) cause message

### LINKED ISSUE #1776: Allow assertion on the root cause message
#### Summary

It is currently not directly possible to assert on the root cause message. There is a workaround possible by using `hasStackTraceContaining`, but it would be nice if this was supported directly.

See also https://stackoverflow.com/questions/38943346/assertj-assert-on-the-cause-message

#### Example

I would like to be able to do this:

```java
assertThatExceptionOfType(PersistenceException.class)
    .isThrownBy(() -> {
        // test code here
    })
   .withRootCauseInstanceOf(PSQLException.class)
   .withRootCauseMessageMatching("Key (user_id)=(.*) is not present in table \"book\"");
```

So to be consistent, probably the following new methods are needed:

* `withRootCauseMessageStartingWith(String description)`
* `withRootCauseMessageContaining(String description)` 
* `withRootCauseMessageContainingAll(CharSequence... values)`
* `withRootCauseMessageNotContaining(String content)`
* `withRootCauseMessageNotContainingAny(CharSequence... values)`
* `withRootCauseMessageMatching(String regex)`
* `withRootCauseMessageEndingWith(String description)`

wimdeblauwe:
#### Check List:
* Fixes #1776
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
@@ -159,6 +159,30 @@ public abstract class AbstractThrowableAssert<SELF extends AbstractThrowableAsse
   }
 
   /**
+   * Verifies that the actual {@code Throwable} has a cause.
+   *
+   * @return this assertion object.
+   * @throws AssertionError if the actual {@code Throwable} is {@code null}.
+   * @throws AssertionError if the actual {@code Throwable} does not have a cause.
+   */
+  public SELF hasCause() {
+    throwables.assertHasCause(info, actual);
+    return myself;
+  }
+
+  /**
+   * Verifies that the actual {@code Throwable} has a root cause.
+   *
+   * @return this assertion object.
+   * @throws AssertionError if the actual {@code Throwable} is {@code null}.
+   * @throws AssertionError if the actual {@code Throwable} does not have a root cause.
+   */
+  public SELF hasRootCause() {
+    throwables.assertHasRootCause(info, actual);
+    return myself;
+  }
+
+  /**
    * Verifies that the message of the actual {@code Throwable} starts with the given description.
    * <p>
    * Examples:
--- a/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
@@ -171,6 +171,18 @@ public abstract class AbstractThrowableAssert<SELF extends AbstractThrowableAsse
   }
 
   /**
+   * Returns a new assertion object that uses the cause of the current Throwable as the actual Throwable
+   *
+   * @return a new assertion object
+   * @throws AssertionError if the actual {@code Throwable} is {@code null}.
+   * @throws AssertionError if the actual {@code Throwable} does not have a cause.
+   */
+  public AbstractThrowableAssert<?,?> getCause() {
+    hasCause();
+    return new ThrowableAssert(actual.getCause());
+  }
+
+  /**
    * Verifies that the actual {@code Throwable} has a root cause.
    *
    * @return this assertion object.
@@ -180,6 +192,18 @@ public abstract class AbstractThrowableAssert<SELF extends AbstractThrowableAsse
   public SELF hasRootCause() {
     throwables.assertHasRootCause(info, actual);
     return myself;
+  }
+
+  /**
+   * Returns a new assertion object that uses the root cause of the current Throwable as the actual Throwable
+   *
+   * @return a new assertion object
+   * @throws AssertionError if the actual {@code Throwable} is {@code null}.
+   * @throws AssertionError if the actual {@code Throwable} does not have a root cause.
+   */
+  public AbstractThrowableAssert<?,?> getRootCause() {
+    hasRootCause();
+    return new ThrowableAssert(org.assertj.core.util.Throwables.getRootCause(actual));
   }
 
   /**
--- a/src/main/java/org/assertj/core/error/ShouldHaveCause.java
+++ b/src/main/java/org/assertj/core/error/ShouldHaveCause.java
@@ -31,6 +31,10 @@ public class ShouldHaveCause extends BasicErrorMessageFactory {
     return new ShouldHaveCause(actualCause, expectedCause);
   }
 
+  public static ErrorMessageFactory shouldHaveCause(Throwable actualCause) {
+    return new BasicErrorMessageFactory("expecting %s to have a cause but it did not", actualCause);
+  }
+
   private ShouldHaveCause(Throwable actualCause, Throwable expectedCause) {
     super("%n" +
           "Expecting a cause with type:%n" +
--- a/src/main/java/org/assertj/core/error/ShouldHaveRootCause.java
+++ b/src/main/java/org/assertj/core/error/ShouldHaveRootCause.java
@@ -42,6 +42,10 @@ public class ShouldHaveRootCause extends BasicErrorMessageFactory {
     return new ShouldHaveRootCause(actual, actualCause, expectedCause);
   }
 
+  public static ErrorMessageFactory shouldHaveRootCause(Throwable actualCause) {
+    return new BasicErrorMessageFactory("expecting %s to have a root cause but it did not", actualCause);
+  }
+
   private ShouldHaveRootCause(Throwable actual, Throwable actualCause, Throwable expectedCause) {
     super("%n" +
           "Expecting a root cause with type:%n" +
--- a/src/test/java/org/assertj/core/api/ThrowableAssertAlternative_havingCause_Test.java
+++ b/src/test/java/org/assertj/core/api/ThrowableAssertAlternative_havingCause_Test.java
@@ -1,0 +1,30 @@
+package org.assertj.core.api;
+
+import org.assertj.core.util.AssertionsUtil;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+@DisplayName("ThrowableAssertAlternative havingCause")
+class ThrowableAssertAlternative_havingCause_Test {
+
+  @Test
+  void should_return_cause_if_throwable_has_cause() {
+    Throwable throwable = new Throwable("top level message", new Throwable("cause message"));
+    new ThrowableAssertAlternative<>(throwable)
+      .havingCause()
+      .withMessage("cause message");
+  }
+
+  @Test
+  void should_fail_if_throwable_has_no_cause() {
+    //WHEN
+    ThrowableAssert.ThrowingCallable code = () -> {
+      Throwable throwable = new Throwable("top level message");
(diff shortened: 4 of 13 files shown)


src/test/java/org/assertj/core/api/ThrowableAssertAlternative_havingRootCause_Test.java
```diff
@@ -0,0 +1,30 @@
+package org.assertj.core.api;
+
+import org.assertj.core.util.AssertionsUtil;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+@DisplayName("ThrowableAssertAlternative havingRootCause")
+class ThrowableAssertAlternative_havingRootCause_Test {
```
joel-costigliola: same comments as `ThrowableAssertAlternative_havingCause_Test`

src/test/java/org/assertj/core/api/ThrowableAssertAlternative_havingCause_Test.java
```diff
@@ -0,0 +1,30 @@
...
+    //WHEN
+    ThrowableAssert.ThrowingCallable code = () -> {
+      Throwable throwable = new Throwable("top level message");
+      new ThrowableAssertAlternative<>(throwable)
+        .havingCause();
+    };
+    //THEN
+    AssertionsUtil.assertThatAssertionErrorIsThrownBy(code).withMessage("expecting java.lang.Throwable: top level message to have a cause but it did not");
+  }
+
```
joel-costigliola: add a test for when actual Throwable is null

src/test/java/org/assertj/core/api/ThrowableAssertAlternative_havingCause_Test.java
```diff
@@ -0,0 +1,30 @@
...
+  @Test
+  void should_fail_if_throwable_has_no_cause() {
+    //WHEN
+    ThrowableAssert.ThrowingCallable code = () -> {
+      Throwable throwable = new Throwable("top level message");
+      new ThrowableAssertAlternative<>(throwable)
+        .havingCause();
+    };
+    //THEN
+    AssertionsUtil.assertThatAssertionErrorIsThrownBy(code).withMessage("expecting java.lang.Throwable: top level message to have a cause but it did not");
```
joel-costigliola: this should build the error message with `shouldHaveCause`, here's an example from an `OptionalAssert` test
```java
  @Test
  public void should_fail_if_optional_is_empty() {
    // GIVEN
    String expectedValue = "something";
    // WHEN
    AssertionError error = expectAssertionError(() -> assertThat(Optional.empty()).contains(expectedValue));
    // THEN
    assertThat(error).hasMessage(shouldContain(expectedValue).create());
  }
```

src/main/java/org/assertj/core/error/ShouldHaveRootCause.java
```diff
@@ -42,6 +42,10 @@ public static ErrorMessageFactory shouldHaveRootCause(Throwable actual, Throwabl
     return new ShouldHaveRootCause(actual, actualCause, expectedCause);
   }
 
+  public static ErrorMessageFactory shouldHaveRootCause(Throwable actualCause) {
```
joel-costigliola: add a unit test in `ShouldHaveRootCause_create_Test`

src/main/java/org/assertj/core/error/ShouldHaveCause.java
```diff
@@ -31,6 +31,10 @@ public static ErrorMessageFactory shouldHaveCause(Throwable actualCause, Throwab
     return new ShouldHaveCause(actualCause, expectedCause);
   }
 
+  public static ErrorMessageFactory shouldHaveCause(Throwable actualCause) {
```
joel-costigliola: Add a unit test in a new  test class `ShouldHaveCause_create_Test`, take `ShouldHaveRootCause_create_Test` as an example (sorry `ShouldHaveCause_create_Test` should have already existed!)

src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
```diff
@@ -158,6 +158,30 @@ public SELF hasNoCause() {
...
   }
 
+  /**
+   * Verifies that the actual {@code Throwable} has a cause.
+   *
+   * @return this assertion object.
+   * @throws AssertionError if the actual {@code Throwable} is {@code null}.
+   * @throws AssertionError if the actual {@code Throwable} does not have a cause.
+   */
+  public SELF hasCause() {
```
scordio: What if we add `getCause()` instead, which checks if there is a cause and returns a new `ThrowableAssert`? This would be symmetrical with `havingCause()`, which could become:

```java
  public ThrowableAssertAlternative<?> havingCause() {
    AbstractThrowableAssert<?, ?> causeAssert = delegate.getCause();
    return new ThrowableAssertAlternative<>(causeAssert.actual);
  }
```

Same would be for `getRootCause()` instead of `hasRootCause()`.

I propose this option also because `hasCause()` would have a slightly different purpose compared to the existing `hasCause(Throwable)`, i.e., checking the cause existence vs. comparison with the input value.

src/test/java/org/assertj/core/api/ThrowableAssertAlternative_havingCause_Test.java
```diff
@@ -0,0 +1,30 @@
...
+    new ThrowableAssertAlternative<>(throwable)
+      .havingCause()
+      .withMessage("cause message");
+  }
+
+  @Test
+  void should_fail_if_throwable_has_no_cause() {
+    //WHEN
+    ThrowableAssert.ThrowingCallable code = () -> {
+      Throwable throwable = new Throwable("top level message");
```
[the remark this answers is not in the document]
wimdeblauwe: Do you mean this?

```
  void should_fail_if_throwable_has_no_cause() {
    //GIVEN
    Throwable throwable = new Throwable("top level message");
    ThrowableAssertAlternative<Throwable> taa = new ThrowableAssertAlternative<>(throwable);
    //WHEN
    ThrowableAssert.ThrowingCallable code = taa::havingCause;
    //THEN
    AssertionsUtil.expectAssertionError(code);//.withMessage("expecting java.lang.Throwable: top level message to have a cause but it did not");
  }
```

But how do I check the assertion message?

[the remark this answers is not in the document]
wimdeblauwe: Ignore the above, found it.

src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
wimdeblauwe: Would it not be better to _add_ the `getCause()` method in `AbstractThrowableAssert` on top? In that case, it would be like this:

```
  public ThrowableAssert getCause() {
    hasCause();
    return new ThrowableAssert(actual.getCause());
  }
```

And `havingCause` in `ThrowableAssertAlternative` becomes:

```
  public ThrowableAssertAlternative<?> havingCause() {
    ThrowableAssert causeAssert = delegate.getCause();
    return new ThrowableAssertAlternative<>(causeAssert.actual);
  }
```

I return a `ThrowableAssert` since that is the type I am using. Returning `AbstractThrowableAssert<?,?>` like you propose also works, but not sure why that would be better?

scordio: > Would it not be better to add the getCause() method in AbstractThrowableAssert on top?

Looks good to me!

> Returning AbstractThrowableAssert<?,?> like you propose also works, but not sure why that would be better?

IIRC this is a pattern used in AssertJ to allow subclasses to return a different assertion type. @joel-costigliola what do you think about this?

joel-costigliola: yes `getCause` in `AbstractThrowableAssert` would be better as `havingCause()` looks a bit like to `hasCause(cause)` which may confuse users.

```java
assertThat(e).getCause()
             .hasMessage("cause message")
             .getRootCause()
             .hasMessage("root cause message");
```

We can return `AbstractThrowableAssert<?,?>` in case someone wants their own `MyThrowableAssert` which only need to extend `AbstractThrowableAssert<?,?>` (to be fair I don't know if people are doing that or not)

wimdeblauwe: Ok. PR updated for this now.

joel-costigliola: @wimdeblauwe we also need to add `"getCause"` and `"getRootCause"`  to `SoftProxies.METHODS_CHANGING_THE_OBJECT_UNDER_TEST` for soft assertions to work properly.

A test must be added in `SoftAssertionsTest` like `map_soft_assertions_should_work_with_navigation_methods` but for throwable (name it `throwable_soft_assertions_should_work_with_navigation_methods`).

We want to prove in that test that after navigating to cause and root cause soft assertions work and capture properly all errors.

wimdeblauwe: Thanks for the tip. Test is added. It failed without updating `SoftProxies` and adding both methods made it work.

wimdeblauwe:
I have addressed all remarks so far. I probably need to update documentation as well. Can you give me some pointers about that?

joel-costigliola:
@wimdeblauwe yes that would be great!
You can create a PR on assertj/doc and add a _Checking cause and root cause_ section like `====== Checking the exception message` in  https://assertj.github.io/doc/#assertj-core-exception-assertions-reference (don't worry about the title style, https://github.com/assertj/doc/issues/33 is going to address it)

joel-costigliola:
Now that we have `getCause` and `getRootCause`, @wimdeblauwe @scordio do you think we still need  `hasCause()` and `hasRootCause()`?

wimdeblauwe:
> Now that we have `getCause` and `getRootCause`, @wimdeblauwe @scordio do you think we still need `hasCause()` and `hasRootCause()`?

I have no preference, but if you decide to remove it, I would make it private methods for clarity.

wimdeblauwe:
Documentation PR can be found at https://github.com/assertj/doc/pull/34

scordio:
> Now that we have `getCause` and `getRootCause`, @wimdeblauwe @scordio do you think we still need `hasCause()` and `hasRootCause()`?

I would keep `getCause()` and not `hasCause()`, or if we want to keep both I'd rename `hasCause(Throwable)` to something else to highlight the different behavior (e.g., `hasCauseLike(Throwable)` or `hasCauseSimilarTo(Throwable)`).

Same for `getRootCause()` vs `hasRootCause()`.

joel-costigliola:
I'm currently integrating the PR, as part of it I have inlined `hasCause()` and `hasRootCause()`.

joel-costigliola:
Integrated thanks @wimdeblauwe!

PR #1784, opened 2020-02-12
Soft Assertion Extension

bibibiu2017:
#### Check List:
* Fixes NO
* Unit tests : YES 
* Javadoc with a code example (on API only) :  NO


joel-costigliola:
Interesting thanks @bibibiu2017, we will have a look at it shortly!

src/test/java/org/assertj/core/extention/SoftlyExtensionExecutionTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+@DisplayName("SoftlyExtension runs")
+@ExtendWith({SoftlyExtension.class, MockitoExtension.class})
+class SoftlyExtensionExecutionTest {
+
+  @Softly(assertAfter = true)
+  @InjectMocks
+  private SoftAssertions softlyAssertAfterTrue;
+
+  @Softly(assertAfter = false)
```
fduminy: I don't see the interest of this use case

src/test/java/org/assertj/core/extention/SoftlyExtensionExecutionTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import static org.assertj.core.util.AssertionsUtil.assertThatAssertionErrorIsThrownBy;
+
+@DisplayName("SoftlyExtension runs")
+@ExtendWith({SoftlyExtension.class, MockitoExtension.class})
+class SoftlyExtensionExecutionTest {
+
+  @Softly(assertAfter = true)
+  @InjectMocks
```
fduminy: Why do you add this mockito annotation ?

src/test/java/org/assertj/core/extention/SoftlyExtensionInstanceTest.java
```diff
@@ -0,0 +1,73 @@
...
+class SoftlyExtensionInstanceTest {
+
+  private SoftAssertions softlyNotAnnotated1;
+  private SoftAssertions softlyNotAnnotated2;
+
+  @Softly
+  private SoftAssertions softlyAnnotated1;
+
+  @Softly
+  private SoftAssertions softlyAnnotated2;
```
fduminy: I don't see a use case where more than one SoftAssertions is needed.

fduminy:
@bibibiu2017 Your approach of using an annotation on a field is interesting (like spring's @AutoWired) but I prefer constructor injection like in my PR #1771. 

Anyway, as I said to @joel-costigliola in my PR, I think we should support different styles of junit 5 tests. So, we should merge your and my PR. What do you think ?

joel-costigliola:
@fduminy why do you think construction injection is better?

If we have a simple `@Softly` field annotation (without parameter), is that simpler as you don't even need to define a constructor.

src/test/java/org/assertj/core/extention/SoftlyExtensionExecutionTest.java
bibibiu2017: This is a mistake that I will rectify. I was trying to write a test that would assert that #assertAll () would be called after the test run. I was thinking of somehow using Mockito to get the test instance then verify that the method was called. But it was not possible. Is it possible to test this?

src/test/java/org/assertj/core/extention/SoftlyExtensionInstanceTest.java
bibibiu2017: Me too, I just put it there because I thought someone might need it. But I can revert to an earlier implementation where I threw an exception when more that one field had the `@Softly` annotation.

src/test/java/org/assertj/core/extention/SoftlyExtensionExecutionTest.java
bibibiu2017: Someone might just need the `@Sofly` field injections but they want control of when `#assertAll()` is invoked .

fduminy:
@joel-costigliola I didn't say "it's better" but "I prefer". That preference was due to my experiences with spring framework but the context is different.

If we must choose between constructor vs field injection, here are some points to help :  
- constructor injection 
  * more consistent with existing `SoftAssertionsExtension` (doing method injection) since a constructor is only a special method
- field injection
  * closer to junit 4 since one need to replace `@Rule JUnitSoftAssertions softly = new JUnitSoftAssertions();` by `@Softly SoftAssertions softly;`
  * since I don't see any usage of `@Softly`, I would even make it simpler by removing this annotation

src/test/java/org/assertj/core/extention/SoftlyExtensionExecutionTest.java
fduminy: I don't see any use case and it's error prone since the user might forget to call `assertAll` (or remove it later by error) : **that would give false positive where a test is passing instead of failing**.

bibibiu2017: I think the annotation has some value this is because it shows intent. I think for field injection intent is very important. For example, when using field dependency injection, we must provide the `@Autowired` annotation. Also, I was inspired by Mockito's `@Captor` annotation.

src/test/java/org/assertj/core/extention/SoftlyExtensionInstanceTest.java
joel-costigliola: I agree, let's remove one of the annotated field and rename the other as `softly`

src/test/java/org/assertj/core/extention/SoftlyExtensionInstanceTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.util.AssertionsUtil.assertThatAssertionErrorIsThrownBy;
+
+@ExtendWith(SoftlyExtension.class)
+@DisplayName("SoftlyExtension hasInstance")
+class SoftlyExtensionInstanceTest {
+
+  private SoftAssertions softlyNotAnnotated1;
+  private SoftAssertions softlyNotAnnotated2;
```
joel-costigliola: Keep only one non annotated field only

src/test/java/org/assertj/core/extention/SoftlyExtensionInstanceTest.java
```diff
@@ -0,0 +1,73 @@
...
+      add(softlyNotAnnotated1);
+      add(softlyNotAnnotated2);
+    }};
+
+    //WHEN
+    ThrowingCallable code = () -> assertThat(softAssertions).doesNotContainNull();
+
+    //THEN
+    assertThatAssertionErrorIsThrownBy(code).withMessage(ShouldNotContainNull.shouldNotContainNull(softAssertions).create());
+
```
joel-costigliola: minor: remove empty lines

src/test/java/org/assertj/core/extention/SoftlyExtensionExecutionTest.java
joel-costigliola: I don't see the point of `assertAfter = false`, I mean if people want to control soft assertions verification phase, they can just create a regular SoftAssertions with `new` and call `assertAll()` when needed.

src/main/java/org/assertj/core/extention/SoftlyExtension.java
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * <p>
+ * Inspired by Mockito @Captor annotation.
+ * </p>
+ * @author Arthur Mita
+ **/
+public class SoftlyExtension implements TestInstancePostProcessor, BeforeTestExecutionCallback, AfterTestExecutionCallback {
+
+  private static final Logger LOGGER = Logger.getLogger(SoftlyExtension.class.getName());
+  private static final Namespace ASSERT_J = Namespace.create("org.assertj.core");
```
joel-costigliola: Rename to `ASSERTJ_CORE`

src/main/java/org/assertj/core/extention/SoftlyExtension.java
```diff
@@ -0,0 +1,104 @@
...
+    extensionContext.getStore(ASSERT_J).put(TEST_INSTANCE, testInstance);
+  }
+
+  @Override
+  public void beforeTestExecution(ExtensionContext extensionContext) throws Exception {
+    Object testInstance = extensionContext.getStore(ASSERT_J).get(TEST_INSTANCE);
+    LOGGER.log(Level.INFO,
+      "\n==========================================================" +
+        "\nGetting Getting SoftAssertion Field annotated with @Softly" +
+        "\n==========================================================");
```
joel-costigliola: remove this log as this is the happy path, no need to pollute the console when things are working as expected

src/main/java/org/assertj/core/extention/SoftlyExtension.java
```diff
@@ -0,0 +1,104 @@
...
+      return;
+    }
+
+    for (Field field : softlyFields) {
+      if (field.getType() != SoftAssertions.class) {
+        throw failures.failure("@Softly field must be of the type " + SoftAssertions.class.getName());
+      }
+      field.setAccessible(true);
+      field.set(testInstance, new SoftAssertions());
+
```
joel-costigliola: remove empty line

src/main/java/org/assertj/core/extention/SoftlyExtension.java
```diff
@@ -0,0 +1,104 @@
...
+ * Inspired by Mockito @Captor annotation.
+ * </p>
+ * @author Arthur Mita
+ **/
+public class SoftlyExtension implements TestInstancePostProcessor, BeforeTestExecutionCallback, AfterTestExecutionCallback {
+
+  private static final Logger LOGGER = Logger.getLogger(SoftlyExtension.class.getName());
+  private static final Namespace ASSERT_J = Namespace.create("org.assertj.core");
+  private static final String TEST_INSTANCE = "testInstance";
+  private final List<Field> fields = new ArrayList<>();
```
joel-costigliola: I think it should be only one field, not sure why we would need several

fduminy:
@joel-costigliola Shouldn't that MR be merged into existing class org.assertj.core.api.junit.jupiter.SoftAssertionsExtension ?

src/main/java/org/assertj/core/extention/SoftlyExtension.java
bibibiu2017: A lot of the inspiration came from Mockito's `@Captor` which allows one to have more that one `ArgumentCapture` field but then `ArgumentCapture` accepts generics, so it makes sense to have more than one instance. I will remove it.

src/main/java/org/assertj/core/extention/SoftlyExtension.java
bibibiu2017: OK

src/main/java/org/assertj/core/extention/SoftlyExtension.java
bibibiu2017: OK

src/main/java/org/assertj/core/extention/SoftlyExtension.java
bibibiu2017: OK

src/main/java/org/assertj/core/extention/SoftlyExtension.java
```diff
@@ -0,0 +1,112 @@
...
+    extensionContext.getStore(ASSERTJ_CORE).put(TEST_INSTANCE, testInstance);
+  }
+
+  @Override
+  public void beforeEach(ExtensionContext extensionContext) throws Exception {
+    Optional<ExtensionContext> currentContext = Optional.of(extensionContext);
+
+    while (currentContext.isPresent()) {
+      ExtensionContext context = currentContext.get();
+      Object testInstance = context.getStore(ASSERTJ_CORE).remove(TEST_INSTANCE);
```
joel-costigliola: why removing it instead of simply getting it? is there some concurrency issue I'm not seeing?

src/main/java/org/assertj/core/extention/SoftlyExtension.java
```diff
@@ -0,0 +1,112 @@
...
+          if (softlyField.getType() != SoftAssertions.class) {
+            throw failures.failure("@Softly field must be of the type " + SoftAssertions.class.getName());
+          }
+          softlyField.setAccessible(true);
+          softlyField.set(testInstance, new SoftAssertions());
+
+          this.softlies.add((SoftAssertions) softlyField.get(testInstance));
+        }
+      }
+      currentContext = context.getParent();
```
joel-costigliola: is that for nested tests?

bibibiu2017: Yes. Nested test classes that have parents whose test instance lifecycle is not `LifeCycle.PER_CLASS` cannot use the SoftAssertions field declared in the parent cause they do not share the same instance. One would get an `IllegalArgumentException`

src/main/java/org/assertj/core/extention/SoftlyExtension.java
bibibiu2017: I figured we do not need the instance since we have already had the `SoftAssertions` instance set. There was no concurrency issue that I could detect.

src/main/java/org/assertj/core/extention/SoftlyExtension.java
```diff
@@ -0,0 +1,112 @@
...
+ * </p>
+ *
+ * @author Arthur Mita
+ **/
+public class SoftlyExtension implements TestInstancePostProcessor, BeforeEachCallback, AfterTestExecutionCallback {
+
+  private static final Logger LOGGER = Logger.getLogger(SoftlyExtension.class.getName());
+  private static final Namespace ASSERTJ_CORE = Namespace.create("org.assertj.core");
+  private static final String TEST_INSTANCE = "testInstance";
+  private final Set<SoftAssertions> softlies = new HashSet<>();
```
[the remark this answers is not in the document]
joel-costigliola: @bibibiu2017 any reason for this?

src/main/java/org/assertj/core/extention/SoftlyExtension.java
joel-costigliola: could we use https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/ExtensionContext.html#getTestInstance() instead of accessing the store?

joel-costigliola: If I understand correctly context are per tests (can access the parent context if needed), so in BeforeTestExecutionCallback we could simply store in the built soft assertion in the current ExtensionContext  and use it in afterTestExecution, WDYT? 
execution

bibibiu2017: Yes I think that would be cleaner. I think the store should be used to store objects that need to be accessed across multiple test instances which is not our current use case. Let me work on this.

src/main/java/org/assertj/core/annotations/Softly.java
```diff
@@ -0,0 +1,32 @@
...
+ * Signifies that a {@link SoftAssertions} field in test
+ * class should be managed by {@link SoftlyExtension}
+ *
+ * @author Arthur Mita
+ * @see SoftlyExtension
+ */
+@Target({ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface Softly {
```
joel-costigliola: I'm playing with a solution that does not require this annotation, after all to find the field that is a SoftAssertions you can simply check the field type. I'll see what I can do with that approach

joel-costigliola:
@bibibiu2017 I have been thinking about this feature and unfortunately I believe there is a fundamental issue: if we run tests in parallel, the soft assertions field is going to be reset while test instances are being initialized, leading to mix soft assertions errors collected.

The current soft assertions extension avoids this issue by using parameter injection, guaranteeing that the provided soft assertions instance is not shared between tests.

I would be happy if you can prove me wrong.

@fduminy I think this issue is also relevant for constructor injection #1771.

joel-costigliola:
@bibibiu2017 I'm closing this issue because of the concurrent issue when running tests in parallel.
I will be happy to reopen it if we find a way to avoid that issue.

joel-costigliola:
It seems we could use that approach except for classes combining `Lifecycle.PER_CLASS` with `@Execution(CONCURRENT)` (https://github.com/joel-costigliola/assertj-core/pull/1771#issuecomment-596756422)

kriegfrj:
> It seems we could use that approach except for classes combining `Lifecycle.PER_CLASS` with `@Execution(CONCURRENT)` ([#1771 (comment)](https://github.com/joel-costigliola/assertj-core/pull/1771#issuecomment-596756422))

Actually, the same comments apply here as apply in https://github.com/joel-costigliola/assertj-core/pull/1771#issuecomment-608992674 - `Lifecycle.PER_CLASS` will be problematic for field injection, regardless of whether it's single threaded or concurrent. Again, we can throw a configuration exception of some sort if we detect that we're being asked to inject a `SoftAssertions` object into a test instance with `PER_CLASS` lifecycle semantics, which (along with decent Javadoc) will avoid any nasty surprises.

bibibiu2017:
`java.lang.instrument.IllegalClassFormatException: Error while instrumenting java/sql/Timestamp.` is this an issue with the pull request I just made?

joel-costigliola:
Thanks @bibibiu2017, I have just pushed a commit with a refactored `SoftlyExtension` approach without annotating fields, you get the commit credit since it was your idea ;-).

Feel free to play with it and to let me know if you see some possible improvements.

thanks again.

PR #1789, opened 2020-02-17
No need to use FQN for imported names

pzygielo:

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -20,7 +20,6 @@ import static org.assertj.core.error.ShouldNotBeNull.shouldNotBeNull;
 import static org.assertj.core.util.Preconditions.checkArgument;
 import static org.assertj.core.util.Strings.formatIfArgs;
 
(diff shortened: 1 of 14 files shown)


joel-costigliola:
Integrated thanks @pzygielo!

pzygielo:
Thanks for checking.
I see some leftovers however: https://github.com/joel-costigliola/assertj-core/compare/master...pzygielo:fqn
Are they missed or kept on purpose?
GIVEN is already imported and used without class qualification in other places in that file.
The same for Arrays.stream.

joel-costigliola:
I left them as I felt the code was less clear with GIVEN only and for stream it would have changed the code formatting not to align the chained methods.

pzygielo:
> I left them as I felt the code was less clear with GIVEN only

I see. But for example in the same file - https://github.com/joel-costigliola/assertj-core/blob/63caf5d81afcef58db11717a1eff14e80c797f60/src/main/java/org/assertj/core/api/AbstractPredicateAssert.java#L65 is clear enough, right?

joel-costigliola:
good point, I'll change that for consistency

Issue #1270, opened 2018-07-06
ByteArrayAssert should provide the specific implementation for isEquals/containsExactly


luengnat:
#### Summary
ByteArrayAssert currently simply calls StandardComparisonStrategy.areEqual where it can be very slow to compare large byte array.

#### Example
The slowest part is Array.get call.
```java
	at java.lang.reflect.Array.get(Native Method)
	at org.assertj.core.util.Objects.areEqualArrays(Objects.java:59)
	at org.assertj.core.util.Objects.areEqual(Objects.java:44)
	at org.assertj.core.internal.StandardComparisonStrategy.areEqual(StandardComparisonStrategy.java:75)
	at org.assertj.core.internal.Objects.areEqual(Objects.java:348)
	at org.assertj.core.internal.Objects.assertEqual(Objects.java:319)
	at org.assertj.core.api.AbstractByteArrayAssert.containsExactly(AbstractByteArrayAssert.java:662)
```

ardmn:
Hello @luengnat . I think it works same as ```containsExactly``` for int array or same as calling to ```arrays. assertContainsExactly()``` . Have you any proposals ? :)

luengnat:
@ardmn it's just a performance issue where Array.get() is a reflection call into native call which can be quite slow as it involves boxing

ardmn:
@luengnat if performance is critical for you can override assertion's methods . It is not nice but it works. Unfortunately, current assertj's api  does not provide the possibility to change/put/insert custom implementation for asserting . I mean that user can't use org.assertj.core.api.Assertions with custom org.assertj.core.api.ByteArrayAssert.

```java
import org.assertj.core.api.AbstractByteArrayAssert;
import org.assertj.core.api.ByteArrayAssert;
import org.assertj.core.internal.AbstractComparisonStrategy;
import org.assertj.core.internal.Failures;
import org.assertj.core.internal.StandardComparisonStrategy;
import org.junit.Test;

import java.util.Arrays;

import static org.assertj.core.error.ShouldBeEqual.shouldBeEqual;

public class MyByteArrayAssertTest {

    @Test
    public void test_isEqualTo_true() {
        MyByteArrayAssert.assertThat(new byte[]{1, 2, 3})
                .isEqualTo(new byte[]{1, 2, 3});
    }

    @Test
    public void test_isEqualTo_false() {
        MyByteArrayAssert.assertThat(new byte[]{1, 2, 3})
[...]

PR #1794, opened 2020-03-01
1270 primitive array consistency

masterav10:
#### Check List:
* Prerequisite to #1270
* Unit tests : YES

While working on 1270, I noticed some inconsistencies in the array implementations. To prevent change overload, here are my proposed changes to the abstract array base classes in support for 1270.

Changes so far:
- [X] AbstractByteArrayAssert.containsExactly.containsExactly() should use arrays internally
- [X] AbstractShortArrayAssert added int methods for like the byte array version

Potential other options I wasn't planning on doing:
- AbstractCharArrayAssert add int methods
- AbstractLongArrayAssert add int methods
- AbstractFloatArrayAssert add double methods


src/test/java/org/assertj/core/api/shortarray/ShortArrayAssertNullTest.java
```diff
@@ -0,0 +1,28 @@
...
+ * Should be used for methods with an int[] or int... parameter.
+ * 
+ * @author Dan Avila
+ */
+public abstract class ShortArrayAssertNullTest extends ShortArrayAssertBaseTest {
+
+  @Test
+  public void should_throw_exception_on_null_argument() {
+    assertThatExceptionOfType(NullPointerException.class).isThrownBy(() -> {
+      int[] content = null;
```
joel-costigliola: Let's use `catchThrowableOfType` and the GIVEN WHEN THEN pattern to make the test easier to read.  
shoud look like:
```java
    // GIVEN
    int[] nullContent = null;
    // WHEN
    NullPointerException npe = catchThrowableOfType(() -> invoke_api_with_null_value(assertThat(new short[] {}), nullContent),
                                                    NullPointerException.class);
    // THEN
    then(npe).hasMessage("The array of values to look for should not be null");
```

src/test/java/org/assertj/core/api/shortarray/ShortArrayAssert_containsAnyOf_with_Integer_Argument_Test.java
```diff
@@ -0,0 +1,45 @@
...
+import org.assertj.core.api.AbstractShortArrayAssert;
+import org.assertj.core.api.ShortArrayAssert;
+import org.junit.jupiter.api.DisplayName;
+
+/**
+ * Tests for <code>{@link ShortArrayAssert#containsAnyOf(int...)}</code>.
+ * 
+ * @author Dan Avila
+ */
+@DisplayName("ShortArrayAssert containsAnyOf (ints)")
```
joel-costigliola: nice!

masterav10: Glad you liked it.

masterav10:
Should be good to go. I've made pretty good progress on the parent branch as well.

joel-costigliola:
@masterav10 I'm not sure what you mean by the parent branch but this PR looks good so I'm integrating it.

PR #1802, opened 2020-03-06
ByteArrayAssert.containsExactly(byte...) error message will mention …

### LINKED ISSUE #1801: ByteArrayAssert#containsExactly(byte...) error message should state which elements were not found and not expected
#### Summary
`AbstractByteArrayAssert#containsExactly(byte...)` produces a sub-optimal error message where the expected and actual values are printed out as memory addresses of those arrays, while I would expect a summary of which elements were not found and which were not expected within the potentially long array.

#### Example

```
org.opentest4j.AssertionFailedError: 
Expecting:
 <[1, 2, 4]>
to be equal to:
 <[1, 2, 3]>
but was not.
Expected :[B@764faa6
Actual   :[B@4c1f22f3
```
In case of `int` arrays the error message is much more helpful:
```
java.lang.AssertionError: 
Expecting:
  <[1, 2, 4]>
  <[1, 2, 3]>
  <[3]>
and others were not expected:
  <[4]>
```

From the source code it's visible that the `byte` array specific `containsExactly` method calls  `objects.assertEqual(info, actual, values)`, but the `int` array specific `containsExactly` method calls `arrays.assertContainsExactly(info, actual, arrays.toByteArray(values))`. 

It hints that the `byte` array specific `containsExactly` method should have also called  `arrays.assertContainsExactly(info, actual, values);`

#### Reproduction steps:
AssertJ version 3.15.0
```
assertThat(new byte[] { 1, 2, 4 }).containsExactly((byte) 1, (byte) 2, (byte) 3);
```

indrekpr:
… not found and not expected items (#1801)

#### Check List:
* Fixes #1801 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractByteArrayAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractByteArrayAssert.java
@@ -768,7 +768,7 @@ public abstract class AbstractByteArrayAssert<SELF extends AbstractByteArrayAsse
    *                              or values are the same but the order is not.
    */
   public SELF containsExactly(byte... values) {
-    objects.assertEqual(info, actual, values);
+    arrays.assertContainsExactly(info, actual, values);
     return myself;
   }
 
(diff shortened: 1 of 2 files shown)


joel-costigliola:
Thanks for the PR @indrekpr!

PR #1804, opened 2020-03-11
Support four arguments for satisfiesAnyOf()

Jakzi666:
In a recent project this was necessary/useful and since there was no particular reason not to support it other than a missing use case, this seemed a good reason to support it from now on.

 Fixes : NA
 Unit tests : YES
 Javadoc with a code example (on API only) : YES


joel-costigliola:
Nice PR thanks @Jakzi666!

Issue #1805, opened 2020-03-14
Base64 support for AbstractStringAssert


scordio:
#### Summary
Java 8 has built-in support for Base64 with `java.util.Base64`.

What about adding decoding assertions to `AbstractStringAssert`?

#### Example

```java
assertThat("THVrZSBTa3l3YWxrZXI=")
  .isValidBase64()  // relies on java.util.Base64.getDecoder().decode(String) which throws IllegalArgumentException in case of wrong input
  .decodedAsBase64() // Changes the object under assertion, returning a new ByteArrayAssert
  ... // additional byte array assertions
```

joel-costigliola:
Fine by me!

PR #1811, opened 2020-03-21
Add isValidBase64 to AbstractStringAssert

scordio:
#### Check List:
* Fixes partially #1805
* Unit tests : YES

### DIFF
--- a/src/test/java/org/assertj/core/error/ShouldBeValidBase64_create_Test.java
+++ b/src/test/java/org/assertj/core/error/ShouldBeValidBase64_create_Test.java
@@ -14,7 +14,7 @@ package org.assertj.core.error;
 
 import static java.lang.String.format;
 import static org.assertj.core.api.BDDAssertions.then;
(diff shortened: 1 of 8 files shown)


src/test/java/org/assertj/core/error/ShouldBeValidBase64_create_Test.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @author Stefano Cordio
+ */
+@DisplayName("ShouldBeValidBase64 create")
+class ShouldBeValidBase64_create_Test {
+
+  @Test
+  void should_create_error_message() {
+    // WHEN
+    String message = shouldBeValidBase64("string").create(new TestDescription("TEST"), STANDARD_REPRESENTATION);
```
joel-costigliola, line 37: add a `%` to make sure this does not break a later String.format (this should be fine but we had quite few of those)

PR #1812, opened 2020-03-22
Add LongAdderAssert

pivovarit:


pivovarit:
Waiting for some initial feedback and then I will add examples

joel-costigliola:
@pivovarit  thanks for the PR. Hope you are doing ok in these crazy times 🤞

src/test/java/org/assertj/core/api/atomic/longadder/LongAdderAssert_isNull_Test.java
```diff
@@ -0,0 +1,30 @@
...
+package org.assertj.core.api.atomic.longadder;
+
+import org.junit.jupiter.api.Test;
+
+import java.util.concurrent.atomic.LongAdder;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.BDDAssertions.then;
+
+class LongAdderAssert_isNull_Test {
```
scordio: Any particular reason why you want to test `isNull()`? `LongAdderAssert` doesn't ovverride it from `AbstractAssert` so I'd probably skip it

src/main/java/org/assertj/core/api/LongAdderAssert.java
```diff
@@ -0,0 +1,462 @@
...
+
+import static org.assertj.core.error.ShouldHaveValue.shouldHaveValue;
+import static org.assertj.core.error.ShouldNotContainValue.shouldNotContainValue;
+
+/**
+ * Assertions for {@link java.util.concurrent.atomic.LongAdder}.
+ *
+ * @author Grzegorz Piwowarek
+ */
+public class LongAdderAssert extends AbstractAssert<LongAdderAssert, LongAdder> {
```
scordio: Being `LongAdder` a `java.lang.Number`, should `LongAdderAssert` implement `NumberAssert`?

pivovarit: Good idea, I made it happen: https://github.com/joel-costigliola/assertj-core/pull/1812/commits/74c66ab398be8cf2d347651b01f8758216d75abd

src/test/java/org/assertj/core/api/atomic/longadder/LongAdderAssert_isNull_Test.java
pivovarit: The `isNull` assertion or just the test?

scordio: Just the test, the assertion should come from `AbstractAssert` and it is covered by `AbstractAssert_isNull_Test`

src/main/java/org/assertj/core/api/AbstractLongAdderAssert.java
```diff
@@ -0,0 +1,552 @@
...
+   * @param startInclusive the start value (inclusive).
+   * @param endInclusive   the end value (inclusive).
+   *
+   * @return this assertion object.
+   *
+   * @throws AssertionError if the actual adder is {@code null}.
+   * @throws AssertionError if the actual sum value is not in [start, end] range.
+   * @since 2.16.0 / 3.16.0
+   */
+  public LongAdderAssert hasValueBetween(long startInclusive, long endInclusive) {
```
scordio: The return type should be `SELF` (same for the other assertion methods)
```suggestion
  public SELF hasValueBetween(long startInclusive, long endInclusive) {
```

scordio: Considering that now the class implements `NumberAssert`, do you still want to keep the `hasValue` variants? (e.g., `hasValueBetween` vs `isBetween`)

src/main/java/org/assertj/core/api/AbstractLongAdderAssert.java
```diff
@@ -0,0 +1,552 @@
...
+/**
+ * Base class for all implementations of assertions for {@link LongAdder}s.
+ *
+ * @param <SELF> the "self" type of this assertion class. Please read &quot;<a href="http://bit.ly/1IZIRcY"
+ *               target="_blank">Emulating 'self types' using Java Generics to simplify fluent API implementation</a>&quot;
+ *               for more details.
+ *
+ * @author Grzegorz Piwowarek
+ */
+class AbstractLongAdderAssert<SELF extends AbstractLongAdderAssert<SELF>> extends AbstractAssert<LongAdderAssert, LongAdder> implements NumberAssert<LongAdderAssert, Long> {
```
scordio: I think this should be:
```suggestion
public class AbstractLongAdderAssert<SELF extends AbstractLongAdderAssert<SELF>> extends AbstractAssert<SELF, LongAdder> implements NumberAssert<SELF, Long> {
```

src/main/java/org/assertj/core/api/AbstractLongAdderAssert.java
```diff
@@ -0,0 +1,552 @@
...
+   *
+   * @return this assertion object.
+   *
+   * @throws AssertionError if the actual adder is {@code null}.
+   * @throws AssertionError if the actual sum value is not in [start, end] range.
+   * @since 2.16.0 / 3.16.0
+   */
+  public LongAdderAssert hasValueBetween(long startInclusive, long endInclusive) {
+    isNotNull();
+    longs.assertIsBetween(getWritableAssertionInfo(), actual.sum(), startInclusive, endInclusive);
```
scordio: You can access `info` directly
```suggestion
    longs.assertIsBetween(info, actual.sum(), startInclusive, endInclusive);
```

src/main/java/org/assertj/core/api/AbstractLongAdderAssert.java
pivovarit: Yes, good catch!

src/main/java/org/assertj/core/api/AbstractLongAdderAssert.java
pivovarit: Not really, feel like they are redundant now. I will go over the API and GC what's no longer needed

pivovarit: Done (https://github.com/joel-costigliola/assertj-core/pull/1812/commits/c6a164897868d351c3ed6b0ce2b197e8a313c928)

However, I believe we could remove even more custom methods if there was a way to access `ComparableAssert` assertions. 

Do you have any workarounds for objects that are comparable but don't implement `Comparable`?

scordio: Assuming `longValue()` or `sum()` of `LongAdder` are always used before comparison, the class could also implement `ComparableAssert<SELF, Long>` and have an additional field:

```java
Comparables comparables = new Comparables();
```

Would this help?

src/main/java/org/assertj/core/api/AbstractLongAdderAssert.java
```diff
@@ -0,0 +1,331 @@
...
+/**
+ * Base class for all implementations of assertions for {@link LongAdder}s.
+ *
+ * @param <SELF> the "self" type of this assertion class. Please read &quot;<a href="http://bit.ly/1IZIRcY"
+ *               target="_blank">Emulating 'self types' using Java Generics to simplify fluent API implementation</a>&quot;
+ *               for more details.
+ *
+ * @author Grzegorz Piwowarek
+ */
+class AbstractLongAdderAssert<SELF extends AbstractLongAdderAssert<SELF>> extends AbstractAssert<SELF, LongAdder> implements NumberAssert<SELF, Long> {
```
scordio: Missing `public`

```suggestion
public class AbstractLongAdderAssert<SELF extends AbstractLongAdderAssert<SELF>> extends AbstractAssert<SELF, LongAdder> implements NumberAssert<SELF, Long> {
```

src/main/java/org/assertj/core/api/AbstractLongAdderAssert.java
```diff
@@ -0,0 +1,331 @@
...
+   * @param other the given value to compare the actual value to.
+   *
+   * @return {@code this} assertion object.
+   *
+   * @throws AssertionError if the actual adder is {@code null}.
+   * @throws AssertionError if the actual sum value is greater than the given one.
+   * @since 2.16.0 / 3.16.0
+   */
+  public SELF hasValueLessThanOrEqualTo(long other) {
+    isNotNull();
```
scordio: I think `isNotNull()` can be avoided as `longs.assertLessThanOrEqualTo()` calls `Objects.instance().assertNotNull()`.

Same for the other methods.

src/main/java/org/assertj/core/api/AbstractLongAdderAssert.java
```diff
@@ -0,0 +1,331 @@
...
+   *
+   * // assertion will fail
+   * assertThat(actual).hasValue(0);</code></pre>
+   *
+   * @param expectedValue the value not expected .
+   *
+   * @return {@code this} assertion object.
+   *
+   * @throws AssertionError if the actual adder is {@code null}.
+   * @throws AssertionError if the actual sum value is not non negative.
```
scordio: Maybe copy-paste typo?

src/main/java/org/assertj/core/api/AbstractLongAdderAssert.java
```diff
@@ -0,0 +1,331 @@
...
+import static org.assertj.core.error.ShouldNotContainValue.shouldNotContainValue;
+
+/**
+ * Base class for all implementations of assertions for {@link LongAdder}s.
+ *
+ * @param <SELF> the "self" type of this assertion class. Please read &quot;<a href="http://bit.ly/1IZIRcY"
+ *               target="_blank">Emulating 'self types' using Java Generics to simplify fluent API implementation</a>&quot;
+ *               for more details.
+ *
+ * @author Grzegorz Piwowarek
```
scordio: I would add `@since` on top of the class instead of the single methods. Also, AssertJ 2.x is in maintenance mode so it shouldn't receive new features.

```suggestion
 * @author Grzegorz Piwowarek
 * @since 3.16.0
```

src/main/java/org/assertj/core/api/AbstractLongAdderAssert.java
```diff
@@ -0,0 +1,331 @@
...
+   *
+   * // assertion will fail
+   * assertThat(actual).doesNotHaveValue(42);</code></pre>
+   *
+   * @param expectedValue the value not expected .
+   *
+   * @return {@code this} assertion object.
+   *
+   * @throws AssertionError if the actual adder is {@code null}.
+   * @throws AssertionError if the actual sum value is not non negative.
```
scordio: Maybe copy-paste typo?

src/main/java/org/assertj/core/api/AbstractLongAdderAssert.java
pivovarit: Sounds like a plan - I will give it a try and let you know!

pivovarit: That actually worked: https://github.com/joel-costigliola/assertj-core/pull/1812/commits/6bea60f84c9df94103a0ab4a9fbb2c3f217149fe

scordio:
@pivovarit thanks for the changes! And sorry for the delay, I should be able to go through them during the weekend

src/main/java/org/assertj/core/api/AbstractLongAdderAssert.java
```diff
@@ -0,0 +1,237 @@
...
+   * // assertion will fail
+   * assertThat(actual).hasValue(0);</code></pre>
+   *
+   * @param expectedValue the value not expected .
+   *
+   * @return {@code this} assertion object.
+   *
+   * @throws AssertionError if the actual adder is {@code null}.
+   */
+  public SELF hasValue(long expectedValue) {
```
scordio, line 66: Let's try to shoot on the remaining `hasValue`/`doesNotHaveValue` pair 🙂
Would it be too aggressive to replace them with `isEqualTo(Long)`/`isNotEqualTo(Long)` similarly to `AbstractLongAssert`, which delegate to `longs.assertEqual()`/`longs.assertNotEqual()` respectively?

@joel-costigliola thoughts?

scordio: As discussed with @joel-costigliola, let's keep `hasValue`/`doesNotHaveValue`.

scordio:
Integrated, thanks @pivovarit!

PR #1817, opened 2020-03-30
Restructure to allow combining soft assertions entry points

### LINKED ISSUE #1569: Composable SoftAssertions entry-point classes
### Summary

Soft assertions are good.

I like being able to use soft assertions with my custom exceptions in my code.

One of the selling points of AssertJ is the single entry point, which makes it easy to simply type assertThat() in your IDE and have it automatically give you the list of applicable assertions.

I'd like to be able to easily combine soft assertions entry classes from multiple sources. This is not currently possible as Java doesn't support multiple inheritance for classes. The workaround is clunky - each project must maintain its own soft assertions entry point with all of the assertThats in it that they want.

Switching to a model based on interfaces and Java 8 default methods would make multiple inheritance possible.

#### Current situation

```java
public class Project1SoftAssertions extends SoftAssertions {
    public MyCustomAssert assertThat(MyCustom a) {
        return proxy(MyCustomAssert.class, MyCustom.class, a);
    }
}

public class Project2SoftAssertions extends SoftAssertions {
    public MyCustom2Assert assertThat(MyCustom2 a) {
        return proxy(MyCustom2Assert.class, MyCustom2.class, a);
    }
}

// Can't do this - multiple inheritance not permitted for classes
public class UberSoftAssertions extends Project1SoftAssertions, Project2SoftAssertions {
}

// Have to do this - lots of repetition if you've got lots of assertions!
public class UberSoftAssertions extends SoftAssertions {
    public MyCustomAssert assertThat(MyCustom a) {
        return proxy(MyCustomAssert.class, MyCustom.class, a);
    }
    public MyCustom2Assert assertThat(MyCustom2 a) {
        return proxy(MyCustom2Assert.class, MyCustom2.class, a);
    }
}

// Also can't do this:
public class JUnitJupiterUberSoftAssertions
    extends JUnitJupiterSoftAssertions,
        Project1SoftAssertions, Project2SoftAssertions {}
```
#### Proposed

There will be a single functional interface that all soft assertions classes must implement. This has one default method that is used to generate the proxy classes. Then subinterfaces of this interface can call this method in their `default assertThat()` implementations. Because the soft assertion entry points can be interfaces, you can combine several of them into a single class.

```java

public interface SoftAssertionsBase {
    <V extends Assert<V, T>, T> V proxy(Class<V> assertClass, Class<T> actualClass, T actual);
}

// Existing implementation with the SoftAssertionsBase implementation added
public class AbstractSoftAssertions implements SoftAssertionsBase {
    // Note that there is also a TODO in the source for proxy() that adds the bound on V,
    // as per below.
    public <V extends Assert<V, T>, T> V proxy(Class<V> assertClass, Class<T> actualClass, T actual) {
        return proxies.createSoftAssertionProxy(assertClass, actualClass, actual);
    }
    ....
}

public interface StandardSoftAssertions extends SoftAssertionsBase {
    // All of the standard assertThat assertions go in here as default methods.
}

public interface Project1SoftAssertions extends SoftAssertionsBase {
    default MyCustomAssert assertThat(MyCustom a) {
        return proxy(MyCustomAssert.class, MyCustom.class, a);
    }
}

public class Project2SoftAssertions extends SoftAssertionsBase {
    public MyCustom2Assert assertThat(MyCustom2 a) {
        return proxy(MyCustom2Assert.class, MyCustom2.class, a);
    }
}

// Now we can do this
public class UberSoftAssertions extends AbstractStandardSoftAssertions
    implements Project1SoftAssertions, Project2SoftAssertions {
}

// And this
public class JUnitJupiterUberSoftAssertions
    extends JUnitJupiterSoftAssertions
    implements Project1SoftAssertions, Project2SoftAssertions {}

// And this, which allows us to get rid of that annoying "softly." prefix on all assertions:
public class MyTest extends SoftAssertions implements Project1SoftAssertions, Project2SoftAssertions {

    @AfterAll
    public void after() {
        assertAll();
    }
}
```

Please note that, although JDK < 1.8 is dying rapidly, nonetheless the above is still JDK <1.8 compatible because it's only the custom extension entry points that need to be interfaces with default methods. If you're pre JDK 1.8, simply define your custom extension entry points as classes (as you always have) and you're back where we started.

kriegfrj:
Changes soft assertions entry point implementations to be interfaces with default methods, rather than classes. This allows you to combine multiple of them into a single entry point class.

#### Check List:
* Fixes #1569
* Unit tests : YES
* Javadoc with a code example (on API only) : Not yet (will add if core team is happy with the implementation)

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
@@ -18,9 +18,10 @@ import static java.util.stream.Collectors.toList;
 import java.lang.reflect.Constructor;
 import java.util.List;
 
+import org.assertj.core.error.AssertionErrorCreator;
 import org.assertj.core.internal.Failures;
 
-public class AbstractSoftAssertions implements InstanceOfAssertFactories {
+public class AbstractSoftAssertions implements SoftAssertionsProvider, InstanceOfAssertFactories {
 
   protected final SoftProxies proxies;
 
@@ -28,32 +29,23 @@ public class AbstractSoftAssertions implements InstanceOfAssertFactories {
     proxies = new SoftProxies();
   }
 
+  private AssertionErrorCreator assertionErrorCreator = new AssertionErrorCreator();
+
+  @Override
+  public void assertAll() {
+    List<AssertionError> errors = assertionErrorsCollected();
+    if (!errors.isEmpty()) throw assertionErrorCreator.multipleSoftAssertionsError(errors);
+  }
+
+  @Override
   public <SELF extends Assert<? extends SELF, ? extends ACTUAL>, ACTUAL> SELF proxy(Class<SELF> assertClass,
                                                                                     Class<ACTUAL> actualClass, ACTUAL actual) {
     return proxies.createSoftAssertionProxy(assertClass, actualClass, actual);
   }
 
-  /**
-   * Catch and collect assertion errors coming from standard and <b>custom</b> assertions.
-   * <p>
-   * Example :
-   * <pre><code class='java'> SoftAssertions softly = new SoftAssertions();
-   * softly.check(() -&gt; Assertions.assertThat(…).…);
-   * softly.check(() -&gt; CustomAssertions.assertThat(…).…);
-   * softly.assertAll(); </code></pre>
-   *
-   * @param assertion an assertion call.
-   */
-  public void check(ThrowingRunnable assertion) {
-    try {
-      assertion.run();
-    } catch (AssertionError error) {
-      proxies.collectError(error);
-    } catch (RuntimeException runtimeException) {
-      throw runtimeException;
-    } catch (Exception exception) {
-      throw new RuntimeException(exception);
-    }
+  @Override
+  public void collectAssertionError(AssertionError error) {
+    proxies.collectError(error);
   }
 
   /**
@@ -125,7 +117,16 @@ public class AbstractSoftAssertions implements InstanceOfAssertFactories {
    * Returns a copy of list of soft assertions collected errors.
    * @return a copy of list of soft assertions collected errors.
    */
-  public List<AssertionError> errorsCollected() {
+  @Override
+  public List<AssertionError> assertionErrorsCollected() {
+    return decorateErrorsCollected(proxies.errorsCollected());
+  }
+
+  /**
+   * Returns a copy of list of soft assertions collected errors.
+   * @return a copy of list of soft assertions collected errors.
+   */
+  public List<Throwable> errorsCollected() {
     return decorateErrorsCollected(proxies.errorsCollected());
   }
 
@@ -224,9 +225,5 @@ public class AbstractSoftAssertions implements InstanceOfAssertFactories {
 
   private boolean isProxiedAssertionClass(String className) {
     return className.contains("$ByteBuddy$");
-  }
-
-  public interface ThrowingRunnable {
-    void run() throws Exception;
   }
 }
--- a/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
@@ -125,16 +125,17 @@ public class AbstractSoftAssertions implements InstanceOfAssertFactories {
    * Returns a copy of list of soft assertions collected errors.
    * @return a copy of list of soft assertions collected errors.
    */
-  public List<Throwable> errorsCollected() {
+  public List<AssertionError> errorsCollected() {
     return decorateErrorsCollected(proxies.errorsCollected());
   }
 
   /**
    * Modifies collected errors. Override to customize modification.
+   * @param <T> the supertype to use in the list return value
    * @param errors list of errors to decorate
    * @return decorated list
   */
-  protected List<Throwable> decorateErrorsCollected(List<Throwable> errors) {
+  protected <T extends Throwable> List<T> decorateErrorsCollected(List<? extends T> errors) {
     return addLineNumberToErrorMessages(errors);
   }
 
@@ -154,13 +155,13 @@ public class AbstractSoftAssertions implements InstanceOfAssertFactories {
     return proxies.wasSuccess();
   }
 
-  private List<Throwable> addLineNumberToErrorMessages(List<Throwable> errors) {
+  private <T extends Throwable> List<T> addLineNumberToErrorMessages(List<? extends T> errors) {
     return errors.stream()
                  .map(this::addLineNumberToErrorMessage)
                  .collect(toList());
   }
 
-  private Throwable addLineNumberToErrorMessage(Throwable error) {
+  private <T extends Throwable> T addLineNumberToErrorMessage(T error) {
     StackTraceElement testStackTraceElement = getFirstStackTraceElementFromTest(error.getStackTrace());
     if (testStackTraceElement != null) {
       try {
@@ -170,12 +171,14 @@ public class AbstractSoftAssertions implements InstanceOfAssertFactories {
     return error;
   }
 
-  private Throwable createNewInstanceWithLineNumberInErrorMessage(Throwable error,
-                                                                  StackTraceElement testStackTraceElement) throws ReflectiveOperationException {
-    Constructor<? extends Throwable> constructor = error.getClass().getConstructor(String.class, Throwable.class);
-    Throwable errorWithLineNumber = constructor.newInstance(buildErrorMessageWithLineNumber(error.getMessage(),
-                                                                                            testStackTraceElement),
-                                                            error.getCause());
+  private <T extends Throwable> T createNewInstanceWithLineNumberInErrorMessage(T error,
+                                                                                StackTraceElement testStackTraceElement) throws ReflectiveOperationException {
+    @SuppressWarnings("unchecked")
+    Constructor<? extends T> constructor = (Constructor<? extends T>) error.getClass().getConstructor(String.class,
+                                                                                                      Throwable.class);
+    T errorWithLineNumber = constructor.newInstance(buildErrorMessageWithLineNumber(error.getMessage(),
+                                                                                    testStackTraceElement),
+                                                    error.getCause());
     errorWithLineNumber.setStackTrace(error.getStackTrace());
     for (Throwable suppressed : error.getSuppressed()) {
       errorWithLineNumber.addSuppressed(suppressed);
--- a/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
@@ -28,7 +28,8 @@ public class AbstractSoftAssertions implements InstanceOfAssertFactories {
     proxies = new SoftProxies();
   }
 
-  public <T, V> V proxy(Class<V> assertClass, Class<T> actualClass, T actual) {
+  public <SELF extends Assert<? extends SELF, ? extends ACTUAL>, ACTUAL> SELF proxy(Class<SELF> assertClass,
+                                                                                    Class<ACTUAL> actualClass, ACTUAL actual) {
     return proxies.createSoftAssertionProxy(assertClass, actualClass, actual);
   }
 
--- a/src/main/java/org/assertj/core/api/SoftAssertionsRule.java
+++ b/src/main/java/org/assertj/core/api/SoftAssertionsRule.java
@@ -14,24 +14,13 @@ package org.assertj.core.api;
 
 import static org.assertj.core.api.SoftAssertionsStatement.softAssertionsStatement;
 
+import org.junit.rules.TestRule;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
 
-/**
- * @deprecated For Android compatible assertions use the latest assertj 2.x version which is based on Java 7 only.
- * <p>
- * JUnitSoftAssertions rule compatible with Android. Duplicated from {@link JUnitSoftAssertions}.
- *
- * @see JUnitSoftAssertions
- *
- * @since 2.5.0 / 3.5.0
- */
-@Deprecated
-public class Java6JUnitSoftAssertions extends AbstractSoftAssertions
-    implements Java6AbstractStandardSoftAssertions, SoftAssertionsRule {
-
+public interface SoftAssertionsRule extends SoftAssertionsProvider, TestRule {
   @Override
-  public Statement apply(final Statement base, Description description) {
+  default public Statement apply(final Statement base, Description description) {
     return softAssertionsStatement(this, base);
   }
 }
--- a/src/test/java/org/assertj/core/api/abstract_/SoftAssertionsMultipleProjectsTest.java
+++ b/src/test/java/org/assertj/core/api/abstract_/SoftAssertionsMultipleProjectsTest.java
@@ -1,0 +1,108 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2020 the original author or authors.
+ */
+package org.assertj.core.api.abstract_; // Make sure that package-private access is lost
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.List;
+
+import org.assertj.core.api.AbstractAssert;
+import org.assertj.core.api.AbstractSoftAssertions;
+import org.assertj.core.api.AbstractStandardSoftAssertions;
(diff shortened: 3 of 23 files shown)


kriegfrj:
One thing I'm not sold on is the name of the superinterface. `SoftAssertion` is too similar to `SoftAssertions`. I'm open to suggestions, but perhaps `SoftAssertionBase`? Thoughts @joel-costigliola ?

kriegfrj:
> One thing I'm not sold on is the name of the superinterface. `SoftAssertion` is too similar to `SoftAssertions`. I'm open to suggestions, but perhaps `SoftAssertionBase`? Thoughts @joel-costigliola ?

Actually, I meant `SoftAssertionsBase`. Another candidate is `ISoftAssertions`.

src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
```diff
@@ -18,43 +18,35 @@
...
+  private AssertionErrorCreator assertionErrorCreator = new AssertionErrorCreator();
+
+  @Override
+  public void assertAll() {
+    List<Throwable> errors = errorsCollected();
+    if (!errors.isEmpty()) throw assertionErrorCreator.multipleSoftAssertionsError(errors);
+  }
+  
+  @Override
+  public <V extends Assert<? extends V, ? extends T>, T> V proxy(Class<V> assertClass, Class<T> actualClass, T actual) {
```
joel-costigliola: use `ASSERT` instead of `V` and `ACTUAL` instead of `T` for readability

src/main/java/org/assertj/core/api/ThrowableCollector.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Copyright 2012-2020 the original author or authors.
+ */
+package org.assertj.core.api;
+
+import java.util.List;
+
+public interface ThrowableCollector {
+
+	void collectError(Throwable t);
```
joel-costigliola: the existing `ErrorCollector` has an `addError(Throwable error)` but actually it think it could have been `addError(AssertionError error)`. Let's try here to define void `collectAssertionError(AssertionError t);`.

rename `ThrowableCollector` to `AssertionErrorCollector` to be consistent.

src/main/java/org/assertj/core/api/ThrowableCollector.java
```diff
@@ -0,0 +1,23 @@
...
+
+	List<Throwable> errorsCollected();
```
joel-costigliola: Let's return `List<AssertionError>` and rename this method to `assertionErrorsCollected()`

src/main/java/org/assertj/core/api/SoftAssertion.java
```diff
@@ -0,0 +1,136 @@
...
+  * @throws MultipleFailuresError if possible or SoftAssertionError if any proxied assertion objects threw an {@link AssertionError}
+  * @since 3.16.0
+  */
+  public static <S extends SoftAssertion> void assertSoftly(Class<S> type, Consumer<S> softly) {
+    S assertions;
+	try {
+		assertions = type.newInstance();
+	} catch (InstantiationException | IllegalAccessException e) {
+		throw new RuntimeException(e);
+	}
```
joel-costigliola: formatting

src/main/java/org/assertj/core/api/SoftAssertion.java
```diff
@@ -0,0 +1,136 @@
...
+   * {@link #collectError(Throwable) collectError()}.
+   * 
+   * @param <V> The type of the assertion class
+   * @param <T> The type of the object-under-test
+   * @param assertClass Class instance for the assertion type.
+   * @param actualClass Class instance for the type of the object-under-test.
+   * @param actual The actual object-under-test. 
+   * @return A proxied assertion class for the given object-under-test.
+   */
+  <V extends Assert<? extends V, ? extends T>, T> V proxy(Class<V> assertClass, Class<T> actualClass, T actual);
```
joel-costigliola: use `ASSERT` instead of `V` and `ACTUAL` instead of `T`

src/main/java/org/assertj/core/api/SoftAssertion.java
```diff
@@ -0,0 +1,136 @@
...
+import org.opentest4j.MultipleFailuresError;
+
+/**
+ * Parent interface for soft assertion implementations.
+ * 
+ * @author Fr Jeremy Krieg
+ * @see AbstractSoftAssertions
+ * @see SoftAssertions
+ */
+public interface SoftAssertion extends ThrowableCollector {
```
joel-costigliola: my renaming suggestion to follow up on your comment:
* `SoftAssertionProvider`
* `SoftAssertionExtensionPoint`
* `SoftAssertionExtension`

Nothing stands out to me, I could live with `ISoftAssertion` although I'm not fond of that pattern naming it makes it clear this is an interface.

src/main/java/org/assertj/core/api/SoftAssertionsRule.java
```diff
@@ -0,0 +1,26 @@
...
+
+import static org.assertj.core.api.SoftAssertionsStatement.softAssertionsStatement;
+
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+public interface SoftAssertionsRule extends SoftAssertion, TestRule {
+  @Override
+  default public Statement apply(final Statement base, Description description) {
```
joel-costigliola: no need for `public`

src/main/java/org/assertj/core/api/SoftAssertion.java
```diff
@@ -0,0 +1,136 @@
...
+  * }</code></pre>
+  *
+  * @param <S> the concrete type of soft assertions to use.
+  * @param type the class object of the concrete type of soft assertions to use.
+  * @param softly the Consumer containing the code that will make the soft assertions.
+  *     Takes one parameter (the SoftAssertion instance used to make the assertions).
+  * @throws MultipleFailuresError if possible or SoftAssertionError if any proxied assertion objects threw an {@link AssertionError}
+  * @since 3.16.0
+  */
+  public static <S extends SoftAssertion> void assertSoftly(Class<S> type, Consumer<S> softly) {
```
joel-costigliola: no need for `public` (I think)

src/test/java/org/assertj/core/api/abstract_/SoftAssertionsMultipleProjectsTest.java
```diff
@@ -0,0 +1,101 @@
...
+	  
+  }
+  
+  static class MyClass1Assert extends AbstractAssert<MyClass1Assert, MyClass1> {
+	public MyClass1Assert(MyClass1 actual) {
+		super(actual, MyClass1Assert.class);
+	}
+	
+	public MyClass1Assert isClassy() {
+		throw new AssertionError("Not classy");
```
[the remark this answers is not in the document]
kriegfrj: As implemented, this method always throws an assertion, so the return would be redundant (in fact, it would be a compile error).

src/main/java/org/assertj/core/api/SoftAssertion.java
```diff
@@ -0,0 +1,23 @@
...
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2020 the original author or authors.
+ */
+package org.assertj.core.api;
+
+import java.util.List;
+
+public interface SoftAssertion {
+	<V extends Assert<? extends V, ? extends T>, T> V proxy(Class<V> assertClass, Class<T> actualClass, T actual);
```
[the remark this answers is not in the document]
kriegfrj: Done. I'm also going to put this change (and related) into a separate commit, because adding bounds to these type parameters is not strictly part of the soft assertions restructure.

src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
kriegfrj: Done.

src/main/java/org/assertj/core/api/ThrowableCollector.java
kriegfrj: I'm not sold on this one.

The point of soft assertions is that if one of them fails it doesn't stop the others from running. This is useful regardless of whether the test failed due to an error (ie, an exception other than `AssertionError`) or an assertion failure (ie, an `AssertionError`).

I have personally had experience when using soft assertions that I had an error and it masked a whole bunch of soft assertion failures. I was lulled into a false sense of security thinking that I had only the one error - when I fixed it, a whole bunch of assertion failures appeared. :smile:

On the other hand, it will be easier to restrict this to `AssertionError` now and find we need to expand it to include `Throwable`, rather than include `Throwable` now and find we need to restrict it to `AssertionError`.

In the end I guess this is your call - I'll go with what you want, I just wanted to put forward an alternative viewpoint for leaving it how it is.

src/main/java/org/assertj/core/api/ThrowableCollector.java
kriegfrj: See comments above.

src/main/java/org/assertj/core/api/SoftAssertion.java
kriegfrj: I think we should avoid the `*Extension*` variants as they are easy to confuse with the JUnit Jupiter concept of an "Extension" (indeed, we've already got `org.assertj.core.api.junit.jupiter.SoftAssertionsExtension`).

I don't mind the `ISoftAssertion` pattern of naming, but I prefer consistency and we haven't used that pattern elsewhere in the code.

So I think by a process of elimination that leaves us with `SoftAssertionProvider`. I actually quite like this.

src/main/java/org/assertj/core/api/ThrowableCollector.java
joel-costigliola: Fair point, to do that `ErrorCollector` would have to change, at the moment its `intercept` method only catch `AssertionError`.

I would rather go with `AssertionError` for now to keep changes in this PR not to be too big.

I have created https://github.com/joel-costigliola/assertj-core/issues/1821 to think about it further.

kriegfrj: Actually @joel-costigliola, I started going down this path and it's bit more complicated.

`List<Throwable> errorsCollected()` was an existing public method of `AbstractSoftAssertions` - I simply copied that signature into `Throwable/AssertionErrorCollector`. If I change the signature of `AssertionErrorCollector`, I'll probably still have to leave the old method in place for backward compatibility.

src/main/java/org/assertj/core/api/SoftAssertion.java
joel-costigliola: let's go with SoftAssertionProvider then

src/main/java/org/assertj/core/api/ThrowableCollector.java
kriegfrj: I've had a look at this and it's slightly complicated by the fact that `List<Throwable> errorsCollected()` was already a public method of `AbstractSoftAssertions`, so I can't just get rid of it. I'll add `List<AssertionError> assertionErrorsCollected()` and delegate from the existing method signature to the other.

src/test/java/org/assertj/core/api/abstract_/SoftAssertionsMultipleProjectsTest.java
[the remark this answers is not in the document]
joel-costigliola: I was not clear, I was thinking of replacing the code throwing the exception as MyClassAssert2

src/main/java/org/assertj/core/api/SoftAssertionsRule.java
kriegfrj: Done.

src/main/java/org/assertj/core/api/SoftAssertion.java
kriegfrj: Done.

src/main/java/org/assertj/core/api/SoftAssertion.java
kriegfrj: Done. Actually, I went with the plural `SoftAssertionsProvider` because it seemed to fit better (more than one soft assertion is typically provided).

src/main/java/org/assertj/core/api/SoftAssertion.java
kriegfrj: Done. Because this was a problem that I inherited I separated this change into a separate commit.

src/main/java/org/assertj/core/api/SoftAssertion.java
kriegfrj: Done.

src/test/java/org/assertj/core/api/abstract_/SoftAssertionsMultipleProjectsTest.java
```diff
@@ -0,0 +1,101 @@
...
+  public void return_empty_list_of_errors_for_passing_assertions() {
+    softly.assertThat(class1ForTesting).isNotClassy(); // No errors to collect
+    softly.assertThat(class2ForTesting).isClassy(); // No errors to collect
+    assertThat(softly.errorsCollected()).isEmpty();
+  }
+
+  @Test
+  public void returns_list_of_errors_in_the_order_they_were_asserted() {
+    softly.assertThat(class1ForTesting).isClassy();
+    softly.assertThat(class2ForTesting).isNotClassy();
```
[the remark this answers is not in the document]
kriegfrj: Done.

src/main/java/org/assertj/core/api/ThrowableCollector.java
kriegfrj: This has been done. Depending on the outcome of #1821, we may want to deprecate `errorsCollected()`.

src/test/java/org/assertj/core/api/abstract_/SoftAssertionsMultipleProjectsTest.java
```diff
@@ -0,0 +1,101 @@
...
+import org.assertj.core.api.AbstractStandardSoftAssertions;
+import org.assertj.core.api.SoftAssertion;
+import org.assertj.core.api.SoftAssertions;
+import org.junit.jupiter.api.Test;
+
+/**
+ * This tests that classes extended from {@link AbstractStandardSoftAssertions} will have access to the list of
+ * collected errors that the various proxies have collected.
+ */
+public class SoftAssertionsMultipleProjectsTest {
```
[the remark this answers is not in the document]
kriegfrj: Fixed (I hope).

src/test/java/org/assertj/core/api/abstract_/SoftAssertionsMultipleProjectsTest.java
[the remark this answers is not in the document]
kriegfrj: Terribly sorry, but I'm still not sure that I follow. I wanted "MyClass1" to be "not classy" and "MyClass2" to be "classy".

kriegfrj:
One thing that occurred to me is that the usage of interfaces with default methods means abandoning Java 6 compatibility - is that going to be a problem?

scordio:
> One thing that occurred to me is that the usage of interfaces with default methods means abandoning Java 6 compatibility - is that going to be a problem?

I don't think so as AssertJ Core 3.x requires Java 8 or higher

src/test/java/org/assertj/core/api/abstract_/SoftAssertionsMultipleProjectsTest.java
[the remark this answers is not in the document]
joel-costigliola: that's minor, I just found a bit confusing that a method named `isClassy` would throw an exception saying it's not classy. You can leave like that.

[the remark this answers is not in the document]
kriegfrj: That's because it's emulating a failing assertion - the error message of a faied *should* say the opposite of the method name,  no?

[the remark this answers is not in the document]
kriegfrj: I admit that it took me a bit to get my head around this when I first wrote it...

joel-costigliola:
@kriegfrj integrated, thanks for the work.
I have changed a few things:
* added `AutoCloseableSoftAssertionsProvider` instead of having `SoftAssertionsProvider`  being `AutoCloseable`, this raised a compiler warning for regular `SoftAssertions` because it was not explicitly closed! 
* renamed `AbstractBDDSoftAssertions` to `BDDSoftAssertionsProvider`
* renamed `AbstractStandardSoftAssertions` to `StandardSoftAssertionsProvider`
* renamed `Java6AbstractBDDSoftAssertions` to `Java6BDDSoftAssertionsProvider`
* renamed `Java6AbstractStandardSoftAssertions` to `Java6StandardSoftAssertionsProvider`
* made `SoftAssertionsMultipleProjectsTest` even more classy ;-)

kriegfrj:
I'm all good with those changes! Explains why auto close was not put on the main class originally. And the abstract classes - i didn't change those for backward compatibility,  but i suppose this would have been a breaking change anyway.

Thanks for your hard work... I'll polish up #1819 now.

joel-costigliola:
If you don't close or run an AutoCloseable in try with resources block, you get a compiler warning.
so basically using SoftAssertions led to this warning.

Renaming is a breaking but we already had one by moving the classes to interfaces.

PR #1820, opened 2020-03-31
Support custom soft assertions in SoftAssertionExtension

### LINKED ISSUE #1819: SoftAssertionExtension to support custom soft assertions
#### Summary

The current implementation of `SoftAssertionExtension` is hardcoded to support `SoftAssertions` and `BDDSoftAssertions`. If you have your own custom soft assertions entry point (with your own custom assertions), then you're out of luck.

#### Example

```java
class MySoftAssertions extends SoftAssertions {
  public CustomAssert assertThat(Custom actual) {
    return proxy(CustomAssert.class, Custom.class, actual);
  }
}
@ExtendWith(SoftAssertionsExtension.class)
public class MyTest {
  @Test 
  void my_test_method(MySoftAssertions softly) {
    // Parameter "softly" will not resolve
  }
}
```

kriegfrj:
Note that this PR depends on #1817.

With this implementation, any class that:
* is concrete implementation of the `SoftAssertion` interface
* has a default constructor
...can be injected as a parameter.

This mechanism has been retro-fitted to `SoftAssertions` and `BDDSoftAssertions`, and will also support any other soft assertion implementations anyone might conjure up (either within AssertJ or users of AssertJ). 

#### Check List:
* Fixes #1819
* Unit tests : YES 
* Javadoc with a code example (on API only) : Some - will ensure that this is complete if the PR is accepted.

### DIFF
--- a/src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
+++ b/src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
@@ -17,11 +17,12 @@ import static org.junit.platform.commons.support.AnnotationSupport.isAnnotated;
 
 import java.lang.reflect.Executable;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.lang.reflect.Parameter;
 import java.util.Optional;
-import java.util.function.Consumer;
 
 import org.assertj.core.api.BDDSoftAssertions;
+import org.assertj.core.api.SoftAssertionsProvider;
 import org.assertj.core.api.SoftAssertions;
 import org.junit.jupiter.api.extension.AfterTestExecutionCallback;
 import org.junit.jupiter.api.extension.ExtensionContext;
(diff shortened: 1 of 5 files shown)


kriegfrj:
This PR will need updating once #1817 is finalised and merged.

joel-costigliola:
I'll have a look at it after #1817 (one thing at a time)

src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
```diff
@@ -88,41 +100,45 @@
   @Override
   public boolean supportsParameter(ParameterContext parameterContext, ExtensionContext extensionContext) {
     // Abort if parameter type is unsupported.
-    if (isUnsupportedParameterType(parameterContext.getParameter())) return false;
-
+    Class<? extends SoftAssertionsProvider> type = getSupportedParameterType(parameterContext.getParameter());
+    if (type == null) {
+    	return false;
+    }
```
joel-costigliola, line 106: I would prefer to keep `isUnsupportedParameterType` and later call `Modifier.isAbstract(parameterContext.getParameter().getType()` as I think we don't need to cast `parameterContext.getParameter().getType()`  to `Class<? extends SoftAssertionsProvider>` for this call

src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
```diff
@@ -88,41 +100,45 @@
...
+                executable));
+    }
     return true;
   }
 
   @Override
   public Object resolveParameter(ParameterContext parameterContext, ExtensionContext extensionContext) {
-    // The parameter type is guaranteed to be either SoftAssertions or BDDSoftAssertions
-    return getStore(extensionContext).getOrComputeIfAbsent(parameterContext.getParameter().getType());
+    // The parameter type is guaranteed to be an instance of SoftAssertion
```
joel-costigliola, line 124: typo: SoftAssertion -> SoftAssertionsProvider

src/test/java/org/assertj/core/api/junit/jupiter/CustomSoftAssertionsExtensionIntegrationTest.java
```diff
@@ -0,0 +1,140 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+/**
+ * Integration tests for {@link SoftAssertionsExtension} with a custom implementation of {@link SoftAssertionsProvider}.
+ *
+ * @author Sam Brannen
+ * @author Fr Jeremy Krieg
+ * @since 3.13
```
joel-costigliola, line 42: 3.16

src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
```diff
@@ -88,41 +100,45 @@
...
     Executable executable = parameterContext.getDeclaringExecutable();
     // @Testable is used as a meta-annotation on @Test, @TestFactory, @TestTemplate, etc.
     boolean isTestableMethod = executable instanceof Method && isAnnotated(executable, Testable.class);
     if (!isTestableMethod) {
-      throw new ParameterResolutionException(format("Configuration error: cannot resolve SoftAssertions or BDDSoftAssertions for [%s]. Only test methods are supported.",
+      throw new ParameterResolutionException(format("Configuration error: cannot resolve SoftAssertionsProvider instances for [%s]. Only test methods are supported.",
                                                     executable));
     }
+    if (Modifier.isAbstract(type.getModifiers())) {
+        throw new ParameterResolutionException(format("Configuration error: cannot resolve abstract SoftAssertionsProvider instances [%s].",
```
joel-costigliola, line 116: the error is not super clear, I would phrase it more like:
```
the resolved SoftAssertionsProvider instance is actually an abstract class (%s) and thus can't be instantiated. 
```

src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
kriegfrj: Done.

src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
kriegfrj: I replaced "instance" with "type", as we don't actually have the instance (indeed we can't).

kriegfrj:
Thanks - all done, and I added an extra sanity check for `SoftAssertionProvider`s that don't have a default constructor.

Ready for round 2!

joel-costigliola:
Integrated thanks @kriegfrj!

kriegfrj:
> Integrated thanks @kriegfrj!

Thanks @joel-costigliola!

I think I might have a couple more suggestions coming up soon - in particular, to make it possible for `SoftAssertionsExtension` to be used by other extensions. We've experimented with something similar in https://github.com/osgi/osgi-test in the last couple of days and it seems to work well.

PR #1828, opened 2020-04-11
Enable CustomSoftAssertionsLineNumberTest on non-MacOs operating systems

PascalSchumacher:


PascalSchumacher:
Only fails on macOs.

PR #1829, opened 2020-04-11
Fix ShouldContain/ShouldContainOnly use a more specific name in error message

### LINKED ISSUE #1645: ShouldContainXXX: allow to replace "elements" by a more specific name
#### Summary

`ShouldContainOnly` is used in a few places where we know what kind of elements we are dealing with, in this case instead of using the term element we could use a more descriptive name. 

#### Example

```java
Spliterator<?> actual = createSpliterator(SORTED | ORDERED);
spliterators.assertHasOnlyCharacteristics(INFO, actual, DISTINCT, SORTED);
```
fails with this error
```
Expecting:
  <["ORDERED", "SORTED"]>
to contain only:
  <["DISTINCT", "SORTED"]>
elements not found:
  <["DISTINCT"]>
and elements not expected:
  <["ORDERED"]>
```
but we could have 
```
Expecting spliterator characteristics:
  <["ORDERED", "SORTED"]>
to contain only:
  <["DISTINCT", "SORTED"]>
characteristics not found:
  <["DISTINCT"]>
and characteristics not expected:
  <["ORDERED"]>
```
Same thing for `ShouldContain`.

WuYff:
#### Check List:
* Fixes #1645 
* Unit tests : YES

I implemented ``getSpecificName()`` in ``BasicErrorMessageFactory`` class and called it in the ``ShouldContain`` and ``ShouldContainOnly`` class.

Changing the error message content caused a few existing tests to fail. So I corrected them by replacing with the more specific message.

Please let me know If you have any feedback.

### DIFF
--- a/src/main/java/org/assertj/core/error/BasicErrorMessageFactory.java
+++ b/src/main/java/org/assertj/core/error/BasicErrorMessageFactory.java
@@ -22,6 +22,7 @@ import static org.assertj.core.util.Objects.hashCodeFor;
 import static org.assertj.core.util.Strings.quote;
 
 import java.util.Arrays;
+import java.util.Map;
 
 import org.assertj.core.description.Description;
 import org.assertj.core.presentation.Representation;
@@ -161,4 +162,31 @@ public class BasicErrorMessageFactory implements ErrorMessageFactory {
                   CONFIGURATION_PROVIDER.representation().toStringOf(arguments));
   }
 
+  /**
+   * Return the specific name that will be used in the error message.
+   *
+   * @param actual the object that is being verified .
+   * @return an String array of length two. The first string is the object (class) name, the second string is the element name.
+   */
+  public static String[] getSpecificName(Object actual) {
+    String name = actual.getClass().getSimpleName();
+    String[] s = { "", "elements" };
+    StackTraceElement[] stElements = Thread.currentThread().getStackTrace();
+    if (stElements[5].getClassName().contains("Spliterators")) {
+      s[0] = "spliterator characteristics";
+      s[1] = "characteristics";
+    } else if (actual instanceof Map) {
+      s[0] = name;
+      s[1] = "map entries";
+    } else if (actual.getClass().isArray()) {
+      s[0] = "array " + name;
+      if(actual.getClass().getComponentType().isPrimitive()){
+        s[1] = name.substring(0, name.length() - 2);
+      };// get "float" from "float[]"
+    } else {
+      s[0] = name;
+    }
+    s[0] = " " + s[0];
+    return s;
+  }
 }
--- a/src/main/java/org/assertj/core/error/BasicErrorMessageFactory.java
+++ b/src/main/java/org/assertj/core/error/BasicErrorMessageFactory.java
@@ -22,7 +22,6 @@ import static org.assertj.core.util.Objects.hashCodeFor;
 import static org.assertj.core.util.Strings.quote;
 
 import java.util.Arrays;
-import java.util.Map;
 
 import org.assertj.core.description.Description;
 import org.assertj.core.presentation.Representation;
@@ -30,7 +29,7 @@ import org.assertj.core.util.VisibleForTesting;
 
 /**
  * A factory of error messages typically shown when an assertion fails.
- * 
+ *
  * @author Alex Ruiz
  */
 public class BasicErrorMessageFactory implements ErrorMessageFactory {
@@ -94,7 +93,7 @@ public class BasicErrorMessageFactory implements ErrorMessageFactory {
 
   /**
    * Creates a new <code>{@link BasicErrorMessageFactory}</code>.
-   * 
+   *
    * @param format the format string.
    * @param arguments arguments referenced by the format specifiers in the format string.
    */
@@ -123,7 +122,7 @@ public class BasicErrorMessageFactory implements ErrorMessageFactory {
 
   /**
    * Return a string who will be unquoted in message format (without '')
-   * 
+   *
    * @param string the string who will be unquoted.
    * @return an unquoted string in message format.
    */
@@ -162,31 +161,4 @@ public class BasicErrorMessageFactory implements ErrorMessageFactory {
                   CONFIGURATION_PROVIDER.representation().toStringOf(arguments));
   }
 
-  /**
-   * Return the specific name that will be used in the error message.
-   *
-   * @param actual the object that is being verified .
-   * @return an String array of length two. The first string is the object (class) name, the second string is the element name.
-   */
-  public static String[] getSpecificName(Object actual) {
-    String name = actual.getClass().getSimpleName();
-    String[] s = { "", "elements" };
-    StackTraceElement[] stElements = Thread.currentThread().getStackTrace();
-    if (stElements[5].getClassName().contains("Spliterators")) {
-      s[0] = "spliterator characteristics";
-      s[1] = "characteristics";
-    } else if (actual instanceof Map) {
-      s[0] = name;
-      s[1] = "map entries";
-    } else if (actual.getClass().isArray()) {
-      s[0] = "array " + name;
-      if(actual.getClass().getComponentType().isPrimitive()){
-        s[1] = name.substring(0, name.length() - 2);
-      };// get "float" from "float[]"
-    } else {
-      s[0] = name;
-    }
-    s[0] = " " + s[0];
-    return s;
-  }
 }
--- a/src/main/java/org/assertj/core/error/GroupTypeDescription.java
+++ b/src/main/java/org/assertj/core/error/GroupTypeDescription.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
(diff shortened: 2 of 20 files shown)


src/main/java/org/assertj/core/error/BasicErrorMessageFactory.java
```diff
@@ -161,4 +162,31 @@ public String toString() {
                   CONFIGURATION_PROVIDER.representation().toStringOf(arguments));
   }
 
+  /**
+   * Return the specific name that will be used in the error message.
+   *
+   * @param actual the object that is being verified .
+   * @return an String array of length two. The first string is the object (class) name, the second string is the element name.
+   */
+  public static String[] getSpecificName(Object actual) {
```
joel-costigliola, line 171: let's return a custom type as `String[]` does not express what it contains, I suggest `GroupTypeDescription` with two fields `groupTypeName` and `elementTypeName`.
and rename to `getGroupTypeDescription` (or whatever class we'll agree on to return)

src/main/java/org/assertj/core/error/BasicErrorMessageFactory.java
```diff
@@ -161,4 +162,31 @@ public String toString() {
...
+    String name = actual.getClass().getSimpleName();
+    String[] s = { "", "elements" };
+    StackTraceElement[] stElements = Thread.currentThread().getStackTrace();
+    if (stElements[5].getClassName().contains("Spliterators")) {
```
joel-costigliola, line 175: Instead of if-elseif-else, I prefer the coding style where you return the value as soon as possible.
ex: 
```
if (actual instanceof Map) return new GroupTypeDescription("map", "map entries");
```
and yes no block needed if the if-return is formatted on the same line

src/main/java/org/assertj/core/error/ShouldContain.java
```diff
@@ -63,8 +64,8 @@ public static ErrorMessageFactory directoryShouldContain(Path actual, List<Strin
     return new ShouldContain(actual, directoryContent, filterDescription);
   }
 
-  private ShouldContain(Object actual, Object expected, Object notFound, ComparisonStrategy comparisonStrategy) {
-    super("%nExpecting:%n <%s>%nto contain:%n <%s>%nbut could not find:%n <%s>%n%s", actual, expected, notFound,
+  private ShouldContain(Object actual, Object expected, Object notFound, ComparisonStrategy comparisonStrategy, String[] name) {
+    super("%nExpecting" + name[0] + ":%n <%s>%nto contain:%n <%s>%nbut could not find the following "+name[1]+":%n <%s>%n%s", actual, expected, notFound,
```
joel-costigliola, line 68: use `%s` instead of string concatenation

joel-costigliola: this comment applies on other files

src/main/java/org/assertj/core/error/BasicErrorMessageFactory.java
joel-costigliola: I also wonder if we could put that code in an interface default method and let `ShouldContain` and co implement it instead of the common base class.
interface could be `WithGroupTypeDescription` (happy to get a better name)

WuYff: Yes, I think an interface will be better.

WuYff:
Thanks for the review, @joel-costigliola  I'm working on it.

src/main/java/org/assertj/core/error/BasicErrorMessageFactory.java
WuYff: I think that the method ``getGroupTypeDescription`` should be `` static`` rather than ``default`` because non-static method cannot be invoked in static method and ShouldContain is a Class with only static methods.
In this case, an interface may not be needed.  I think we can  make this a static method in ``GroupTypeDescription ``  class  and if other ShouldChontainXXX classes need different implementation, they write their own ``getGroupTypeDescription`` method.

src/main/java/org/assertj/core/error/ShouldContain.java
WuYff: I find use %s here and pass a String in this super() will show the double quotes, which can be a little strange to read

WuYff:
commit f624a9e is the new one.  ( commit c9ac81f is the old one showing again after rebase..

src/main/java/org/assertj/core/error/ShouldContain.java
joel-costigliola: yeah, fair enough let's go back to your initial implementation, just add a comment to say it will avoid quoted strings

src/main/java/org/assertj/core/error/BasicErrorMessageFactory.java
joel-costigliola: Fine by me!

src/test/java/org/assertj/core/internal/spliterator/Spliterators_assertHasCharacteristics_Test.java
```diff
@@ -77,12 +78,18 @@ public void should_fail_if_spliterator_does_not_have_characteristic() {
...
-    expectAssertionError(() -> spliterators.assertHasCharacteristics(INFO, actual, DISTINCT));
+    AssertionError error = expectAssertionError(() -> spliterators.assertHasCharacteristics(INFO, actual, DISTINCT));
     // THEN
-    verify(failures).failure(INFO, shouldContain(singleton("SORTED"), array("DISTINCT"), singleton("DISTINCT")));
+    assertThat(error).hasMessage(format("%nExpecting spliterator characteristics:%n"
+                                 + " <[\"SORTED\"]>%n"
+                                 + "to contain:%n"
+                                 + " <[\"DISTINCT\"]>%n"
+                                 + "but could not find the following characteristics:%n"
+                                 + " <[\"DISTINCT\"]>%n"));
```
joel-costigliola, line 88: this test (and the other similar ones) should not have changed, here we only want to verify that `shouldContain` is called with the proper arguments.
checking the actual message should be done in `ShouldContain_create_Test`

src/main/java/org/assertj/core/error/GroupTypeDescription.java
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Creates a new <code>{@link GroupTypeDescription}</code> for a group of elements.
+   *
+   * @param actual the group of elements.
+   * @return the created GroupTypeDescription object
+   */
+  public static GroupTypeDescription getGroupTypeDescription(Object actual) {
+
+    Class<?> c = actual.getClass();
+    if (Thread.currentThread().getStackTrace()[5].getClassName().contains("Spliterators"))
```
joel-costigliola, line 47: why 5 here? let's extract that to a constant to avoid a magic number.

src/main/java/org/assertj/core/error/GroupTypeDescription.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+package org.assertj.core.error;
+
+import java.util.Map;
+
+/**
+ * Generates a description for the type of a group of elements. The description is used in the error message.
+ *
+ */
+public class GroupTypeDescription {
```
joel-costigliola, line 21: add a `GroupTypeDescription_getGroupTypeDescription_Test` test for this class to check `getGroupTypeDescription` with a parameterized test (see https://www.baeldung.com/parameterized-tests-junit-5#6-method)

src/test/java/org/assertj/core/error/ShouldContain_create_Test.java
```diff
@@ -48,11 +48,11 @@ public void should_create_error_message() {
     String message = factory.create(new TextDescription("Test"));
     // THEN
     then(message).isEqualTo(format("[Test] %n" +
-                                   "Expecting:%n" +
+                                   "Expecting ArrayList:%n" +
                                    " <[\"Yoda\"]>%n" +
                                    "to contain:%n" +
                                    " <[\"Luke\", \"Yoda\"]>%n" +
-                                   "but could not find:%n" +
+                                   "but could not find the following elements:%n" +
```
joel-costigliola, line 55: add tests in this class to covert the different kind of GroupTypeDescription (map, arrays, ...)

src/test/java/org/assertj/core/internal/spliterator/Spliterators_assertHasCharacteristics_Test.java
WuYff: After implementing ``getGroupTypeDescription`` method,
```java
Spliterator<?> actual = createSpliterator(SORTED);
spliterators.assertHasCharacteristics(INFO, actual, DISTINCT);
```
fails with this error
```
Expecting spliterator characteristics:
 <["SORTED"]>
to contain:
 <["DISTINCT"]>
but could not find the following characteristics:
 <["DISTINCT"]>
```
But
```java
shouldContain(singleton("SORTED"), array("DISTINCT"), singleton("DISTINCT"));
```
fails with this error
```
Expecting SingletonSet:
 <["SORTED"]>
to contain:
 <["DISTINCT"]>
 <["DISTINCT"]>
```
spliterator characteristics are converted to a Ser\<String\> in ``spliterators.assertHasCharacteristics`` method, so I use stack trace to check whether ``shouldContain`` method is called through Spliterator class. Thus, a direct calling of ``shouldContain`` method will display a different type name. And that's why I need to change this test

joel-costigliola:
@WuYff let me know when you want another code review ;-)

WuYff:
> @WuYff let me know when you want another code review ;-)

@joel-costigliola  Oh, yes. Please review the code !
Thanks for your suggestions and the reminder :)

joel-costigliola:
Integrated thanks @WuYff!

PR #1830, opened 2020-04-11
Add decodedAsBase64 to AbstractStringAssert

### LINKED ISSUE #1805: Base64 support for AbstractStringAssert

scordio:
#### Check List:
* Fixes #1805
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractStringAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractStringAssert.java
@@ -15,6 +15,7 @@ package org.assertj.core.api;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 
+import java.util.Base64;
 import java.util.Comparator;
 
 import org.assertj.core.internal.Comparables;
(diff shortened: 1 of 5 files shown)


src/main/java/org/assertj/core/api/AbstractStringAssert.java
```diff
@@ -242,6 +243,31 @@ public SELF isValidBase64() {
     return myself;
   }
 
+  /**
+   * Decodes the actual value as a Base64 encoded string, the decoded value becoming the new object under test.
+   * <p>
+   * Examples:
+   * <pre><code class='java'> // assertions succeeds
```
joel-costigliola, line 250: assertions -> assertion

src/main/java/org/assertj/core/api/AbstractStringAssert.java
```diff
@@ -242,6 +243,31 @@ public SELF isValidBase64() {
     return myself;
   }
 
+  /**
+   * Decodes the actual value as a Base64 encoded string, the decoded value becoming the new object under test.
```
joel-costigliola, line 247: `the decoded value becoming ` -> `the decoded value (a byte[]) becoming `

scordio: What about `the decoded bytes becoming the new array under test`?

joel-costigliola:
Integrated thanks @scordio!

PR #1833, opened 2020-04-14
Add method for custom failures with actual/expected

### LINKED ISSUE #1826: AbstractAssert method for asserting with expected/actual
#### Summary

I think that #1731 was great for a certain class of custom assertions. However, i think it would also be a useful addition to `AbstractAssert` if it had a variant of the `failWithMessage()` method that you could call directly that would populate expected and actual in the generated `AssertionError` (where available). 

#### Example

```java
failWithExpectedActual(expected,  actual, "My message %d", myArg);
```

kriegfrj:
#### Check List:
* Fixes #1826
* Unit tests : YES


joel-costigliola:
Integrated thanks @kriegfrj!

PR #1843, opened 2020-04-19
Fix issue #1796 / add a method isEqualToWithSortedQueryParameters

### LINKED ISSUE #1796: URL should test for logical equality
#### Summary
The following test case returns false. I think that it should return true, as these two URLs are logically and practically equal:

#### Example

```java
        URL urlA = new URL("http://localhost?a=b&c=d");
        URL urlB = new URL("http://localhost?c=d&a=b");

        Assertions.assertThat(urlA).isEqualTo(urlB);
```

Sunt-ing:
#### Check List:

* Fixes #1796 
* Unit tests : YES

This part is designed for URL. And after fully finish it, I think I can copy it to URI.

To compare actual and expected, we can separately compare non-query parts and query parts, or assemble non-query parts and sorted querys into new URLs and then compare them. Personally, I choose the former one.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractUrlAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractUrlAssert.java
@@ -419,11 +419,10 @@ public abstract class AbstractUrlAssert<SELF extends AbstractUrlAssert<SELF>> ex
   }
 
   /**
-   * Verifies that the URL {@code actual} and URL {@code expected} are actually equivalent after <b>their parameters are sorted</b>.
+   * Verifies that the URL {@code expected} is equivalent to URL actual after <b>their parameters are sorted</b>.
    * <p>
    * Example:
    * <pre><code class='java'> URL expected = new URL("http://localhost?a=b&c=d");
-   *       URL actual2 = new URL("http://example.com?a=b&c=e");
    *
    * // this assertion succeeds ...
    * assertThat(expected).isEqualToWithSortedQueryParameters(new URL("http://localhost?c=d&a=b"));
--- a/src/main/java/org/assertj/core/api/AbstractUrlAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractUrlAssert.java
@@ -418,6 +418,7 @@ public abstract class AbstractUrlAssert<SELF extends AbstractUrlAssert<SELF>> ex
     return myself;
   }
 
+
   /**
    * Verifies that the URL {@code url1} and URL {@code url2} are actually equivalent after <b>their parameters are sorted</b>.
    * <p>
--- a/src/main/java/org/assertj/core/api/AbstractUrlAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractUrlAssert.java
@@ -431,7 +431,7 @@ public abstract class AbstractUrlAssert<SELF extends AbstractUrlAssert<SELF>> ex
    * assertThat(expected).isEqualToWithSortedQueryParameters(new URL("http://example.com?a=b&amp;c=e"));
    *
    * //... and this case fails as even their domains are different.
-   * assertThat(expected).isEqualToWithSortedQueryParameters(new URL("http://example2.com?amp;a=b&c=d"));
+   * assertThat(expected).isEqualToWithSortedQueryParameters(new URL("http://example2.com?amp;a=b&amp;c=d"));
    * </code></pre>
    *
    * @param expected the expected URL of the actual {@code URL}.
--- a/src/main/java/org/assertj/core/api/AbstractUrlAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractUrlAssert.java
@@ -422,16 +422,16 @@ public abstract class AbstractUrlAssert<SELF extends AbstractUrlAssert<SELF>> ex
    * Verifies that the URL {@code expected} is equivalent to URL actual after <b>their parameters are sorted</b>.
    * <p>
    * Example:
-   * <pre><code class='java'> URL expected = new URL("http://localhost?a=b&c=d");
+   * <pre><code class='java'> URL expected = new URL("http://example.com?a=b&c=d");
    *
    * // this assertion succeeds ...
-   * assertThat(expected).isEqualToWithSortedQueryParameters(new URL("http://localhost?c=d&a=b"));
+   * assertThat(expected).isEqualToWithSortedQueryParameters(new URL("http://example.com?c=d&a=b"));
    *
    * // ... but these cases fail as their parameters do not match.
-   * assertThat(expected).isEqualToWithSortedQueryParameters(new URL("http://localhost?a=b&c=e"));
+   * assertThat(expected).isEqualToWithSortedQueryParameters(new URL("http://example.com?a=b&c=e"));
    *
    * //... and this case fails as even their domains are different.
-   * assertThat(expected).isEqualToWithSortedQueryParameters(new URL("http://example.com?a=b&c=d"));
+   * assertThat(expected).isEqualToWithSortedQueryParameters(new URL("http://example2.com?a=b&c=d"));
    * </code></pre>
    *
    * @param expected the expected URL of the actual {@code URL}.
--- a/src/main/java/org/assertj/core/api/AbstractUrlAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractUrlAssert.java
@@ -419,7 +419,7 @@ public abstract class AbstractUrlAssert<SELF extends AbstractUrlAssert<SELF>> ex
   }
 
   /**
-   * Verifies that the URL {@code url1} and URL {@code url2} are actually equivalent after <b>their parameters are sorted</b>.
+   * Verifies that the URL {@code actual} and URL {@code expected} are actually equivalent after <b>their parameters are sorted</b>.
    * <p>
    * Example:
    * <pre><code class='java'> URL expected = new URL("http://localhost?a=b&c=d");
--- a/src/main/java/org/assertj/core/api/AbstractUrlAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractUrlAssert.java
@@ -441,7 +441,7 @@ public abstract class AbstractUrlAssert<SELF extends AbstractUrlAssert<SELF>> ex
    * @throws AssertionError if the actual {@code CharSequence} is {@code null}.
    * @throws AssertionError if the actual {@code CharSequence} does not contain all the given strings <b>in the given order</b>.
    * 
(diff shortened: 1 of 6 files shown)


src/main/java/org/assertj/core/internal/Urls.java
```diff
@@ -138,4 +138,21 @@ public void assertHasNoParameter(AssertionInfo info, URL actual, String name, St
...
+    if (expected.getQuery() == null || actual.getQuery() == null)
+      if (expected.getQuery() == null && actual.getQuery() == null)
+        return;
+      else throw failures.failure(info, shouldBeEqualToWithSortedQueryParameters(actual, expected));
+    String expected_non_query = expected.toString().substring(0, expected.toString().length() - expected.getQuery().length());
+    String actual_non_query = actual.toString().substring(0, actual.toString().length() - actual.getQuery().length());
+    if (!expected_non_query.equals(actual_non_query))
+      throw failures.failure(info, shouldBeEqualToWithSortedQueryParameters(actual, expected));
+    String[] expected_query_params = expected.getQuery().split("&");
+    String[] actual_query_params = actual.getQuery().split("&");
```
joel-costigliola, line 152: You would have a fragment if the URL has one [with query params and fragments](https://en.wikipedia.org/wiki/URL) :
```
URI = scheme:[//authority]path[?query][#fragment]
```

src/main/java/org/assertj/core/internal/Urls.java
```diff
@@ -138,4 +138,21 @@ public void assertHasNoParameter(AssertionInfo info, URL actual, String name, St
...
+    java.util.Arrays.sort(expected_query_params);
+    java.util.Arrays.sort(actual_query_params);
+    if (!java.util.Arrays.toString(expected_query_params).equals(Arrays.toString(actual_query_params)))
```
joel-costigliola, line 155: use `Objects.deepEquals` to compare `String[]` directly

Sunt-ing:
I've finished all you mentioned yet, please check them. Thanks.

src/main/java/org/assertj/core/api/AbstractUrlAssert.java
```diff
@@ -417,4 +417,34 @@ public SELF hasNoParameter(String name, String value) {
...
+   * </code></pre>
+   *
+   * @param expected the expected URL of the actual {@code URL}.
+   * @return {@code this} assertion object.
+   * @throws NullPointerException if the given values is {@code null}.
+   * @throws IllegalArgumentException if the given values is empty.
+   * @throws AssertionError if the actual {@code CharSequence} is {@code null}.
+   * @throws AssertionError if the actual {@code CharSequence} does not contain all the given strings <b>in the given order</b>.
+   * 
+   * @since 3.15.0
```
joel-costigliola, line 444: 3.16

src/main/java/org/assertj/core/internal/Urls.java
```diff
@@ -48,6 +47,15 @@ public static Urls instance() {
 
   Urls() {}
 
+  static String extractNonQueryPart(URL url) {
```
joel-costigliola, line 50: can this be private?

src/main/java/org/assertj/core/internal/Urls.java
```diff
@@ -48,6 +47,15 @@ public static Urls instance() {
...
+    return url.toString().substring(0, url.toString().length()
+                                       - (url.getQuery() == null ? 0
+                                           : url.getQuery().length())
+                                       - (url.getRef() == null ? 0
+                                           : url.getRef().length() + 1))
+           + " " + (url.getRef() == null ? "" : url.getRef());
```
joel-costigliola, line 56: this piece of code is hard to understand!  
My first suggestion would be to see this method can be implemented simply by calling  `StringUtils.remove(url.toString(), url.getQuery());` if that does not work then a few things could make it more readable:
* introduce local variables for `url.getQuery()` and `url.getRef()`
* `s == null ? 0 : s.length()` is code repeated twice for `url.getQuery()` and `url.getRef()` and should be replaced by `StringUtils.length`

src/main/java/org/assertj/core/internal/Urls.java
```diff
@@ -138,4 +146,16 @@ public void assertHasNoParameter(AssertionInfo info, URL actual, String name, St
     }
   }
 
+  public void assertIsEqualToWithSortedQueryParameters(AssertionInfo info, URL actual, URL expected) {
+    assertNotNull(info, actual);
+    if (!extractNonQueryPart(expected).equals(extractNonQueryPart(actual)))
+      throw failures.failure(info, shouldBeEqualToWithSortedQueryParameters(actual, expected));
+
+    String[] expectedQueryParams = (expected.getQuery() == null ? "" : expected.getQuery()).split("&");
+    String[] actualQueryParams = (actual.getQuery() == null ? "" : actual.getQuery()).split("&");
```
joel-costigliola, line 155: remove unnecessary `()` and extract to a static method call `extractQueryParams`

src/main/java/org/assertj/core/internal/Urls.java
```diff
@@ -138,4 +146,16 @@ public void assertHasNoParameter(AssertionInfo info, URL actual, String name, St
...
+    java.util.Arrays.sort(expectedQueryParams);
+    java.util.Arrays.sort(actualQueryParams);
```
joel-costigliola, line 157: static imports `sort`

src/test/java/org/assertj/core/internal/urls/Urls_assertIsEqualToWithSortedQueryParameters_Test.java
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Tests for
+ * <code>{@link org.assertj.core.internal.Urls#assertIsEqualToWithSortedQueryParameters(AssertionInfo, URL, URL)} (org.assertj.core.api.AssertionInfo, java.net.URL, java.net.URL)}  </code>
+ * .
+ *
+ * @author SUN Ting
+ */
+
+public class Urls_assertIsEqualToWithSortedQueryParameters_Test extends UrlsBaseTest {
```
joel-costigliola, line 36: add GIVEN WHEN THEN to the tests

src/test/java/org/assertj/core/internal/urls/Urls_assertIsEqualToWithSortedQueryParameters_Test.java
```diff
@@ -0,0 +1,120 @@
...
+
+  @Test
+  public void should_pass_if_only_query_parameters_orders_are_different() throws MalformedURLException {
+    URL actual = new URL("https://example.com/path/to/page?color=purple&name=ferret#hello");
+    URL expected = new URL("https://example.com/path/to/page?name=ferret&color=purple#hello");
+    urls.assertIsEqualToWithSortedQueryParameters(info, actual, expected);
+  }
```
joel-costigliola, line 43: replace the 3 methods `should_pass...` by a parameterized test like:
```
  @ParameterizedTest
  @CsvSource({ 
        "https://example.com/path/to/page?color=purple&name=ferret#hello, https://example.com/path/to/page?color=purple&name=ferret#hello", 
        "https://example.com/path/to/page#hello", "https://example.com/path/to/page#hello", 
        "https://example.com/path/to/page", "https://example.com/path/to/page", 
       ... other tests params
  })
  public void should_pass_using_primitive_comparison(String actual, Float expected) {
   // GIVEN
    URL actualUrl = new URL(actual);
    URL expectedUrl = new URL(expected);
   // WHEN-THEN
     urls.assertIsEqualToWithSortedQueryParameters(info, actualUrl, expectedUrl);
  }
```

src/test/java/org/assertj/core/internal/urls/Urls_assertIsEqualToWithSortedQueryParameters_Test.java
```diff
@@ -0,0 +1,120 @@
...
+
+  @Test
+  public void should_fail_if_sorted_query_parameters_are_different() throws MalformedURLException {
+    AssertionInfo info = someInfo();
+    URL actual = new URL("https://example.com/path/to/page?color=purple&name=ferret#hello");
+    URL expected = new URL("https://example.com/path/to/page?color=red&name=jackson#hello");
+
+    Throwable error = catchThrowable(() -> urls.assertIsEqualToWithSortedQueryParameters(info, actual, expected));
```
joel-costigliola, line 67: use `AssertionsUtil.expectAssertionError` instead of `catchThrowable` (static import it) so that you don't have to execute 
```
assertThat(error).isInstanceOf(AssertionError.class);
````

Comment (joel-costigliola) [reply src/test/java/org/assertj/core/internal/urls/Urls_assertIsEqualToWithSortedQueryParameters_Test.java:67]: Same as `should_pass...`, all the `should_fail...` can be parameterized

File: src/main/java/org/assertj/core/internal/Urls.java (hunk 5 above)
Comment (Sunt-ing) [reply src/main/java/org/assertj/core/internal/Urls.java:56]: Evidently you are right. How about this?

```java
  public void assertIsEqualToWithSortedQueryParameters(AssertionInfo info, URL actual, URL expected) {
    assertNotNull(info, actual);
    String expectedQuery = expected.getQuery() == null ? "" : expected.getQuery();
    String actualQuery = actual.getQuery() == null ? "" : actual.getQuery();
    String expectedNonQueryPart = expected.toString().replace(expectedQuery, " ");
    String actualNonQueryPart = actual.toString().replace(actualQuery, " ");

    if (!expectedNonQueryPart.equals(actualNonQueryPart))
      throw failures.failure(info, shouldBeEqualToWithSortedQueryParameters(actual, expected));

    String[] expectedQueryParams = extractQueryParams(expected);
    String[] actualQueryParams = extractQueryParams(actual);
    Arrays.sort(expectedQueryParams);
    Arrays.sort(actualQueryParams);
    if (!Objects.deepEquals(expectedQueryParams, actualQueryParams))
      throw failures.failure(info, shouldBeEqualToWithSortedQueryParameters(actual, expected));
  }
```

Sunt-ing: All my test cases are passed by this code snippet.

Sunt-ing:
Please check again. @joel-costigliola

joel-costigliola:
Integrated thanks @Sunt-ing!  I have done a bit more refactoring if you are interested this is here https://github.com/joel-costigliola/assertj-core/commit/c614174af9bd7ae668bdb8c2900dfe4c13f0f3e5

Sunt-ing:
Oh, thanks very much! You are so nice! @joel-costigliola

PR #1844, opened 2020-04-20
Fix #1643, add an api isNotEqualToComparingFieldByField()

### LINKED ISSUE #1643: Add reverse of isEqualToComparingFieldByField
#### Summary
I'd like do check if Object's fields were updated sucessfully, therefore does not contain old data, and `isNotEqualToComparingFieldByField` could be handy.

#### Example

```java
//given
final User user = User();
user.setAddress(oldAddress);

//when
updateUserAddress(user, newAddressDto);

//then
assertThat(user.getAddress())
  .isEqualToComparingFieldByField(newAddressDto)
  .isNotEqualToComparingFieldByField(oldAddress); //<-please add this
```
I can not use `isNotEqualTo`, because `address` reference did not change (its fields were updated by mapper)

darkliang:
#### Check List:
* Fixes #1643 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/Java6Assertions.java
+++ b/src/main/java/org/assertj/core/api/Java6Assertions.java
@@ -126,18 +126,6 @@ public class Java6Assertions {
   }
 
   /**
-   * Create assertion for {@link LongAdder}.
-   *
-   * @param actual the actual value.
-   *
-   * @return the created assertion object.
-   * @since 3.16.0
-   */
-  public static LongAdderAssert assertThat(LongAdder actual) {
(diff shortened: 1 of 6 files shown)


joel-costigliola:
@darkliang I'd rather we add this feature as a `isNotEqual` to the new `RecursiveComparisonAssert` that has taken over field by field comparison.

darkliang:
> @darkliang I'd rather we add this feature as a `isNotEqual` to the new `RecursiveComparisonAssert` that has taken over field by field comparison.

OK, I think I can do that.

PR #1845, opened 2020-04-20
Introduce ProxyableObjectChangingMethodTest for soft assertions tests

scordio:
First try to break down `SoftAssertionTests`. The idea is to enhance the test classes of single methods with some centralized tests, avoiding the repetitive writing every time a new assertion method is added.

* `ProxyableObjectChangingMethodTest`: for public methods which change the object under test and are proxied according to [`SoftProxies`](https://github.com/joel-costigliola/assertj-core/blob/master/src/main/java/org/assertj/core/api/SoftProxies.java#L42)
* `ObjectChangingMethodTest`: for non-public methods like [`AbstractAssert#extracting`](https://github.com/joel-costigliola/assertj-core/blob/master/src/main/java/org/assertj/core/api/AbstractAssert.java#L877-L931) which change the method under assertion but are not proxied

As demonstration, just few tests have been moved to the new setup, but more should be refactored.

Open issues with `ProxyableObjectChangingMethodTest`:

- [ ] When applied to:
  * `ObjectAssert_extracting_with_String_and_InstanceOfAssertFactory_Test`
  * `ObjectAssert_extracting_with_Function_and_InstanceOfAssertFactory_Test`

  as:
  ```java
  implements ProxyableObjectChangingMethodTest<ObjectAssert<Employee>>
  ```

  and providing the soft assertions as:

  ```java
  @Override
  public ObjectAssert<Employee> getSoftAssertion(AbstractSoftAssertions softly) {
    return softly.proxy(ObjectAssert.class, Employee.class, luke);
  }
  ``` 
  it fails with:

  ```
  java.lang.RuntimeException: java.lang.NoSuchMethodException: org.assertj.core.api.ObjectAssert$ByteBuddy$it09GaKX.<init>(org.assertj.core.test.Employee)

	at org.assertj.core.api.SoftProxies.createSoftAssertionProxy(SoftProxies.java:121)
	at org.assertj.core.api.AbstractSoftAssertions.proxy(AbstractSoftAssertions.java:43)
	at org.assertj.core.api.object.ObjectAssert_extracting_with_String_and_InstanceOfAssertFactory_Test.getSoftAssertion(ObjectAssert_extracting_with_String_and_InstanceOfAssertFactory_Test.java:128)
	at org.assertj.core.api.object.ObjectAssert_extracting_with_String_and_InstanceOfAssertFactory_Test.getSoftAssertion(ObjectAssert_extracting_with_String_and_InstanceOfAssertFactory_Test.java:47)
	at org.assertj.core.api.ProxyableObjectChangingMethodTest.soft_assertions_should_collect_errors_after_changing_the_object_under_test(ProxyableObjectChangingMethodTest.java:50)
  ```
  This could be caused by the use of `ObjectAssert` instead of `ProxyableObjectAssert`, but at the moment I don't understand the exact cause.

- [ ] When applied to `MapAssert_extractingByKeys_Test` as:
  ```java
  implements ProxyableObjectChangingMethodTest<MapAssert<Object, Object>>
  ```

   and providing the soft assertions as:
  ```java
  @Override
  public MapAssert<Object, Object> getSoftAssertion(AbstractSoftAssertions softly) {
    return softly.proxy(MapAssert.class, Map.class, map);
  }
  ```

  it stops at the first assertion error in `soft_assertions_should_collect_errors_after_changing_the_object_under_test`.

  Most likely this is due to `extractingByKeys` being final and an instance of `ProxyableMapAssert` should be used instead, but it would require a separate test class with:
  ```java
  implements ProxyableObjectChangingMethodTest<ProxyableMapAssert<Object, Object>>
  ```

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
@@ -666,7 +666,8 @@ public abstract class AbstractObjectAssert<SELF extends AbstractObjectAssert<SEL
    */
   @CheckReturnValue
   public AbstractObjectAssert<?, ?> extracting(String propertyOrField) {
-    return super.extracting(propertyOrField, this::newObjectAssert);
+    // eclipse needs the explicit cast or fails to compile :(
+    return super.extracting(propertyOrField, (AssertFactory<Object, AbstractObjectAssert<?, ?>>) this::newObjectAssert);
   }
 
   /**
@@ -712,7 +713,9 @@ public abstract class AbstractObjectAssert<SELF extends AbstractObjectAssert<SEL
   @CheckReturnValue
   public <ASSERT extends AbstractAssert<?, ?>> ASSERT extracting(String propertyOrField,
                                                                  InstanceOfAssertFactory<?, ASSERT> assertFactory) {
-    return super.extracting(propertyOrField, this::newObjectAssert).asInstanceOf(assertFactory);
+    // eclipse needs the explicit cast or fails to compile :(
+    return super.extracting(propertyOrField,
+                            (AssertFactory<Object, AbstractAssert<?, ?>>) this::newObjectAssert).asInstanceOf(assertFactory);
   }
 
   /**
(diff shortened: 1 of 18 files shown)


joel-costigliola:
That's a good approach, I have run into an Eclipse compiler error (but maven compiles fine) in `AbstractObjectAssert` L669:
```
  public AbstractObjectAssert<?, ?> extracting(String propertyOrField) {
    return super.extracting(propertyOrField, this::newObjectAssert);
  }
```
error
```
The type of newObjectAssert(Object) from the type AbstractObjectAssert<SELF,ACTUAL> is AbstractObjectAssert<?,Object>, 
this is incompatible with the descriptor's return type: AbstractObjectAssert<AbstractObjectAssert<AbstractObjectAssert<?,Object>,Object>,Object>	
``` 
This error prevents me from running the `should_keep_existing_assertion_state` tests :(

I'll raise an issue to eclipse.

PR #1856, opened 2020-04-29
Fix issue #1289 by adding the annotation @deprecated to method temporaryFolder

### LINKED ISSUE #1289: Files#temporaryFolder() will fail if System temporary folder does not exist
#### Summary

We noticed this issue when working on a docker based build. We found that the Java defined property for `java.io.tmpdir` was defined but the folder did _not exist_. This caused calls to Files#temporaryFolder to fail with the exception:

`org.assertj.core.api.exception.RuntimeIOException: Unable to find temporary directory`

#### Example

```java
    public static void main(String ... args) {
        System.setProperty("java.io.tmpdir", "badger");
        assertThat(new File(System.getProperty("java.io.tmpdir"))).doesNotExist();
        Files.temporaryFolder(); // Will fail
    }
```

The following code sample would resolve this issue. I am happy to raise this as a pull request if this issue is considered valid.

```java
    public static File temporaryFolder() {
        File temp = new File(Files.temporaryFolderPath());
        if (!temp.isDirectory()) {
            throw new RuntimeIOException("Unable to find temporary directory");
        }
        if (!temp.exists() && !temp.mkdirs()) {
            throw new RuntimeIOException("Failed to create temporary directory");
        }
        return temp;
    }
```

Sunt-ing:
#### Check List:
* Fixes #1289
* Unit tests : YES

The annotation @deprecated is added before the method  `File temporaryFolder()`. Also, the alternatives in JUnit4+JUnit5 are added in javadoc part of the method.

### DIFF
--- a/src/main/java/org/assertj/core/util/Files.java
+++ b/src/main/java/org/assertj/core/util/Files.java
@@ -82,7 +82,10 @@ public class Files {
    * 
    * @return the system's temporary directory.
    * @throws RuntimeException if this method cannot find or create the system's temporary directory.
+   *
+   * @deprecated For JUnit4, use TemporaryFolder and @Rule; for JUnit5, use TempDirectory extension and @TempDir instead.
    */
+  @Deprecated
   public static File temporaryFolder() {
     File temp = new File(temporaryFolderPath());
     if (!temp.isDirectory()) {
--- a/src/main/java/org/assertj/core/util/Files.java
+++ b/src/main/java/org/assertj/core/util/Files.java
@@ -83,7 +83,7 @@ public class Files {
    * @return the system's temporary directory.
    * @throws RuntimeException if this method cannot find or create the system's temporary directory.
    *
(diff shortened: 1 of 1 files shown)


src/main/java/org/assertj/core/util/Files.java
```diff
@@ -82,7 +82,10 @@
    * 
    * @return the system's temporary directory.
    * @throws RuntimeException if this method cannot find or create the system's temporary directory.
+   *
+   * @deprecated For JUnit4, use TemporaryFolder and @Rule; for JUnit5, use TempDirectory extension and @TempDir instead.
```
scordio, line 86: I would refer to the rule/annotation directly: 

```suggestion
   * @deprecated Use either {@link org.junit.jupiter.api.io.TempDir} or {@link org.junit.rules.TemporaryFolder}
```

Sunt-ing: No problem. I will change that.

scordio:
Integrated, thanks @Sunt-ing!

PR #1857, opened 2020-04-30
Add assertion HasOnlyFields

### LINKED ISSUE #1496: Provide a way to assert all the fields a given class can contain
#### Summary

It would be nice if `AssertJ` provides a way to _check_ that a given class contains **only** a certain set of fields/members. I (and probably others) usually test creation/instantiation workflows for POJOs; for instance:

```kotlin
import nl.jqno.equalsverifier.EqualsVerifier
import org.assertj.core.api.Assertions
import org.junit.jupiter.api.Test
import java.time.Clock
import java.time.ZonedDateTime

internal class IssueTest {
  @Test
  fun `create proper instance`() {
    val start = ZonedDateTime.now(Clock.systemUTC())
    val instance = Issue("the title", "some description", "full content")
    val end = ZonedDateTime.now(Clock.systemUTC())
    Assertions.assertThat(instance)
        .hasNoNullFieldsOrPropertiesExcept("id")
        .satisfies {
          Assertions.assertThat(it.title).isEqualTo("the title")
          Assertions.assertThat(it.description).isEqualTo("some description")
          Assertions.assertThat(it.content).isEqualTo("full content")
          Assertions.assertThat(it.status).isEqualTo(Issue.Status.CREATED)
          Assertions.assertThat(it.createdAt).isBetween(start, end)
          Assertions.assertThat(it.modifiedAt).isBetween(start, end)
          Assertions.assertThat(it.version).isEqualTo(0)
        }
  }

  @Test
  fun `contract between equals and hashCode must be fulfilled`() {
    EqualsVerifier.forClass(Issue::class.java)
        .withIgnoredFields("createdAt", "modifiedAt", "id", "version")
        .verify()
  }
}
```

In that realm, we keep adding/removing/updating/modifying members, so having that kind of assertion(s) will let you have more confidence when you do such things.

I see there is already a `hasFieldOrProperty("fieldName")`, so I'm thinking about something like `hasOnlyFieldsOrProperties("field1", "field2", ..., "fieldN")`.

In the end, I'm not sure if this is super useful for some others, but I thought I would start somewhere :sunglasses:

wht-github:
#### Check List:
* Fixes #1496
* Unit tests : YES 

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
@@ -574,7 +574,47 @@ public abstract class AbstractObjectAssert<SELF extends AbstractObjectAssert<SEL
     objects.assertHasFieldOrPropertyWithValue(info, actual, name, value);
     return myself;
   }
-
+  /**
+   * Asserts that the actual object has only the specified fields or properties.
+   * <p>
+   * Private fields are matched by default but this can be changed by calling {@link Assertions#setAllowExtractingPrivateFields(boolean) Assertions.setAllowExtractingPrivateFields(false)}.
+   * <p>
+   *
+   * Example:
+   * <pre><code class='java'> public class TolkienCharacter {
+   *
+   *   private String name;
+   *   private int age;
+   *   // constructor omitted
+   *
+   *   public String getName() {
+   *     return this.name;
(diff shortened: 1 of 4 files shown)


joel-costigliola:
Integrated thanks @wht-github!

PR #1859, opened 2020-05-03
Explicitly set Javadoc locale to English

### LINKED ISSUE #1825: Explicitly set Javadoc locale to English.
#### Summary

exactly that https://github.com/spring-projects/spring-data-build/commit/88a625abc1c1dd52c6f14a3af52b1e9c25b2a91a

epeee:
This ensures that the html tag always contains language "en" and therefore does not default to the local used locale.

In order to verify that it works I've change the value to `de_DE` locally and the html tag did contain the de one afterwards.

#### Check List:
* Fixes #1825
* Unit tests : YES / NO / **NA**
* Javadoc with a code example (on API only) : YES / NO / **NA**

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -433,6 +433,8 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
         <configuration>
+          <!-- Explicitly set Javadoc locale to English -->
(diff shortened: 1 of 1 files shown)


joel-costigliola:
thanks @epeee, you could have merged the PR directly after a green build

PR #1863, opened 2020-05-05
Provide comparable assertions for non comparable class when given a comparator

### LINKED ISSUE #1492: Provide comparable assertions for non comparable class when given a comparator
#### Summary

Comparable assertions are restricted to `Comparable` types, this could extended for non `Comparable` types if a proper `Comparator` is given.

This is issue was created after this discussion https://stackoverflow.com/questions/55811121/assertj-fluent-ordering-tests-with-custom-comparator/55828748?noredirect=1#comment98322628_55828748.

#### Example

```java
Foo foo1 = ...;
Foo foo2 = ...;
FooComparator fooComparator = ...;
assertThat(foo1).usingComparator(fooComparator).isGreaterThan(foo2);
```

We want to avoid  to copy comparable assertions in `AbstractAssert` as it would pollute the API.

The idea of this improvement is to return a `ComparableAssert` from `usingComparator` and then chain comparable assertions.  Moreover we would ideally keep the original type assertions.
Whether this is technically possible is something to explore as part of this improvement.

wettera:
#### Check List:
* Fixes #1492 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -301,6 +301,44 @@ public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, 
     return myself;
   }
 
+  /**
+   * Given the Comparator<T> instance, verifies that the actual value is greater than the given one
+   * <p>
+   * Example:
+   * <pre><code class='java'> // assertions will pass
+   * assertThat('b').usingComparator(compara).isGreaterThan('a');
+   *
+   * // assertions will fail
+   * assertThat('a').usingComparator(compara).isGreaterThan('b');
+   *
+   * @param other the given value to compare the actual value to.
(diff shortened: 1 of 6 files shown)


joel-costigliola:

joel-costigliola:
The idea of this improvement was to return a `ComparableAssert` from `usingComparator` to avoid copying comparable assertions in `AbstractAssert`.

Whether this is technically possible is something to explore.

wettera:
> @wettera sorry the issue was not properly described, the idea of this improvement was to return a `ComparableAssert` from `usingComparator` to avoid copying comparable assertions in `AbstractAssert`, if we did that we would expose `isLessThan` to all objects when it does not make sense for non comparable ones (the API would be polluted).
> 
> Whether this is technically possible is something to explore.

I get your idea, I will try it later

joel-costigliola:
closed in favor of https://github.com/assertj/assertj-core/issues/2041

PR #1865, opened 2020-05-07
Fix javadoc

eddumelendez:


joel-costigliola:
Good catch @eddumelendez thanks

PR #1867, opened 2020-05-07
filteredOn function

### LINKED ISSUE #1620: filteredOn with Function for arrays and iterables
#### Summary

Extend the fluent filter API for arrays and iterables with a method to filter based on a `Function<? super ELEMENT, T>` parameter (e.g., a method reference to a getter) and a `T` parameter (for matching). 

The approach is similar to the `extracting(Function<? super ELEMENT, V>)` method. It will allow a save filtering without introspection.

#### Example

```java
assertThat(fellowshipOfTheRing).filteredOn(TolkienCharacter::getRace, HOBBIT)
                                   .containsOnly(sam, frodo, pippin, merry);

assertThat(fellowshipOfTheRing).filteredOn(character -> character.getRace().getName(), "Hobbit")
                                   .containsOnly(sam, frodo, pippin, merry);
```

mgrafl:
Implementation for `filteredOn(Function<? super ELEMENT, T>, T)` for `AbstractIterableAssert`, `AbstractObjectArrayAssert`, and `AtomicReferenceArrayAssert`.

#### TODO:
* Also add `filteredOn(Function, FilterOperator)`.

#### Questions: 
* The filteredOn implementation is currently located directly in `AbstractIterableAssert` etc. 
Should the implementation be moved to `org.assertj.core.api.filter.Filters` (i.e., in a public method `having(Function, String)`)?

#### Check List:
* Fixes #1620
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
@@ -2380,7 +2380,7 @@ public abstract class AbstractIterableAssert<SELF extends AbstractIterableAssert
   /**
    * Filter the iterable under test keeping only elements for which the result of the {@code function} is equal to {@code expectedValue}.
    * <p>
-   * It allows to filter elements in more safe way than by using {@link #filterOn(String, Object)}, as it
+   * It allows to filter elements in more safe way than by using {@link #filteredOn(String, Object)}, as it
    * doesn't utilize introspection.
    * <p>
    *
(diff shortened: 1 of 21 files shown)


mgrafl:
Upon further inspection of the source code, both the additionally proposed `filteredOn(Function, FilterOperator)` method and the relocation of the implementation into the `Filters` class would require substantial modifications. 
I would rather have a separate feature for `filteredOn(Function, FilterOperator)` (or maybe `filteredOn(Function, Condition)`) in the future.

mgrafl:
Note that the new method ```filteredOn(Function<? super ELEMENT, T>, T)``` can create ambiguity with ```filteredOn(String, Object)``` when called as ```filteredOn(null, null)``` (as shown by the disambiguation in the Unit tests). 
I hope this is not a show-stopper, @joel-costigliola.

joel-costigliola:
@mgrafl likely not as I'm not sure anyone will call `filteredOn(null, null)`

joel-costigliola:
Integrated thanks @mgrafl!

PR #1870, opened 2020-05-08
Add isEven/isOdd to AbstractByteAssert, AbstractShortAssert, AbstractIntegerAssert and AbstractLongAssert

### LINKED ISSUE #1807: Consider adding isEven/isOdd to AbstractByteAssert, AbstractShortAssert, AbstractIntegerAssert and AbstractLongAssert
#### Summary
It is currently not possible to directly assert whether integers are odd or even.

#### Example

```java
assertThat(someInteger).isEven();
assertThat(someInteger).isOdd();
```

Cal027:
#### Check List:
* Fixes #1807 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractByteAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractByteAssert.java
@@ -195,7 +195,7 @@ public abstract class AbstractByteAssert<SELF extends AbstractByteAssert<SELF>> 
    * @return this assertion object.
    * @throws AssertionError if the actual value is {@code null}.
    * @throws AssertionError if the actual value is not positive.
-   * @since 3.16.0
+   * @since 3.17.0
    */
   public SELF isEven() {
     bytes.assertIsEven(info, actual);
@@ -217,7 +217,7 @@ public abstract class AbstractByteAssert<SELF extends AbstractByteAssert<SELF>> 
    * @return this assertion object.
    * @throws AssertionError if the actual value is {@code null}.
    * @throws AssertionError if the actual value is not positive.
-   * @since 3.16.0
+   * @since 3.17.0
    */
   public SELF isOdd() {
     bytes.assertIsOdd(info, actual);
--- a/src/main/java/org/assertj/core/api/AbstractByteAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractByteAssert.java
@@ -35,6 +35,7 @@ import org.assertj.core.util.VisibleForTesting;
  * @author Alex Ruiz
  * @author Mikhail Mazursky
  * @author Nicolas François
+ * @author Cal027
  */
 public abstract class AbstractByteAssert<SELF extends AbstractByteAssert<SELF>> extends AbstractComparableAssert<SELF, Byte>
     implements NumberAssert<SELF, Byte> {
@@ -194,9 +195,32 @@ public abstract class AbstractByteAssert<SELF extends AbstractByteAssert<SELF>> 
    * @return this assertion object.
    * @throws AssertionError if the actual value is {@code null}.
    * @throws AssertionError if the actual value is not positive.
+   * @since 3.16.0
    */
   public SELF isEven() {
     bytes.assertIsEven(info, actual);
+    return myself;
+  }
+
+  /**
+   * Verifies that the actual value is odd.
+   * <p>
+   * Example:
+   * <pre><code class='java'> // assertions will pass
+   * assertThat((byte) 3).isOdd();
+   * assertThat((byte) -17).isOdd();
+   *
+   * // assertions will fail
+   * assertThat((byte) 2).isOdd();
+   * assertThat((byte) -24).isOdd();</code></pre>
+   *
+   * @return this assertion object.
+   * @throws AssertionError if the actual value is {@code null}.
+   * @throws AssertionError if the actual value is not positive.
+   * @since 3.16.0
+   */
+  public SELF isOdd() {
+    bytes.assertIsOdd(info, actual);
     return myself;
   }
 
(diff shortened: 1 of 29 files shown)


src/main/java/org/assertj/core/internal/Bytes.java
```diff
@@ -63,4 +65,29 @@ protected boolean isGreaterThan(Byte value, Byte other) {
...
+   * Asserts that the actual Byte value is even.
+   *
+   * @param info contains information about the assertion.
+   * @param actual the actual value.
+   * @throws AssertionError if the actual value is {@code null}.
+   * @throws AssertionError if the actual value is negative.
+   */
+  public void assertIsEven(AssertionInfo info, Byte actual) {
+    Byte lastDigit = (byte) (actual & one());
+    assertIsZero(info, lastDigit);
```
joel-costigliola, line 78: Although the `assertIsZero` check is correct, it will give a confusing error message:
```
"%nExpecting:%n <1>%nto be equal to:%n <0>%nbut was not."
``` 
The better approach is to define a specific error message for even assertion and one for odd.
I suggest:
```
"%nExpecting <%s> to be odd"
``` 

These error messages must be defined in `ShouldBeOdd`/`ShouldBeEven`, use `ShouldBeBase64`/`ShouldBeBase64_create_Test` for an implementation guide.

Having a specific error message requires to define `assertIsOdd`/`assertIsEven`,  where to do that?
* in `org.assertj.core.internal.Numbers` 
* in a new interface `WholeNumbers` with default methods

I have given a quick try to this second option, 
```
public interface WholeNumbers<NUMBER extends Number> {

  boolean isEven(NUMBER number);

  default boolean isOdd(NUMBER number) {
    return !isEven(number);
  }

  default void assertIsOdd(NUMBER number, AssertionInfo info) {
    if (!isOdd(number)) throw Failures.instance().failure(info, (number));
  }

  default void assertIsEven(NUMBER number, AssertionInfo info) {
    if (!isEven(number)) throw Failures.instance().failure(info, shouldBeEven(number));
  }
}
```
This approach needs to be evaluated, let me know if you need clarifications for that.

src/test/java/org/assertj/core/internal/bytes/Bytes_assertIsEven_Test.java
```diff
@@ -0,0 +1,93 @@
...
+import org.assertj.core.internal.BytesBaseTest;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Tests for <code>{@link Bytes#assertIsEven(AssertionInfo, Byte)}</code>.
+ *
+ * @author Cal027
+ */
+public class Bytes_assertIsEven_Test extends BytesBaseTest {
```
joel-costigliola, line 31: Prefer using `expectAssertionError` over `assertThatExceptionOfType(AssertionError.class)` and introduce GIVEN WHEN THEN as indicated in https://github.com/joel-costigliola/assertj-core/blob/master/CONTRIBUTING.md.
`InputStreams_assertIsEmpty_Test` is a good example to follow too.

Most of these tests can be parameterized (thanks JUnit 5), ex in `Classes_assertHasSuperclass_Test`:
```
  @ParameterizedTest
  @ValueSource(classes = { Object.class, Comparable.class, String.class })
  void should_fail_if_actual_has_not_given_class_as_direct_superclass(Class<?> superclass) {
    // GIVEN
    Class<?> actual = Integer.class;
    // WHEN
    AssertionError assertionError = expectAssertionError(() -> classes.assertHasSuperclass(someInfo(), actual, superclass));
    // THEN
    then(assertionError).hasMessage(shouldHaveSuperclass(actual, superclass).create());
  }
```
You can use short, byte, int and long in `ValueSource`, ex:
```
  @ParameterizedTest
  @ValueSource(bytes = { 0, 2, -2, 0x02, -0x02 })
  void should_succeed_since_actual_is_even(byte b) {
    bytes.assertIsEven(someInfo(), b);
  }
```

joel-costigliola: I'm only commenting this class but apply the same changes to other similar tests.

src/main/java/org/assertj/core/api/AbstractByteAssert.java
```diff
@@ -179,6 +180,50 @@ public SELF isNotPositive() {
...
+   * assertThat((byte) 46).isEven();
+   *
+   * // assertions will fail
+   * assertThat((byte) 3).isEven();
+   * assertThat((byte) 15).isEven();</code></pre>
+   *
+   * @return this assertion object.
+   * @throws AssertionError if the actual value is {@code null}.
+   * @throws AssertionError if the actual value is not positive.
+   * @since 3.16.0
```
joel-costigliola, line 198: By lack of time I could not review it for 3.16.0 so it's 3.17.0 (promise!)

Cal027: Thanks for the review! I will keep working this week.

src/main/java/org/assertj/core/internal/WholeNumbers.java
```diff
@@ -0,0 +1,36 @@
+package org.assertj.core.internal;
+
+import org.assertj.core.api.AssertionInfo;
+import static org.assertj.core.error.ShouldBeEven.shouldBeEven;
+import static org.assertj.core.error.ShouldBeOdd.shouldBeOdd;
+
+public interface WholeNumbers<NUMBER extends Number> {
```
joel-costigliola, line 7: nice work!

joel-costigliola:
Integrated thanks @Cal027!

PR #1878, opened 2020-05-19
In the event that an iterable is only singly-passable, ie can only be…

rpolton:
… read once (and is destroyed after reading) then the assertion was failing because the iterable was first read into a list and secondly parsed for equal order. Instead read the copied list as the original iterable may have been destroyed.

#### Check List:
* Fixes #??? -- of course :D :D Change 'actual' to 'actualAsList' in Iterables.java and add a test
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/internal/Iterables.java
+++ b/src/main/java/org/assertj/core/internal/Iterables.java
@@ -338,7 +338,7 @@ public class Iterables {
 
   private void assertIterableContainsGivenValues(Iterable<?> actual, Object[] values, AssertionInfo info) {
     Set<Object> notFound = stream(values).filter(value -> !iterableContains(actual, value))
-                                         .collect(toCollection(LinkedHashSet::new));
+      .collect(toCollection(LinkedHashSet::new));
     if (notFound.isEmpty())
       return;
     throw failures.failure(info, shouldContain(actual, values, notFound, comparisonStrategy));
@@ -387,8 +387,8 @@ public class Iterables {
 
     if (!unexpectedValues.isEmpty() || !missingValues.isEmpty()) {
       throw failures.failure(info, shouldContainOnly(actual, expectedValues,
-                                                     missingValues, unexpectedValues,
-                                                     comparisonStrategy));
+        missingValues, unexpectedValues,
+        comparisonStrategy));
     }
   }
 
@@ -568,7 +568,7 @@ public class Iterables {
     assertNotNull(info, actual);
     checkIterableIsNotNull(values);
     List<Object> extra = stream(actual).filter(actualElement -> !iterableContains(values, actualElement))
-                                       .collect(toList());
+      .collect(toList());
     if (extra.size() > 0) throw failures.failure(info, shouldBeSubsetOf(actual, values, extra, comparisonStrategy));
   }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1083,7 +1083,7 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
     if (!diff.differencesFound()) {
       // actual and values have the same elements but are they in the same order ?
       int i = 0;
-      for (Object elementFromActual : actual) {
+      for (Object elementFromActual : actualAsList) {
```
joel-costigliola, line 1086: don't reuse `actualAsList`, although it's not the case it could have modified in `diff`, simply create another list

src/test/java/org/assertj/core/internal/iterables/Iterables_assertContainsExactly_Test.java
```diff
@@ -47,6 +48,37 @@ public void should_pass_if_actual_contains_exactly_given_values() {
...
+    return new Iterable<String>() {
+      private boolean isIteratorCreated = false;
+
+      @Override
+      public Iterator<String> iterator() {
+        if (isIteratorCreated)
+          throw new IllegalArgumentException("Cannot create two iterators on a singly-iterable sequence");
+        isIteratorCreated = true;
+        return new Iterator<String>() {
+          private final Iterator<String> l_it = values.iterator();
```
joel-costigliola, line 66: rename to `listIterator`

src/test/java/org/assertj/core/internal/iterables/Iterables_assertContainsExactly_Test.java
```diff
@@ -68,19 +100,19 @@ public void should_fail_if_array_of_values_to_look_for_is_empty_and_actual_is_no
   @Test
   public void should_throw_error_if_array_of_values_to_look_for_is_null() {
     assertThatNullPointerException().isThrownBy(() -> iterables.assertContainsExactly(someInfo(), emptyList(), null))
-                                    .withMessage(valuesToLookForIsNull());
+      .withMessage(valuesToLookForIsNull());
```
joel-costigliola, line 103: general comment: use assertj code formatting settings, the changes in this PR is 90% code formatting which is a lot of noise. thanks!

joel-costigliola:
@rpolton reviewed, a few more changes and we good

src/test/java/org/assertj/core/internal/iterables/Iterables_assertContainsExactly_Test.java
rpolton: Sorry, I didn't notice my IDE had changed these. I'd even made sure the 'reformat' option was not selected. Will try to undo.

src/main/java/org/assertj/core/internal/Iterables.java
rpolton: I'm not sure about this suggestion because I don't believe we should be encouraging methods with side-effects. Also I note that there is an explicit comment on the IterableDiff class that reads "immutable" so, are you sure you want a second copy of the data?

joel-costigliola:
@rpolton let me know when your PR is ready for another review

rpolton:
> @rpolton let me know when your PR is ready for another review

Yup, it's ready. I've effectively resubmitted so all the formatting changes should have gone

src/main/java/org/assertj/core/internal/Iterables.java
joel-costigliola: fair enough

joel-costigliola:
Integrated thanks @rpolton! (last commit message made me laugh :smile: )

PR #1880, opened 2020-05-22
Fix issue #1879: allow failures to be recorded from multiple threads

### LINKED ISSUE #1879: There appears no way to add failures to SoftAssertions from multiple threads
#### Summary

Consider I need to collect multiple failures produced in multiple threads. Currently, `SoftAssertions` doesn't allow me to do so w/o extra synchronizers.

#### Example

For example, this code is broken because changes made to the `SoftAssertions` instance from other threads might not be visible from the main thread:

```java
        final SoftAssertions softly = new SoftAssertions();

        final ForkJoinPool forkJoinPool = ForkJoinPool.commonPool();
        forkJoinPool.execute(() -> {
            softly.assertThat("foo").hasToString("foo");
        });
        forkJoinPool.execute(() -> {
            softly.assertThat("bar").hasToString("bar");
        });

        softly.assertAll();
```

What is the best approach to solve the above problem?

Guarding `SoftAssertions` with a synchronized block wherever I'm accessing the instance doesn't seem like a good solution (especially from test readability standpoint).

Probably you could introduce something like `SynchronizedSoftAssertions` or least make the list of failures a _synchronized list_. At the same time, it's not clear how to guard the whole call chain like

```java
softly.assertThat("foo").hasToString("foo");
```

with a single lock.

unix-junkie:
#### Check List:
* Fixes #1879 
* Unit tests : NA

### DIFF
--- a/src/main/java/org/assertj/core/api/ErrorCollector.java
+++ b/src/main/java/org/assertj/core/api/ErrorCollector.java
@@ -34,8 +34,9 @@ public class ErrorCollector {
   private static final String INTERCEPT_METHOD_NAME = "intercept";
   private static final String CLASS_NAME = ErrorCollector.class.getName();
 
+  // The list is synchronized in case errors arrive from more than a single thread.
   // scope : the current softassertion object
-  private final List<AssertionError> errors = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


joel-costigliola:
Integrated thanks @unix-junkie!

PR #1882, opened 2020-05-24
Implement lazy withFailMessage / overridingErrorMessage

### LINKED ISSUE #1837: lazy withFailMessage /  overridingErrorMessage
#### Summary

junit provide a lazy fail message description by using a supplier. The supplier is called only if an error occurs, otherwise it is not called.

```java
        Assertions.assertTrue(true, () -> "I won't be called");
        Assertions.assertTrue(false, () -> "I will be called");
```
It is useful for at least two reason : 

- performance optimization if computing the custom error message is costly.
- getting error message may not be possible if there is not an error, however assertThat try to get the error message even if there is no error.

Suggestion, add the same capability to withFailMEssage or overridingErrorMessage, example : 

```java
        assertThat(response.isError())
            .withFailMessage(() -> "my error")
            .isFalse();
```

Eveneko:
#### Check List:
* Fixes #1837  
* Unit tests : YES

We implement a lazy fail message description by using a supplier to improve the performance.
We add a new variable `Supplier<String> overridingErrorMessageSupplier` in `WritableAssertionInfo` and implement `overridingErrorMessage(Supplier)`
Also the method `overridingErrorMessage()` that gets the overridden message will get the String message from supplier `verridingErrorMessageSupplier`
We override `withFailMessage(Supplier)` and `overridingErrorMessage(Supplier)` in `AbstractAssert` they will set the supplier in `WritableAssertionInfo`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -170,7 +170,7 @@
     <dependency>
       <groupId>nl.jqno.equalsverifier</groupId>
       <artifactId>equalsverifier</artifactId>
-      <version>3.1.13</version>
+      <version>3.2</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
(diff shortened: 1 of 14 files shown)


joel-costigliola:
Integrated thanks @Eveneko!

Issue #1182, opened 2018-02-12
Multidimensional arrays assertion


radistao:
#### Summary

Pursuing SO question [AssertJ multidimensional arrays assertion (deep equals)
](https://stackoverflow.com/questions/48652678/assertj-multidimensional-arrays-assertion-deep-equals/) i'd like to try to make an assertion for multidimensional arrays.

I think i'd implement it similar to native Java `Arrays#deepEquals()`, just "extending" it with nicer assertion exceptions, including indexes and actual/expected values.

#### Example

```java
int[][] expected = {
    new int[]{1, 0, 1},
    new int[]{0, 0, 0},
    new int[]{1, 1, 1}
};
 
int[][] actual = gameOfLife.play(board);

assertThat(actual).isDeepEqualTo(expected);
```

joel-costigliola:
That would be a nice addition, bear in mind that it would need to be added for all primitive types and Object which is some substantial work.

mwajcht:
Two years and nothing happened to this issue, so I took it and results can be seen in this quite a big pull request.

joel-costigliola:
@mwajcht let's start with one of the array type (int for example) and then when this is done, we can replicate it to other types

mwajcht:
Ok, I created the second, much smaller pull request. It contains only changes for int two-dimensional arrays.

joel-costigliola:
thanks, that will make it much easier!

mwajcht:
@joel-costigliola As assertions for two-dimensional int arrays are already merged, I can apply your comments to other types. Do you prefer separate pull requests for each type or all-in-one?

joel-costigliola:
@mwajcht I would prefer separate PRs, I might have missed things to improve (hopefully mot many) and seeing more than 100 files ina PR just sucks the energy out of me ;-)

mwajcht:
I added a few pull requests with assertions for other types.

joel-costigliola:
@mwajcht waht would you think of renaming `hasSize` to `hasDimensions`, the latter feels more natural and shows we are checking for multiple dimensions whereas the singular `size`?

BTW, I have renamed the master `branch` to `main`, your PR have been updated accordingly in this repo but not yours, don't be surprised.

mwajcht:
[...]

PR #1883, opened 2020-05-24
Added two-dimensional arrays assertions

mwajcht:
#### Check List:
* Fixes #1182
* Unit tests : YES


mwajcht:
Will be split into multiple pull requests - one for each type

PR #1892, opened 2020-05-30
Add satisfy in AbstractIterableAssert

### LINKED ISSUE #1621: Consumer-based "satisfy" assertions for elements in arrays and iterables
#### Summary

Consider extending the assertions API for arrays and iterables with `satisfy*` methods accepting multiple  `Consumer<? super ELEMENT>`s. The consumers probe the individual elements of the tested collection.

The additional methods could be: 
* `satisfy(Consumer<? super ELEMENT>... consumers)`: For each element of the tested collection the (remaining) consumers are called; if a consumer accepts the element, then the element and that consumer are removed from their collections and the next element is probed.
* `satisfyExactly(Consumer<? super ELEMENT>... consumers)`: The tested collection must have the same size as the consumers. Each element of the tested collection is fed to the corresponding consumer.
* `satisfyExactlyInAnyOrder(Consumer<? super ELEMENT>... consumers)`: The tested collection must have the same size as the consumers. For each element of the tested collection the (remaining) consumers are called; if a consumer accepts the element, then the element and that consumer are removed from their collections and the next element is probed.

This would allow more flexible assertions on arrays and iterables. 

#### Example

```java
assertThat(trilogy).satisfy(movie -> {
    // the two Towers
    assertThat(movie.getReleaseDate()).isEqualTo(parse("2002-12-18"));
    assertThat(movie.getTitle()).contains("Towers");
  }, 
  movie -> {
    // the fellowship of the Ring
    assertThat(movie.getReleaseDate()).isBefore(parse("2002-12-17"));
    assertThat(movie.xrated).isFalse();
  }
);

assertThat(trilogy).satisfyExactly(movie -> {
    // the fellowship of the Ring
    assertThat(movie.getReleaseDate()).isBefore(parse("2002-12-17"));
    assertThat(movie.xrated).isFalse();
  }, 
  movie -> {
    // the two Towers
     assertThat(movie.getReleaseDate()).isEqualTo(parse("2002-12-18"));
     assertThat(movie.getTitle()).contains("Towers");
  }, 
  movie -> {
    // the Return of the King
    assertThat(movie.getTitle()).endsWith("King");
  }
);

assertThat(trilogy).satisfyExactlyInAnyOrder(movie -> {
    // the two Towers
    assertThat(movie.getReleaseDate()).isEqualTo(parse("2002-12-18"));
    assertThat(movie.getTitle()).contains("Towers");
  }, 
  movie -> {
    // the fellowship of the Ring
    assertThat(movie.getReleaseDate()).isBefore(parse("2002-12-17"));
    assertThat(movie.xrated).isFalse();
  }, 
  movie -> {
    // the Return of the King
    assertThat(movie.getTitle()).endsWith("King");
  }
);
```

#### Alternatives

The current API provides the following alternatives (for the proposed `satisfyExactlyInAnyOrder` method): 

1. `extracting` multiple fields and assert matching `tuple`s:
```java
assertThat(trilogy)
  .extracting(Movie::getTitle, Movie::getReleaseDate, movie -> movie.xrated)
  .containsExactlyInAnyOrder(
    tuple("the Return of the King", parse("2003-12-17"), false),
    tuple("the fellowship of the Ring", parse("2001-12-19"), false), 
    tuple("the two Towers", parse("2002-12-18"), false)
  );
```
This approach requires all values to match exactly, requires all elements to be tested for the same fields, is not type safe, and gets unreadable above three fields. 

2. Individual `assertThat` for each desired element with `filterOn` distinguishing characteristic of that element: 
```java
assertThat(trilogy).hasSize(3);

assertThat(trilogy)
  .filterOn(movie -> parse("2002-12-17").compareTo(movie.getReleaseDate()) > 0)
  .hasSize(1)
  .element(0)
  .satisfies(movie -> {
    assertThat(movie.xrated).isFalse();
  });

assertThat(trilogy)
  .filterOn("releaseDate", parse("2002-12-18"))
  .hasSize(1)
  .element(0)
  .satisfies(movie -> {
    assertThat(movie.getTitle()).contains("Towers");
  });

assertThat(trilogy)
  .filterOn(movie -> movie.getTitle().endsWith("King"))
  .hasSize(1);
```
While this approach is more flexible than the first, it is more verbose and does usually not reflect the mental model of how a list is tested.

Sunt-ing:
#### Check List:
* Fixes #1621 
* Unit tests : YES

This PR is for `iterables.satisfy`. If this one is completely finished, I'll finish `iterables.satisfyExactly`, `iterables.satisfyExactlyInAnyOrder`, and then `arrays.` part.


Sunt-ing:
I have completed my code, please review them at your convenience. @joel-costigliola

src/main/java/org/assertj/core/api/AbstractIterableAssert.java
```diff
@@ -2754,6 +2754,50 @@ public SELF allSatisfy(Consumer<? super ELEMENT> requirements) {
...
+   *       assertThat(s).contains("L");
+   *     }, s --&gt; {
+   *       assertThat(s).contains("L");
+   *     }, s --&gt; {
+   *       assertThat(s).contains("L");
+   *     }));</code></pre>
+   *
+   * @param consumers the consumers that are expected to be satisfied by the elements of the given {@code Iterable}.
+   * @return this assertion object.
+   *
```
mgrafl, line 2792: As this is a new feature, should we add a "since" tag?
```suggestion
   * @since 3.17.0
```

Sunt-ing: Yes, sure. Thanks!

src/main/java/org/assertj/core/api/AbstractIterableAssert.java
```diff
@@ -2754,6 +2754,51 @@ public SELF allSatisfy(Consumer<? super ELEMENT> requirements) {
...
+   *       assertThat(s).doesNotContain("L");
+   *     }));
+   *
+   * assertThat(actual).satisfy((s --&gt; {
+   *       assertThat(s).contains("L");
+   *     }, s --&gt; {
+   *       assertThat(s).contains("L");
+   *     }, s --&gt; {
+   *       assertThat(s).contains("L");
+   *     }));</code></pre>
```
mgrafl, line 2788: Would it be more readable to name the list "starWarsCharacterNames" or "names" instead of "actual"?
Similarly, could the lambda arguments renamed from "s" to "name"?

In the code examples, the ```consumers``` argument of the ```satisfy``` method has two parentheses. Can the inner parenthesis be removed (i.e., ```satisfy(s --&gt; ``` instead of ```satisfy((s --&gt; ```)?

src/main/java/org/assertj/core/api/AbstractIterableAssert.java
```diff
@@ -2754,6 +2754,51 @@ public SELF allSatisfy(Consumer<? super ELEMENT> requirements) {
...
   }
 
+  /**
+   * Verifies that the all the given consumers can be satisfied by elements in the iterable with
+   * an element at most satisfies one consumer. No order requirement.
+   * <p>
+   * This assertion is to test whether there are enough elements in the iterable that can satisfy all the consumers.
+   * <p>
+   * Grouping assertions example :
+   * <pre><code class='java'> // second constructor parameter is the light saber color
```
mgrafl, line 2764: Since the examples is not primarily about grouping, maybe just call them "Examples"?
Since we just create a list of Strings, is there a copy/paste error about the second constructor parameter? 

```suggestion
   * Examples:
   * <pre><code class='java'>
```

src/main/java/org/assertj/core/api/AbstractIterableAssert.java
```diff
@@ -2754,6 +2754,51 @@ public SELF allSatisfy(Consumer<? super ELEMENT> requirements) {
     return myself;
   }
 
+  /**
+   * Verifies that the all the given consumers can be satisfied by elements in the iterable with
+   * an element at most satisfies one consumer. No order requirement.
```
mgrafl, line 2759: Typos
```suggestion
   * Verifies that all given consumers can be satisfied by elements in the iterable under test with
   * an element at most satisfying one consumer. No order requirement.
```

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1108,6 +1109,51 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
       throw failures.failure(info, elementsShouldSatisfy(actual, unsatisfiedRequirements, info));
   }
 
+  /**
+   * Verifies that the all the given consumers can be satisfied by elements in the iterable with
+   * an element at most satisfies one consumer. No order requirement.
```
mgrafl, line 1114: Typos
```suggestion
   * Verifies that all given consumers can be satisfied by elements in the iterable under test with
   * an element at most satisfying one consumer. No order requirement.
```

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1108,6 +1109,51 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
...
+          consumer.accept(element);
+        } catch (AssertionError ex) {
+          return false;
+        }
+        return true;
+      }).collect(toList());
+    }
+    if (!isSatisfied(stasfiedElementsLists, 0))
+      throw failures.failure(info, shouldSatisfy(actual));
+  }
```
mgrafl, line 1142: Typo in variable name "stasfiedElementsList". 
```suggestion
    List<E>[] satisfiedElementsLists = new ArrayList[consumers.length];
    for (int i = 0; i < consumers.length; i++) {
      Consumer<? super E> consumer = consumers[i];
      satisfiedElementsLists[i] = stream(actual).filter(element -> {
        try {
          consumer.accept(element);
        } catch (AssertionError ex) {
          return false;
        }
        return true;
      }).collect(toList());
    }
    if (!isSatisfied(satisfiedElementsLists, 0))
      throw failures.failure(info, shouldSatisfy(actual));
  }
```

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1108,6 +1109,51 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
...
+   * @param consumers the consumers that are expected to be satisfied by the elements of the given {@code Iterable}.
+   * @throws NullPointerException if the given consumers is {@code null}.
+   * @throws AssertionError if the given {@code Iterable} is {@code null}.
+   * @throws AssertionError if any {@code Consumer} in the consumers cannot be satisfied by elements in the given {@code Iterable}.
+   */
+  @SafeVarargs
+  public final <E> void assertSatisfy(AssertionInfo info, Iterable<? extends E> actual, Consumer<? super E>... consumers) {
+    assertNotNull(info, actual);
+    requireNonNull(consumers, "The Consumer<? super E>... expressing the assertions consumers must not be null");
+    requireNonNull(consumers[0], "The Consumer<? super E>... expressing the assertions consumers must not be null");
```
mgrafl, line 1127: Is there a particular reason why the first consumer must be non-null? If so, why is it necessary that the consumers is non-empty (and is the length requirement checked somewhere else)? Should other consumers also be explicitly checked for non-nullness?

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1108,6 +1109,51 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
...
+      throw failures.failure(info, shouldSatisfy(actual));
+  }
+
+  private static <E> boolean isSatisfied(List<E>[] stasfiedElementsLists, int begin) {
+    if (begin == stasfiedElementsLists.length) return true;
+    if (stasfiedElementsLists[begin].size() == 0) return false;
+
+    for (E element : stasfiedElementsLists[begin]) {
+      List<E>[] listsCopy = stasfiedElementsLists.clone();
+      for (int i = begin + 1; i < stasfiedElementsLists.length; i++)
```
mgrafl, line 1150: Typo in variable name "stasfiedElementsList".
```suggestion
  private static <E> boolean isSatisfied(List<E>[] satisfiedElementsLists, int begin) {
    if (begin == satisfiedElementsLists.length) return true;
    if (satisfiedElementsLists[begin].size() == 0) return false;

    for (E element : satisfiedElementsLists[begin]) {
      List<E>[] listsCopy = satisfiedElementsLists.clone();
      for (int i = begin + 1; i < satisfiedElementsLists.length; i++)
```

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1108,6 +1109,51 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
...
+          return false;
+        }
+        return true;
+      }).collect(toList());
+    }
+    if (!isSatisfied(stasfiedElementsLists, 0))
+      throw failures.failure(info, shouldSatisfy(actual));
+  }
+
+  private static <E> boolean isSatisfied(List<E>[] stasfiedElementsLists, int begin) {
```
mgrafl, line 1144: What does the ```isSatisfied``` method do? Maybe some documentation would help, or the method could somehow be simplified (there seems to be a quite a bit of array cloning and recursion going on)?

On array cloning: since we clone a ```List<E>[]```, it might not be immediately clear whether the individual lists are cloned as well.

Sunt-ing: Maybe we can replace it with another approach:
```java
  private static <E> boolean isSatisfied(List<E>[] satisfiedElementsLists, int begin) {
    // recursively test whether we can find any specific matching strategy that can meet the requirements
    if (begin == satisfiedElementsLists.length) return true;   // all consumers have been satisfied
    if (satisfiedElementsLists[begin].size() == 0) return false;   // no satisfied element for consumers[begin]

    for (E element : satisfiedElementsLists[begin]) {
      List<Integer> indeces = newArrayList();
      for (int i = begin + 1; i < satisfiedElementsLists.length; i++) {
        if (satisfiedElementsLists[i].contains(element)) {
          indeces.add(i);
          satisfiedElementsLists[i].remove(element);
        }
      }

      if (isSatisfied(satisfiedElementsLists, begin + 1)) return true;
      for (int i : indeces)
        satisfiedElementsLists[i].add(element); // restore the array
    }
    return false;
  }

```

src/main/java/org/assertj/core/internal/Iterables.java
Sunt-ing: "Is there a particular reason why the first consumer (`consumers[0]`) must be non-null?": Oh, it's my mistake! Thanks! 
"why is it necessary that the `consumers` is non-empty": Maybe the test can be directly passed if the `consumers` is empty or null?

mgrafl: Could it be that the wrong line (```consumers``` instead of ```consumers[0]```) was accidentally removed?

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1108,6 +1109,58 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
...
+      }).collect(toList());
+    }
+    if (!isSatisfied(satisfiedElementsLists, 0))
+      throw failures.failure(info, shouldSatisfy(actual));
+  }
+
+  private static <E> boolean isSatisfied(List<E>[] satisfiedElementsLists, int begin) {
+    // recursively test whether we can find any specific matching strategy that can meet the requirements
+    if (begin == satisfiedElementsLists.length) return true;   // all consumers have been satisfied
+    if (satisfiedElementsLists[begin].size() == 0) return false;   // no satisfied element for consumers[begin]
```
mgrafl, line 1146: Would ```isEmpty()``` be more readable here? 
```suggestion
    if (satisfiedElementsLists[begin].isEmpty()) return false;   // no satisfied element for consumers[begin]
```

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1108,6 +1109,58 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
...
+
+    for (E element : satisfiedElementsLists[begin]) {
+      List<Integer> indeces = newArrayList();
+      for (int i = begin + 1; i < satisfiedElementsLists.length; i++) {
+        if (satisfiedElementsLists[i].contains(element)) {
+          indeces.add(i);
+          satisfiedElementsLists[i].remove(element);
+        }
```
mgrafl, line 1154: The ```List.remove(Object)``` method returns a boolean indicating whether the list contained the element. 
The ```if``` block could be shortened this way (but I am not sure whether it would be more readable). 
What is your preference?

```suggestion
        if (satisfiedElementsLists[i].remove(element)) {
          indeces.add(i);
        }
```

src/main/java/org/assertj/core/internal/Iterables.java
Sunt-ing: I am thinking about: if a consumer is empty or null, what will happen?

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1108,6 +1109,60 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
...
+        } catch (AssertionError ex) {
+          return false;
+        }
+        return true;
+      }).collect(toList());
+    }
+    if (!isSatisfied(satisfiedElementsLists, 0))
+      throw failures.failure(info, shouldSatisfy(actual));
+  }
+
```
mgrafl, line 1145: There is already a method ```byPassingAssertions```, which takes care of the try/catch-block around ```consumer.accept```. 
Should we extract some methods to make the intend of the main method clearer?

```suggestion
    List<?>[] satisfiedElementsLists = Arrays.stream(consumers)
                                             .map(listFilteredBySatisfiedElements(actual))
                                             .toArray(List[]::new);
    
    if (!isSatisfied(satisfiedElementsLists))
      throw failures.failure(info, shouldSatisfy(actual));
  }

  private <E> Function<? super Consumer<? super E>, ?> listFilteredBySatisfiedElements(Iterable<? extends E> actual) {
    return consumer -> stream(actual).filter(byPassingAssertions(consumer)).collect(toList());
  }

  private static <E> boolean isSatisfied(List<E>[] satisfiedElementsLists) {
    return isSatisfied(satisfiedElementsLists, 0);
  }


```

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1108,6 +1109,60 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
...
+    for (E element : satisfiedElementsLists[begin]) {
+      List<Integer> indeces = newArrayList();
+      for (int i = begin + 1; i < satisfiedElementsLists.length; i++) {
+        if (satisfiedElementsLists[i].remove(element)) {
+          indeces.add(i);
+        }
+      }
+
+      if (isSatisfied(satisfiedElementsLists, begin + 1)) return true;
+      for (int i : indeces)
```
mgrafl, line 1160: Would the variable name "removedElementRowIndices" be a better fit (also fixing the typo)?

```suggestion
      List<Integer> removedElementRowIndices = newArrayList();
      for (int i = begin + 1; i < satisfiedElementsLists.length; i++) {
        if (satisfiedElementsLists[i].remove(element)) {
          removedElementRowIndices.add(i);
        }
      }

      // Check whether remaining consumers can be satisfied
      if (isSatisfied(satisfiedElementsLists, begin + 1)) return true;
      
      for (int i : removedElementRowIndices)
```

src/main/java/org/assertj/core/internal/Iterables.java
Sunt-ing: The stream processing part is great, but I'm not sure it is worth having another overload for `isSatisfied`.

Sunt-ing: Would it be better to change the type of `satisfiedElementsLists` from `List<?>[]` to `List<E>[]` or `List[]`?

mgrafl: I understand your concern about overloading ```isSatisfied```. I often use such overloading to have a dedicated entry point for recursive methods. But I agree, the overloaded method might distract more than it helps. Feel free to remove it. 

I think the best type of ```satisfiedElementsLists``` is ```List<E>[]```. To avoid compiler warnings, it could be written as: 
```
    List<E>[] satisfiedElementsLists = Arrays.stream(consumers)
                                             .map(listFilteredBySatisfiedElements(actual))
                                             .<List<E>>toArray(List[]::new);
```
```suggestion
    List<E>[] satisfiedElementsLists = Arrays.stream(consumers)
                                             .map(listFilteredBySatisfiedElements(actual))
                                             .<List<E>>toArray(List[]::new);
    
    if (!isSatisfied(satisfiedElementsLists, 0))
      throw failures.failure(info, shouldSatisfy(actual));
  }
  private <E> Function<? super Consumer<? super E>, List<E>> listFilteredBySatisfiedElements(Iterable<? extends E> actual) {
    return consumer -> stream(actual).filter(byPassingAssertions(consumer)).collect(toList());
  }

```

Sunt-ing: maybe it would be better to static import `stream`, just as we did in method `Iterables.assertAllSatisfy`

src/test/java/org/assertj/core/internal/iterables/Iterables_assertSatisfy_Test.java
```diff
@@ -0,0 +1,148 @@
...
+    AssertionError assertionError = expectAssertionError(() -> assertThat(actual).satisfy(consumer));
+    // THEN
+    then(assertionError).hasMessage(actualIsNull());
+  }
+
+  @Test
+  public void should_throw_error_if_consumer_array_is_null() {
+    // GIVEN
+    Consumer<String>[] consumers = null;
+    Object o = null;
```
mgrafl, line 86: Object ```o``` is never used.
```suggestion
```

src/test/java/org/assertj/core/internal/iterables/Iterables_assertSatisfy_Test.java
```diff
@@ -0,0 +1,148 @@
...
+      assertThat(s).doesNotContain("L");
+    };
+    Consumer<String> consumer2 = s -> {
+      assertThat(s).doesNotContain("a");
+    };
+
+    // WHEN/THEN
+    iterables.assertSatisfy(info, actual, consumer1, consumer2);
+  }
+
```
mgrafl, line 67: Should we add unit tests for element order independence, duplicate elements, and no consumers? 
```suggestion

  @Test
  public void should_pass_there_are_multiple_consumers_and_can_be_satisfied_in_any_order() {
    // GIVEN
    Consumer<String> consumer1 = s -> assertThat(s).contains("L"); // Matches "Luke" and "Leia"
    Consumer<String> consumer2 = s -> assertThat(s).doesNotContain("a"); // Matches "Luke"

    // WHEN/THEN
    iterables.assertSatisfy(info, actual, consumer1, consumer2);
  }
  
  @Test
  public void should_pass_if_iterable_contains_multiple_equal_elements() {
    // GIVEN
    List<String> names = newArrayList("Luke", "Luke");
    Consumer<String> consumer1 = s -> assertThat(s).contains("L");
    Consumer<String> consumer2 = s -> assertThat(s).contains("u");

    // WHEN/THEN
    iterables.assertSatisfy(info, names, consumer1, consumer2);
  }

  @Test
  public void should_pass_if_there_is_no_consumer() {
    // WHEN/THEN
    iterables.assertSatisfy(info, actual);
  }

```

src/test/java/org/assertj/core/internal/iterables/Iterables_assertSatisfy_Test.java
```diff
@@ -0,0 +1,148 @@
...
+public class Iterables_assertSatisfy_Test extends IterablesBaseTest {
+
+  private List<String> actual = newArrayList("Luke", "Leia", "Yoda");
+
+  @Test
+  public void should_pass_if_there_is_only_one_consumer_and_can_be_satisfied() {
+    // GIVEN
+    Consumer<String> consumer = s -> {
+      assertThat(s.length()).isEqualTo(4);
+    };
```
mgrafl, line 47: If a consumer contains only a single assertion statement, should we use the shorthand notion without brackets? 
(Applies to other unit tests as well.)

```suggestion
    Consumer<String> consumer = s -> assertThat(s.length()).isEqualTo(4);
```

Sunt-ing: Great advice! I will refactor my test, thanks！

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1108,6 +1109,62 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
...
+        if (satisfiedElementsLists[i].remove(element)) {
+          removedElementRowIndices.add(i);
+        }
+      }
+
+      // Check whether remaining consumers can be satisfied
+      if (isSatisfied(satisfiedElementsLists, begin + 1)) return true;
+      
+      for (int i : removedElementRowIndices)
+        satisfiedElementsLists[i].add(element); // restore the array
```
mgrafl: Use of brackets vs. shorthand notion for single-statement if/for blocks: 
I noticed that the assertj-core code base is not consistent on this, but we should be consistent at least within the PR. 
My personal preference is to use brackets, but I recognize the neatness of a single-line ```if (condition) return true;```. 

How should we handle single-statement if/for blocks?

Sunt-ing: Indeed. I also noticed it. I have no preference for this. I think you can format the code as you like.

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1108,6 +1110,58 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
...
+        if (satisfiedElementsLists[i].remove(element)) {
+          removedElementRowIndices.add(i);
+        }
+      }
+
+      // Check whether remaining consumers can be satisfied
+      if (isSatisfied(satisfiedElementsLists, begin + 1)) return true;
+
+      for (int i : removedElementRowIndices)
+        satisfiedElementsLists[i].add(element); // restore the array
```
mgrafl, line 1160: Most of the assertj-core code based suggests that bracketless shorthand notations are only used for trivial ```if (condition) return x;``` and ```if (condition) throw e;``` cases. 
Usually, all other ```if``` blocks and all ```for``` blocks use brackets.

```suggestion
      for (int i : removedElementRowIndices) {
        satisfiedElementsLists[i].add(element); // restore the array
      }
```

Sunt-ing:
Thanks for your great effort!

joel-costigliola:
Labeled as "not ready" as the assertion name is not clear enough and can lead to different behaviour interpretation.

joel-costigliola:
I think we close this as `satisfyExactly` and `satisfyExactlyInAnyOrder` have been added

PR #1893, opened 2020-05-30
Add isEmpty, isNotEmpty in AbstractInputStreamAssert 

### LINKED ISSUE #1284: AbstractInputStreamAssert should assert isEmpty, isNotEmpty
#### Example

```
       static Supplier<InputStream> getDocumentDataById( long id ) {...}

        assertThat(  DocumentRepository.getDocumentDataById( 0 ).get() )
            .hasSameContentAs( new ByteArrayInputStream( new byte[0] ) );
```
thinking it'd be nice to write

```
        assertThat(  DocumentRepository.getDocumentDataById( 0 ).get() ).isEmpty();
```
also looking at verifying that contents of some things aren't empty or maybe `contentIsSize( int size )`

also, I don't appear to get at `hasSameContentAs` when running in list context.

        assertThat( ids )
            .extracting( DocumentRepository::getDocumentDataById )
            .extracting( Supplier::get )
            .doesNotContain( bais )
            .isNotNull();

also maybe a 

```
.hasSize( bytes )
```

TendoPW:
#### Check List:
* Fixes #1284 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractInputStreamAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractInputStreamAssert.java
@@ -81,6 +81,46 @@ public abstract class AbstractInputStreamAssert<SELF extends AbstractInputStream
   }
 
   /**
+   * Verifies whether the content of the actual {@code InputStream} is empty.
+   * <p>
+   * Example:
+   * <pre><code class='java'> // assertion will pass
+   * assertThat(new ByteArrayInputStream(new byte[] {})).isEmpty());
+   *
+   * // assertions will fail
+   * assertThat(new ByteArrayInputStream(new byte[] {0xa})).isEmpty();
+   *
+   * @return {@code this} assertion object.
+   * @throws NullPointerException if the given {@code InputStream} is {@code null}.
+   * @throws AssertionError if the content of the actual {@code InputStream} is not empty.
+   * @throws InputStreamsException if an I/O error occurs.
+   */
+  public SELF isEmpty(){
+    inputStreams.assertIsEmpty(info, actual);
+    return myself;
+  }
+
+  /**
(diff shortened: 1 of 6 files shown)


TendoPW:
I have a question is how should I judge that a `InputStream` is empty? Here I use  `read()` return value is -1 and seems it works.

TendoPW:
In tests of `assertIsEmpty` and `assertIsNotEmpty`, SpotBugs think I should not write to the static field `actual` like 
```java
actual = mock(InputStream.class); // actual is static field
```
However, the same code is also used in InputStreams_assertIsEmpty_Test.java

One solution is declaring a new field `private actual;` in each file

joel-costigliola:
Thanks, we will review your PR for 3.17.0 !

joel-costigliola:
I had a quick look at the PR @TendoPW, could you follow the https://github.com/joel-costigliola/assertj-core/blob/master/CONTRIBUTING.md especially the use of GIVEN WHEN THEN, AssertionUtil.expectAssertionError and @DisplayName.

The rest of the review will come later but if you address that first that would be great, thanks!

TendoPW:
> I had a quick look at the PR @TendoPW, could you follow the https://github.com/joel-costigliola/assertj-core/blob/master/CONTRIBUTING.md especially the use of GIVEN WHEN THEN, AssertionUtil.expectAssertionError and @DisplayName.
> 
> The rest of the review will come later but if you address that first that would be great, thanks!

Thanks for your review! I will work on this.

TendoPW:
In the latest 2 commits, I have refactored the test code to meet the requirement in the contributing document.

joel-costigliola:
Integrated thanks for the good work @TendoPW!

PR #1895, opened 2020-06-01
Avoid infinite recursion in StandardRepresentation

### LINKED ISSUE #1894: Avoid infinite recursion in StandardRepresentation
#### Summary

There's a risk of stack overflow when when representing containers (array, iterable, map) having  elements referring back to itself.

#### Example

The following test leads to a stack overflow.

```java
  @Test
  public void should_format_iterable_having_itself_as_element() {
    List<Object> list1 = list("Hello", "World");
    List<Object> list2 = list(list1);
    list1.set(1, list2);
    assertThat(STANDARD_REPRESENTATION.toStringOf(list1)).isEqualTo("[[\"Hello\", (this iterable)]]");
  }
```

Sunt-ing:
#### Check List:
* Fixes #1894 
* Unit tests : YES 

Only useful for `array` case. For `iterable` case, it doesn't work now because neither `HashSet` (hashcode issue leads to a stack overflow) nor `TreeSet` (lead to ClassCastException) can work. I think maybe we can convert an iterable to an array at the beginning. @joel-costigliola

### DIFF
--- a/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
+++ b/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
@@ -41,12 +41,11 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
-import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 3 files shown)


filiphr:
@Sunt-ing you might want to look at `IdentityHashMap` both for the arrays and iterable case. An object is visited if it's instance is visited. Otherwise you might come into problems when objects which are equal, but not identical are not properly shown.

Sunt-ing:
> @Sunt-ing you might want to look at `IdentityHashMap` both for the arrays and iterable case. An object is visited if it's instance is visited. Otherwise you might come into problems when objects which are equal, but not identical are not properly shown.

Sounds great, Thanks!

joel-costigliola:
@Sunt-ing I'm working on this one and yes I'm using `IdentitHashMap`

Sunt-ing:
> @Sunt-ing I'm working on this one and yes I'm using `IdentitHashMap`

OK, it doesn't matter.

PR #1916, opened 2020-06-16
Assertions for two-dimensional arrays of boolean

mwajcht:
#### Check List:
* Fixes #1182 
* Unit tests : YES


joel-costigliola:
Integrated thanks @mwajcht!

PR #1928, opened 2020-06-21
Renamed hasSize(int, int) to hasDimensions(int, int) for two-dimensio…

mwajcht:
…nal array asserts

#### Check List:
* Fixes #1182 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/Array2DAssert.java
+++ b/src/main/java/org/assertj/core/api/Array2DAssert.java
@@ -79,20 +79,20 @@ public interface Array2DAssert<SELF extends Array2DAssert<SELF, ELEMENT>, ELEMEN
    * <p>
    * Example:
    * <pre><code class='java'> // assertion will pass
-   * assertThat(new int[][] { {1, 2, 3}, {4, 5, 6} }).hasSize(2, 3);
+   * assertThat(new int[][] { {1, 2, 3}, {4, 5, 6} }).hasDimensions(2, 3);
    *
    * // assertions will fail
    * assertThat(new int[][] { }).hasSize(1, 1);
-   * assertThat(new int[][] { {1, 2, 3}, {4, 5, 6} }).hasSize(3, 2);
-   * assertThat(new int[][] { {1, 2, 3}, {4, 5, 6, 7} }).hasSize(2, 3);
+   * assertThat(new int[][] { {1, 2, 3}, {4, 5, 6} }).hasDimensions(3, 2);
+   * assertThat(new int[][] { {1, 2, 3}, {4, 5, 6, 7} }).hasDimensions(2, 3);
    * </code></pre>
    *
    * @param expectedFirstDimensionSize the expected number of values in first dimension of the actual array.
    * @param expectedSecondDimensionSize the expected number of values in second dimension of the actual array.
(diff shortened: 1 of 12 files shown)


src/main/java/org/assertj/core/api/Array2DAssert.java
```diff
@@ -79,20 +79,20 @@
    * <p>
```
joel-costigliola, line 79: `Verifies that the number of values in the actual array is equal to the given one.`
should be reworded to:
`Verifies that the actual 2D array has the same dimensions as the given one.`

src/main/java/org/assertj/core/api/Array2DAssert.java
```diff
@@ -79,20 +79,20 @@
...
    * </code></pre>
    *
    * @param expectedFirstDimensionSize the expected number of values in first dimension of the actual array.
    * @param expectedSecondDimensionSize the expected number of values in second dimension of the actual array.
    * @return {@code this} assertion object.
-   * @throws AssertionError if the number of values of the actual array is not equal to the given one.
+   * @throws AssertionError if the actual array's dimensions are not equal to the given ones.
    */
-  SELF hasSize(int expectedFirstDimensionSize, int expectedSecondDimensionSize);
+  SELF hasDimensions(int expectedFirstDimensionSize, int expectedSecondDimensionSize);
```
joel-costigliola, line 95: rename:
* `expectedFirstDimensionSize` -> `expectedFirstDimension` 
* `expectedSecondDimensionSize` -> `expectedSecondDimension`

src/test/java/org/assertj/core/internal/char2darrays/Char2DArrays_assertHasSize_Test.java
```diff
@@ -24,7 +24,7 @@
 import org.junit.jupiter.api.Test;
 
 /**
- * Tests for <code>{@link Char2DArrays#assertHasSize(AssertionInfo, char[][], int, int)}</code>.
+ * Tests for <code>{@link Char2DArrays#assertHasDimensions(AssertionInfo, char[][], int, int)}</code>.
  *
```
joel-costigliola, line 28: the class name should be renamed to `Char2DArrays_assertHasDimensions_Test`

joel-costigliola: the same applies for other tests

src/main/java/org/assertj/core/api/Array2DAssert.java
mwajcht: done

src/main/java/org/assertj/core/api/Array2DAssert.java
mwajcht: done

src/test/java/org/assertj/core/internal/char2darrays/Char2DArrays_assertHasSize_Test.java
mwajcht: done

joel-costigliola:
Integrated thanks @mwajcht!

PR #1930, opened 2020-06-24
Add succeedsWithin to FutureAssert

### LINKED ISSUE #1929: Move succeedsWithin to FutureAssert
`succeedsWithin` is currently only defined on `CompletableFutureAssert` but is compatible with a regular `Future` also.

This means if I have a method that returns a `Future` (but I know it is a `CompletableFuture`), I have to cast it in order to use this assertion.

Placing this assertion on `FutureAssert` instead would let it work in either case.

rupert-madden-abbott:
#### Check List:
* Fixes #1929
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractCompletableFutureAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractCompletableFutureAssert.java
@@ -12,6 +12,16 @@
  */
 package org.assertj.core.api;
 
+import org.assertj.core.internal.Failures;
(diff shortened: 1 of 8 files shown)


joel-costigliola:
Integrated thanks @rupert-madden-abbott!
FYI I had to refactor a bit the code as soft assertions were failing (not your fault, it's just the soft assertions mechanism that is a bit tricky to deal with), I added a bunch of tests for soft assertions to cover it properly

PR #1936, opened 2020-07-05
fix references to main branch

stevencrockett:
#### Check List:
* Fixes N/A
* Unit tests : NA

Seems there was a branch renaming, but some things didn't get renamed, causing broken links and being generally misleading. Not sure if this needs a related issue; nothing in the contrib guide so I just made the PR.


joel-costigliola:
Awesome thanks @stevencrockett!

PR #1937, opened 2020-07-05
simplify Percentage impl + tests as it no longer takes a Number

stevencrockett:
#### Check List:
* Fixes N/A
* Unit tests : YES

It seems that at some point `withPercentage` accepted a `Number`, which led to null checks and test with different subtypes. These days it only takes a `double`, so some things can be simplified.

Again, not sure if an issue was needed here.


src/test/java/org/assertj/core/data/Percentage_withPercentage_Test.java
```diff
@@ -14,27 +14,17 @@
...
- * Tests for {@link Percentage#withPercentage(Number)}.
+ * Tests for {@link Percentage#withPercentage(double)}.
  *
  * @author Alexander Bischof
  */
-public class Percentage_withPercentage_with_Double_Test {
-
-  @SuppressWarnings("null")
-  @Test
-  public void should_throw_error_if_value_is_null() {
```
stevencrockett, line 32: This test had actually still been working, but for the wrong reasons. I think the runtime tries to get a Double from the null, which it can then unbox to the double that the method actual requires.

So the runtime was throwing the NPE, instead of the `requireNonNull` check in the code (which is now deleted).

scordio:
Thank you @stevencrockett, I applied similar changes in ab8de52734dacaeb972fd0ff454c9bd9ccd22f99.

PR #1938, opened 2020-07-05
Compare sequences when "actual" is infinite or singly-traversable

rpolton:
#### Check List:
* Fixes #???
* Unit tests : YES / NO / NA

### DIFF
--- a/src/main/java/org/assertj/core/internal/Iterables.java
+++ b/src/main/java/org/assertj/core/internal/Iterables.java
@@ -458,31 +458,30 @@ public class Iterables {
     // break if actual were only singly-traversable.
     checkNotNullIterables(info, actual, sequence);
     // store the elements from actual that have been visited (because we don't know we can look ahead - the 'actual'
-    // might be singly-traversable) in a cyclic list, which can then be rolled for each new element
-    // until a match is found or until the 'actual' is exhausted. Of course if 'actual' really is infinite then this could
-    // take a while :-D
-    // once it's full, check if it equals sequence for each new element
-    RingBuffer buffer = new RingBuffer(sequence.length);
-    int howMany=0;
-    final Iterator<?> actual_iterator = actual.iterator();
+    // might be singly-traversable) in a fixed-length buffer having what is in effect a sliding window.
+    // So we store each element and slide for each new element until a match is found or until the 'actual' is
+    // exhausted. Of course if 'actual' really is infinite then this could take a while :-D
+    // Note that we only ever store sequence.length entries from actual in the buffer. Once it's full, check if it
+    // equals sequence for each new element
+    final Iterator<?> actualIterator = actual.iterator();
     // ... perform the remaining "normal" checks explicitly
-    if(!actual_iterator.hasNext() && sequence.length==0) return;
+    if(!actualIterator.hasNext() && sequence.length==0) return;
     failIfEmptySinceActualIsNotEmpty(sequence);
-    while(actual_iterator.hasNext()) {
-      final Object next = actual_iterator.next();
+    Lifo buffer = new Lifo(sequence.length);
+    while(actualIterator.hasNext()) {
+      final Object next = actualIterator.next();
       buffer.add(next);
-      howMany++;
-      if(howMany>=sequence.length && buffer.equals(sequence)) return;
+      if(buffer.equals(sequence)) return;
     }
     throw actualDoesNotContainSequence(info, actual, sequence);
   }
 
-  class RingBuffer {
+  private class Lifo {
     private int writePosition;
     private int length;
     private Object[] buffer;
 
-    public RingBuffer(int length) {
+    public Lifo(int length) {
       this.length = length;
       this.buffer = new Object[length];
       this.writePosition = 0;
@@ -495,7 +494,7 @@ public class Iterables {
           buffer[i] = buffer[i + 1];
         }
       }
-      buffer[writePosition++]=t;
+      buffer[writePosition++] = t;
       return true;
     }
 
--- a/src/main/java/org/assertj/core/internal/Iterables.java
+++ b/src/main/java/org/assertj/core/internal/Iterables.java
@@ -477,7 +477,7 @@ public class Iterables {
     throw actualDoesNotContainSequence(info, actual, sequence);
   }
 
-  class RingBuffer<T> {
+  class RingBuffer {
     private int writePosition;
     private int length;
     private Object[] buffer;
@@ -485,10 +485,10 @@ public class Iterables {
     public RingBuffer(int length) {
       this.length = length;
       this.buffer = new Object[length];
-      this.writePosition =0;
+      this.writePosition = 0;
     }
 
-    public boolean add(final T t) {
+    public boolean add(final Object t) {
       if(writePosition==length) { // ie we've already filled the buffer
         writePosition--;
         for (int i = 0; i < length - 1; ++i) {
@@ -1129,7 +1129,8 @@ public class Iterables {
     if (!diff.differencesFound()) {
       // actual and values have the same elements but are they in the same order ?
       int i = 0;
-      for (Object elementFromActual : actualAsList) {
+      // use actualAsList instead of actual in case actual is a singly-passable iterable
+        for (Object elementFromActual : actualAsList) {
         if (!areEqual(elementFromActual, values[i])) {
           throw failures.failure(info, elementsDifferAtIndex(elementFromActual, values[i], i, comparisonStrategy));
         }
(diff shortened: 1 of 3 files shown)


rpolton:
NB somehow I have failed to merge this current change into a branch that already contains my previous change, for singly-traversable sequences. If there is a different branch to which I should be merging please let me know. Thanks

joel-costigliola:
@rpolton it should be `main`, rebase your branch to `origin/main` and then you should be fine.

rpolton:
> @rpolton it should be `main`, rebase your branch to `origin/main` and then you should be fine.

Okay, thanks. It's ready to go now :-)

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -453,16 +453,62 @@ public void assertContainsOnlyNulls(AssertionInfo info, Iterable<?> actual) {
...
+      howMany++;
+      if(howMany>=sequence.length && buffer.equals(sequence)) return;
     }
     throw actualDoesNotContainSequence(info, actual, sequence);
   }
 
+  class RingBuffer {
+    private int writePosition;
+    private int length;
+    private Object[] buffer;
```
joel-costigliola, line 483: have you considered using a linked list? it exposes `addLast` and `removeFirst` methods which can simplify the `add` method and get rid of `writePosition` (I haven't tried it so I might be wrong but worth considering)

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -453,16 +453,62 @@ public void assertContainsOnlyNulls(AssertionInfo info, Iterable<?> actual) {
...
+    // break if actual were only singly-traversable.
+    checkNotNullIterables(info, actual, sequence);
+    // store the elements from actual that have been visited (because we don't know we can look ahead - the 'actual'
+    // might be singly-traversable) in a cyclic list, which can then be rolled for each new element
+    // until a match is found or until the 'actual' is exhausted. Of course if 'actual' really is infinite then this could
+    // take a while :-D
+    // once it's full, check if it equals sequence for each new element
+    RingBuffer buffer = new RingBuffer(sequence.length);
+    int howMany=0;
+    final Iterator<?> actual_iterator = actual.iterator();
```
joel-costigliola, line 467: minor: `actual_iterator` -> `actualIterator`

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -453,16 +453,62 @@ public void assertContainsOnlyNulls(AssertionInfo info, Iterable<?> actual) {
...
+    // perform the checks that would have been done in commonCheckThatIterableAssertionSucceeds
+    // but do them explicitly without having to create a new iterator on actual - which would
+    // break if actual were only singly-traversable.
+    checkNotNullIterables(info, actual, sequence);
+    // store the elements from actual that have been visited (because we don't know we can look ahead - the 'actual'
+    // might be singly-traversable) in a cyclic list, which can then be rolled for each new element
+    // until a match is found or until the 'actual' is exhausted. Of course if 'actual' really is infinite then this could
+    // take a while :-D
+    // once it's full, check if it equals sequence for each new element
+    RingBuffer buffer = new RingBuffer(sequence.length);
```
joel-costigliola, line 465: Move the RingBuffer creation after the call to `failIfEmptySinceActualIsNotEmpty` as we don't need until the while loop.

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -453,16 +453,62 @@ public void assertContainsOnlyNulls(AssertionInfo info, Iterable<?> actual) {
...
+    int howMany=0;
+    final Iterator<?> actual_iterator = actual.iterator();
+    // ... perform the remaining "normal" checks explicitly
+    if(!actual_iterator.hasNext() && sequence.length==0) return;
+    failIfEmptySinceActualIsNotEmpty(sequence);
+    while(actual_iterator.hasNext()) {
+      final Object next = actual_iterator.next();
+      buffer.add(next);
+      howMany++;
```
joel-costigliola, line 474: I don't think we need howMany if RingBuffer gives how many elements it contains.
the if condition could be: 
```
if(buffer.size()>=sequence.length && buffer.equals(sequence))
```
We can even go further and let simply use `if(buffer.equals(sequence))` as buffer.equals checks the sizes of the sequence against the buffer's number of elements. I'm not against keeping `buffer.size()>=sequence.length` in the if condition as it makes it clear that there is no point of doing a comparison if buffer does not have enough elements but that's a detail.

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -453,16 +453,62 @@ public void assertContainsOnlyNulls(AssertionInfo info, Iterable<?> actual) {
...
-    List<?> actualAsList = newArrayList(actual);
-    for (int i = 0; i < actualAsList.size(); i++) {
-      // look for given sequence in actual starting from current index (i)
-      if (containsSequenceAtGivenIndex(actualAsList, sequence, i)) return;
+    // perform the checks that would have been done in commonCheckThatIterableAssertionSucceeds
+    // but do them explicitly without having to create a new iterator on actual - which would
+    // break if actual were only singly-traversable.
+    checkNotNullIterables(info, actual, sequence);
+    // store the elements from actual that have been visited (because we don't know we can look ahead - the 'actual'
+    // might be singly-traversable) in a cyclic list, which can then be rolled for each new element
```
joel-costigliola, line 461: that comment confused me a little as it made me thinks we were storing all actual elements, I think an important point here is to mention that we won't take no more that sequence.length elements, but hey good on you for all the comments in this method!

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -453,16 +453,62 @@ public void assertContainsOnlyNulls(AssertionInfo info, Iterable<?> actual) {
...
+    while(actual_iterator.hasNext()) {
+      final Object next = actual_iterator.next();
+      buffer.add(next);
+      howMany++;
+      if(howMany>=sequence.length && buffer.equals(sequence)) return;
     }
     throw actualDoesNotContainSequence(info, actual, sequence);
   }
 
+  class RingBuffer {
```
joel-costigliola, line 480: I think this should be renamed `Lifo` as when we add an element we remove the first one.

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -453,16 +453,62 @@ public void assertContainsOnlyNulls(AssertionInfo info, Iterable<?> actual) {
...
-      // look for given sequence in actual starting from current index (i)
-      if (containsSequenceAtGivenIndex(actualAsList, sequence, i)) return;
+    // perform the checks that would have been done in commonCheckThatIterableAssertionSucceeds
+    // but do them explicitly without having to create a new iterator on actual - which would
+    // break if actual were only singly-traversable.
+    checkNotNullIterables(info, actual, sequence);
+    // store the elements from actual that have been visited (because we don't know we can look ahead - the 'actual'
+    // might be singly-traversable) in a cyclic list, which can then be rolled for each new element
+    // until a match is found or until the 'actual' is exhausted. Of course if 'actual' really is infinite then this could
+    // take a while :-D
```
joel-costigliola, line 463: indeed, made me think of Woody Allen quote: _Eternity is a very long time, especially towards the end_ :smile:

src/main/java/org/assertj/core/internal/Iterables.java
rpolton: Hey, interesting idea. I confess that I haven't considered that in detail. I do quite like the RingBuffer because it does exactly what I want to show, nothing more and nothing less. That would certainly not be true of the linked list, which would also need (possibly) the tail pointing to the head. My feeling is that there would be work involved in making the list function as required which would detract from the issue at hand. What do you think?

src/main/java/org/assertj/core/internal/Iterables.java
rpolton: That's a good spot actually. I definitely think that buffer.equals(sequence) will obviously return false if they are of a different length so let's trim that howMany reference out ...

src/main/java/org/assertj/core/internal/Iterables.java
rpolton: I was going to grumble about this but now, on reflection, I realise this actually isn't a RingBuffer any more lol

rpolton:
Hi, it's ready for re-review. Thanks

joel-costigliola:
Integrated thanks @rpolton!

PR #1941, opened 2020-07-14
assertContains and assertContainsOnly fail when the actual iterable is not restartable

rpolton:
#### Check List:
* Fixes #??? absolutely, assertContains and assertContainsOnly fail if the iterable is not restartable
* Unit tests : YES 


joel-costigliola:
Integrated with a "bit" of delay, thanks @rpolton!

rpolton:
Hey, that's okay. I guess you do this in your spare time so it's good of
you to do this anyway :)

Thanks

PR #1955, opened 2020-08-01
Test improvement: removed resource optimism (test smell)

eas5:
#### Check List:
* Fixes NA
* Unit tests: YES
* Javadoc with a code example (on API only): NA

This is a test refactoring.

The resource optimism occurs when a test method makes an optimistic assumption that the external resource (e.g., file), utilized by the test method, exists.

As this test presumes the existence of a file with content to be tested for, It is necessary to ensure the file exists and has content before proceeding with the test steps.

### DIFF
--- a/src/test/java/org/assertj/core/api/Assertions_linesOf_Test.java
+++ b/src/test/java/org/assertj/core/api/Assertions_linesOf_Test.java
@@ -15,6 +15,7 @@ package org.assertj.core.api;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.linesOf;
 import static org.assertj.core.util.Lists.newArrayList;
+import static org.junit.Assume.*;
 
 import java.io.File;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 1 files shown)


src/test/java/org/assertj/core/api/Assertions_linesOf_Test.java
```diff
@@ -29,6 +30,7 @@
   @Test
   public void should_read_lines_of_file_with_UTF8_charset() {
     File file = new File("src/test/resources/utf8.txt");
+    assumeTrue("src/test/resources/utf8.txt file is blank. Skipping test", file.length() > 0);
```
joel-costigliola, line 33: I'm not sure to understand @eas5, we don't need to assume the file is blank, we know it is not and it it was the following assertion would fail anyway. 

Can you clarify?

Cheers

eas5: Sure!

The assumption is `file.length() > 0`, so we always expect the file to exist and have content. The error message _"src/test/resources/utf8.txt file is blank. Skipping test"_ will be passed to the AssumptionViolatedException in case of failure, which will halt and skip the test instead of failing it.

scordio: In my eyes, this assumption would hide a potential problem. The test should fail if the file does not exist, we should not skip it.

(BTW `org.assertj.core.api.Assumptions.assumeThat` should be favored)

joel-costigliola: I agree wih @scordio and thus closing this PR.

PR #1957, opened 2020-08-04
fix typo in AssertionErrorMessagesAggregrator

ahmadmoawad:
#### Check List:

* Fix typo in `AssertionErrorMessagesAggregrator` class and `aggregrateErrorMessages()` method
* Unit tests :  NO

### DIFF
--- a/src/main/java/org/assertj/core/error/AssertionErrorMessagesAggregator.java
+++ b/src/main/java/org/assertj/core/error/AssertionErrorMessagesAggregator.java
@@ -14,13 +14,9 @@ package org.assertj.core.error;
 
 import java.util.List;
 
-/**
- * @deprecated use {@link AssertionErrorMessagesAggregator} instead
- */
-@Deprecated
-public class AssertionErrorMessagesAggregrator {
+public class AssertionErrorMessagesAggregator {
 
-  public static String aggregrateErrorMessages(List<String> errors) {
+  public static String aggregateErrorMessages(List<String> errors) {
     StringBuilder msg = new StringBuilder("%nThe following ");
     countAssertions(errors, msg);
     msg.append(" failed:%n");
--- a/src/main/java/org/assertj/core/error/AssertionErrorMessagesAggregator.java
+++ b/src/main/java/org/assertj/core/error/AssertionErrorMessagesAggregator.java
@@ -14,9 +14,9 @@ package org.assertj.core.error;
 
 import java.util.List;
 
-public class AssertionErrorMessagesAggregrator {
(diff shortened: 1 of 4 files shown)


src/main/java/org/assertj/core/error/AssertionErrorMessagesAggregator.java
```diff
@@ -14,9 +14,9 @@
 
 import java.util.List;
 
-public class AssertionErrorMessagesAggregrator {
+public class AssertionErrorMessagesAggregator {
```
scordio, line 17: Not being in the internal package, this class might be used by other projects/libraries and the renaming would be a breaking change. We should probably deprecate this one and create a new class with the correct namings.

@joel-costigliola do you agree?

joel-costigliola: I agree!

ahmadmoawad:
@scordio ,

Should I mark `AssertionErrorMessagesAggregrator` class as deprecated and put a message referring to the new one `AssertionErrorMessagesAggregator`?

scordio:
@ahmadmoawad yes, thank you. Summarizing, I would suggest the following changes:

- Create a new `AssertionErrorMessagesAggregator` with `aggregateErrorMessages()`
- Let `SoftAssertionError` and `MultipleAssertionsError` use the new class
- Deprecate `AssertionErrorMessagesAggregrator` with javadoc referring to the new class

Please let us know if you have any questions.

ahmadmoawad:
Okay @scordio , I will start implementing them

ahmadmoawad:
@scordio , Could you please review the changes

scordio:
Merged, thanks @ahmadmoawad!

src/main/java/org/assertj/core/error/AssertionErrorMessagesAggregrator.java
```diff
@@ -14,6 +14,10 @@
 
 import java.util.List;
 
+/**
+ * @deprecated use {@link AssertionErrorMessagesAggregator} instead
+ */
+@Deprecated
 public class AssertionErrorMessagesAggregrator {
 
   public static String aggregrateErrorMessages(List<String> errors) {
```
pzygielo, line 23: Perhaps method from replacement class could be called here to stay DRY, and have no other code in this class?

scordio: Done in 45eb16da4a2290a0d3f9c77f7806372507705af6, thanks @pzygielo!

PR #1959, opened 2020-08-06
Remove commented variables, update equals and fix typo method's name 

ahmadmoawad:
1. Remove commented variables
2. Update `equals()` method to check if an object equals to itself 
3. Mark `fielLocation()` method as deprecated and create a new one with the correct name `fieldLocation()`
4. Add unit tests for `equals()`, `hashCode()` and `fieldLocation()`

#### Check List:
* Unit tests:  YES
* Javadoc with a code example (on API only): NO

### DIFF
--- a/src/main/java/org/assertj/core/api/recursive/comparison/FieldLocation.java
+++ b/src/main/java/org/assertj/core/api/recursive/comparison/FieldLocation.java
@@ -22,8 +22,6 @@ import java.util.stream.Stream;
 public class FieldLocation implements Comparable<FieldLocation> {
 
   private String fieldPath;
-  // private boolean whereverInGraph = false;
-  // private boolean matches(Field field, Field parent); ?
 
   @Override
   public int compareTo(final FieldLocation other) {
@@ -32,6 +30,7 @@ public class FieldLocation implements Comparable<FieldLocation> {
 
   @Override
   public boolean equals(final Object other) {
+    if (this == other) { return true; }
     if (!(other instanceof FieldLocation)) return false;
     FieldLocation castOther = (FieldLocation) other;
     return Objects.equals(fieldPath, castOther.fieldPath);
@@ -64,7 +63,15 @@ public class FieldLocation implements Comparable<FieldLocation> {
     return Stream.of(fieldPaths).map(FieldLocation::new).collect(toList());
   }
 
+  /**
+   * @deprecated use {@link #fieldLocation} instead
+   * */
+  @Deprecated
   public static FieldLocation fielLocation(String fieldPath) {
+    return fieldLocation(fieldPath);
+  }
+
+  public static FieldLocation fieldLocation(String fieldPath) {
     return new FieldLocation(fieldPath);
   }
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/assertj/core/api/recursive/comparison/FieldLocation.java
```diff
@@ -22,8 +22,6 @@
 public class FieldLocation implements Comparable<FieldLocation> {
```
scordio, line 22: @joel-costigliola does this class really need to be `public`? (same question for all the non overridden methods)

src/main/java/org/assertj/core/api/recursive/comparison/FieldLocation.java
```diff
@@ -64,7 +63,15 @@ public boolean matches(String concatenatedPath) {
     return Stream.of(fieldPaths).map(FieldLocation::new).collect(toList());
   }
 
+  /**
+   * @deprecated use {@link #fieldLocation} instead
+   * */
+  @Deprecated
   public static FieldLocation fielLocation(String fieldPath) {
```
scordio, line 70: If this class and especially this method can be `package-private`, then the renaming could be directly applied and the deprecation is not needed. Let's wait for the feedback from @joel-costigliola.

src/main/java/org/assertj/core/api/recursive/comparison/FieldLocation.java
```diff
@@ -32,6 +30,7 @@ public int compareTo(final FieldLocation other) {
 
   @Override
   public boolean equals(final Object other) {
```
scordio, line 32: I would introduce a unit test for this `equals`, like the one for `Tuple`. See:

https://github.com/joel-costigliola/assertj-core/blob/45eb16da4a2290a0d3f9c77f7806372507705af6/src/test/java/org/assertj/core/api/Tuple_Test.java#L85-L90

ahmadmoawad: Okay, Let me implement it

src/main/java/org/assertj/core/api/recursive/comparison/FieldLocation.java
joel-costigliola: The intention was to model field location for the recursive comparison, I realized later that it would a pain for users to build instances of this class instead of passing `String`.

`FieldLocation` is still used in [RecursiveComparisonConfiguration.registerComparatorForField](https://www.javadoc.io/static/org.assertj/assertj-core/3.16.1/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration.html#registerComparatorForField(java.util.Comparator,org.assertj.core.api.recursive.comparison.FieldLocation)) so we can't make package private yet (although I have deprecated the method it in 3.17.0).

src/main/java/org/assertj/core/api/recursive/comparison/FieldLocation.java
joel-costigliola: Since the class is exposed through `RecursiveComparisonConfiguration`, I'm going to assume it is used, we can make `matches` package private as it was never meant to be exposed to users but the constructor and factory method should still be public.

scordio:
Integrated, thanks @ahmadmoawad! I did some more refactoring following the comments from @joel-costigliola.

PR #1960, opened 2020-08-08
Update equals() and hashCode() for Percentage class

ahmadmoawad:
1. As the `equals()` uses a `double` value to check the equality, no need to use `deepEquals()`
2. As the `value` variable is a `double` we can use `Double.hashCode()` to get the hashCode for the double value and no Need to use `hashCodeFor()` [Oracle#Double##hashCode](https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#hashCode-double-)

#### Check List:
* Unit tests: NO
* Javadoc with a code example (on API only): NO


scordio:
Thanks @ahmadmoawad! I merged your changes with some refactoring in the tests.

PR #1963, opened 2020-08-11
Add assertThatNoException() method

### LINKED ISSUE #1962: Consider adding an assertThatNoException method
#### Summary
The `assertThatExceptionOfType` method and related helpers (`assertThatNullPointerException`, `assertThatIllegalArgumentException`, `assertThatIOException` & `assertThatIllegalStateException`) provide a nice way to check that a specific exception has been thrown.

If, however, you want to check that an exception was not thrown you need to use `assertThatCode` and follow it with `doesNotThrowAnyException`. It would be really nice to have an `assertThatNoException` method, so that exception tests can consistently lead with the assertion.

#### Example

Current:
```java
assertThatIllegalArgumentException().isThrownBy(() -> new Foo(null, "bar"));
assertCode(() -> new Foo("foo", null)).doesNotThrowAnyException();
```

Wanted:
```java
assertThatIllegalArgumentException().isThrownBy(() -> new Foo(null, "bar"));
assertThatNoException().isThrownBy(() -> new Foo("foo", null));
```

philwebb:
Add an `assertThatNoException()` method that can be used as an
alternative to `assertThatCode(...).doesNotThrowAnyException().

#### Check List:
* Fixes #1962
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -1368,6 +1368,20 @@ public class Assertions implements InstanceOfAssertFactories {
   }
 
   /**
+   * Entry point to check that no exception of any type is thrown by a given {@code throwingCallable}.
+   * <p>
+   * Example:
+   * <pre><code class='java'>assertThatNoException().isThrownBy(() -&gt; { System.out.println("OK"); });</code></pre>
+   *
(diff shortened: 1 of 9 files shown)


scordio:
Integrated with some cosmetic changes in the tests. Thanks @philwebb!

philwebb:
Nice one. Thanks for the fast turnaround @scordio!

PR #1968, opened 2020-08-14
Better handle soft proxies for custom assert classes in OSGi bundles

bjhargrave:
We use a composite class loader for defining soft proxies when the
assert class is from a different class loader than the AssertJ classes.

This can occur in OSGi when the assert class is from a bundle. The
composite class loader provides access to the internal, non-exported
types of AssertJ. ByteBuddy will define the proxy class in the
CompositeClassLoader rather than in the class loader of the assert
class. This means the assert class cannot assume package private access
to super types, interfaces, etc. since the proxy class is defined in a
different class loader (the CompositeClassLoader) than the assert class.

Fixes https://github.com/joel-costigliola/assertj-core/issues/1964


joel-costigliola:
Integrated thanks @bjhargrave!

PR #1973, opened 2020-08-23
Speed up reflective comparison

AndreyNudko:
We mostly rely on RecursiveComparisonDifferenceCalculator in our tests (which is awesome feature!)
However our naming convention is a little unusual - we prefer 'bare' methods instead of getters, i.e. 'someProperty()' instead of 'getSomeProperty()'
Turns this causes a large number of exceptions to be thrown during introspection, which add considerable runtime overhead and slow tests down

This PR speeds up reflective comparison in 3 ways:
1. Introspection::findMethod now caches result of Method lookup - both positive and negative; this eliminates NoSuchMethodException being thrown all the time for us
   This does increases memory consumption slightly, however I don't believe this is an issue in practice
   In my experiments a cache with 30000 entries - which is the estimation of a pathological domain with 500 classes, each having 20 attributes, and no getters at all - takes only about 4m; which is pretty small amount of memory nowadays
   In more typical cases (<1000 entries) it should take something like 128K or less
2. Introspection::getPropertyGetter uses conditions instead of exceptions
   Checking for null instead of catching NullPointerException is a lot cheaper
   So is checking if method is public instead of catching IllegalAccessException (although this may be less useful).
   It also eliminates some duplicate logic in propertyNotFoundErrorMessage which had to do more-or-less same analysis as the actual method just to choose correct error message
3. RecursiveComparisonDifferenceCalculator calculates field differences in a single pass, instead of checking for ignored fields first
   With caching in place introspection is a lot cheaper, and for non-ignored fields (which is probably most common case) getNonIgnoredActualFieldNames is not that efficient anyway
   It has to do essentially same work as comparison logic to build child path and DualValue - which is repeated immediately afterwards by comparison code; and this involves not just introspection, but also copying lists and concatenating strings, so is not totally free
   And then subsequent logic compared two sets first - which is not totally free either
   For our test suites single-pass actually performs marginally better - although the improvement is not anywhere close to what caching does

These changes improve run time for some of our test suites by about 50-60%

#### Check List:
* Fixes: NA
* Unit tests : NA


AndreyNudko:
Thanks a lot for looking into this. I've just found another case in RecursiveComparisonDifferenceCalculator when filtering ignored fields upfront adds overhead for non-ignored ones compared to actually building DualValue immediately after. I'm adding the change to the same PR - but the original logic is a little fiddly, this one definitely deserves a second pair of eyes. Totally understand if you would prefer to deal with that separately.

joel-costigliola:
@AndreyNudko I would prefer a separate PR indeed, as you said some part of the code are quite tricky

src/main/java/org/assertj/core/util/introspection/Introspection.java
```diff
@@ -102,19 +108,47 @@ private static boolean isValidGetter(Method method) {
   }
 
   private static Method findMethod(String name, Object target) {
-    Class<?> clazz = target.getClass();
+    final MethodKey methodKey = new MethodKey(name, target.getClass());
+    return METHOD_CACHE.computeIfAbsent(methodKey, Introspection::findMethodByKey).orElse(null);
```
joel-costigliola, line 112: I believe the code can be simplified if `findMethodByKey` returned `Method`, you wouldn't need `.orElse(null)`.
Optional are useful to force client to deal with the absence of value but since we are returning in this method `findMethodByKey` could also do that.

joel-costigliola (commented):
@AndreyNudko can you keep in this PR the `Introspection` changes which I only have a minor comment on and I can integrate it shortly.
The rest of the changes will take more time to review and I don't want to block the `Introspection` changes because of that, can you create another PR with these changes? thanks

src/main/java/org/assertj/core/util/introspection/Introspection.java
AndreyNudko: So the reason I went for Optional is to cache negative results. Otherwise in our case - bare getters - every subsequent lookup for 'getXYZ' executes the search logic all over again, with all exceptions being thrown etc
Since ConcurrentHashMap does not allow null values, I needed something to represent missing value - empty optional looked like the most obvious choice.
I certainly don't want to compromise thread-safety by using regular HashMap
It's possible to use some sentinel value (Object::toString?) - that would actually slightly reduce memory footprint, but not necessarily improve the code
It's possible to inspect the class as a whole and cache some richer object similar to BeanInfo per class (and that would potentially benefit field reflection as well) - that's a bigger change, but I'm prepared to give it a go
Happy to hear other ideas

joel-costigliola: that's alright, leave the optional, I did not remember concurrent hash map required non null values, add a comment for that for me when I and my gold fish memory read the code again in 3 months. :D

AndreyNudko:
Thanks for merging the Introspection change!

I think I'm going to close this PR unless you think RecursiveComparisonDifferenceCalculator changes are useful
Performance improvement in our tests is very small, and the more I study both comparison and reflection code the less convinced I am that there's real value in that. Caching more reflection data to help ClassUtils, FieldUtils, and Objects::getDeclaredFieldsIncludingInherited/getFieldsNames may be a bigger win - I'm going to experiment with those to see how that changes test run time

joel-costigliola:
> Thanks for merging the Introspection change!

We will release 3.17.1 over the week end, it will include this improvement, won't have to wait too much to get it ;-)

> I think I'm going to close this PR unless you think RecursiveComparisonDifferenceCalculator changes are useful
Performance improvement in our tests is very small, and the more I study both comparison and reflection code the less convinced I am that there's real value in that. Caching more reflection data to help ClassUtils, FieldUtils, and Objects::getDeclaredFieldsIncludingInherited/getFieldsNames may be a bigger win - I'm going to experiment with those to see how that changes test run time

Let's close this PR and if in your experiment you find significant performance improvements, simply raise another PR.

Issue #1970, opened 2020-08-20
Softly/SoftAssertionsExtension revamp


kriegfrj:
I have a number of suggestions regarding these two extensions that I wanted to float - they could be done as separate issues but they are tightly coupled so I'm going to include them in one issue for now and we can split them up later if you feel that this is appropriate.

## SoftlyExtension doesn't support custom SoftAssertionsProvider

I find the field-injection style of `SoftlyExtension` much preferable to parameter injections. Not only does it simplify the method signatures of the tests, but you can easily have common methods that can perform a common set of (soft) assertions on some objects without having to pass the `SoftAssertions` reference around. I find that in many of my `SoftAssertionExtension` implementations, I have a line at the top of them all that says "this.softly = softly", which is boilerplate code polluting every test. The alternative is to declare `SoftAssertions` as a parameter to all of the common methods too, which is even less attractive.

However, `SoftlyExtension` has the drawback that it is hard-coded to use `SoftAssertions`. This means you miss out on the ability to use 

## Combine SoftlyExtension and SoftAssertionsExtension

In osgi-test, we've settled upon a pattern of extensions that support both field and parameter injection. This allows them to make use of common code (eg, instantiating the object to be injected) and it makes it easier for the user as they don't have to make a special effort to remember which one to use.

I can't see any good reason not to merge these two, but perhaps someone else can. Failing that, I suggest that we merge them (which will go a significant part of the way to solving the first point). Given that `SoftAssertionsExtension` has been around longer, I'm going to suggest merging `SoftlyExtension` into it and then deprecating `SoftlyExtension`.

## Adding API for third-party extensions

If you're using multiple soft assertion-type libraries, it is nice if you can combine the two into a common "multiple failure". This was the thrust behind #1459.

@fduminy came up with a novel way to combine it with Mockito soft verification (see https://github.com/fduminy/assertj-mockito). The only problem with this implementation is that it relies on what is effectively private API of `SoftAssertionsExtension` to work, as it maintains its own copy of has to reach into the extension context and peek into `SoftAssertionExtension`'s private store to get the current soft assertion instance to re-use. It would be nice if this aspect of it was publicised so that `AssertJMockitoExtension` and other 3rd-party extensions can get access to the soft assertion object for the current context.

[...]

PR #1977, opened 2020-08-26
Consolidate error collection for soft assertions into one place

kriegfrj:
Refer to the comment in https://github.com/joel-costigliola/assertj-core/issues/1800#issuecomment-680697890

This is a quick spike a put together as a proof-of-concept to consolidate the error collection in one publicly-accessible place. This will make it much easier to make AssertJ interoperate with other soft assertion frameworks (including itself, when it comes to custom soft assertion providers). If it is ok, I will add test cases to directly test the new `AssertionErrorCollectorImpl` class.

One thing I'm not sure about here is the `wasSuccess()` functionality. The changes I've made pass all the tests so hopefully it should all be good.


joel-costigliola:
@kriegfrj this can be closed as #1970 was addressed, right?

PR #1978, opened 2020-08-26
Speed up reflective comparison

AndreyNudko:
Second iteration of https://github.com/joel-costigliola/assertj-core/pull/1973
Only introduces method cache w/o changing comparison algorithm


joel-costigliola:
Integrated thanks @AndreyNudko!

PR #1980, opened 2020-08-26
Handle soft proxies for custom assert classes in OSGi bundles (#1979)

bjhargrave:
When using a composite class loader for defining soft proxies when the
assert class is from a different class loader than the AssertJ classes,
we now use the composite class loader as the ClassLoadingStrategy for
ByteBuddy. This is necessary for Java 9+ to avoid the
ClassLoadingStrategy.UsingLookup default strategy used by AssertJ for
Java 9+. That strategy always defines the proxy classes in the class
loader defining the assert class ignoring the specified class loader.
This change is also beneficial for Java 8 as it avoids the need to
reflectively call the composite class loader to define the proxy
classes.

We refactor the CompositeClassLoader into the
ClassLoadingStrategyFactory class so that we can centralize the logic
for class loader and class loading strategy determination. Assumptions
is updated to use these changes which prepares it for a future where it
could support defining custom assumptions.

Finally, we update the class loader used for the ByteBuddy class caches
to use the class loader of the assert class rather than AssertJ's class
loader as the key. This is important in the OSGi case where multiple
bundles could define custom assertions with the same class name and we
must allow for unique proxy class generation for each bundle.

Fixes https://github.com/joel-costigliola/assertj-core/issues/1979
Fixes #1888

PR #1982, opened 2020-08-26
Add OSGi integration testing to build

bjhargrave:
We define a simple test class to verify basic operations and a
test case to verify class loading of soft proxies for custom assertions
from OSGi bundles. The latter tests has been a problem area for which
regular testing will help avoid regressions.

Fixes https://github.com/joel-costigliola/assertj-core/issues/1888


bjhargrave:
I moved the fix into #1980 as a separate commit.

Issue #1985, opened 2020-09-04
Assumptions are broken in JUnit4 when opentest4j is in the classpath with 3.17.1


JoergSiebahn:
#### Summary

#1983 causes assumptions to fail instead of ignoring the test in edge cases.

This only occurs with JUnit4 when opentest4j is in the classpath, e.g. as transitive dependency from other test utilities like [Json-Unit](https://github.com/lukas-krecan/JsonUnit).

JUnit4 does not recognize `org.opentest4j.TestAbortedException` as expected and fails.

#### Example

I created a [test case for Json-Unit](https://github.com/JoergSiebahn/JsonUnit/commit/f0d62310a1864095747d49d8b321d421c8ca7154).

```java
package net.javacrumbs.jsonunit.test.junit4;

import org.assertj.core.api.Assumptions;
import org.junit.Test;

public class AssumptionsWithAssertJTest {
    @Test
    public void shouldBeIgnoredButFails() {
        Assumptions.assumeThat("Test").isEqualTo("NotEqualToTest");
    }
}
```

See also https://github.com/lukas-krecan/JsonUnit/issues/276

scordio:
Thanks for raising this @JoergSiebahn.

@lukas-krecan is there a particular reason that JsonUnit is providing opentest4j as a transitive dependency instead of optional?

lukas-krecan:
JsonUnit supports multiple modes of operation, it can work as standalone assertion library, as Hamcrest macher or as an extension of AssertJ. In some scenarios it throws `AssertionFailedError` from opentest4j to report multiple failures (I believe that it's better displayed by IDEs if it's AssertionFailedError).

joel-costigliola:
The ideal fix would be for JUnit4 to support `org.opentest4j.TestAbortedException` but that's not going to happen since the 4.13 is the final JUnit 4 release (unless they changed their mind).

Possible way forward 

1. Revert the fix done in 3.17.1 but I'm not super keen to support a library that is being replaced by JUnit 5 TBH.

2. Make `opentest4j` optional in JsonUnit: JsonUnit would only raise `org.opentest4j.MultipleFailuresError` or `org.opentest4j.AssertionFailedError` when `opentest4j` is in the classpath and fall back to the previous mechanism if not. That's for @lukas-krecan to decide, not me ;-)

3. Migrate to JUnit 5: Might be a chunky piece of work for projects but that's the future.

scordio:
[...]

Issue #1961, opened 2020-08-10
Assumptions.assumeThat incompatible with TestNG 7.1.1+


C-Otto:
See https://github.com/cbeust/testng/issues/2352

scordio:
Thanks for reporting the issue @C-Otto.

We could increase the precedence of `org.testng.SkipException` over `org.junit.AssumptionViolatedException`, but I would first wait for a feedback on the TestNG issue you opened.

PR #1989, opened 2020-09-13
Add assumptions it tests

scordio:
This covers the cases mentioned in #1961 and #1985. Any other relevant use case to cover?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,6 +34,7 @@
   <properties>
     <javadocAdditionalOptions>-html5 --allow-script-in-comments --no-module-directories</javadocAdditionalOptions>
     <byte-buddy.version>1.10.14</byte-buddy.version>
+    <opentest4j.version>1.2.0</opentest4j.version>
     <bnd.version>5.1.2</bnd.version>
     <!-- Dependency versions overriding -->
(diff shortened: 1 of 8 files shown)


joel-costigliola:
I think that should be enough as JUnit 5 + opentest4j is done in the main project. 
Maybe add JUnit 4 + 5 for projects that migrate slowly to JUnit 5? not so sure it is 100% relevant

scordio:
> Maybe add JUnit 4 + 5 for projects that migrate slowly to JUnit 5? not so sure it is 100% relevant though

At least it would protect against unintended changes, forcing it to be on purpose. I will add this case.

scordio:
>  Maybe add JUnit 4 + 5 for projects that migrate slowly to JUnit 5? not so sure it is 100% relevant though

On second thought, JUnit 4 + 5 would lead to the same case as JUnit 4 + opentest4j.

Also, projects that are migrating slowly to JUnit 5 would probably use JUnit 5 together with the `junit-vintage-engine`, so they would probably avoid JUnit 4 in the classpath (and if they don't, they are falling back to the case above).

joel-costigliola:
Alright, let's integrate that PR then

PR #1999, opened 2020-09-21
SonarCloud fixes for SoftAssertionsExtension

kriegfrj:
None of these were bugs in practice, but it is easy to keep SonarCloud happy so let's do it. That way real bugs won't be masked by the false alarms.

In particular, the alleged bug in the sanity checking test - that the test class was private - was actually intentional so that JUnit 5 *wouldn't* run it directly - it was meant to be run via its subclasses. Made it abstract instead which achieves the same thing.


joel-costigliola:
Fair enough, thanks @kriegfrj.

PR #2003, opened 2020-09-23
Add assertion factory methods to AbstractAssert.

### LINKED ISSUE #2001: AbstractAssert to add assertion factory methods
#### Summary

Some of the utility methods in `AbstractAssert` that are commonly used by custom assertion implementations will always throw an exception. Specifically:

https://github.com/joel-costigliola/assertj-core/blob/3c6bb8a5d4f452c60cf3c8e336a03feba5e9b207/src/main/java/org/assertj/core/api/AbstractAssert.java#L143

https://github.com/joel-costigliola/assertj-core/blob/3c6bb8a5d4f452c60cf3c8e336a03feba5e9b207/src/main/java/org/assertj/core/api/AbstractAssert.java#L185

Unfortunately, methods like this are a pain for the compiler and for code coverage tools because they have no way of knowing that calls to these methods will never return normally. Hence you get code coverage maps that look like this:

https://codecov.io/gh/osgi/osgi-test/src/main/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/bundlecontext/AbstractBundleContextAssert.java

...with the invocations of the `failXXX()` methods all showing up as uncovered. This happens because the way Codecov works is to insert probes (little bits of test code) before and after each line of source code - because the probe after the call to `failXXX()` never gets called, the code coverage tool concludes (incorrectly) that the line before it wasn't executed.

In order to fix this, it would be better if instead of throwing the exception directly, the two methods had counterparts which returned an exception that the caller could throw directly. So instead of:

```java
if (!actual.getName().equals(name)) {
     failWithMessage("Expected character's name to be <%s> but was <%s>" name, actual.getName());
}
```

...you'd do something like this:

```java
if (!actual.getName().equals(name)) {
     throw failureWithMessage("Expected character's name to be <%s> but was <%s>" name, actual.getName());
}
```

This way, the compiler and code coverage tools would be in on the secret and can handle appropriately, and our code coverage checks for our custom assertions would be showing 100% coverage instead of 87%. :smile:

This is more than just a "code coverage" nicety - the compiler can also work better (eg, spot dead code) with this pattern rather than the pattern of calling a method that always throws.

kriegfrj:
Fixes #2001.

I couldn't use the same name (`failure()`) for both methods because the arguments are too similar for common use cases and lead to ambiguities that the compiler can't resolve. So I went with `failure()` and `failureWithActualExpected`. I'm open to alternatives.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -49,6 +49,7 @@ import org.assertj.core.presentation.PredicateDescription;
 import org.assertj.core.presentation.Representation;
 import org.assertj.core.util.CheckReturnValue;
 import org.assertj.core.util.VisibleForTesting;
+import org.opentest4j.AssertionFailedError;
 
 /**
  * Base class for all assertions.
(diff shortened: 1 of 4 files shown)


joel-costigliola:
Fair enough.

scordio:
Merged with c9b57150923677d6d57c4c64295a83c544c2088c.

PR #2009, opened 2020-10-02
FEATURE #1652 : assert return value of callable throwing exception

### LINKED ISSUE #1652: Assert return value of callable throwing exception
#### Summary

It would be nice if the `assertThatCode().doesNotThrowException()` API would be able execute functions that return a value and then return that value for further inspection.
This would make it easy to provide a nice description to an otherwise maybe nondescript exception that is thrown.

#### Example

```java
@Test 
public void intendedApi() {
  int x = assertThatCode(this::doStuff)
    .describedAs("doing things")
    .doesNotThrowAnyException();
  assertThat(x).isEqualTo(3);
}

@Test
public void currentSolution1() {
  int x;
  try {
    x = doStuff();
  } catch (Exception e) {
    fail("doing things");
    return; // never reached; needed because fail() is not recognized by the java compiler as method exit
  }
  assertThat(x).isEqualTo(3);
}

@Test
public void currentSolution2() {
  AtomicReference<Integer> x = new AtomicReference<>();
  assertThatCode(() -> x.set(doStuff()))
    .describedAs("doing things")
    .doesNotThrowAnyException();
  assertThat(x.get()).isEqualTo(3);
}

// utility for demonstration purposes
public int doStuff() throws Exception {
  if (new Random().nextBoolean()) {
    throw new Exception();
  }
  return 3;
}
```

hosuaby:
#### Check List:
* Fixes #1652
* Unit tests : YES

The goal of that feature is to allow the tests of code that returns a value using method **Assertions.assertThatReturningCode**. For example:

```java
assertThatReturningCode(() -> 42)
    .doesNotThrowAnyExceptionAndReturns()
    .isEqualTo(42);
```

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractTryAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractTryAssert.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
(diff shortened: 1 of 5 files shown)


scordio:
Thanks for raising the PR, @hosuaby, and sorry for not providing any feedback for a long time!

We decided to take a different direction, currently under investigation with #2519. For this reason, I'll close this PR.

PR #2019, opened 2020-10-15
Added "hasPackage" assertions for Class

### LINKED ISSUE #1947: Add an assertion for checking a class package
#### Summary
I'm working on a code generation module and I need to check that generated classes have the right package.
Currently, this is the best AssertJ lets me express:

```java
Class<?> aClass;
String packageName;

assertThat(aClass.getPackageName()).isEqualTo(packageName);
```
I would like a more expressive assertion such as:

#### Example

```java
assertThat(aClass).hasPackage(packageName);
```
There could be an overload for a `Package` argument as well.

polarene:
This PR adds the `hasPackage()` assertion for checking if a `Class` declares a given package.
For example given: 

    package one.two;
   
    class MyClass {}

this assertion succeeds:

    assertThat(MyClass.class).hasPackage("one.two");

There's also an overload with the `java.lang.Package` argument.

#### Check List:
* Fixes #1947
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractClassAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractClassAssert.java
@@ -647,4 +647,59 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
     classes.assertHasPublicMethods(info, actual, methodNames);
     return myself;
   }
+
+  /**
+   * Verifies that the actual {@code Class} has the given package name.
+   * <p>
+   * Example:
+   * <pre><code>
+   * package one.two;
+   *
+   * class MyClass {}
+   *
+   * // this assertions succeeds:
+   * assertThat(MyClass.class).hasPackage("one.two");
+   *
+   * // these assertions fail:
+   * assertThat(MyClass.class).hasPackage("one");
+   * assertThat(MyClass.class).hasPackage("");
+   * assertThat(MyClass.class).hasPackage("java.lang");</code></pre>
+   *
+   * @param packageName the package name which the class must declare
+   * @return {@code this} assertions object
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} doesn't declare the given package.
+   */
+  public SELF hasPackage(String packageName) {
+    classes.assertHasPackage(info, actual, packageName);
+    return myself;
+  }
+
+  /**
+   * Verifies that the actual {@code Class} has the given package.
+   * <p>
+   * Example:
+   * <pre><code>
+   * package one.two;
+   *
+   * class MyClass {}
+   **
+   * // these assertions succeed:
+   * assertThat(MyClass.class).hasPackage(Package.getPackage("one.two"));
+   * assertThat(MyClass.class).hasPackage(MyClass.class.getPackage());
+   *
+   * // these assertions fail:
+   * assertThat(MyClass.class).hasPackage(Package.getPackage("one"));
+   * assertThat(MyClass.class).hasPackage(Package.getPackage(""));
+   * assertThat(MyClass.class).hasPackage(Object.class.getPackage());</code></pre>
+   *
+   * @param aPackage the package which the class must declare
+   * @return {@code this} assertions object
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} doesn't declare the given package.
+   */
+  public SELF hasPackage(Package aPackage) {
+    classes.assertHasPackage(info, actual, aPackage);
+    return myself;
+  }
 }
--- a/src/main/java/org/assertj/core/api/AbstractClassAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractClassAssert.java
@@ -665,10 +665,10 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
    * assertThat(MyClass.class).hasPackage("");
    * assertThat(MyClass.class).hasPackage("java.lang");</code></pre>
    *
-   * @param packageName the package name which the class must declare
+   * @param packageName the package name the class should have
    * @return {@code this} assertions object
    * @throws AssertionError if {@code actual} is {@code null}.
-   * @throws AssertionError if the actual {@code Class} doesn't declare the given package.
+   * @throws AssertionError if the actual {@code Class} does not have the given package.
    *
    * @since 3.18.0
    */
@@ -695,10 +695,10 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
    * assertThat(MyClass.class).hasPackage(Package.getPackage(""));
    * assertThat(MyClass.class).hasPackage(Object.class.getPackage());</code></pre>
    *
-   * @param aPackage the package which the class must declare
+   * @param aPackage the package the class should have
    * @return {@code this} assertions object
    * @throws AssertionError if {@code actual} is {@code null}.
-   * @throws AssertionError if the actual {@code Class} doesn't declare the given package.
+   * @throws AssertionError if the actual {@code Class} does not have the given package.
    *
    * @since 3.18.0
    */
@@ -706,4 +706,5 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
     classes.assertHasPackage(info, actual, aPackage);
     return myself;
   }
+
 }
--- a/src/main/java/org/assertj/core/api/AbstractClassAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractClassAssert.java
@@ -652,8 +652,7 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
    * Verifies that the actual {@code Class} has the given package name.
    * <p>
    * Example:
-   * <pre><code>
-   * package one.two;
+   * <pre><code class='java'> package one.two;
    *
    * class MyClass {}
    *
@@ -682,8 +681,7 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
    *
    * <p>
    * Example:
-   * <pre><code>
-   * package one.two;
+   * <pre><code class='java'> package one.two;
    *
    * class MyClass {}
    **
--- a/src/main/java/org/assertj/core/api/AbstractClassAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractClassAssert.java
@@ -649,7 +649,8 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
   }
 
   /**
-   * Verifies that the actual {@code Class} has the given package name.
+   * Verifies that the actual {@code Class} has the given package name (as in {@link Class#getPackage()}).
+   *
    * <p>
    * Example:
    * <pre><code class='java'> package one.two;
--- a/src/main/java/org/assertj/core/api/AbstractClassAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractClassAssert.java
@@ -669,6 +669,8 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
    * @return {@code this} assertions object
    * @throws AssertionError if {@code actual} is {@code null}.
    * @throws AssertionError if the actual {@code Class} doesn't declare the given package.
+   *
+   * @since 3.18.0
    */
   public SELF hasPackage(String packageName) {
     classes.assertHasPackage(info, actual, packageName);
@@ -676,7 +678,8 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
   }
 
   /**
-   * Verifies that the actual {@code Class} has the given package.
+   * Verifies that the actual {@code Class} has the given package (as in {@link Class#getPackage()}).
+   *
    * <p>
    * Example:
    * <pre><code>
@@ -697,6 +700,8 @@ public abstract class AbstractClassAssert<SELF extends AbstractClassAssert<SELF>
    * @return {@code this} assertions object
    * @throws AssertionError if {@code actual} is {@code null}.
    * @throws AssertionError if the actual {@code Class} doesn't declare the given package.
+   *
+   * @since 3.18.0
    */
   public SELF hasPackage(Package aPackage) {
     classes.assertHasPackage(info, actual, aPackage);
--- a/src/main/java/org/assertj/core/error/ShouldHavePackage.java
+++ b/src/main/java/org/assertj/core/error/ShouldHavePackage.java
@@ -1,3 +1,15 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2020 the original author or authors.
+ */
 package org.assertj.core.error;
 
 import java.util.StringJoiner;
--- a/src/main/java/org/assertj/core/error/ShouldHavePackage.java
+++ b/src/main/java/org/assertj/core/error/ShouldHavePackage.java
@@ -4,6 +4,7 @@ import java.util.StringJoiner;
 
 /**
  * Creates an error message indicating that a {@link Class} should have a given package.
+ *
  * @author Matteo Mirk
  */
 public class ShouldHavePackage extends BasicErrorMessageFactory {
@@ -26,10 +27,7 @@ public class ShouldHavePackage extends BasicErrorMessageFactory {
    */
   public static ErrorMessageFactory shouldHavePackage(Class<?> actual, String packageName) {
     final Package actualPackage = actual.getPackage();
-    if (actualPackage == null) {
-      return new ShouldHavePackage(actual, packageName);
-    }
-    return new ShouldHavePackage(actual, packageName, actualPackage.getName());
+    return actualPackage == null ? new ShouldHavePackage(actual, packageName) : new ShouldHavePackage(actual, packageName, actualPackage.getName());
   }
 
   private ShouldHavePackage(Class<?> actual, String expectedPackage) {
--- a/src/main/java/org/assertj/core/error/ShouldHavePackage.java
+++ b/src/main/java/org/assertj/core/error/ShouldHavePackage.java
@@ -20,14 +20,28 @@ public class ShouldHavePackage extends BasicErrorMessageFactory {
                                                                         .toString();
 
   /**
-   * Creates a new <code>ShouldHavePackage</code>.
-   * @param actual the actual value in the failed assertion.
+   * Creates a new <code>ShouldHavePackage</code> with a {@link Package} instance.
+   *
+   * @param actual  the actual value in the failed assertion.
+   * @param aPackage the expected package
+   * @return the created {@code ErrorMessageFactory}.
+   */
+  public static ErrorMessageFactory shouldHavePackage(Class<?> actual, Package aPackage) {
+    return shouldHavePackage(actual, aPackage.getName());
+  }
+
+  /**
+   * Creates a new <code>ShouldHavePackage</code> with a package name.
+   *
+   * @param actual      the actual value in the failed assertion.
    * @param packageName the expected package name
    * @return the created {@code ErrorMessageFactory}.
    */
   public static ErrorMessageFactory shouldHavePackage(Class<?> actual, String packageName) {
     final Package actualPackage = actual.getPackage();
-    return actualPackage == null ? new ShouldHavePackage(actual, packageName) : new ShouldHavePackage(actual, packageName, actualPackage.getName());
+    return (actualPackage == null)
+        ? new ShouldHavePackage(actual, packageName)
+        : new ShouldHavePackage(actual, packageName, actualPackage.getName());
   }
 
   private ShouldHavePackage(Class<?> actual, String expectedPackage) {
@@ -37,4 +51,5 @@ public class ShouldHavePackage extends BasicErrorMessageFactory {
   private ShouldHavePackage(Class<?> actual, String expectedPackage, String actualPackage) {
     super(SHOULD_HAVE_PACKAGE + BUT_HAD, actual, expectedPackage, actualPackage);
   }
+
 }
--- a/src/main/java/org/assertj/core/error/ShouldHavePackage.java
+++ b/src/main/java/org/assertj/core/error/ShouldHavePackage.java
@@ -1,0 +1,42 @@
+package org.assertj.core.error;
+
+import java.util.StringJoiner;
+
+/**
+ * Creates an error message indicating that a {@link Class} should have a given package.
+ * @author Matteo Mirk
+ */
+public class ShouldHavePackage extends BasicErrorMessageFactory {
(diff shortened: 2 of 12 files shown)


src/main/java/org/assertj/core/internal/Classes.java
```diff
@@ -581,4 +583,36 @@ private static void assertNotNull(AssertionInfo info, Class<?> actual) {
...
+   * @param info contains information about the assertion.
+   * @param actual the "actual" {@code Class}.
+   * @param aPackage the package that must be declared in the class.
+   */
+  public void assertHasPackage(AssertionInfo info, Class<?> actual, Package aPackage) {
+    assertNotNull(info, actual);
+    requireNonNull(aPackage, shouldNotBeNull("package").create());
+    Package actualPackage = actual.getPackage();
+
+    if (!java.util.Objects.equals(actualPackage, aPackage)) {
```
scordio, line 614: As `aPackage` is non-null already at this point, this could be:

```suggestion
    if (!aPackage.equals(actualPackage)) {
```

src/main/java/org/assertj/core/internal/Classes.java
```diff
@@ -581,4 +583,36 @@ private static void assertNotNull(AssertionInfo info, Class<?> actual) {
...
+   * @param info contains information about the assertion.
+   * @param actual the "actual" {@code Class}.
+   * @param packageName the package that must be declared in the class.
+   */
+  public void assertHasPackage(AssertionInfo info, Class<?> actual, String packageName) {
+    assertNotNull(info, actual);
+    requireNonNull(packageName, shouldNotBeNull("package").create());
+    Package actualPackage = actual.getPackage();
+
+    if (actualPackage == null || !actualPackage.getName().equals(packageName)) {
```
scordio, line 598: The `actualPackage == null` check made me think that for completeness we should have another `hasNoPackage()` assertion, which would cover the cases mentioned in the [`getPackage()`](https://docs.oracle.com/javase/9/docs/api/java/lang/Class.html#getPackage--) javadoc: array types, primitive types and void.

However, I wouldn't include `hasNoPackage()` in this PR but as a separate change. I will create an issue to track it.

src/main/java/org/assertj/core/internal/Classes.java
```diff
@@ -581,4 +583,36 @@ private static void assertNotNull(AssertionInfo info, Class<?> actual) {
...
+    }
+  }
+
+  /**
+   * Verifies that the actual {@code Class} has the given {@code Package}.
+   * @param info contains information about the assertion.
+   * @param actual the "actual" {@code Class}.
+   * @param aPackage the package that must be declared in the class.
+   */
+  public void assertHasPackage(AssertionInfo info, Class<?> actual, Package aPackage) {
```
scordio, line 609:
```suggestion
  public void assertHasPackage(AssertionInfo info, Class<?> actual, Package expected) {
```

src/test/java/org/assertj/core/api/classes/ClassAssert_hasPackage_Package_Test.java
```diff
@@ -0,0 +1,28 @@
...
+import org.assertj.core.api.ClassAssert;
+import org.assertj.core.api.ClassAssertBaseTest;
+import org.junit.jupiter.api.DisplayName;
+
+/**
+ * Tests for <code>{@link ClassAssert#hasPackage(Package)}</code>.
+ *
+ * @author Matteo Mirk
+ */
+@DisplayName("Classes hasPackage")
```
scordio, line 14:
```suggestion
@DisplayName("ClassAssert hasPackage(Package)")
```

src/test/java/org/assertj/core/api/classes/ClassAssert_hasPackage_Package_Test.java
```diff
@@ -0,0 +1,28 @@
...
+class ClassAssert_hasPackage_Package_Test extends ClassAssertBaseTest {
+
+  static final Package PACKAGE = Package.getPackage("org.assertj.core.api");
```
scordio, line 17: As `Package.getPackage()` is deprecated from java 9 onwards, I'd just use a mock here:
```suggestion
  static final Package PACKAGE = mock(Package.class);
```

src/test/java/org/assertj/core/api/classes/ClassAssert_hasPackage_String_Test.java
```diff
@@ -0,0 +1,28 @@
...
+import org.assertj.core.api.ClassAssert;
+import org.assertj.core.api.ClassAssertBaseTest;
+import org.junit.jupiter.api.DisplayName;
+
+/**
+ * Tests for <code>{@link ClassAssert#hasPackage(String)}</code>.
+ *
+ * @author Matteo Mirk
+ */
+@DisplayName("Classes hasPackage")
```
scordio, line 14:
```suggestion
@DisplayName("ClassAssert hasPackage(String)")
```

src/test/java/org/assertj/core/api/classes/ClassAssert_hasPackage_String_Test.java
```diff
@@ -0,0 +1,28 @@
...
+class ClassAssert_hasPackage_String_Test extends ClassAssertBaseTest {
```
scordio, line 15:
```suggestion
class ClassAssert_hasPackage_with_String_Test extends ClassAssertBaseTest {
```

src/test/java/org/assertj/core/api/classes/ClassAssert_hasPackage_Package_Test.java
```diff
@@ -0,0 +1,28 @@
...
+import org.assertj.core.api.ClassAssertBaseTest;
+import org.junit.jupiter.api.DisplayName;
+
+/**
+ * Tests for <code>{@link ClassAssert#hasPackage(Package)}</code>.
+ *
+ * @author Matteo Mirk
+ */
+@DisplayName("Classes hasPackage")
+class ClassAssert_hasPackage_Package_Test extends ClassAssertBaseTest {
```
scordio, line 15:
```suggestion
class ClassAssert_hasPackage_with_Package_Test extends ClassAssertBaseTest {
```

src/test/java/org/assertj/core/internal/classes/Classes_assertHasPackage_Test.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.stream.Stream;
+
+import org.assertj.core.internal.ClassesBaseTest;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+@DisplayName("Classes assertHasPackage")
+class Classes_assertHasPackage_Test extends ClassesBaseTest {
```
scordio, line 20: Could you split this into two separate test classes, one for each assertion method? (e.g., `Classes_assertHasPackage_with_Package_Test` and `Classes_assertHasPackage_with_String_Test`)

src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -647,4 +647,59 @@ public SELF hasPublicMethods(String... methodNames) {
     classes.assertHasPublicMethods(info, actual, methodNames);
     return myself;
   }
+
+  /**
+   * Verifies that the actual {@code Class} has the given package name.
```
scordio, line 652: As we are still java 8 based and `Class#getPackageName()` comes from java 9+, I'd explicitly mention the relationship with `Class#getPackage()` behavior:

```suggestion
   * Verifies that the actual {@code Class} has the given package name (as in {@link Class#getPackage()}).
   *
```

src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -647,4 +647,59 @@ public SELF hasPublicMethods(String... methodNames) {
...
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} doesn't declare the given package.
+   */
+  public SELF hasPackage(String packageName) {
+    classes.assertHasPackage(info, actual, packageName);
+    return myself;
+  }
+
+  /**
+   * Verifies that the actual {@code Class} has the given package.
```
scordio, line 679: As we are still java 8 based and `Class#getPackageName()` comes from java 9+, I'd explicitly mention the relationship with `Class#getPackage()` behavior:

```suggestion
   * Verifies that the actual {@code Class} has the given package (as in {@link Class#getPackage()}).
   *
```

src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -647,4 +647,59 @@ public SELF hasPublicMethods(String... methodNames) {
...
+   *
+   * // these assertions fail:
+   * assertThat(MyClass.class).hasPackage("one");
+   * assertThat(MyClass.class).hasPackage("");
+   * assertThat(MyClass.class).hasPackage("java.lang");</code></pre>
+   *
+   * @param packageName the package name which the class must declare
+   * @return {@code this} assertions object
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} doesn't declare the given package.
```
scordio, line 671:
```suggestion
   * @throws AssertionError if the actual {@code Class} doesn't declare the given package.
   *
   * @since 3.18.0
```

src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -647,4 +647,59 @@ public SELF hasPublicMethods(String... methodNames) {
...
+   *
+   * // these assertions fail:
+   * assertThat(MyClass.class).hasPackage(Package.getPackage("one"));
+   * assertThat(MyClass.class).hasPackage(Package.getPackage(""));
+   * assertThat(MyClass.class).hasPackage(Object.class.getPackage());</code></pre>
+   *
+   * @param aPackage the package which the class must declare
+   * @return {@code this} assertions object
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} doesn't declare the given package.
```
scordio, line 699:
```suggestion
   * @throws AssertionError if the actual {@code Class} doesn't declare the given package.
   *
   * @since 3.18.0
```

src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -647,4 +647,59 @@ public SELF hasPublicMethods(String... methodNames) {
...
+    classes.assertHasPackage(info, actual, packageName);
+    return myself;
+  }
+
+  /**
+   * Verifies that the actual {@code Class} has the given package.
+   * <p>
+   * Example:
+   * <pre><code>
+   * package one.two;
```
scordio, line 683: Join lines:
```suggestion
   * <pre><code class='java'> package one.two;
```

src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -647,4 +647,59 @@ public SELF hasPublicMethods(String... methodNames) {
     classes.assertHasPublicMethods(info, actual, methodNames);
     return myself;
   }
+
+  /**
+   * Verifies that the actual {@code Class} has the given package name.
+   * <p>
+   * Example:
+   * <pre><code>
+   * package one.two;
```
scordio, line 656: Join lines:

```suggestion
   * <pre><code class='java'> package one.two;
```

src/main/java/org/assertj/core/error/ShouldHavePackage.java
```diff
@@ -0,0 +1,42 @@
+package org.assertj.core.error;
+
+import java.util.StringJoiner;
+
+/**
+ * Creates an error message indicating that a {@link Class} should have a given package.
+ * @author Matteo Mirk
```
scordio, line 7:
```suggestion
 * Creates an error message indicating that a {@link Class} should have a given package.
 *
 * @author Matteo Mirk
```

src/main/java/org/assertj/core/internal/Classes.java
```diff
@@ -581,4 +583,36 @@ private static void assertNotNull(AssertionInfo info, Class<?> actual) {
   private static void classParameterIsNotNull(Class<?> clazz) {
     requireNonNull(clazz, "The class to compare actual with should not be null");
   }
+
+  /**
+   * Verifies that the actual {@code Class} has the given {@code packagename}.
+   * @param info contains information about the assertion.
```
scordio, line 589:
```suggestion
   * Verifies that the actual {@code Class} has the given {@code packageName}.
   *
   * @param info contains information about the assertion.
```

src/main/java/org/assertj/core/internal/Classes.java
```diff
@@ -581,4 +583,36 @@ private static void assertNotNull(AssertionInfo info, Class<?> actual) {
...
+   * @param actual the "actual" {@code Class}.
+   * @param packageName the package that must be declared in the class.
+   */
+  public void assertHasPackage(AssertionInfo info, Class<?> actual, String packageName) {
+    assertNotNull(info, actual);
+    requireNonNull(packageName, shouldNotBeNull("package").create());
```
scordio, line 595:
```suggestion
    requireNonNull(packageName, shouldNotBeNull("packageName").create());
```

src/main/java/org/assertj/core/internal/Classes.java
```diff
@@ -581,4 +583,36 @@ private static void assertNotNull(AssertionInfo info, Class<?> actual) {
...
+
+  /**
+   * Verifies that the actual {@code Class} has the given {@code Package}.
+   * @param info contains information about the assertion.
+   * @param actual the "actual" {@code Class}.
+   * @param aPackage the package that must be declared in the class.
+   */
+  public void assertHasPackage(AssertionInfo info, Class<?> actual, Package aPackage) {
+    assertNotNull(info, actual);
+    requireNonNull(aPackage, shouldNotBeNull("package").create());
```
scordio, line 611:
```suggestion
    requireNonNull(aPackage, shouldNotBeNull("aPackage").create());
```

scordio (changes requested):
Thanks @polarene! The PR is already in good shape, just a few changes and it would be ready for merging.

src/main/java/org/assertj/core/error/ShouldHavePackage.java
```diff
@@ -0,0 +1,42 @@
...
+   * @param actual the actual value in the failed assertion.
+   * @param packageName the expected package name
+   * @return the created {@code ErrorMessageFactory}.
+   */
+  public static ErrorMessageFactory shouldHavePackage(Class<?> actual, String packageName) {
+    final Package actualPackage = actual.getPackage();
+    if (actualPackage == null) {
+      return new ShouldHavePackage(actual, packageName);
+    }
+    return new ShouldHavePackage(actual, packageName, actualPackage.getName());
```
joel-costigliola, line 32: could use ternary operator:
```suggestion
    return actualPackage == null ? new ShouldHavePackage(actual, packageName) : new ShouldHavePackage(actual, packageName, actualPackage.getName());
```

src/main/java/org/assertj/core/internal/Classes.java
polarene: Oh right, didn't think about that! I tried to follow the Javadoc closely but didn't try to cover those cases. Thanks for raising the issue, maybe I'll work on it as soon as I can.

src/test/java/org/assertj/core/api/classes/ClassAssert_hasPackage_String_Test.java
polarene: I tried to follow the other tests convention and didn't see "with" used in the name, am I wrong?

src/main/java/org/assertj/core/api/AbstractClassAssert.java
polarene: Ok, but doesn't this mess up the indentation? In Intellij preview it seems so.

src/main/java/org/assertj/core/api/AbstractClassAssert.java
polarene: Fine, I thought that references to inner logic should be avoided in API docs. But if you think it help understand the assertion let's add it. Should it be "{@link Class#getPackage().getName()}" anyway?

src/test/java/org/assertj/core/internal/classes/Classes_assertHasPackage_Test.java
polarene: Sure I'll do it. About the naming convention, as I said in the other comment, I didn't find "with" to separate the arguments in other tests around, can you confirm this? Or if this is a new convention we're introducing I'll use it for sure.

src/test/java/org/assertj/core/api/classes/ClassAssert_hasPackage_Package_Test.java
polarene: I see, since we're just testing internal interactions, this way the code will be future-proof when it's migrated past Java 8?

src/main/java/org/assertj/core/internal/Classes.java
polarene: On second thought, now that I'm reviewing this modification in the IDE, I don't like it very much because it's not consistent with the other method in the class. For the String variant, it's named "packageName" which is explicative and makes sense, and the related error message is constructed like `shouldNotBeNull("packageName")`. But here changing it to "expected" loses some information and it's not coherent with the other methods in the class, where this semantic is not used. I would change it back to "aPackage", "pkg", or I'm accepting better name suggestions.

src/test/java/org/assertj/core/internal/classes/Classes_assertHasPackage_Test.java
polarene: Sorry, my bad I missed those tests following the convention. I'll rename my classes immediately, ignore the other comments!

polarene:
Could you review the changes? I think it should be ok now, thanks.

scordio:
Thank you @polarene and sorry for the slow feedback, I will have a look during this week.

polarene:
Hi @scordio no problems at all, take your time! Would you mind to tag the PR with `hacktoberfest-accepted` so it counts for this Hacktoberfest? Thank you

scordio:
`assertj-core` has now the `hacktoberfest` topic so this PR should be considered valid once it is merged.

polarene:
Thank you very much!

scordio (approved):

scordio:
Integrated, thanks @polarene!

polarene:
Thanks for merging @scordio! Only a small disappointment is that I would have liked to be notified about those last changes in the tests because I had a rationale behind, but it's fine anyway. Thanks for your support, looking forward to contributing again to Assertj!

scordio:
I noticed that one test was missing and I spotted some changes I'd have done around. As I promised to merge the PR by the end of the week I did them by myself, but nothing is written in stone: what would you do differently?

joel-costigliola:
@polarene we wanted to release 3.18.0 this last week end, even if the PR was merged, we are keen to hear what you have to say.

polarene:
Hi guys, sorry if my words didn't come out right. I really appreciate your support and hard work, in fact, congratulations on letting this PR in time for 3.18.0, great job!

All the latest changes from Stefano are ok indeed, I just wanted to point out that the modifications in the `Classes_assertHasPackage_*` tests:

- erased a specific test case I wrote, which checked the mismatch between packages that have a common prefix (like "com.foo" and "com.foo.bar")
- eliminated a parametric test where the rationale was to check against a Package instance obtained both from `Class#getPackage()` and `Package#getPackage()` -- since the latter is deprecated in Java 9+, when AssertJ migrates from 8 it should be checked against `ClassLoader.getDefinedPackage(java.lang.String)`
- replaced the Class instances used in the test cases with Object and Collection, but I think those I chose initially made the tests more entertaining (Jedi), following the spirit of this library, like the examples with LoTR characters, and more specific to the point (MyClass) as its package and package prefix were more evident for the cases
- eliminated some // GIVEN blocks which I think made the tests more readable

That's it, thanks for reading so far.

src/main/java/org/assertj/core/internal/Classes.java
polarene: > However, I wouldn't include `hasNoPackage()` in this PR but as a separate change. I will create an issue to track it.

Hey @scordio I don't know if you created this but I would like to contribute this assertion too if you don't mind, thanks.

scordio:
Hi @polarene,

> * erased a specific test case I wrote, which checked the mismatch between packages that have a common prefix (like "com.foo" and "com.foo.bar")

Here I assumed that classes have only one package which is uniquely identified, that's why I generalized it to `should_fail_if_packageName_does_not_match`. Do you see it differently?

> * eliminated a parametric test where the rationale was to check against a Package instance obtained both from Class#getPackage() and Package#getPackage() -- since the latter is deprecated in Java 9+, when AssertJ migrates from 8 it should be checked against ClassLoader.getDefinedPackage(java.lang.String)

Here I didn't want to test how the Package instance retrieval is done, which is JDK business, unless we have a reason for it. To the best of my knowledge, we should expect the same result between `Class#getPackage()`, `Package#getPackage(String)` and the Java 9+ `ClassLoader.getDefinedPackage(String)`.

Also, I just realized we could even write:

```java
  @ParameterizedTest
  @ValueSource(strings = "java.util")
  void should_fail_if_package_does_not_match(Package aPackage) {
    ...
  }
```

and JUnit 5 will resolve it with its implicit argument converter. The latter is actually my preference and I would update the test with it.

What do you think?

> * replaced the Class instances used in the test cases with Object and Collection, but I think those I chose initially made the tests more entertaining (Jedi), following the spirit of this library, like the examples with LoTR characters, and more specific to the point (MyClass) as its package and package prefix were more evident for the cases

I preferred to use well-known JDK types to avoid jumping back and forth in the classes and understand where the Jedi class is located. In general I like the entertaining test style, but I think it's better suited for the javadoc and for [`assertj-examples`](https://github.com/assertj/assertj-examples). If there is an option to avoid the usage of custom code in the tests, I vote for it.

> * eliminated some // GIVEN blocks which I think made the tests more readable

In general I tend to shorten tests where local variables are single values and can be inline, but I see your point. I'll put them back.

polarene:
Hi @scordio, thank you very much for answering my points, I see your reasons and generally agree on everything.

> Here I assumed that classes have only one package which is uniquely identified, that's why I generalized it to should_fail_if_packageName_does_not_match. Do you see it differently?

Basically you're right, packages are unique for a given class, but what I wanted to show with those tests was to make clear that `hasPackage()` would fail even for a "sub-package", that is one that has an existing package as a prefix. Although Java considers a sub-package to be different from its ancestor, I wanted to make it clear both for users and contributors of AssertJ that the assertion is following the language specification, in fact, this is reflected in one of the Javadoc examples. 
Tell me if it makes sense for you to include such a test or if the Javadoc suffices.

> Here I didn't want to test how the Package instance retrieval is done (which is JDK business) unless we have a reason for it. To the best of my knowledge, we should expect the same result between Class#getPackage(), Package#getPackage(String) and the Java 9+ ClassLoader.getDefinedPackage(String).

Ok maybe I confused test responsibilities, but I really wanted to assert that the instances retrieved in different ways were checked consistently, to cover corner cases, and to be future-proof whenever something changes in the next JDK versions, so those tests could signal it. I wrote it exactly to assert that we expect the same result from the various retrieval methods.

> JUnit 5 conversions
Nice, let's discuss that in another issue, it seems a feature worth investigating. Only thing is that I wanted to make the check explicit depending on how you retrieve the package, see above. But yes, let's discuss this in a separate thread.

> I preferred to use well-known JDK types [...]

Got it, I understand your reason, thanks for clarifying. Now I know for future PRs: funny stuff only in Javadoc. The thing was, after looking around the code base I found some custom classes uses for tests here and there, so I thought it was a legit use, also I didn't write Jedi, found it in ClassesBaseTest.

Thanks for restoring the GIVEN blocks, I appreciate it!

PR #2022, opened 2020-10-19
bnd: Clean up some pom dependencies now that Bnd 5.2.0 is being used

bjhargrave:
Now that Bnd 5.2.0 is released and being used by the project, we can
clean up some pom dependencies which were necessary to avoid a bad
interaction between Bnd and the enforcer plugin.


scordio:
Thanks @bjhargrave!

PR #2027, opened 2020-10-29
Get the current actual

### LINKED ISSUE #1931: Provide a way to the get the current actual
#### Summary

Some assertions have side effects or traverse the objects under test.

Sometimes, it is desirable to assert something, then perform additional manipulations on the object under test, before asserting more things.

In these cases, you have 2 choices:

1. Perform the traversal outside of assertj and then assert as you reach the relevant pieces.
2. Duplicate the traversal both with and without assertj

It would be useful instead if all assertions allowed for the actual object to be retrieved, removing this duplication.

Please let me know if there is already a way to do this (or a better approach) but I can't see one.

Below examples show this with succeedsWithin but the problem is not specific to succeedsWithin.

### Inconvenient Alternative 1

```java
  //no way to access what the future returned
  assertThat(future).succeedsWithin(10, TimeUnit.SECONDS);

  //Duplicate the traversal logic
  Object foo = future.get();

  //manipulate object outside of assertj
  //perform further assertions
```

### Inconvenient Alternative 2

```java
  //Get the underlying object first. Don't use assertj and lose the assertion failures if something goes wrong
  Object foo = future.get()  
  
  //manipulate object outside of assertj
  //perform further assertions
```

### Desired Outcome
```java
  Object foo = assertThat(future)
    .succeedsWithin(10, TimeUnit.SECONDS)
    .getActual();

  //manipulate object outside of assertj
  //perform further assertions
```

### Proposed Solution

Just add a getter to AbstractAssert:

```java
  public ACTUAL getActual() {
    return actual;
  }
```

Zomzog:
This is the solution proposed in the #1931 to be able to get the current value.

I know it will stop the assertion flow but sometimes it's easier for test readability.

#### Check List:
* Fixes #1931
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -1084,4 +1084,20 @@ public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, 
     return (ASSERT) assertFactory.createAssert(extractedValue).withAssertionState(myself);
   }
 
+  /**
+   * Return the actual object under test.
+   * <p>
+   * This allows users to retrieve the current object under test. This method will break assertion chain.
+   * <p>
+   * Example:
+   * <pre><code class='java'> TolkienCharacter smaug = new TolkienCharacter("Smaug", DRAGON);
+   * assertThat(smaug).getActual() // return smaug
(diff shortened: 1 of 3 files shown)


joel-costigliola:
We decided not to expose `getActual` thus I'm closing this PR, thanks anyway @Zomzog.

PR #2034, opened 2020-11-06
Add AbstractDateAssert.isBetween for Instant bounds

### LINKED ISSUE #2032: Add method isBetween(Instant, Instant) to AbstractDateAssert
#### Summary

I think it will be convenient to have in AbstractDateAssert<T> method `isBetween(start: Instant, end: Instant)` because `Data` and `Instant` are the same in meaning (unix time moment). But it's more comfortable to calculate times as `Instant` and don't use `Date.from()` to convert.

#### Example

```java
assertThat(validatingDate).isBetween(Instant.now().minusSeconds(2),  Instant.now().plusSeconds(2))
```

ars-java:
#### Check List:
* Fixes #2032
* Unit tests : YES

Because it's not possible to add assertionInvocationWithInstantArg to base class and don't override it in other derived classes (then I must add other methods for Instant) I had added it in the derived classes. At the next step, when there will be Instant methods for all derived classes it can be pulled up to the base class.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractDateAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractDateAssert.java
@@ -30,6 +30,7 @@ import static org.assertj.core.util.Lists.newArrayList;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.time.Instant;
 import java.util.Calendar;
 import java.util.Collection;
 import java.util.Comparator;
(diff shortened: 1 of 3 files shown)


src/main/java/org/assertj/core/api/AbstractDateAssert.java
```diff
@@ -1330,6 +1350,29 @@ public SELF isBetween(String start, String end, boolean inclusiveStart, boolean
     return myself;
   }
 
+  /**
+   * Same assertion as {@link #isBetween(Date, Date, boolean, boolean)} but given date is represented as
+   * {@code java.time.Instant}.
+   * <p>
+   * Example:
+   * <pre><code class='java'>
+   * assertThat(new Date()).isBetween(Instant.now().minusSeconds(5), Instant.now().plusSeconds(5), true, true);</code></pre>
```
joel-costigliola, line 1359: join lines

src/main/java/org/assertj/core/api/AbstractDateAssert.java
```diff
@@ -1330,6 +1350,29 @@ public SELF isBetween(String start, String end, boolean inclusiveStart, boolean
     return myself;
   }
 
+  /**
+   * Same assertion as {@link #isBetween(Date, Date, boolean, boolean)} but given date is represented as
+   * {@code java.time.Instant}.
```
joel-costigliola, line 1355: `but given date` -> `but given period`

joel-costigliola (commented):
> Because it's not possible to add assertionInvocationWithInstantArg to base class and don't override it in other derived classes (then I must add other methods for Instant) I had added it in the derived classes. At the next step, when there will be Instant methods for all derived classes it can be pulled up to the base class.

I agree with the approach, I have just minor comments on this PR, once addressed I'll merge it and then if you can create another PR for all other assertions and pull up `assertionInvocationWithInstantArg` and `should_verify_assertion_with_instant_arg` to `AbstractDateAssertWithDateArg_Test` that will be great. Let me know if that's not too much work.

src/main/java/org/assertj/core/api/AbstractDateAssert.java
joel-costigliola: When joining the lines leave one space before `assertThat`:
```
<pre><code class='java'> assertThat(new Date())...
```

joel-costigliola (commented):
just minor comments on javadoc but better getting them right before doing the remaining assertions.

joel-costigliola:
Integrated thanks @ars-java! I just finished cleaning up the javadoc

PR #2038, opened 2020-11-07
Fix error mesage of ShouldNotHaveSameClass and test naming convention of same class assertions

lykims:
#### Check List:
* Fixes NA
* Unit tests : NA

The error message of `ShouldNotHaveSameClass` is fixed because it was double negative.

Some inconsistencies were also found for the test naming convention of same class assertions while I was working on https://github.com/assertj/assertj-core/issues/2028.


joel-costigliola:
Nice catch, thanks for the fix @lykims!

PR #2040, opened 2020-11-15
Use dedicated executorservice for every testcase

### LINKED ISSUE #2030: Investigate future flaky tests
#### Summary

Some tests were annotated with `@DisabledOnOs` because they failed in CI but not locally, we haven't yet understood why.

epeee:
fixes #2030

Basically, this pr creates a separate [executorservice](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ExecutorService.html) for every test case and additionally a [shutdown now](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ExecutorService.html#shutdownNow()) is performed.
This will make our future tests more stable (especially on nodes with low number of cores).

Previously, in some cases tasks from other test cases were still running.

We might go for a junit jupiter parameterresolver based solution in the future.


joel-costigliola:
looks good, thanks @epeee!

PR #2063, opened 2020-12-08
Use org.junit.ComparisonFailure when available and org.opentest4j.AssertionFailedError is not.

### LINKED ISSUE #2062: Make .isEqualToNormalizingXxx failures easier to analyze.
#### Summary

IntelliJ IDEA (and most likely other IDEs) have a nice support for `org.junit.ComparisonFailure`, where it process the exception and generate a nice report with the diff, which I find very useful when comparing large objects. `isEqualTo ()` will already throw this error. I believe all `isEqualToNormalizingXxxx` methods should do the same instead of  throwing a raw `AssertionError`.

#### Example

Compare the result of the below (failling) tests in IntelliJ:

```java
public class AssertionFailureTest {

  @Test
  public void isEqualTo () {
    String expected = IntStream.range (0, 200)
      .mapToObj (String::valueOf)
      .collect (Collectors.joining ("\n"));
    String actual = IntStream.range (0, 200)
      .filter (i -> i != 42)
      .mapToObj (String::valueOf)
      .collect(Collectors.joining ("\n"));

    assertThat (actual).isEqualTo (expected);
  }

  @Test
  public void isEqualToNormalizingXxxx () {
    String expected = IntStream.range (0, 200)
      .mapToObj (String::valueOf)
      .collect (Collectors.joining ("\n"));
    String actual = IntStream.range (0, 200)
      .filter (i -> i != 42)
      .mapToObj (String::valueOf)
      .collect(Collectors.joining ("\n"));

    assertThat (actual).isEqualToNormalizingNewlines (expected);
  }

}
```

#### NB

The goal of this issue is to make it easy to find the issue in case an assertion fails. Throwing a ComparisonFailure seems me to be an easy way to do that, but feel free to suggest any other way.

EtienneMiret:
As I saw that the "show difference" link was available in IntelliJ for any failure when OpenTest4J is on the classpath, I ended up using an `org.junit.ComparisonFailure` for any failed assertions, not only `isEqualToNormalizingXxx()`.

For the special case of `isEqualToNormalizingXxx()` the `expected` and `actual` string representations are now that of normalized objects (see 5e2ef20122c91ddb643612505d4d244e85ed3573). If that is not the desired behavior, that commit can easily be removed.

#### Check List:
* Fixes #2062.
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -237,7 +237,7 @@ public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, 
     String errorMessage = Optional.ofNullable(info.overridingErrorMessage())
                                   .orElse(format(errorMessageFormat, arguments));
     String description = MessageFormatter.instance().format(info.description(), info.representation(), errorMessage);
-    AssertionError assertionError = assertionErrorCreator.assertionError(description, actual, expected);
+    AssertionError assertionError = assertionErrorCreator.assertionError(description, actual, expected, info.representation ());
     Failures.instance().removeAssertJRelatedElementsFromStackTraceIfNeeded(assertionError);
     removeCustomAssertRelatedElementsFromStackTraceIfNeeded(assertionError);
     return assertionError;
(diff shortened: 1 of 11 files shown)


joel-costigliola:
Integrated thanks a lot @EtienneMiret!

EtienneMiret:
You're welcome @joel-costigliola!

And sorry about the formatting. I did install the 'Eclipse Code Formatter' plugin in my IntelliJ, and configured it to use the `assertj-eclipse-formatter.xml`, but it seems this works only when asking IntelliJ to reformat a file. When using auto-completion, IntelliJ kept using my default code style. And I'm so used to it that I didn't noticed the code I submitted wasn't complying with AssertJ's convention.

joel-costigliola:
No problem, I don't mind minor reformatting and going through the PR code
from my IDE is a good way of reviewing the contribution.
Thanks again!

PR #2071, opened 2020-12-11
Fix #2068 isValidGetter checks if return type of method is not void

### LINKED ISSUE #2068: Introspection.findGetter find wrong getter method with return type void
#### Summary
Introspection does not work if target class has a method that has same name as the field I like to extract but with return type void

#### Example
```java
org.apache.xerces.dom.events.EventImpl event = ...
assertThat(event)
        .extracting("preventDefault")
        .asInstanceOf(InstanceOfAssertFactories.BOOLEAN)
        .isTrue();
```
The class EventImpl has a method 
```java
public void preventDefault()
```
The method findGetter of class org.assertj.core.util.introspection.Introspection will return this method as getter. Therefore, the field is not introspected at all. The found getter method will return null since it has void as return type.

To fix this the method isValidGetter should check if the given method has a return type that is not void.

#### Workaround
Introspection.setExtractBareNamePropertyMethods(false);

weissreto:
Method isValidGetter of class Introspection checks if return type of
given method is not void

Added a test on Introspection_getProperty_Test that assures a exception
is thrown if getter method returns void

Added a test on PropertyOrFieldSupport_getValueOf_Test that assures a
field value is read even if a getter method with void return type exists

#### Check List:
* Fixes #2068
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/util/introspection/Introspection.java
+++ b/src/main/java/org/assertj/core/util/introspection/Introspection.java
@@ -107,7 +107,7 @@ public final class Introspection {
   }
 
   private static boolean isValidGetter(Method method) {
-    return method != null && !Modifier.isStatic(method.getModifiers());
+    return method != null && !Modifier.isStatic(method.getModifiers()) && !Void.TYPE.equals(method.getReturnType());
   }
 
   private static Method findMethod(String name, Object target) {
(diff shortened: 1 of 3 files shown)


scordio (approved):

scordio:
Integrated, thanks @weissreto!

weissreto:
Thanks @scordio for integrating and your work assertj

PR #2073, opened 2020-12-14
Fix long overflow in AbstractTemporalAssert

stachenov:
If the difference between the actual and expected values was too large to fit in the specified `ChronoUnit`, an exception would be thrown. Fix that by falling back to `Duration.between()` instead of `ChronoUnit.between()`. It leads to ugly assertion messages, but it's better than an `ArithmeticException`.

An example would be comparing Instant.MIN (e. g. used to signify an “uninitialized” value) to something reasonable `within(1, MILLIS)`.

The appropriate unit test is added.

I'm not sure if some thing can happen with, say, `LocalDate`. (That is, can the number of `DAYS` be too large to fit in a `long`? Probably not.)


joel-costigliola:
Good catch @stachenov!

joel-costigliola:
Integrated thanks @stachenov!

PR #2074, opened 2020-12-14
[Conditions] recalculate Description - nested Conditions for Not and Join

stbischof:
recalculate the description of nested Conditions like `Not` and `Join`

1 - Respect change of Description after Construct

It could happen, that the Description changes while executing for example the matches Method.
At the moment there is the description in JoinConditions/(maybe also Not) only set in the constructor. It does not show the later changed Description.

2 - Not should be shown properly 
```
    allNot = allOf(not(new TestCondition<>("AllNot")));
```
shoult look like
```
  <all of:
    not :<AllNot>
 >
```
but was:
```
  <"all of:[
   Not
]">
```

#### Check List:
* Fixes #???
* Unit tests : YES 

### DIFF
--- a/src/main/java/org/assertj/core/condition/DoesNotHave.java
+++ b/src/main/java/org/assertj/core/condition/DoesNotHave.java
@@ -36,11 +36,6 @@ public class DoesNotHave<T> extends Negative<T> {
 
   private DoesNotHave(Condition<? super T> condition) {
     super(condition);
+    describedAs("does not have :<%s>", condition);
   }
-
-  @Override
-  public String toString() {
-    return String.format("does not have :<%s>", condition);
-  }
-
 }
--- a/src/main/java/org/assertj/core/condition/Join.java
+++ b/src/main/java/org/assertj/core/condition/Join.java
@@ -58,14 +58,12 @@ public abstract class Join<T> extends Condition<T> {
    */
   protected Join(Iterable<? extends Condition<? super T>> conditions) {
     this(Streams.stream(checkNotNullConditions(conditions)));
+    calculateDescription();
   }
 
   private Join(Stream<? extends Condition<? super T>> stream) {
     this.conditions = stream.map(Join::notNull).collect(toList());
-    List<Description> descriptions = this.conditions.stream()
-                                                    .map(Condition::description)
-                                                    .collect(toList());
-    this.describedAs(new JoinDescription(descriptionPrefix() + ":[", "]", descriptions));
+    calculateDescription();
   }
 
   private static <T> T checkNotNullConditions(T conditions) {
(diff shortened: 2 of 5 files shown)


joel-costigliola:
@stbischof what is the root cause issue you are trying to solve? it's not super clear.

stbischof:
It could happen, that the Description changes while executing for example the Match. 

At the moment there is the description in JoinConditions only set in the constructor.

```
class ButWasCondtion{
	private ButWasCondtion(Function<T, O> operation, EXPECTED expectedResult, BiPredicate<O, EXPECTED> predicate,
		String predicatText, Function<EXPECTED, ?> transformExpected, Function<O, ?> transformWas) {
		super("");
		this.operation = operation;
		this.biPredicatet = predicate;
		this.predicatText = predicatText;
		this.expectedResult = expectedResult;
		this.transformExpected = transformExpected;
		this.transformWas = transformWas;
	}

	@Override
	public boolean matches(T value) {
		O wasObject = operation.apply(value);
		boolean match = biPredicatet.test(wasObject, expectedResult);
		if (match) {
			describedAs("\n[OK] " + predicatText + " <%s>\n", transformIf(transformExpected, expectedResult));
		} else {
			describedAs("\n[FAIL] " + predicatText + " <%s> but was <%s>\n",
				transformIf(transformExpected, expectedResult), transformIf(transformWas, wasObject));
		}
		return match;
	}

	private static <E> Object transformIf(Function<E, ?> transform, E object) {
		return transform == null ? object : transform.apply(object);
	}
```

stbischof:
Do you wish some changes?

Is `ButWasCondtion` or the `MappedCondition` (see below) something you would like to add to assertj in any way?



```
public class MappedCondition<FROM, T> extends Condition<FROM> {

	private Function<FROM, T>	mapping;
	private Condition<T>		condition;
	private String				mappedToText;

	public static <FROM, T> Condition<FROM> mappedCondition(Function<FROM, T> mapping, Condition<T> condition,
		String mappedToText) {
		return new MappedCondition<FROM, T>(mapping, condition, mappedToText);
	}

	private MappedCondition(Function<FROM, T> mapping, Condition<T> condition, String mappedToText) {
		super("");
		this.mapping = mapping;
		this.condition = condition;
		this.mappedToText = mappedToText;
		calcDescribedAs();
	}

	@Override
	public boolean matches(FROM value) {
		T mappedObject = mapping.apply(value);
		boolean matchResult = condition.matches(mappedObject);
		calcDescribedAs();
		return matchResult;
	}

	protected void calcDescribedAs() {
		describedAs(mappedToText + " [\n" + "   %s\n" + "]", condition);
	}

}
```

joel-costigliola:
thanks @stbischof I'll integrate the PR soonish.

> If we iterate over all SubConditions in AllOff we could have something like an SoftAssert. At the moment it stops testing on the first fail.

We could indeed, this would an improvement but let's address it in another issue/PR.

> Is ButWasCondtion or the MappedCondition (see below) something you would like to add to assertj in any way?

Let's discuss that in a separate issue, it's not super clear what it the problem you try to solve with these, some examples would be welcome.

joel-costigliola:
Integrated thanks @stbischof!
I refactored the code a bit:

- `calculateDescription()` in `Join` is called in the overriden `description()` insteaf of `toString()`
- `toString()` in `Condition` uses `description()` insteaf of the `description` field
- moved the test involving `allOf` in `Not_toString_Test` to `AllOf_toString_Test`

PR #2076, opened 2020-12-15
Satisfies exactly in any order

### LINKED ISSUE #2054: Consumer-based "satisfyExactlyInAnyOrder" assertions for elements in arrays and iterables
#### Summary

Based on #1621, extend the assertions API for arrays and iterables with a `satisfyExactlyInAnyOrder` method accepting multiple  `Consumer<? super ELEMENT>`s. 
The tested collection must have the same size as the consumers. For each element of the tested collection the (remaining) consumers are called; if a consumer accepts the element, then the element and that consumer are removed from their collections and the next element is probed.

The additional method would have the signature: 
```java
public final SELF satisfyExactlyInAnyOrder(Consumer<? super ELEMENT>... consumers)
```

Additionally, the method would be overloaded with an assert factory as first parameter and consumers operating not on the collection ELEMENT but on an assertion over that element (as suggested by @jbytecoder):
```java
public final <A extends AbstractAssert<?, ELEMENT>> 
SELF satisfyExactlyInAnyOrder(Function<ELEMENT, A> assertFactory, Consumer<? super A>... consumers)
```

Note: the implementation from #1892 (originally for the `satisfy` method, aka. `containsElementsSatisfyingExactlyInAnyOrder`, still pending agreement on clear naming) already performs all necessary logic. It could be combined with a simple length verification.

#### Example

```java
assertThat(trilogy).satisfyExactlyInAnyOrder(movie -> {
    // the fellowship of the Ring
    assertThat(movie.getReleaseDate()).isBefore(parse("2002-12-17"));
    assertThat(movie.xrated).isFalse();
  }, 
  movie -> {
    // the Return of the King
    assertThat(movie.getTitle()).endsWith("King");
  },
  movie -> {
    // the two Towers
     assertThat(movie.getReleaseDate()).isEqualTo(parse("2002-12-18"));
     assertThat(movie.getTitle()).contains("Towers");
  } 
);

assertThat(movieTitles).satisfyExactlyInAnyOrder(
  Assertions::assertThat,
  s -> s.matches(".* Return .* King"),
  s -> s.endsWith("Towers"), 
  s -> s.contains("fellowship")
);
```

This issue (and #2053) was split from #1621 as suggested by @joel-costigliola in https://github.com/assertj/assertj-core/issues/1621#issuecomment-539696963

mgrafl:
#### Check List:
* Fixes #2054
* Unit tests : YES

Note: Method was named `satisfiesExactlyInAnyOrder` to according to https://github.com/assertj/assertj-core/issues/2053#issuecomment-743946171


src/main/java/org/assertj/core/api/ObjectEnumerableAssert.java
```diff
@@ -1311,11 +1311,49 @@
...
   SELF satisfiesExactly(@SuppressWarnings("unchecked") Consumer<? super ELEMENT>... allRequirements);
 
+  /**
+   * Verifies that the actual iterable has the same size as the given consumers and 
+   * that at least one permutation of elements in the iterable exists that satisfies 
+   * the individual consumers in order.
+   * <p>
+   * Examples:
+   * <pre><code class='java'>
+   * List&lt;String&gt; starWarsCharacterNames = list("Luke", "Leia", "Yoda");
```
joel-costigliola, line 1326: join line and one space between them

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1147,6 +1149,33 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
       throw failures.failure(info, elementsShouldSatisfy(actual, unsatisfiedRequirements, info));
   }
 
+  private <E> Function<? super Consumer<? super E>, List<E>> listFilteredBySatisfiedElements(Iterable<? extends E> actual) {
```
joel-costigliola, line 1152: can be static

joel-costigliola: name is confusing, I was expecting a list from this method

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1171,6 +1200,37 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
...
+    assertNotNull(info, actual);
+    requireNonNull(consumers, "The Consumer<? super E>... expressing the assertions consumers must not be null");
+    for (Consumer<? super E> consumer : consumers)
+      requireNonNull(consumer, "The element in consumers must not be null");
+
+    checkSizes(actual, sizeOf(actual), consumers.length, info);
+
+    List<E>[] satisfiedElementsLists = stream(consumers)
+                                                        .map(listFilteredBySatisfiedElements(actual))
+                                                        .<List<E>> toArray(List[]::new);
```
joel-costigliola, line 1227: the code is a bit complex (the issue is too to be fair), I feel it could simplified by introducing a simple class to capture the elements satisfying a consumer, something like:
```java
// boilerplate code omitted
static class ElementsSatisfyingRequirements {
  List<E> elements;
  Consumer<? super E> requirements; 
}
```
it will make the code more intention explicit.

I would also suggest  `listFilteredBySatisfiedElements` to return directly a `List<ElementsSatisfyingRequirements>`

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1171,6 +1200,37 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
...
+
+    if (!isSatisfied(satisfiedElementsLists, 0))
+      throw failures.failure(info, shouldSatisfy(actual));
```
joel-costigliola, line 1230: we better have a  `shouldSatisfyExactlyInAnyOrder` method here that will build an error message explaining why the assertions failed.

src/test/java/org/assertj/core/error/ShouldSatisfy_create_Test.java
```diff
@@ -36,4 +43,17 @@ void should_create_error_message() {
...
+  void should_create_error_message_if_consumers_are_not_all_satisfied() {
+    // GIVEN
+    ErrorMessageFactory factory = shouldSatisfy(newArrayList("Luke", "Leia", "Yoda"));
+    // WHEN
+    String message = factory.create(new TextDescription("Test"), STANDARD_REPRESENTATION);
+    // THEN
+    then(message).isEqualTo(format("[Test] %n"
+                                   + "Expecting:%n"
+                                   + "  <[\"Luke\", \"Leia\", \"Yoda\"]>%n"
+                                   + "to satisfy all the consumers."));
```
joel-costigliola, line 57: maybe add `in any order`?

src/test/java/org/assertj/core/internal/iterables/Iterables_assertSatisfiesExactlyInAnyOrder_Test.java
```diff
@@ -0,0 +1,164 @@
...
+ * @author Ting Sun
+ */
+@DisplayName("Iterables assertSatisfiesExactlyInAnyOrder")
+class Iterables_assertSatisfiesExactlyInAnyOrder_Test extends IterablesBaseTest {
+
+  private List<String> actual = newArrayList("Luke", "Leia", "Yoda");
+
+
+  @Test
+  void should_pass_if_there_are_multiple_consumers_and_can_be_satisfied() {
```
joel-costigliola, line 45: rename to `should_pass_if_all_consumers_are_satisfied_by_different_elements_in_order`?

src/test/java/org/assertj/core/internal/iterables/Iterables_assertSatisfiesExactlyInAnyOrder_Test.java
```diff
@@ -0,0 +1,164 @@
...
+      assertThat(s).doesNotContain("L");
+    }; // Matches "Yoda"
+    Consumer<String> consumer3 = s -> assertThat(s).doesNotContain("a"); // Matches "Luke"
+
+    // WHEN/THEN
+    iterables.assertSatisfiesExactlyInAnyOrder(info, actual, consumer1, consumer2, consumer3);
+  }
+
+  @Test
+  void should_pass_there_are_multiple_consumers_and_can_be_satisfied_in_any_order() {
```
joel-costigliola, line 59: rename to `should_pass_if_all_consumers_are_satisfied_by_different_elements_in_any_order`?

src/test/java/org/assertj/core/internal/iterables/Iterables_assertSatisfiesExactlyInAnyOrder_Test.java
```diff
@@ -0,0 +1,164 @@
...
+    Consumer<String> consumer = null;
+    String message = "The element in consumers must not be null";
+
+    // WHEN/THEN
+    assertThatNullPointerException().isThrownBy(() -> assertThat(actual).satisfiesExactlyInAnyOrder(consumer))
+                                    .withMessage(message);
+  }
+
+  @Test
+  void should_fail_if_there_is_only_one_consumer_and_cannot_be_satisfied() {
```
joel-costigliola, line 129: rename to `should_fail_if_one_of_the_consumers_cannot_be_satisfied`

src/test/java/org/assertj/core/internal/iterables/Iterables_assertSatisfiesExactlyInAnyOrder_Test.java
```diff
@@ -0,0 +1,164 @@
...
+    Consumer<String> consumer2 = s -> assertThat(s).hasSize(4);
+    Consumer<String> consumer3 = s -> assertThat(s).hasSize(4);
+    // WHEN
+    AssertionError assertionError = expectAssertionError(() -> iterables.assertSatisfiesExactlyInAnyOrder(info, actual, consumer1, consumer2, consumer3));
+    // THEN
+    then(assertionError).hasMessage(shouldSatisfy(actual).create());
+  }
+
+  @Test
+  void should_fail_if_there_are_multiple_consumers_and_cannot_be_all_satisfied() {
```
joel-costigliola, line 141: rename to `should_fail_if_multiple_consumers_cannot_be_satisfied`

src/test/java/org/assertj/core/internal/iterables/Iterables_assertSatisfiesExactlyInAnyOrder_Test.java
```diff
@@ -0,0 +1,164 @@
...
+    Consumer<String> consumer3 = s -> assertThat(s).startsWith("L"); // Matches "Luke" or "Leia"
+
+    // WHEN
+    AssertionError assertionError = expectAssertionError(() -> iterables.assertSatisfiesExactlyInAnyOrder(info, actual, consumer1, consumer2, consumer3));
+    // THEN
+    then(assertionError).hasMessage(shouldSatisfy(actual).create());
+  }
+
+  @Test
+  void should_fail_if_consumers_are_more_than_elements() {
```
joel-costigliola, line 154: rename to `should_fail_if_there_are_too_many_consumers`

joel-costigliola (changes requested):
Thanks @mgrafl, I think the code can be made a bit more explicit and the error message more informative.

src/main/java/org/assertj/core/api/ObjectEnumerableAssert.java
mgrafl: Thanks, I will align the line breaks in documentation example based on existing methods.

src/main/java/org/assertj/core/internal/Iterables.java
mgrafl: Yes, I will rename the method accordingly.

src/test/java/org/assertj/core/error/ShouldSatisfy_create_Test.java
mgrafl: Thanks, that's a good idea.

src/test/java/org/assertj/core/internal/iterables/Iterables_assertSatisfiesExactlyInAnyOrder_Test.java
mgrafl: I agree.

src/main/java/org/assertj/core/internal/Iterables.java
mgrafl: Thanks, that sounds like a good idea. I will try to refactor the code accordingly.

mgrafl: @joel-costigliola, would it be okay to place the ```ElementsSatisfyingRequirements``` class in a separate public class (instead of a nested class of ```Iterables```, as your previous comment indicates)?

joel-costigliola: A separate class is fine, try to keep it package private if possible

src/main/java/org/assertj/core/internal/Iterables.java
mgrafl: You are right, I renamed it.

src/main/java/org/assertj/core/internal/Iterables.java
mgrafl: @joel-costigliola, I have refactored it and hope the intention is now clearer. Based on other methods and variable names, I have named to class ```ElementsSatisfyingConsumer```. Please have a look at it and let me know if you still prefer ```ElementsSatisfyingRequirements```.

joel-costigliola:
Integrated thanks @mgrafl!

PR #2080, opened 2020-12-17
AllOf.allOfSoftly

stbischof:
If we iterate over all SubConditions in AllOff we could have something like an SoftAssert. At the moment it stops testing on the first fail. This is the first approach to address this.

Sorry, still not able to run unit tests in my ide.

### DIFF
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -2224,6 +2224,33 @@ public class Assertions implements InstanceOfAssertFactories {
   }
 
   /**
+   * Creates a new <code>{@link AllOf}</code>
+   * All nested conditions will be checked.
+   * @param <T> the type of object the given condition accept.
+   * @param conditions the conditions to evaluate.
+   * @return the created {@code AllOf}.
+   * @throws NullPointerException if the given iterable is {@code null}.
+   * @throws NullPointerException if any of the elements in the given iterable is {@code null}.
+   */
+  public static <T> Condition<T> allOfSoftly(Iterable<? extends Condition<? super T>> conditions) {
+    return  AllOf.allOfSoftly(conditions);
+  }
+
+  /**
+   * Creates a new <code>{@link AllOf}</code>
+   * All nested conditions will be checked.
+   * @param <T> the type of object the given condition accept.
+   * @param conditions the conditions to evaluate.
+   * @return the created {@code AllOf}.
+   * @throws NullPointerException if the given array is {@code null}.
+   * @throws NullPointerException if any of the elements in the given array is {@code null}.
+   */
+  @SafeVarargs
+  public static <T> Condition<T> allOfSoftly(Condition<? super T>... conditions) {
+    return AllOf.allOfSoftly(conditions);
+  }
+
+  /**
    * Only delegate to {@link AnyOf#anyOf(Condition...)} so that Assertions offers a full feature entry point to all
    * AssertJ features (but you can use {@link AnyOf} if you prefer).
    * <p>
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -2223,34 +2223,7 @@ public class Assertions implements InstanceOfAssertFactories {
     return AllOf.allOf(conditions);
   }
 
-  /**
-   * Creates a new <code>{@link AllOf}</code>
-   * All nested conditions will be checked.
-   * @param <T> the type of object the given condition accept.
-   * @param conditions the conditions to evaluate.
-   * @return the created {@code AllOf}.
-   * @throws NullPointerException if the given iterable is {@code null}.
-   * @throws NullPointerException if any of the elements in the given iterable is {@code null}.
-   */
-  public static <T> Condition<T> allOfSoftly(Iterable<? extends Condition<? super T>> conditions) {
-    return  AllOf.allOfSoftly(conditions);
-  }
-
-  /**
-   * Creates a new <code>{@link AllOf}</code>
-   * All nested conditions will be checked.
-   * @param <T> the type of object the given condition accept.
-   * @param conditions the conditions to evaluate.
-   * @return the created {@code AllOf}.
-   * @throws NullPointerException if the given array is {@code null}.
-   * @throws NullPointerException if any of the elements in the given array is {@code null}.
-   */
-  @SafeVarargs
-  public static <T> Condition<T> allOfSoftly(Condition<? super T>... conditions) {
-    return AllOf.allOfSoftly(conditions);
-  }
-
-  /**
+   /**
    * Only delegate to {@link AnyOf#anyOf(Condition...)} so that Assertions offers a full feature entry point to all
    * AssertJ features (but you can use {@link AnyOf} if you prefer).
    * <p>
--- a/src/main/java/org/assertj/core/condition/AllOf.java
+++ b/src/main/java/org/assertj/core/condition/AllOf.java
@@ -12,6 +12,8 @@
  */
 package org.assertj.core.condition;
 
+import java.util.concurrent.atomic.AtomicBoolean;
+
 import org.assertj.core.api.Condition;
 
 /**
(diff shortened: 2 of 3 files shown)


joel-costigliola:
@stbischof I'm not sure we are on the same page on this one, what I had in mind was to report all the failed conditions as `allOf`  default behaviorinstead of stopping at the first one and make the error message to report all of them (like soft assertions).
It's probably a good idea to reuse `UnsatisfiedRequirement` class from the org.assertj.core.error package

stbischof:
> what I had in mind was to report all the failed conditions as `allOf` default behaviorinstead of stopping at the first one and make the error message to report all of them (like soft assertions).

i think just have one AllOf and that tests all conditions (not stopping) would be the best way. but i would not change the default behavior. will have a look at `UnsatisfiedRequirement` . do not know what it is at the moment

joel-costigliola:
I had another look at the code and I think the way to go is to add a new factory method to `ShouldBe`:
```java
  public static <T> ErrorMessageFactory shouldBe(T actual, AllOf<? super T> condition) {
    String collect = condition.conditions().stream()
                              .filter(c -> !c.matches(actual))
                              .map(Condition::description)
                              .map(Description::toString)
                              .collect(joining(format("%n- ")));

    return new ShouldBe(actual, condition, collect);
  }

  private <T> ShouldBe(T actual, AllOf<?> condition, String failures) {
    super("%nExpecting:%n <%s>%nto be <%s>%nbut these conditions failed:%n- " + failures, actual, condition);
  }
```
`conditions()` must be made public in `Join`.

Here's an example message with these changes:
```java
    TestCondition<Object> condition1 = new TestCondition<>("Condition 1");
    TestCondition<Object> condition2 = new TestCondition<>("Condition 2");
    condition1.shouldMatch(true);
    condition2.shouldMatch(false);
    AllOf<Object> allOf = allOf(condition1, condition2);
    ErrorMessageFactory factory = shouldBe("Yoda", allOf);
    // WHEN
    String message = factory.create(new TextDescription("Test"), new StandardRepresentation());
    // THEN
    System.out.println(message);
```
error message:
```
[Test] 
Expecting:
 <"Yoda">
to be <all of:[
   Condition 1,
   Condition 2
]>
but these conditions failed:
- Condition 2
```
Does that make sense @stbischof?
If so you can start from there and improve it further (notably with tests showing more nested conditions)

stbischof:
This ShouldBe looks really good. 
I did not saw this. so i came with this: https://github.com/assertj/assertj-core/pull/2083

stbischof:
this looks good. didnt know about ShouldBe. (thats why i came up with https://github.com/assertj/assertj-core/pull/2083)

would see the
- should be value
- but was value
- state of the condition (NOT_EXECUTED, OK, FAILED)

joel-costigliola:
The way I suggested is better as allOf would benefit from it and we don't need to introduce `allOfSoftly`

stbischof:
sure. it is a really good way. Ist just that you have to know much more about assertj  because all of the ErrorMessageFactory part.
```
   ErrorMessageFactory factory = shouldBe("Yoda", allOf);
    // WHEN
    String message = factory.create(new TextDescription("Test"), new StandardRepresentation());
```

joel-costigliola:
Let's do one thing at a time, I'll review #2083 later on.
I might change the formatting of this one if nested conditions are not clearly displayed.

joel-costigliola:
playing further, here's an example with nested failed conditions:
```java
TestCondition<Object> condition1 = new TestCondition<>("a Sith");
TestCondition<Object> condition2 = new TestCondition<>("very tall");
TestCondition<Object> condition3 = new TestCondition<>("young");
condition1.shouldMatch(false);
condition2.shouldMatch(false);
condition3.shouldMatch(false);
Condition<Object> allOf = allOf(condition1,
                                allOf(condition1, condition2),
                                anyOf(condition2, condition3));
ErrorMessageFactory factory = shouldBe("Yoda", allOf);

String message = factory.create(new TextDescription("Test"), new StandardRepresentation());
```
gives this message:
```
[Test] 
Expecting:
  "Yoda"
to be all of:[
   a Sith,
   all of:[
      a Sith,
      very tall
   ],
   any of:[
      very tall,
      young
   ]
]
but these conditions failed:
- a Sith
- all of:[
   a Sith,
   very tall
]
- any of:[
   very tall,
   young
]
```

stbischof:
@joel-costigliola that looks good!!!

could you post the result is
```
condition1.shouldMatch(true);
condition2.shouldMatch(false);
condition3.shouldMatch(true);
```

joel-costigliola:
For:
```
condition1.shouldMatch(true);
condition2.shouldMatch(false);
condition3.shouldMatch(true);
```
the message is:
```
[Test] 
Expecting:
  "Yoda"
to be all of:[
   a Sith,
   all of:[
      a Sith,
      very tall
   ],
   any of:[
      very tall,
      young
   ]
]
but this condition failed:
- all of:[
   a Sith,
   very tall
]
```

stbischof:
thank you for execting it fore me. 
this shows why it is not really is what expect from a softly.
you cant see:
-  which condition in  `al` fails
- stat a condition in any fails (even it is okay because it is isnside an `any`)

joel-costigliola:
Ok, I see.

stbischof:
for this pr where we talk about the  `but this condition failed: X` section i agree that we  could and should only report the failed conditions in allOf and nothing in `any`.

But I still think that we should have an more Descriptive condition #2083 with better inline output

joel-costigliola:
I'm playing with the code to report which **nested** conditions failed in allOf, I have come up with something that probably needs more testing.
I don't want to show the failed state of the first level conditions since they are described in the failed conditions section but I need to show the state of the nested conditions.
So I can get this from the previous test where only the "very tall" condition was set to  fail:
```
[Test] 
Expecting:
  "Yoda"
to be all of:[
   a Sith,
   all of:[
      a Sith,
      very tall
   ],
   any of:[
      very tall,
      young
   ]
]
but this condition failed:
- all of:[
   a Sith,
   [failed] very tall
]
```

joel-costigliola:
another example with nested allOf failing due to one condition:
```
[Test] 
Expecting actual:
  "Yoda"
to be all of:[
   all of:[
      a Jedi,
      all of:[
         a Jedi,
         all of:[
            very tall,
            old
         ]
      ]
   ]
]
but this condition failed:
- all of:[
   a Jedi,
   [failed] all of:[
      a Jedi,
      [failed] all of:[
         [failed] very tall,
         old
      ]
   ]
]
```

joel-costigliola:
I haven't decided yet whether to use `[failed]` or `[x]` to indicate failures

stbischof:
If i read this i thought:
* we need to indent it for better reading.
* Condition strings of one level should be in one vertical line
* we also should add a [failed] at the first level `all of` 

```
[Test] 
Expecting actual:
  "Yoda"
to be all of:[
   all of:[
      a Jedi,
      all of:[
         a Jedi,
         all of:[
            very tall,
            old
         ]
      ]
   ]
]
but this condition failed:
- [failed] all of:[
               a Jedi,
      [failed] all of:[
                  a Jedi,
         [failed] all of:[
            [failed] very tall,
                     old
      ]
   ]
]
```
* but then we waste a loot of space we can fill a opposite state to calm the eyes

```
[Test] 
Expecting actual:
  "Yoda"
to be all of:[
   all of:[
      a Jedi,
      all of:[
         a Jedi,
         all of:[
            very tall,
            old
         ]
      ]
   ]
]
but this condition failed:
- [failed] all of:[
      [ok]     a Jedi,
      [failed] all of:[
         [ok]     a Jedi,
         [failed] all of:[
            [failed] very tall,
            [ok]     old
      ]
   ]
]
```  
the result `[failing]` and `[ok]` (and may be  `[not executed]` ) are not the same length any my eyes could seperate that all using one symbol


```
[Test] 
Expecting actual:
  "Yoda"
to be all of:[
   all of:[
      a Jedi,
      all of:[
         a Jedi,
         all of:[
            very tall,
            old
         ]
      ]
   ]
]
but this condition failed:
- [✗] all of:[
      [✓] a Jedi,
      [✗] all of:[
         [✓] a Jedi,
         [✗] all of:[
            [✗] very tall,
            [✓] old
      ]
   ]
]
```  

but then the `to be` and the `but this condition failed:` part is still a bit duplicate 

```
[Test] 
Expecting actual:
  "Yoda"
to be 
[✗] all of:[
   [✗] all of:[
      [✓] a Jedi,
      [✗] all of:[
         [✓] a Jedi,
         [✗] all of:[
            [✓] very tall,
            [✗] old
         ]
      ]
   ]
]
```

joel-costigliola:
I find having the status of every condition to be noisy, it impedes reading the condition description, this is why I separated the condition to be met from the failed condition part(s).

I need to take a step back and ask the team for its preference.

joel-costigliola:
team has decided to go with your suggestion:
```
Expecting actual:
  "Yoda"
to be 
[✗] all of:[
   [✗] all of:[
      [✓] a Jedi,
      [✗] all of:[
         [✓] a Jedi,
         [✗] all of:[
            [✓] very tall,
            [✗] old
         ]
      ]
   ]
] 
```

PR #2086, opened 2020-12-22
Add comparator-based Object assert

### LINKED ISSUE #2041: support comparison checks using Comparator instead of Comparable
#### Summary
Currently we have a fluent API for asserting comparison results: `Assertions.assertThat(java.lang.Comparable)`, where the ordering can also be replaced by a custom comparator by the use of `usingComparator(java.lang.Comparator)`.
I'd like to have the same API available through the use of `Assertions.assertThat(Object).usingComparator(java.lang.Comparator)` for classes that don't implement `Comparable`.

#### Example

```java

assertThat(object1).usingComparator(mySpecialOrdering).isLessThan(object2);

```

scordio:
#### Check List:
* Fixes #2041
* Unit tests : NOT YET
* Javadoc with a code example (on API only) : NOT YET

### DIFF
--- a/src/main/java/org/assertj/core/api/ComparatorBasedObjectAssert.java
+++ b/src/main/java/org/assertj/core/api/ComparatorBasedObjectAssert.java
@@ -1,0 +1,428 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2020 the original author or authors.
+ */
+package org.assertj.core.api;
+
+import java.util.Comparator;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+
+import org.assertj.core.description.Description;
+import org.assertj.core.presentation.Representation;
+
+/**
+ * @param <ACTUAL> 
+ *
+ * @author Stefano Cordio
+ * @since 3.19.0
+ */
+public class ComparatorBasedObjectAssert<ACTUAL> extends ObjectAssert<ACTUAL>
+    implements ComparisonBasedAssert<ComparatorBasedObjectAssert<ACTUAL>, ACTUAL> {
+
+  public ComparatorBasedObjectAssert(ACTUAL actual, AbstractAssert<?, ?> sourceAssert,
+                                     Comparator<? super ACTUAL> customComparator, String customComparatorDescription) {
+    super(actual);
+    withAssertionState(sourceAssert).usingComparator(customComparator, customComparatorDescription);
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public ComparatorBasedObjectAssert<ACTUAL> isEqualByComparingTo(ACTUAL other) {
+    objects.assertEqualByComparison(info, actual, other);
+    return this;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public ComparatorBasedObjectAssert<ACTUAL> isNotEqualByComparingTo(ACTUAL other) {
+    objects.assertNotEqualByComparison(info, actual, other);
+    return this;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public ComparatorBasedObjectAssert<ACTUAL> isLessThan(ACTUAL other) {
+    objects.assertLessThan(info, actual, other);
+    return this;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public ComparatorBasedObjectAssert<ACTUAL> isLessThanOrEqualTo(ACTUAL other) {
+    objects.assertLessThanOrEqualTo(info, actual, other);
+    return this;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public ComparatorBasedObjectAssert<ACTUAL> isGreaterThan(ACTUAL other) {
+    objects.assertGreaterThan(info, actual, other);
+    return this;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public ComparatorBasedObjectAssert<ACTUAL> isGreaterThanOrEqualTo(ACTUAL other) {
+    objects.assertGreaterThanOrEqualTo(info, actual, other);
+    return this;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public ComparatorBasedObjectAssert<ACTUAL> isBetween(ACTUAL startInclusive, ACTUAL endInclusive) {
+    objects.assertIsBetween(info, actual, startInclusive, endInclusive, true, true);
+    return this;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public ComparatorBasedObjectAssert<ACTUAL> isStrictlyBetween(ACTUAL startExclusive, ACTUAL endExclusive) {
+    objects.assertIsBetween(info, actual, startExclusive, endExclusive, false, false);
+    return this;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public ComparatorBasedObjectAssert<ACTUAL> as(Description description) {
+    return (ComparatorBasedObjectAssert<ACTUAL>) super.as(description);
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public ComparatorBasedObjectAssert<ACTUAL> as(String description, Object... args) {
+    return (ComparatorBasedObjectAssert<ACTUAL>) super.as(description, args);
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public ComparatorBasedObjectAssert<ACTUAL> hasNoNullFieldsOrProperties() {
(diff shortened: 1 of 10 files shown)


src/main/java/org/assertj/core/api/ComparatorBasedObjectAssert.java
```diff
@@ -0,0 +1,363 @@
...
+public class ComparatorBasedObjectAssert<ACTUAL> extends ObjectAssert<ACTUAL>
+    implements ComparisonBasedAssert<ComparatorBasedObjectAssert<ACTUAL>, ACTUAL> {
+
+  public ComparatorBasedObjectAssert(ACTUAL actual, Comparator<? super ACTUAL> comparator, String comparatorDescription) {
+    super(actual);
+  }
+
+  @Override
+  public ComparatorBasedObjectAssert<ACTUAL> isEqualByComparingTo(ACTUAL other) {
+    // FIXME assertion
```
scordio, line 39: My preference would be to enhance `org.assertj.core.internal.Objects` with some comparator-based methods that work only if `comparisonStrategy` is an instance of `ComparatorBasedComparisonStrategy`

src/main/java/org/assertj/core/api/ComparisonBasedAssert.java
```diff
@@ -0,0 +1,196 @@
...
+ * 
+ * @param <SELF> the "self" type of this assertion class. Please read &quot;<a href="http://bit.ly/1IZIRcY"
+ *          target="_blank">Emulating
+ *          'self types' using Java Generics to simplify fluent API implementation</a>&quot; for more details.
+ * @param <ACTUAL> the type of the "actual" value.
+ * 
+ * @author Stefano Cordio
+ * @since 3.19.0
+ */
+public interface ComparisonBasedAssert<SELF extends ComparisonBasedAssert<SELF, ACTUAL>, ACTUAL> {
```
scordio, line 28: The javadoc in this interface has to be adjusted (currently copy-pasted from `ComparableAssert`)

src/main/java/org/assertj/core/api/ComparatorBasedObjectAssert.java
joel-costigliola: sounds good to me

src/main/java/org/assertj/core/api/ComparatorBasedObjectAssert.java
```diff
@@ -0,0 +1,363 @@
...
+  @Override
+  public ComparatorBasedObjectAssert<ACTUAL> isStrictlyBetween(ACTUAL startExclusive, ACTUAL endExclusive) {
+    // FIXME assertion
+    return this;
+  }
+
+  @Override
+  public ComparatorBasedObjectAssert<ACTUAL> as(Description description) {
+    return (ComparatorBasedObjectAssert<ACTUAL>) super.as(description);
+  }
```
joel-costigliola, line 88: not sure it it is a better approach but we would not need to cast if we were returning `myself`:
```
  @Override
  @CheckReturnValue
  public SELF as(Description description) {
    return super.as(description);
  }
```
Not 100% sure if we need to annotate with `@CheckReturnValue` as the overridden method has already the annotation

src/main/java/org/assertj/core/api/ComparatorBasedObjectAssert.java
```diff
@@ -0,0 +1,363 @@
...
+/**
+ * @param <ACTUAL> 
+ *
+ * @author Stefano Cordio
+ * @since 3.19.0
+ */
+public class ComparatorBasedObjectAssert<ACTUAL> extends ObjectAssert<ACTUAL>
+    implements ComparisonBasedAssert<ComparatorBasedObjectAssert<ACTUAL>, ACTUAL> {
+
+  public ComparatorBasedObjectAssert(ACTUAL actual, Comparator<? super ACTUAL> comparator, String comparatorDescription) {
```
joel-costigliola, line 33: you should assign `comparatorDescription` to a field if it's going to be reused later or get rid of it if we find it's not that useful

src/main/java/org/assertj/core/api/ComparatorBasedObjectAssert.java
scordio: `myself` is tricky in this case. As we inherit from `ObjectAssert` and not from `AbstractObjectAssert`, `myself` has `ObjectAssert<ACTUAL>` type. We could redefine another `myself` field in `ComparatorBasedObjectAssert` or just use `this` and then the pattern would become something like:

```java
  @Override
  public ComparatorBasedObjectAssert<ACTUAL> as(Description description) {
    super.as(description);
    return this;
  }
```

I preferred the casting not to miss any potential manipulation of myself in any of the superclasses, like methods doing:
```java
    return myself.with...();
```
but I'm not fully sure if it is needed as this might happen only for methods returning a new assertion object and these should not be overridden. I need to double-check this.

Inheriting from `AbstractObjectAssert` was not an option because I needed to specialize the return type of `usingComparator` methods, and I think that as long as we have SELF the type invariance of generics is preventing the specialization. That's why I fall back to `ObjectAssert` where we have the concrete type, which can be specialized to a subtype. I'm open to any better approach!

src/main/java/org/assertj/core/api/ComparatorBasedObjectAssert.java
scordio: Right, I need to figure out if this is useful or not. Initially, I was not sure if I needed to define a new class for assertions, something intermediate between `Objects` and `Comparables`, but now I'll try the Objects approach mentioned in the other comment, maybe `comparatorDescription` will be useless here.

scordio:
@joel-costigliola one more step forward on this topic. I'd suggest having a look at the direction as this change is going to be much bigger once everything is properly tested.

joel-costigliola:
will do this week end!

src/main/java/org/assertj/core/api/ObjectAssert.java
```diff
@@ -47,4 +48,15 @@ public ObjectAssert(AtomicReference<ACTUAL> actual) {
     return super.extracting(extractors);
   }
 
+  @Override
+  public ComparatorBasedObjectAssert<ACTUAL> usingComparator(Comparator<? super ACTUAL> customComparator) {
+    return usingComparator(customComparator, null);
```
joel-costigliola, line 53: Minor point: should we try to describe the customComparator here?
if it is defined as a class we could pass the class name (eg: `AbsComparator`) and leave it to null if it's defined as a lambda or an anonymous class.

src/main/java/org/assertj/core/internal/Objects.java
```diff
@@ -932,6 +937,169 @@ public boolean areEqualToComparingOnlyGivenFields(Object actual, Object other,
...
+   * Asserts that two <code>{@link Comparable}</code>s are equal by invoking
+   * <code>{@link Comparable#compareTo(Object)}</code>.<br>
+   *
+   * @param <T> used to guarantee that two objects of the same type are being compared against each other.
+   * @param info contains information about the assertion.
+   * @param actual the actual value.
+   * @param expected the expected value.
+   * @throws AssertionError if the actual value is {@code null}.
+   * @throws AssertionError if the actual value is not equal to the expected one. This method will throw a
+   *           {@code org.junit.ComparisonFailure} instead if JUnit is in the classpath and the expected and actual
```
joel-costigliola, line 950: likely an opentest4j `AssertionFailedError` but I don't think it's important, probably better removing the whole `This method will throw a....`  sentence

src/main/java/org/assertj/core/internal/Objects.java
```diff
@@ -932,6 +937,169 @@ public boolean areEqualToComparingOnlyGivenFields(Object actual, Object other,
     if (actual.hashCode() == other.hashCode()) throw failures.failure(info, shouldNotHaveSameHashCode(actual, other));
   }
 
+  /**
+   * Asserts that two <code>{@link Comparable}</code>s are equal by invoking
+   * <code>{@link Comparable#compareTo(Object)}</code>.<br>
```
joel-costigliola, line 942: Is it? or I guess we are going to use the register comparator since `T` is not extending `Comparable`

src/main/java/org/assertj/core/internal/Objects.java
```diff
@@ -932,6 +937,169 @@ public boolean areEqualToComparingOnlyGivenFields(Object actual, Object other,
...
+   * @param actual the actual value.
+   * @param other the value to compare the actual value to.
+   * @throws AssertionError if the actual value is {@code null}.
+   * @throws AssertionError if the actual value is greater than the other one.
+   */
+  public <T> void assertLessThanOrEqualTo(AssertionInfo info, T actual, T other) {
+    requireComparatorBasedComparisonStrategy(comparisonStrategy);
+    assertNotNull(info, actual);
+    if (!isGreaterThan(actual, other))
+      return;
```
joel-costigliola, line 1007: minor: join lines

src/main/java/org/assertj/core/internal/Objects.java
joel-costigliola: TBH, I'm not sure there is a lot of value having javadoc not methods and classes that are not exposed to users  (I certainly almost never read it and don't write it anymore).
If something is important to note, a regular comment will do.

scordio: I didn't work on javadoc at all in this PR, all you see is the result of copy-paste (and most likely wrong for all the methods).

I fully agree that internal methods don't need trivial javadoc. I'll clean them up.

src/main/java/org/assertj/core/internal/Objects.java
scordio: Same as the other comment, let's drop trivial javadoc for internal methods.

PR #2093, opened 2021-01-04
Fix typo in Javadoc

Bananeweizen:
#### Check List:
* Unit tests : NA


joel-costigliola:
Good catch @Bananeweizen!

PR #2106, opened 2021-01-27
Starting work for CollectionAssert

HaydenMeloche:
#### Check List:
* Fixes 2102
* Unit tests : not yet
* Javadoc with a code example (on API only) : not yet

Hoping I could get some initial overall feedback before I go to deep down the rabbit hole as well as help with my comment here:
https://github.com/assertj/assertj-core/issues/2102#issuecomment-767285633 @scordio 

Note: This PR is still in a rough state, I have not done any sort of documentation, unused imports are leftover etc. This will all be cleaned up by the time this PR is put in a ready state

### DIFF
--- a/src/main/java/org/assertj/core/internal/Collection.java
+++ b/src/main/java/org/assertj/core/internal/Collection.java
@@ -1,3 +1,15 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2021 the original author or authors.
+ */
 package org.assertj.core.internal;
 
 import org.assertj.core.api.AssertionInfo;
--- a/src/main/java/org/assertj/core/internal/Collection.java
+++ b/src/main/java/org/assertj/core/internal/Collection.java
@@ -1,0 +1,46 @@
+package org.assertj.core.internal;
+
+import org.assertj.core.api.AssertionInfo;
+import org.assertj.core.util.VisibleForTesting;
+
+import static org.assertj.core.api.Fail.failBecauseExceptionWasNotThrown;
+
+/**
+ * Reusable assertions for <code>{@link java.util.Collection}</code>
+ */
+public class Collection {
+
+  private static final Collection INSTANCE = new Collection();
+
+  @VisibleForTesting
+  Failures failures = Failures.instance();
+
+  /**
+   * Returns singleton instance of {@link java.util.Collection}
+   * @return the singleton instance of {@link Collection}
+   */
+  public static Collection instance() {
+    return INSTANCE;
+  }
+
+  public void assertIsUnmodifiable(AssertionInfo info, java.util.Collection<?> actual) {
+    assertNotNull(info, actual);
+    assertUnsupportedOperationExceptionIsThrown(info, actual.getClass(), () -> actual.add(null));
+  }
+
+  private void assertUnsupportedOperationExceptionIsThrown(AssertionInfo info, Class<?> clazz, Runnable runnable) {
+    try {
+      runnable.run();
+    } catch (Exception ex) {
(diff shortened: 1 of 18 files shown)


scordio:
Thank you @HaydenMeloche, I'll look into it today.

scordio:
> hey @scordio got a chance to working on this over the weekend. I setup my assertion class as such
> 
> ```java
> public abstract class AbstractCollectionAssert <SELF extends AbstractCollectionAssert<SELF, ACTUAL, ELEMENT, ELEMENT_ASSERT>,
>   ACTUAL extends java.util.Collection<? extends ELEMENT>,
>   ELEMENT,
>   ELEMENT_ASSERT extends AbstractAssert<ELEMENT_ASSERT, ELEMENT>>
>   extends AbstractIterableAssert<SELF, ACTUAL, ELEMENT, ELEMENT_ASSERT>
>   implements IndexedObjectEnumerableAssert<SELF, ELEMENT> {
> ```
> 
> I'm running into an issue now with this setup. Existing test cases that test the `Iterable` type are now looking for a Collection assert rather than a Iterator assert.
> 
> Ex: This test case:
> 
> https://github.com/assertj/assertj-core/blob/4fad9a03993e66fd4e2735352c22c52d206e9a1e/src/test/java/org/assertj/core/api/Assertions_assertThat_with_Iterable_Test.java#L30
> 
> Gives error:
> 
> ```
> Required type:
> AbstractIterableAssert
> <?,
> java.lang.Iterable<? extends java.lang.Object>,
> java.lang.Object,
> org.assertj.core.api.ObjectAssert<java.lang.Object>>
> Provided:
> CollectionAssert
> <Object,
> 
> 
> 
> no instance(s) of type variable(s) ELEMENT, T exist so that Collection<? extends ELEMENT> conforms to Iterable<?>
> ```
> 
> In [Assertions.java](https://github.com/assertj/assertj-core/blob/main/src/main/java/org/assertj/core/api/Assertions.java) I setup the collection assert as
> 
> ```java
>   public static <ELEMENT> CollectionAssert<ELEMENT> assertThat(Collection<? extends ELEMENT> actual) {
>     return AssertionsForInterfaceTypes.assertThat(actual);
>   }
> ```
> 
> Since `Collection` extends `Iterable` I understand why this is happening just not sure what the recommended action would be...
> any ideas/help would be appreciated :) I can push the code up to a branch if it's easier.

I think the easiest way for those tests is to declare a local variable, e.g.:

```java
  @Test
  void should_create_Assert() {
    Iterable<Object> actual = newLinkedHashSet();
    AbstractIterableAssert<?, Iterable<? extends Object>, Object, ObjectAssert<Object>> assertThat = Assertions.assertThat(actual);
    assertThat(assertThat).isNotNull();
  }
```

HaydenMeloche:
@scordio I was thinking I had set something up incorrectly somewhere else. Casting it to an `Iterable<>` works for me :)
Thanks!

HaydenMeloche:
@scordio When you have some time do you mind taking another look? I'm running into an issue where objects that extend `Collection` seem to not be able to find existing asserts... Not sure exactly what is causing the problem.

scordio:
@HaydenMeloche I imagine the issue happens only with lists, right? I tried the following on top of your branch:

```java
    assertThat(Collections.singleton("element")) // CollectionAssert
      .isUnmodifiable(); // exists
    assertThat(Collections.singletonList("element")) // ListAssert
      .isUnmodifiable(); // does not exist
```

Probably making `AbstractListAssert` extend from `AbstractCollectionAssert` instead of `AbstractIterableAssert` should solve it, without introducing breaking changes.

Later on, you might need to add a `ClassBasedNavigableCollectionAssert` and a `ProxyableCollectionAssert` (you could take inspiration from the corresponding iterable ones).

It might help to visualize the class hierarchy for this topic:

class-diagram

src/main/java/org/assertj/core/internal/Collection.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.assertj.core.api.AssertionInfo;
+import org.assertj.core.util.VisibleForTesting;
+
+import static org.assertj.core.api.Fail.failBecauseExceptionWasNotThrown;
+
+/**
+ * Reusable assertions for <code>{@link java.util.Collection}</code>
+ */
+public class Collection {
```
scordio, line 23: To follow the naming convention of similar classes, this should be `Collections`

scordio:
Hi @HaydenMeloche, I have only now noticed that the PR has been closed for several weeks. It is absolutely fine if you don't want to continue or don't have time to do so, but we are happy to help if you would like to return to this or any other topic at a later time. 

As we will participate in the upcoming hack-commit-push event, I will unassign the issue for now and mark it as a possible task for the event.

PR #2113, opened 2021-01-31
Add actual after Expecting to disambiguate error messages

utoniumharsha:
Check List:
Fixes nothing
Unit tests : Updated
Javadoc with a code example (on API only) : NA

### DIFF
--- a/src/main/java/org/assertj/core/error/ClassModifierShouldBe.java
+++ b/src/main/java/org/assertj/core/error/ClassModifierShouldBe.java
@@ -29,7 +29,7 @@ public class ClassModifierShouldBe extends BasicErrorMessageFactory {
   private static final String PACKAGE_PRIVATE = "package-private";
 
   private ClassModifierShouldBe(Class<?> actual, boolean positive, String modifier) {
-    super("%nExpecting:%n  %s%n" + (positive ? "to" : "not to") + " be a %s class but was %s.",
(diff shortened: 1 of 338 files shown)


src/main/java/org/assertj/core/error/ShouldHaveBinaryContent.java
```diff
@@ -56,17 +56,17 @@ public static ErrorMessageFactory shouldHaveBinaryContent(InputStream actual, Bi
   }
 
   private ShouldHaveBinaryContent(File actual, BinaryDiffResult diff) {
-    super("%nFile:%n  %s%ndoes not have expected binary content at offset %s, expecting:%n  %s%nbut was:%n  %s",
+    super("%nFile:%n  %s%ndoes not have expected binary content at offset %s, expecting actual:%n  %s%nbut was:%n  %s",
```
joel-costigliola, line 59: revert the 3 changes in this file, I realize adding `actual` does not improve the error messages in these cases

joel-costigliola (changes requested):
thanks, now you need to update the unit tests ;-)

src/main/java/org/assertj/core/error/ShouldHaveBinaryContent.java
utoniumharsha: Yes, agreed. Reverted changes for this file.

joel-costigliola:
Integrated thanks @utoniumharsha!

PR #2114, opened 2021-01-31
Rework `StandardComparisonStrategy#areEqual` to avoid shortcuts

### LINKED ISSUE #1949: #isEqualTo(Object other) takes invalid shortcut
#### Summary
Before PR #1767 (specifically commit [76762a3664b3af2101d580355a05c6f0b3aa46cc](../commit/76762a3664b3af2101d580355a05c6f0b3aa46cc)) an object x was not automatically considered equal to itself when using `BDDAssertions.assertThat(x).isEqualTo(x);`
By switching from 

```java
    if (o1 == null) {
      return o2 == null;
    }
    if (o1.equals(o2)) {
      return true;
    }
```
to calling `java.util.Objects.deepEquals(o1, o2);` we end up with
```java
    if (o1 == o2)
        return true;
    else if (o1 == null || o2 == null)
        return false;
    else
        // defaults to o1.equals(o2) if type is not an array
        return Arrays.deepEquals0(o1, 2);
```

This means when testing a custom equals implementation of a class, BDDAssertions doesn't actually do what we expect it to do and we may oversee bugs in the application code.

#### Example

```java
// fails with assertj 3.16.1; succeeds with 3.13.2; probably broken since 13.15.1
@Test
void testEquals() {
	Object x = new Object() {
		@Override
		public boolean equals(Object other) {
			return false;
		}
	};
	BDDAssertions.assertThat(x).isNotEqualTo(x);
}
```

scordio:
The previous behavior relied on `java.util.Objects#deepEquals`, which
performs a reference check before invoking `equals` on the first
parameter. While this it not an issue for proper implementations of
the `equals` contract, it was behaving incorrectly in case of
implementations not satisfying the reflexive property. Also, performing
the reference check outside of the `equals` implementation reduced the
overall coverage.

The tests now provide thorough regression to ensure that no shortcut is
taken, especially in case of implementations violating the contract.

#### Check List:
* Fixes #1949
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/internal/AtomicReferenceArrayElementComparisonStrategy.java
+++ b/src/main/java/org/assertj/core/internal/AtomicReferenceArrayElementComparisonStrategy.java
@@ -20,7 +20,7 @@ import java.util.concurrent.atomic.AtomicReferenceArray;
 
 public class AtomicReferenceArrayElementComparisonStrategy<T> extends StandardComparisonStrategy {
 
-  private Comparator<? super T> elementComparator;
+  private final Comparator<? super T> elementComparator;
 
   public AtomicReferenceArrayElementComparisonStrategy(Comparator<? super T> elementComparator) {
     this.elementComparator = elementComparator;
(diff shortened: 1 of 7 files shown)


scordio:
Test executions with:

* `java.util.Objects#deepEquals`

Screenshot 2021-01-31 at 15 48 02

* Proposed changes:

Screenshot 2021-01-31 at 15 49 16

joel-costigliola (approved):

PR #2122, opened 2021-02-10
Extract value only if the map key exists in PropertyOrFieldSupport

### LINKED ISSUE #2111: Assertions ignore missing entry on maps
#### Summary
A custom assertion extending `MapAssert` that performs a `org.assertj.core.api.AbstractObjectAssert.hasFieldOrProperty(String)` on an `actual` that neither has a getter or field, but _is_ a `Map` silently ignores missing map entries.

#### Example
Given JSON:
```json
{
  "startLevel" : 1,
  "initialBundleStartLevel" : 1
}
```
The following will silently succeed
```java
net.javacrumbs.jsonunit.assertj.JsonAssertions.assertThatJson(
    json
).isObject().hasFieldOrProperty(
    "lastModified"
);
```
#### Reason
The method `org.assertj.core.util.introspection.PropertyOrFieldSupport.getSimpleValue(String, Object)` throws in the case of propertySupport not having a getter method, and fieldSupport not finding a field. However in the case of the Map fallback it does not fail for a missing entry.
```
    // try to get name as a property, then try as a field, then try as a map key
    try {
      return propertySupport.propertyValueOf(name, Object.class, input);
    } catch (IntrospectionError propertyIntrospectionError) {
      // no luck as a property, let's try as a field
      try {
        return fieldSupport.fieldValue(name, Object.class, input);
      } catch (IntrospectionError fieldIntrospectionError) {
        // neither field nor property found with given name

        // if the input object is a map, try name as a map key
        if (input instanceof Map) {
          Map<?, ?> map = (Map<?, ?>) input;
          return map.get(name);
        }

        // no value found with given name, it is considered as an error
        String message = format("%nCan't find any field or property with name '%s'.%n" +
                                "Error when introspecting properties was :%n" +
                                "- %s %n" +
                                "Error when introspecting fields was :%n" +
                                "- %s",
                                name, propertyIntrospectionError.getMessage(),
                                fieldIntrospectionError.getMessage());
        throw new IntrospectionError(message, fieldIntrospectionError);
      }
    }
```

scordio:
#### Check List:
* Fixes #2111
* Unit tests : YES


scordio:
@joel-costigliola raising it as a draft just to sync on the strategy. If we like the approach, I'll use the opportunity for some additional BDD refactoring in the test classes.

src/test/java/org/assertj/core/api/map/MapAssert_flatExtracting_Test.java
```diff
@@ -45,6 +47,7 @@ void beforeEachTest() {
     map.put("job", jobs);
     map.put("city", cities);
     map.put("rank", ranks);
+    map.put("id", asList(null, null));
```
joel-costigliola, line 50: minor: could use `org.assertj.core.util.Lists.list`

joel-costigliola (approved):

PR #2125, opened 2021-02-18
Add isNotFinite and isNotInfinite assertions for float and double

### LINKED ISSUE #2120: Add isNotFinite and isNotInfinite for floating-point types (double, float)
#### Summary

`isFinite()` and `isInfinite()` assertions were handily added via #2005. In the comments on PR #2006, it looks like the original author believed `isNotFinite()` and `isNotInfinite()` were unnecessary, perhaps because they appear to be opposites.

However, `Double.isFinite()` and `Double.isInfinite()` (and corresponding `Float` methods) aren't exact opposites. `isFinite()` means:
* Not `POSITIVE_INFINITY`, _and_
* Not `NEGATIVE_INFINITY`, _and_
* Not `NaN`

`isInfinite()` means:
* `POSITIVE_INFINITY`, _or_
* `NEGATIVE_INFINITY`

While it's possible to achieve the same result as `isNotInfinite()` with `isNotEqualTo(Double.POSITIVE_INFINITY).isNotEqualTo(Double.NEGATIVE_INFINITY)`, that's a little cumbersome. 

`isNotFinite()` is much trickier to achieve--the best I've been able to do is `satisfiesAnyOf(...)`, but that's very cumbersome, indeed.

jcttrll:
#### Check List:
* Fixes #2120
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractDoubleAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractDoubleAssert.java
@@ -464,7 +464,7 @@ public abstract class AbstractDoubleAssert<SELF extends AbstractDoubleAssert<SEL
   public SELF isEqualTo(double expected) {
     if (noCustomComparatorSet()) {
       // use primitive comparison since the parameter is a primitive.
-      if (expected == actual.doubleValue()) return myself;
+      if (expected == actual) return myself;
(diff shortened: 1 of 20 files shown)


scordio:
Integrated, thanks @jcttrll!

PR #2126, opened 2021-02-20
Fix StackOverflowError when usingRecursiveComparison of Path on Windows

### LINKED ISSUE #1855: StackoverflowError when usingRecursiveComparison for an Object with a nio path
#### Summary
I get a StackoverflowError when usingRecursiveComparison for an Object which contains a nio path.

#### Example

```java

import java.nio.file.Path;
import java.nio.file.Paths;

public class AssertJBugReport {

    @Test
    public void pathThrowsStackoverflowError() {
        final Container container1 = new Container("/tmp/example");
        final Container container2 = new Container("/tmp/example");

        Assertions.assertThat(container1)
                .usingRecursiveComparison()
                .isEqualTo(container2);
    }

    public static class Container {
        private Path path;

        public Container(String path) {
            this.path = Paths.get(path);
        }

        public Path getPath() {
            return path;
        }
    }

}

```

VanRoy:
#### Check List:
* Fixes #1855
* Unit tests : YES 

### DIFF
--- a/src/main/java/org/assertj/core/util/NaturalOrderComparator.java
+++ b/src/main/java/org/assertj/core/util/NaturalOrderComparator.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2021 the original author or authors.
+ */
+package org.assertj.core.util;
+
+import java.nio.file.Path;
+import java.util.Comparator;
+
+import static java.lang.String.format;
+
+public class NaturalOrderComparator<T extends Comparable<? super T>> extends NullSafeComparator<T> {
+
+  private String description;
+
+  public NaturalOrderComparator(Class<T> clazz) {
+    this.description = format("lexicographic comparator (%s naturalComparator)", clazz.getSimpleName());
+  }
+
+  @Override
+  protected int compareNonNull(T o1, T o2) {
+    return Comparator.<T>naturalOrder().compare(o1, o2);
+  }
+
+  @Override
+  public String toString() {
+    return this.description;
+  }
+
+}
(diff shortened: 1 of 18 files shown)


src/test/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration_multiLineDescription_Test.java
```diff
@@ -331,6 +331,7 @@ void should_show_a_complete_multiline_description() {
                "  - java.lang.Double -> DoubleComparator[precision=1.0E-15]%n" +
                "  - java.lang.Float -> FloatComparator[precision=1.0E-6]%n" +
                "  - java.lang.Integer -> AbsValueComparator%n" +
+               "  - java.nio.file.Path -> INSTANCE%n" +
```
joel-costigliola, line 334: it would be better to have a description instead of `INSTANCE`, like `naturalComparator` or even better: `lexicographic comparator (Path naturalComparator)`

Thw way to go is creating a Comparator class in order to override `toString`  as in `FloatComparator`

joel-costigliola (changes requested):

src/test/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration_multiLineDescription_Test.java
VanRoy: Thanks, it should be better now.

src/main/java/org/assertj/core/util/NaturalOrderComparator.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Comparator;
+
+import static java.lang.String.format;
+
+public class NaturalOrderComparator<T extends Comparable<? super T>> extends NullSafeComparator<T> {
+
+  private String description;
+
+  public NaturalOrderComparator(Class<T> clazz) {
+    this.description = format("lexicographic comparator (%s naturalComparator)", clazz.getSimpleName());
```
scordio, line 25: What about the following?
```suggestion
    this.description = format("lexicographic comparator (%s natural order)", clazz.getSimpleName());
```

joel-costigliola: happy with that, @VanRoy I have started integrating your PR so I'll make the change (you'll get full credit of course ;-) )

joel-costigliola: one thing that I'm changing is removing `lexicographic ` from here since it's for Path only and subclassing `NaturalOrderComparator` with a `NaturalPathComparator`

VanRoy: Indeed, it's better like that. 
Thanks.

scordio:
Applied in 8912e7324fab207e2863891d499cd111eb10f129.

PR #2145, opened 2021-03-13
Maps.assertContainsOnly(info, actual, entries) has inconsistent behaviour if `entries` is empty

### LINKED ISSUE #2107: Maps.assertContainsOnly(info, actual, entries) has inconsistent behaviour if `entries` is empty
https://github.com/assertj/assertj-core/blob/f27b022a369ba731ed72413f6c10c82a9e539495/src/main/java/org/assertj/core/internal/Maps.java#L745
The call mentioned in the link fails with IllegalArgumentException if `entries` is empty, and `actual` is not. This, however, would work if `actual` is empty as well (the mentioned line is not executed in this case). Thus, execution success depends on `actual` input, which is the most frequently changing part of the test.
I would expect this code to fail with AssertionError with proper message (`actual should be empty, but was XXX`)

abhijeetshuklaoist:
#### Check List:
* Fixes #2107 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/internal/Maps.java
+++ b/src/main/java/org/assertj/core/internal/Maps.java
@@ -63,7 +63,6 @@ import java.util.function.Consumer;
 
 import org.assertj.core.api.AssertionInfo;
 import org.assertj.core.api.Condition;
-import org.assertj.core.error.ShouldBeEmpty;
 import org.assertj.core.error.ShouldContainAnyOf;
 import org.assertj.core.error.UnsatisfiedRequirement;
 import org.assertj.core.util.VisibleForTesting;
@@ -889,7 +888,7 @@ public class Maps {
   private <K, V> void doEntriesEmptyCheck(AssertionInfo info, Map<K, V> actual,
                                             Map.Entry<? extends K, ? extends V>[] entries) {
     if(entries.length == 0) {
-      throw failures.failure(info, ShouldBeEmpty.shouldBeEmpty(actual));
+      throw failures.failure(info, shouldBeEmpty(actual));
     }
   }
 
--- a/src/main/java/org/assertj/core/internal/Maps.java
+++ b/src/main/java/org/assertj/core/internal/Maps.java
@@ -63,6 +63,7 @@ import java.util.function.Consumer;
 
 import org.assertj.core.api.AssertionInfo;
 import org.assertj.core.api.Condition;
+import org.assertj.core.error.ShouldBeEmpty;
 import org.assertj.core.error.ShouldContainAnyOf;
 import org.assertj.core.error.UnsatisfiedRequirement;
 import org.assertj.core.util.VisibleForTesting;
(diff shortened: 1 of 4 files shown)


abhijeetshuklaoist:
@joel-costigliola Hey, I have raised this PR for #2107 
Please review it, and also I am not sure why one test is failing in CI job while same test is running fine on my local. I will try to debug it further. 

Thanks!

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -732,23 +732,26 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
...
 
     Set<Map.Entry<? extends K, ? extends V>> notFound = new LinkedHashSet<>();
     Set<Map.Entry<? extends K, ? extends V>> notExpected = new LinkedHashSet<>();
 
     compareActualMapAndExpectedEntries(actual, entries, notExpected, notFound);
 
+    if(!actual.isEmpty() && entries.length == 0) {
+      throw failures.failure(info, shouldContainOnly(actual, entries, notFound, notExpected));
+    }
+
```
joel-costigliola, line 754: please do the same for `assertContainsExactly` and move the code in a private method to avoid code duplication

joel-costigliola: you don't need to check that actual is not empty because if entries is empty, actual can't due to the check line 742

joel-costigliola: change the failure to use `ShouldBeEmpty.shouldBeEmpty` and move the `if` before creating `notFound`

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -774,7 +776,6 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
                                            @SuppressWarnings("unchecked") Map.Entry<? extends K, ? extends V>... entries) {
     doCommonContainsCheck(info, actual, entries);
     if (actual.isEmpty() && entries.length == 0) return;
-    failIfEmpty(entries);
```
joel-costigliola, line 777: apply the same check as commented before

joel-costigliola (changes requested):
@abhijeetshuklaoist thanks, there is a bit of work to do ;-)

src/main/java/org/assertj/core/internal/Maps.java
abhijeetshuklaoist: > please do the same for `assertContainsExactly` and move the code in a private method to avoid code duplication

In assertContainsExactly there is already a check to validate that the size of both actual and entries is same, which will take care of scenario where actual is not empty but entries is empty.  So doing the same thing again is of no significance. WDYT?
I have updated the test Maps_assertContainsExactly_Test. should_fail_if_given_entries_array_is_empty to test this scenario.

src/main/java/org/assertj/core/internal/Maps.java
abhijeetshuklaoist: Done

joel-costigliola:
Integrated thanks @abhijeetshuklaoist!

PR #2154, opened 2021-03-19
Support String assertions for InputStream

stefanbirkner:
In some of my tests I have to check InputStreams which contain text. I
cannot check the whole test with assertThat(...).hasContent(...) because
the text contains non-predictable parts (e.g. UUIDs).

This commit adds a new method asString(Charset) for InputStream
assertions. It returns an AbstractStringAssert so that all assertions
for String can be used for the InputStream. Here is an example

    assertThat(new ByteArrayInputStream("abc".getBytes()))
      .asString(UTF_8)
      .startWith("a")
      .endsWith("c");

The method asString(Charset) is similar to the one of ByteArrayAssert.
The test is for the most part a copy of
ByteArrayAssert_asString_with_charset_Test

#### Check List:
* Fixes #???
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractInputStreamAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractInputStreamAssert.java
@@ -12,12 +12,19 @@
  */
 package org.assertj.core.api;
 
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.io.InputStream;
+import java.nio.charset.Charset;
 import java.security.MessageDigest;
 
 import org.assertj.core.internal.InputStreams;
 import org.assertj.core.internal.InputStreamsException;
+import org.assertj.core.util.CheckReturnValue;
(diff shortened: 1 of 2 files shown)


src/main/java/org/assertj/core/api/AbstractInputStreamAssert.java
```diff
@@ -298,4 +335,19 @@ public SELF hasDigest(String algorithm, String expected) {
     inputStreams.assertHasDigest(info, actual, algorithm, expected);
     return myself;
   }
+
+  private String readString(Charset charset) {
```
stefanbirkner, line 339: I didn't add this method to InputStream because InputStreams is a collection of assertions and this is not an assertion.
I added the method to the bottom so that all public methods are on top.

joel-costigliola:
Thanks, we will review your PR shortly!

joel-costigliola:
Integrated thanks @stefanbirkner!

PR #2156, opened 2021-03-28
Added new assertions for URI and URL for exhaustive parameters check

### LINKED ISSUE #2153: Comprehensively checking url query params
#### Summary
With assertThat on URL object isn't possible check all parameters in one method call. For example is possible check whole map by insert expected map. It will be very usefull becouse sometimes depends on exactly params and api not able validate there aren't more params...

#### Example
```java
        // for map is it possible by
       assertThat(Map.of()).containsExactlyInAnyOrderEntriesOf(Map.of(
                "k1", "v1"
        ));

       // for url maybe something like this
       assertThat(new URL(url)).hasExactlyParams(Map.of(
                "qK1", "qV1"
        ))
```

schrek1:
Added new assertions for URI and URL for exhaustive parameters check

#### Check List:
* Fixes #2153
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractUriAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractUriAssert.java
@@ -12,10 +12,13 @@
  */
 package org.assertj.core.api;
 
-import java.net.URI;
-
 import org.assertj.core.internal.Uris;
 import org.assertj.core.util.VisibleForTesting;
+
+import java.net.URI;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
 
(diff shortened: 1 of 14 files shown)


joel-costigliola:
Thanks, we will review your PR shortly (hopefully)!

scordio:
Hi @schrek1, at the moment the team tends to favor the `getParameters()` approach I was describing in https://github.com/assertj/assertj-core/issues/2153#issuecomment-801949339, mainly to avoid duplications and reduce maintenance.

Would this also fit you?

PR #2160, opened 2021-04-06
Do not create new map when testing for containsOnly and containsOnlyKeys for maps

### LINKED ISSUE #2159: Map `containsOnly` assertions don't work correctly with case insensitive keys
#### Summary

When using case insensitive maps the map assertions for containsOnly are not working correctly.
The reason for this is due to the fact that the actual map is copied into a `LinkedHashMap` and thus the checks are not using the map.

#### Example

```java
Map<String, String> map = new LinkedCaseInsensitiveMap<>();
map.put("TEST", "value");
```

The following assertions will pass
```java
assertThat(map).containsKey("test");
assertThat(map).contains(entry("test", "value"));
```

The following assertions will fail
```java
assertThat(map).containsOnlyKeys("test");
assertThat(map).containsOnly(entry("test", "value"));
```

Note: `LinkedCaseInsensitiveMap` is from Spring.

If you agree I can work on a PR to change the mechanism that is used for validating `containsOnly` entries.

filiphr:
#### Check List:
* Fixes #2159
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/internal/Maps.java
+++ b/src/main/java/org/assertj/core/internal/Maps.java
@@ -53,6 +53,8 @@ import static org.assertj.core.util.Objects.areEqual;
 import static org.assertj.core.util.Preconditions.checkArgument;
 
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -796,9 +798,10 @@ public class Maps {
     if (notExpected.isEmpty() && notFound.isEmpty()) {
       // check entries order
       int index = 0;
-      for (K keyFromActual : actual.keySet()) {
-        if (!areEqual(keyFromActual, entries[index].getKey())) {
-          Map.Entry<K, V> actualEntry = entry(keyFromActual, actual.get(keyFromActual));
+      for (Map.Entry<K, V> entryFromActual : actual.entrySet()) {
+        V value = actual.get(entries[index].getKey());
+        if (!areEqual(value, entryFromActual.getValue())) {
+          Map.Entry<K, V> actualEntry = entry(entryFromActual.getKey(), entryFromActual.getValue());
           throw failures.failure(info, elementsDifferAtIndex(actualEntry, entries[index], index));
         }
         index++;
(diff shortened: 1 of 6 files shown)


joel-costigliola:
We have a similar issue for `assertContainsExactly` when it checks that the entries are in the same order and compares keys with `equals`.
@filiphr would you like to fix that too? no problem if you can't, I'm happy to do it.

filiphr:
I didn't have much ideas about the `containsExactly` one, but I just got an idea. I'll try something out shortly

filiphr:
@joel-costigliola I've adapted the PR to include the fix for the `assertContainsExactly` problem as well.

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -796,9 +798,10 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
     if (notExpected.isEmpty() && notFound.isEmpty()) {
       // check entries order
       int index = 0;
-      for (K keyFromActual : actual.keySet()) {
-        if (!areEqual(keyFromActual, entries[index].getKey())) {
-          Map.Entry<K, V> actualEntry = entry(keyFromActual, actual.get(keyFromActual));
+      for (Map.Entry<K, V> entryFromActual : actual.entrySet()) {
+        V value = actual.get(entries[index].getKey());
+        if (!areEqual(value, entryFromActual.getValue())) {
```
joel-costigliola, line 803: comparing only the values is not enough, the following test should fail but succeeds with this implementation because all the keys have the same value:
```java
Map<String, String> actual = new LinkedHashMap<>(2);
actual.put("a", "1");
actual.put("b", "1");
maps.assertContainsExactly(INFO, actual, entry("b", "1"), entry("a", "1"));
```
We could solve the problem if we were able to use the actual map way to compare keys to compare `actualKeys[i]` to `entries[i].key` but I'm not sure this is possible.

To be pragmatic I have created https://github.com/assertj/assertj-core/issues/2165 and I will remove the `containsExactly` changes from this PR when integrating (don't bother pushing a commit for that)

filiphr: You are right, I didn't think about that use case. Thanks for removing the changes in the contains exactly method.

filiphr: btw, it would be good to add the test case you shared in this commit to the codebase :)

joel-costigliola: good idea, I'll add it to the other tests

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -813,38 +816,72 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
...
-    notExpected.addAll(actualEntries.keySet());
+
+    // We now iterate through the actual entries in order to detect the unexpected ones
+    for (Map.Entry<K, V> entry : actual.entrySet()) {
+      boolean keyNotExpected = true;
+      Iterator<K> foundKeysIterator = foundKeys.iterator();
+      while (foundKeysIterator.hasNext()) {
+        K foundKey = foundKeysIterator.next();
+        V foundValue = actual.get(foundKey);
+        if (areEqual(foundValue, entry.getValue())) {
```
joel-costigliola, line 838: I'm not sure to see why we need to compare values when the assertion is only checking keys, am I missing something here?

joel-costigliola: I think we have the same issue as with `containsExactly` since we have to compare keys and we can't access actual's map key comparison strategy.

This test fails as it should but it does not report the correct error:
```java
    actual.put("NAME", "green");
    actual.put("cool", "green");
    // THEN
    maps.assertContainsOnlyKeys(someInfo(), actual, "Name", "Color");
```
error is:
```
Expecting actual:
  {"cool"="green", "name"="green"}
to contain only following keys:
  ["Name", "Color"]
keys not found:
  ["Color"]
and keys not expected:
  ["name"]
```
"keys not found" is correctly reported but "keys not expected" should have shown "cool", it reports "name" instead which actually is a map key.

joel-costigliola: I guess the only way forward is to find a way to clone actual's map which might be possible using introspection to call a a no arg constructor and then populate the cloned map. Question is now: is there always a no arg constructor ...

filiphr: You are right. What we can also do is to support sime well known case insensitive maps behind the scenes, and for everything else create a normal copy. Reusing constructors might lead to using unmodifiable / immutable maps. I can give it another go on Sunday afternoon, no access to a laptop before that

scordio: > Question is now: is there always a no arg constructor ...

[Objenesis](http://objenesis.org/) could deal with it and I think it would be fine to have it as a shaded dependency

joel-costigliola: I was thinking we could try to instantiate the map (I'm fine using Objenesis) and if it fails for whatever reason use a simple `LinkedHashMap` as in the existing implementation, that won't completely solve the issue but will work for a vast majority of use cases, finally we would mention this limitation in the javadoc.

joel-costigliola: I have tried Objenesis and that did not go well with `CaseInsensitiveMap`, creating the instance worked but would not initialize `transient HashEntry<K, V>[] data;` which is accessed  later on when putting actual entries in the new map.

Interestingly it works with `ConstructorInvoker` (some old internal class in our codebase) and it worked.

Here's the stack trace for the record:
```
	at org.apache.commons.collections4.map.AbstractHashedMap.ensureCapacity(AbstractHashedMap.java:627)
	at org.apache.commons.collections4.map.AbstractHashedMap._putAll(AbstractHashedMap.java:325)
	at org.apache.commons.collections4.map.AbstractHashedMap.putAll(AbstractHashedMap.java:304)
	at org.assertj.core.internal.Maps.compareActualMapAndExpectedKeys(Maps.java:823)
```

joel-costigliola:
I think we should close this PR in favor of https://github.com/assertj/assertj-core/pull/2167 which tries the approach with cloning the actual map

joel-costigliola:
closing this PR in favor of #2167

PR #2162, opened 2021-04-06
Prevent compiler warning when using extracting with SoftAssertions

### LINKED ISSUE #2161: Using softAssertions with extracting leads to warning related to unchecked generics array creation for varargs parameter
#### Summary

When I use `SoftAssertions` combined with `extracting`, my IDE complains with the following warning

> Unchecked generics array creation for varargs parameter

When I use the same code with standard `Assertions`, the warning is not there.

#### Example

With warning

```java
    TolkienCharacter legolas = TolkienCharacter.of("Legolas", 1000, ELF);

    softly.assertThat(legolas)
          .extracting(TolkienCharacter::getName, TolkienCharacter::getAge)
          .contains("Legolas", 1000);
```

Without warning

```java
    TolkienCharacter legolas = TolkienCharacter.of("Legolas", 1000, ELF);

    assertThat(legolas)
          .extracting(TolkienCharacter::getName, TolkienCharacter::getAge)
          .contains("Legolas", 1000);
```

jgiovaresco:
I'm not sure about the "fix". 
However, it seems we use the same way in ObjectAssert class: 

https://github.com/assertj/assertj-core/blob/14cd17293c62376abc8c762084819a1e08db1520/src/main/java/org/assertj/core/api/ObjectAssert.java#L43-L48

Because `SoftAssertions` (`Java6StandardSoftAssertionsProvider`) is building a `ProxyableObjectAssert` I added an override on the `extracting` method to be able to use the annotation `@SafeVarargs` on it.

#### Check List:
* Fixes #2161
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
@@ -774,11 +774,9 @@ public abstract class AbstractObjectAssert<SELF extends AbstractObjectAssert<SEL
    * @throws IntrospectionError if one of the given name does not match a field or property
    */
   @CheckReturnValue
-  public AbstractListAssert<?, List<?>, Object, ObjectAssert<Object>> extracting(String... propertiesOrFields) {
-    Tuple values = byName(propertiesOrFields).apply(actual);
-    String extractedPropertiesOrFieldsDescription = extractedDescriptionOf(propertiesOrFields);
-    String description = mostRelevantDescription(info.description(), extractedPropertiesOrFieldsDescription);
-    return newListAssertInstance(values.toList()).withAssertionState(myself).as(description);
+  @SafeVarargs
+  public final AbstractListAssert<?, List<?>, Object, ObjectAssert<Object>> extracting(String... propertiesOrFields) {
+    return extractingForProxy(propertiesOrFields);
   }
 
   /**
@@ -1168,6 +1166,13 @@ public abstract class AbstractObjectAssert<SELF extends AbstractObjectAssert<SEL
   // override for proxyable friendly AbstractObjectAssert
   protected <T> AbstractObjectAssert<?, T> newObjectAssert(T objectUnderTest) {
     return new ObjectAssert<>(objectUnderTest);
+  }
+
+  protected AbstractListAssert<?, List<?>, Object, ObjectAssert<Object>> extractingForProxy(String[] propertiesOrFields) {
+    Tuple values = byName(propertiesOrFields).apply(actual);
+    String extractedPropertiesOrFieldsDescription = extractedDescriptionOf(propertiesOrFields);
+    String description = mostRelevantDescription(info.description(), extractedPropertiesOrFieldsDescription);
+    return newListAssertInstance(values.toList()).withAssertionState(myself).as(description);
   }
 
   @SuppressWarnings({ "rawtypes", "unchecked" })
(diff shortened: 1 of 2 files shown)


jgiovaresco:
It seems that broke something 💥 . I will check that later.

joel-costigliola:
We haven't found a good solution to get rid of these annoying warnings.

Soft assertions implementation is based on proxying and collecting the assertion errors, final method can't be proxied as bytebuddy can't replace (AFAIK) final methods or classes. Now `@SafeVarargs` requires methods to be `final` so we can't use it in the soft assertion context, we used it for regular assertions as you can see in `ObjectAssert` but not for soft assertions that use `ProxyableObjectAssert`.

One might think that `extracting` does not need to be proxied since it does not perform any assertions but it needs to, `extracting` returns an assertion instance (`AbstractListAssert`) that needs to be proxied so that the chained assertions error are captured.

If only `@SafeVarargs` was allowed on non final methods, it would be usable for soft assertions and we would have had to introduce `ProxyableObjectAssert` and co, this is a major pain in the ass for me just to get (partially) rid of these warnings.

filiphr:
Reading these comments I got a potential idea. What if we mark the method as `public final` and annotate it with `@SafeVarargs` in the `AbstractListAssert` and then delegate that to a new method:

```
protected AbstractListAssert<?, List<?>, Object, ObjectAssert<Object>> extractingForProxy(Function<? super ACTUAL, ?>... extractors) { 
   // Do the same logic as before
 } 
```

I think that ByteBuddy can then proxy the `extractingForProxy` method in the class that it creates. There will of course be warnings in the AssertJ codebase, but not in the user code.

We could even do it generically in the place where we create the proxies by looking for a method names `<methodName>ForProx` if the `<methodName>` is annotated with a `@SafeVarargs`

joel-costigliola:
That's an interesting idea, worth a shot! thanks @filiphr

jgiovaresco:
I've tried to implement @filiphr idea. It seems not working but I'm not sure that I understood everything 😅 

I've extracted `extractingForProxy` protected method in `AbstractObjectAssert` to be able to mark the public method `extracting` as final. 
However, the same test is failing 💥  I suppose it is related to ByteBuddy not able to proxy the protected method 🤔

filiphr:
@jgiovaresco have a look in the `SoftProxies` [here](https://github.com/assertj/assertj-core/blob/7fab7ecc29cc9c61a91409542f83522653a0b56e/src/main/java/org/assertj/core/api/SoftProxies.java#L66). I think that you need to tell ByteBuddy not to proxy methods annotated with `@SafeVarargs`, or better not final methods. Not sure whether ByteBuddy handles that automatically or not.

filiphr:
Actually, the problem is in something else. Look at [this](https://github.com/assertj/assertj-core/blob/7fab7ecc29cc9c61a91409542f83522653a0b56e/src/main/java/org/assertj/core/api/SoftProxies.java#L42). There is something special happening for the `extracting` method. I think that you need to add the `extractingForProxy` method there as well, and also make sure that protected such methods are also handled.

You'll need to play with the ByteBuddy interceptors

filiphr:
I had some spare time and went ahead and did https://github.com/assertj/assertj-core/pull/2163. That should cover all possible compiler warnings in user code. Sorry for taking it over @jgiovaresco. As I mentioned in the PR, I am completely fine in giving the entire contribution to you.

PR #2166, opened 2021-04-11
Fix NullPointerException in primitive double assertions

goughy000:
#### Check List:
* Fixes NullPointerException in AbstractDoubleAssert
* Unit tests : YES

The Javadoc for the `isEqualTo(double)` and `isNotEqualTo(double)` (with primitive parameter) states that an `AssertionError` will be thrown if `actual == null`, however it was actually throwing a `NullPointerException` when casting `actual` to primitive in the `==` and `!=` checks.

Javadoc:
```
* @throws AssertionError if the actual value is {@code null}.
```

Reproduction:
```
assertThat((Double)null).isEqualTo(5.0d);
```

Threw NPE here when it tried to cast boxed null actual to primitive double
```
if (expected == actual) return myself;
                ^-- npe
```

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractDoubleAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractDoubleAssert.java
@@ -42,6 +42,7 @@ import org.assertj.core.util.VisibleForTesting;
  * @author Joel Costigliola
  * @author Mikhail Mazursky
  * @author Nicolas François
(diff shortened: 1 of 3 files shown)


joel-costigliola:
Integrated thanks @goughy000!

goughy000:
No problem - thanks for merge @joel-costigliola :)

PR #2167, opened 2021-04-11
Honor map key comparison semantics with `containsOnly` assertions

### LINKED ISSUE #2159: Map `containsOnly` assertions don't work correctly with case insensitive keys

joel-costigliola:
This affects:
- `assertContainsOnly`
- `assertContainsOnlyKeys`

#### Open points:
- [x] Complete `Maps_assertContainsOnlyKeys_Test`
- [x] Complete `Maps_assertContainsOnly_Test`
- [x] Evaluate refactoring of `assertDoesNotContainKeys(AssertionInfo, Map, Object[])`
- [x] Mention in the javadoc that in some cases the key comparison semantics won't be honored

#### Check List:
* Fixes #2159
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractMapAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractMapAssert.java
@@ -1138,7 +1138,7 @@ public abstract class AbstractMapAssert<SELF extends AbstractMapAssert<SELF, ACT
    * Verifies that the actual map contains only the given keys and nothing else, in any order.
    * <p>
    * The verification tries to honor the key comparison semantic of the underlying map implementation, but there might
-   * be some cases where the semantic cannot be honored and not all the unexpected elements are identified  correctly
+   * be some cases where the semantic cannot be honored and not all the unexpected elements are identified correctly
    * (e.g., due to immutability).
    * <p>
    * Examples :
@@ -1176,7 +1176,10 @@ public abstract class AbstractMapAssert<SELF extends AbstractMapAssert<SELF, ACT
 
   /**
    * Verifies that the actual map contains only the given keys and nothing else, in any order.
-   *
+   * <p>
+   * The verification tries to honor the key comparison semantic of the underlying map implementation, but there might
+   * be some cases where the semantic cannot be honored and not all the unexpected elements are identified correctly
+   * (e.g., due to immutability).
    * <p>
    * Examples :
    * <pre><code class='java'> Map&lt;Ring, TolkienCharacter&gt; ringBearers = new HashMap&lt;&gt;();
@@ -1303,7 +1306,7 @@ public abstract class AbstractMapAssert<SELF extends AbstractMapAssert<SELF, ACT
    * Verifies that the actual map contains only the given entries and nothing else, in any order.
    * <p>
    * The verification tries to honor the key comparison semantic of the underlying map implementation, but there might
-   * be some cases where the semantic cannot be honored and not all the unexpected elements are identified  correctly
+   * be some cases where the semantic cannot be honored and not all the unexpected elements are identified correctly
    * (e.g., due to immutability).
    * <p>
    * Examples :
--- a/src/main/java/org/assertj/core/api/AbstractMapAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractMapAssert.java
@@ -1136,7 +1136,10 @@ public abstract class AbstractMapAssert<SELF extends AbstractMapAssert<SELF, ACT
 
   /**
    * Verifies that the actual map contains only the given keys and nothing else, in any order.
-   *
+   * <p>
+   * The verification tries to honor the key comparison semantic of the underlying map implementation, but there might
+   * be some cases where the semantic cannot be honored and not all the unexpected elements are identified  correctly
+   * (e.g., due to immutability).
    * <p>
    * Examples :
    * <pre><code class='java'> Map&lt;Ring, TolkienCharacter&gt; ringBearers = new HashMap&lt;&gt;();
@@ -1298,7 +1301,10 @@ public abstract class AbstractMapAssert<SELF extends AbstractMapAssert<SELF, ACT
 
   /**
    * Verifies that the actual map contains only the given entries and nothing else, in any order.
-   *
+   * <p>
+   * The verification tries to honor the key comparison semantic of the underlying map implementation, but there might
+   * be some cases where the semantic cannot be honored and not all the unexpected elements are identified  correctly
+   * (e.g., due to immutability).
    * <p>
    * Examples :
    * <pre><code class='java'> Map&lt;Ring, TolkienCharacter&gt; ringBearers = new HashMap&lt;&gt;();
--- a/src/main/java/org/assertj/core/api/AbstractMapAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractMapAssert.java
@@ -1137,9 +1137,10 @@ public abstract class AbstractMapAssert<SELF extends AbstractMapAssert<SELF, ACT
   /**
    * Verifies that the actual map contains only the given keys and nothing else, in any order.
    * <p>
-   * The verification tries to honor the key comparison semantic of the underlying map implementation, but there might
-   * be some cases where the semantic cannot be honored and not all the unexpected elements are identified correctly
-   * (e.g., due to immutability).
+   * The verification tries to honor the key comparison semantic of the underlying map implementation.
+   * The map under test has to be cloned to identify unexpected elements, but depending on the map implementation
+   * this may not always be possible. In case it is not possible, a regular map is used and the key comparison strategy
+   * may not be the same as the map under test.
    * <p>
    * Examples :
    * <pre><code class='java'> Map&lt;Ring, TolkienCharacter&gt; ringBearers = new HashMap&lt;&gt;();
@@ -1177,9 +1178,10 @@ public abstract class AbstractMapAssert<SELF extends AbstractMapAssert<SELF, ACT
   /**
    * Verifies that the actual map contains only the given keys and nothing else, in any order.
    * <p>
-   * The verification tries to honor the key comparison semantic of the underlying map implementation, but there might
-   * be some cases where the semantic cannot be honored and not all the unexpected elements are identified correctly
-   * (e.g., due to immutability).
+   * The verification tries to honor the key comparison semantic of the underlying map implementation.
+   * The map under test has to be cloned to identify unexpected elements, but depending on the map implementation
+   * this may not always be possible. In case it is not possible, a regular map is used and the key comparison strategy
+   * may not be the same as the map under test.
    * <p>
    * Examples :
    * <pre><code class='java'> Map&lt;Ring, TolkienCharacter&gt; ringBearers = new HashMap&lt;&gt;();
@@ -1305,9 +1307,10 @@ public abstract class AbstractMapAssert<SELF extends AbstractMapAssert<SELF, ACT
   /**
    * Verifies that the actual map contains only the given entries and nothing else, in any order.
    * <p>
-   * The verification tries to honor the key comparison semantic of the underlying map implementation, but there might
-   * be some cases where the semantic cannot be honored and not all the unexpected elements are identified correctly
-   * (e.g., due to immutability).
+   * The verification tries to honor the key comparison semantic of the underlying map implementation.
+   * The map under test has to be cloned to identify unexpected elements, but depending on the map implementation
+   * this may not always be possible. In case it is not possible, a regular map is used and the key comparison strategy
+   * may not be the same as the map under test.
    * <p>
    * Examples :
    * <pre><code class='java'> Map&lt;Ring, TolkienCharacter&gt; ringBearers = new HashMap&lt;&gt;();
--- a/src/main/java/org/assertj/core/internal/Maps.java
+++ b/src/main/java/org/assertj/core/internal/Maps.java
@@ -66,11 +66,10 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import org.assertj.core.api.AssertionInfo;
 import org.assertj.core.api.Condition;
+import org.assertj.core.data.MapEntry;
 import org.assertj.core.error.UnsatisfiedRequirement;
 import org.assertj.core.util.VisibleForTesting;
 
@@ -377,22 +376,29 @@ public class Maps {
       throw failures.failure(info, shouldContainOnlyKeys(actual, keys, notFound, notExpected));
   }
 
-  private static <K> Set<K> getNotFoundKeys(Map<K, ?> actual, K[] keys) {
-    return Stream.of(keys)
-                 .filter(key -> !actual.containsKey(key))
-                 .collect(Collectors.toCollection(LinkedHashSet::new));
+  private static <K> Set<K> getNotFoundKeys(Map<K, ?> actual, K[] expectedKeys) {
+    Set<K> notFound = new LinkedHashSet<>();
+    for (K expectedKey : expectedKeys) {
+      if (!actual.containsKey(expectedKey)) {
+        notFound.add(expectedKey);
+      }
+    }
+    return notFound;
   }
 
-  private static <K> Set<K> getNotExpectedKeys(Map<K, ?> actual, K[] keys) {
-    List<K> expectedKeys = java.util.Arrays.asList(keys);
+  private static <K> Set<K> getNotExpectedKeys(Map<K, ?> actual, K[] expectedKeys) {
     try {
       Map<K, ?> clonedMap = clone(actual);
-      expectedKeys.forEach(clonedMap::remove);
+      for (K expectedKey : expectedKeys) {
+        clonedMap.remove(expectedKey);
+      }
       return clonedMap.keySet();
     } catch (NoSuchMethodException | UnsupportedOperationException e) {
       // actual cannot be cloned or is unmodifiable, falling back to LinkedHashMap
       Map<K, ?> copiedMap = new LinkedHashMap<>(actual);
-      expectedKeys.forEach(copiedMap::remove);
+      for (K expectedKey : expectedKeys) {
+        copiedMap.remove(expectedKey);
+      }
       return copiedMap.keySet();
     }
   }
@@ -453,28 +459,37 @@ public class Maps {
 
   private static <K, V> Set<Entry<? extends K, ? extends V>> getNotFoundEntries(Map<K, V> actual,
                                                                                 Entry<? extends K, ? extends V>[] entries) {
-    return Stream.of(entries)
-                 .filter(entry -> !containsEntry(actual, entry))
-                 .collect(Collectors.toCollection(LinkedHashSet::new));
+    Set<Entry<? extends K, ? extends V>> notFound = new LinkedHashSet<>();
+    for (Entry<? extends K, ? extends V> entry : entries) {
+      if (!containsEntry(actual, entry)) {
+        notFound.add(entry);
+      }
+    }
+    return notFound;
   }
 
   private static <K, V> Set<Entry<K, V>> getNotExpectedEntries(Map<K, V> actual, Entry<? extends K, ? extends V>[] entries) {
-    return mapWithoutExpectedEntries(actual, entries).entrySet()
-                                                     .stream()
-                                                     .map(entry -> entry(entry.getKey(), entry.getValue()))
-                                                     .collect(Collectors.toCollection(LinkedHashSet::new));
+    Set<Entry<K, V>> notExpected = new LinkedHashSet<>();
+    for (Entry<K, V> entry : mapWithoutExpectedEntries(actual, entries).entrySet()) {
+      MapEntry<K, V> mapEntry = entry(entry.getKey(), entry.getValue());
+      notExpected.add(mapEntry);
+    }
+    return notExpected;
   }
 
-  private static <K, V> Map<K, V> mapWithoutExpectedEntries(Map<K, V> actual, Entry<? extends K, ? extends V>[] entries) {
-    List<Entry<? extends K, ? extends V>> expectedEntries = java.util.Arrays.asList(entries);
+  private static <K, V> Map<K, V> mapWithoutExpectedEntries(Map<K, V> actual, Entry<? extends K, ? extends V>[] expectedEntries) {
     try {
       Map<K, V> clonedMap = clone(actual);
-      expectedEntries.forEach(entry -> clonedMap.remove(entry.getKey(), entry.getValue()));
+      for (Entry<? extends K, ? extends V> expectedEntry : expectedEntries) {
+        clonedMap.remove(expectedEntry.getKey(), expectedEntry.getValue());
+      }
       return clonedMap;
     } catch (NoSuchMethodException | UnsupportedOperationException e) {
       // actual cannot be cloned or is unmodifiable, falling back to LinkedHashMap
       Map<K, V> copiedMap = new LinkedHashMap<>(actual);
-      expectedEntries.forEach(entry -> copiedMap.remove(entry.getKey(), entry.getValue()));
+      for (Entry<? extends K, ? extends V> expectedEntry : expectedEntries) {
+        copiedMap.remove(expectedEntry.getKey(), expectedEntry.getValue());
+      }
       return copiedMap;
     }
   }
--- a/src/main/java/org/assertj/core/internal/Maps.java
+++ b/src/main/java/org/assertj/core/internal/Maps.java
@@ -494,11 +494,12 @@ public class Maps {
     doCommonContainsCheck(info, actual, entries);
     if (actual.isEmpty() && entries.length == 0) return;
     failIfEntriesIsEmptyEmptySinceActualIsNotEmpty(info, actual, entries);
-
-    Set<Entry<? extends K, ? extends V>> notFound = getNotFoundEntries(actual, entries);
-    Set<Entry<K, V>> notExpected = getNotExpectedEntries(actual, entries);
-
     assertHasSameSizeAs(info, actual, entries);
+
+    Set<Entry<? extends K, ? extends V>> notFound = new LinkedHashSet<>();
+    Set<Entry<? extends K, ? extends V>> notExpected = new LinkedHashSet<>();
+
+    compareActualMapAndExpectedEntries(actual, entries, notExpected, notFound);
 
     if (notExpected.isEmpty() && notFound.isEmpty()) {
       // check entries order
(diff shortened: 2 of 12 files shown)


src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -827,12 +833,24 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
     notExpected.addAll(actualEntries.keySet());
   }
 
+  private <V, K> Map<K, V> instantiate(Class<? extends Map> mapClass) {
+    try {
+      Constructor<?> constructor = mapClass.getConstructor();
+      return (Map<K, V>) constructor.newInstance();
+      // return objenesis.newInstance(class1);
```
joel-costigliola, line 840: I have setup objenesis, uncomment to see the error we have when using it

joel-costigliola:
@filiphr feel free to review and contribute to this PR if you have time, this one supersedes https://github.com/assertj/assertj-core/pull/2160.

joel-costigliola:
The build fails due to too many lib to shade, not a blocker right now for us to play with objenesis

src/main/java/org/assertj/core/internal/Maps.java
joel-costigliola: We should create an `InstantiatorStrategy` with different implementation instead of this basic approach.

filiphr:
I am not sure that using `Objenesis` or even creating a copy of the map should be the way forward. At least not for all maps.

Have a look at my additional test cases in https://github.com/filiphr/assertj-core/commit/96360d786af5948a887ab23a2ac92749921752b2. Using `Collections.unmodifiableMap` works, using Guava immutable map works as well. Using `SingletonMap` doesn't work.

A potential (not 100% cases solution) would be to only create a copy if the map is a case insensitive map. e.g. like in https://github.com/filiphr/assertj-core/commit/d7a77b9db070797a12e744ad7a2b687b606ad60b.

joel-costigliola:
thanks for fixing the pom @scordio.

src/main/java/org/assertj/core/internal/Maps.java
goughy000: Just a thought - perhaps trying a `clone()` if the class implements `Cloneable` could be one of the implementations of `InstantiatorStrategy`.  `java.util.HashMap` implements this, so it's likely to be common

scordio:
I try to summarize the topic to see if I got it correctly.

The `containsOnly` / `containsExactly` assertions consist of two steps:
1. Checking if the expected elements are present
2. Finding the unexpected ones

Elements can be either keys or entries depending on the specific assertion.

### Checking the expected elements

The check should always be performed on `actual` to honor the original semantic.

This is mostly relevant with special implementations like `org.apache.commons.collections4.map.CaseInsensitiveMap` or `org.springframework.util.LinkedCaseInsensitiveMap`.

### Finding of unexpected elements

In the ideal case, we want to have a copy of the map and remove the expected elements, so what is left are the unexpected elements.

Being able to do this strongly depends on the map implementation.

1. #### `Cloneable` implementation
    A new instance can be obtained via `clone()`, then we try to `remove()` the expected elements
    * `remove` succeeds &#8594; unexpected elements are identified
    * `remove` fails with `UnsupportedOperationException`, (i.e., unmodifiable) &#8594; new `LinkedHashMap` initialized with the elements of `actual`, `remove()` is repeated again &#8594;  unexpected elements are identified but with the potential loss of the original semantic.

2. #### Non `Cloneable` implementation, but a new instance can be obtained
    A new instance can be obtained via reflection (e.g., plain, Objenesis), then we try to `putAll()` the `actual` elements
    * `putAll()` succeeds &#8594; we try to `remove()` the expected elements as in the `Cloneable` case.
    * `putAll()` fails with `UnsupportedOperationException`, (i.e., unmodifiable) &#8594; new `LinkedHashMap` is initialized with the elements of `actual`, `remove()` is repeated again &#8594;  unexpected elements are identified but with the potential loss of the original semantic.

3. #### No new instance can be obtained
    New `LinkedHashMap` is initialized with the elements of `actual`, the `remove()` of the expected elements is performed &#8594;  unexpected elements are identified but with the potential loss of the original semantic.

@joel-costigliola @filiphr what do you think about this approach? Do you see any other cases to be handled?

joel-costigliola:
Sounds good to me, a few comments:

- `containsExactly` also checks the order of keys/entries 
- `IdentityHashMap` is another map implementation that does not rely on equals but being part of the JDK we could simply instantiate it directly.
- We would need to state in the javadoc that in some cases the key comparison semantics won't honored

filiphr:
I think that your approach @scordio makes sense.

>`IdentityHashMap` is another map implementation that does not rely on equals but being part of the JDK we could simply instantiate it directly.

@joel-costigliola `IdentityHashMap` implements `Cloneable`. Therefore I think that it will be part of the first example from @scordio. I think that we should try and do as little key equal checks, which I think is being only done for `containsExactly`.

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -670,15 +678,59 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
...
+      Class<? extends Map> type = map.getClass();
+      try {
+        // try with copying constructor
+        return type.getConstructor(Map.class).newInstance(map);
+      } catch (NoSuchMethodException e) {
+        // try with default constructor
+        Map<K, V> newMap = type.getConstructor().newInstance();
+        newMap.putAll(map);
+        return newMap;
+      }
```
scordio, line 728: With this, I don't see the need of bringing in Objenesis. Please, let me know if you think otherwise.

joel-costigliola: I agree we don't need objenesis for now

scordio:
@joel-costigliola @filiphr `assertContainsOnly` and `assertContainsOnly` are in a reviewable shape, happy to get any feedback about the new implementation.

Most likely this PR will also include `assertContainsExactly`.

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -370,15 +370,54 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
...
 
     if (!notFound.isEmpty() || !notExpected.isEmpty())
       throw failures.failure(info, shouldContainOnlyKeys(actual, keys, notFound, notExpected));
   }
 
+  private static <K> Set<K> getNotFoundKeys(Map<K, ?> actual, K[] keys) {
+    return Stream.of(keys)
+                 .filter(key -> !actual.containsKey(key))
+                 .collect(Collectors.toCollection(LinkedHashSet::new));
+  }
```
filiphr, line 384: For performance reasons it is better to not use good old iterative style

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -406,13 +445,40 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
...
     if (!notFound.isEmpty() || !notExpected.isEmpty())
       throw failures.failure(info, shouldContainOnly(actual, entries, notFound, notExpected));
   }
 
+  private static <K, V> Set<Entry<? extends K, ? extends V>> getNotFoundEntries(Map<K, V> actual,
+                                                                                Entry<? extends K, ? extends V>[] entries) {
+    return Stream.of(entries)
+                 .filter(entry -> !containsEntry(actual, entry))
+                 .collect(Collectors.toCollection(LinkedHashSet::new));
+  }
```
filiphr, line 459: Same comment about the performance

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -406,13 +445,40 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
...
+    try {
+      Map<K, V> clonedMap = clone(actual);
+      expectedEntries.forEach(entry -> clonedMap.remove(entry.getKey(), entry.getValue()));
+      return clonedMap;
+    } catch (NoSuchMethodException | UnsupportedOperationException e) {
+      // actual cannot be cloned or is unmodifiable, falling back to LinkedHashMap
+      Map<K, V> copiedMap = new LinkedHashMap<>(actual);
+      expectedEntries.forEach(entry -> copiedMap.remove(entry.getKey(), entry.getValue()));
+      return copiedMap;
+    }
```
filiphr, line 479: For performance reasons it is better to use for loops instead of `.forEach`.

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -441,27 +507,22 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
...
-    notExpected.addAll(actualEntries.keySet());
   }
 
   private <K, V> void compareActualMapAndExpectedEntries(Map<K, V> actual, Entry<? extends K, ? extends V>[] entries,
                                                          Set<Entry<? extends K, ? extends V>> notExpected,
                                                          Set<Entry<? extends K, ? extends V>> notFound) {
     Map<K, V> expectedEntries = entriesToMap(entries);
-    Map<K, V> actualEntries = new LinkedHashMap<>(actual);
+    Map<K, V> actualEntries = instantiate(actual.getClass());
+    actualEntries.putAll(actual);
```
filiphr, line 525: Why not use `clone` here as well?

filiphr (commented):
It looks great @scordio. Quite extensive tests and works as I would expect.

I've left some performance comments and one place where you used `instantiate` instead of `clone`.

src/main/java/org/assertj/core/internal/Maps.java
scordio: This part is invoked only by `assertContainsExactly` that I still need to rework, once done it should disappear (and `clone` will be used)

scordio:
Having the same pattern for `assertContainsExactly` will increase the complexity of this PR as the tests are more complicated due to the ordering requirement.

I will focus on `containsOnly` related assertions and update `assertContainsExactly` in a separate PR.

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -370,15 +363,67 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
...
     if (!notFound.isEmpty() || !notExpected.isEmpty())
       throw failures.failure(info, shouldContainOnlyKeys(actual, keys, notFound, notExpected));
   }
 
+  private static <K> Set<K> getFoundKeys(Map<K, ?> actual, K[] expectedKeys) {
+    Set<K> found = new LinkedHashSet<>();
+    for (K expectedKey : expectedKeys) {
+      if (actual.containsKey(expectedKey)) found.add(expectedKey);
+    }
+    return found;
```
joel-costigliola, line 378: could use Stream API filter:
```java
return Stream.of(expectedKeys)
             .filter(actual::containsKey) 
             .collect(toCollection( LinkedHashSet::new));
```

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -370,15 +363,67 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
...
+  }
+
+  private static <K> Set<K> getNotFoundKeys(Map<K, ?> actual, K[] expectedKeys) {
+    Set<K> notFound = new LinkedHashSet<>();
+    for (K expectedKey : expectedKeys) {
+      if (!actual.containsKey(expectedKey)) notFound.add(expectedKey);
+    }
+    return notFound;
```
joel-costigliola, line 386: similar comment here

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -406,13 +451,48 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
...
       throw failures.failure(info, shouldContainOnly(actual, entries, notFound, notExpected));
   }
 
+  private static <K, V> Set<Entry<? extends K, ? extends V>> getNotFoundEntries(Map<K, V> actual,
+                                                                                Entry<? extends K, ? extends V>[] entries) {
+    Set<Entry<? extends K, ? extends V>> notFound = new LinkedHashSet<>();
+    for (Entry<? extends K, ? extends V> entry : entries) {
+      if (!containsEntry(actual, entry)) notFound.add(entry);
+    }
+    return notFound;
```
joel-costigliola, line 467: could  use stream api

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -406,13 +451,48 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
...
+  }
+
+  private static <K, V> Set<Entry<K, V>> getNotExpectedEntries(Map<K, V> actual, Entry<? extends K, ? extends V>[] entries) {
+    Set<Entry<K, V>> notExpected = new LinkedHashSet<>();
+    for (Entry<K, V> entry : mapWithoutExpectedEntries(actual, entries).entrySet()) {
+      MapEntry<K, V> mapEntry = entry(entry.getKey(), entry.getValue());
+      notExpected.add(mapEntry);
+    }
+    return notExpected;
```
joel-costigliola, line 476: same comment

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -406,13 +451,48 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
...
+  }
+
+  private static <K, V> Map<K, V> mapWithoutExpectedEntries(Map<K, V> actual, Entry<? extends K, ? extends V>[] expectedEntries) {
+    try {
+      Map<K, V> clonedMap = clone(actual);
+      for (Entry<? extends K, ? extends V> expectedEntry : expectedEntries) {
+        clonedMap.remove(expectedEntry.getKey(), expectedEntry.getValue());
+      }
+      return clonedMap;
+    } catch (NoSuchMethodException | UnsupportedOperationException e) {
```
joel-costigliola, line 486: should we also catch the `IllegalStateException` thrown by `clone(Map<K, V> map)`?

src/test/java/org/assertj/core/test/Maps.java
```diff
@@ -24,18 +26,19 @@
...
-    return map;
+    return mapOf(LinkedHashMap::new, entries);
   }
 
   @SafeVarargs
   public static <K extends Comparable<? super K>, V> TreeMap<K, V> treeMapOf(MapEntry<K, V>... entries) {
-    TreeMap<K, V> map = new TreeMap<>();
-    for (MapEntry<K, V> entry : entries) {
-      map.put(entry.key, entry.value);
+    return mapOf(TreeMap::new, entries);
```
joel-costigliola, line 34: nice, thanks!

joel-costigliola (commented):
I still have to look at the tests but I thought I'll share my comments on the rest of the code

src/main/java/org/assertj/core/internal/Maps.java
scordio: Initially, I used the Stream API but @filiphr correctly pointed out that the for loop would be better performance-wise. I seem to recall some issues about performance in case of big input, so the proposal seemed valid to me. What do you think about it?

src/main/java/org/assertj/core/api/AbstractMapAssert.java
```diff
@@ -1136,7 +1136,10 @@ protected SELF doesNotContainKeysForProxy(K[] keys) {
 
   /**
    * Verifies that the actual map contains only the given keys and nothing else, in any order.
-   *
+   * <p>
+   * The verification tries to honor the key comparison semantic of the underlying map implementation, but there might
+   * be some cases where the semantic cannot be honored and not all the unexpected elements are identified correctly
+   * (e.g., due to immutability).
```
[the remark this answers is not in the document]
scordio: What about:

> The verification tries to honor the key comparison semantic of the underlying map implementation. The map under test has to be cloned to identify unexpected elements, but depending on the map implementation this may not always be possible. In these cases, a regular map is used and the key comparison strategy may not be the same as the map under test.

src/main/java/org/assertj/core/internal/Maps.java
scordio: I was wondering about the same.

Catching `IllegalStateException` (or even propagate the checked exceptions and catch them here) would make the assertion more robust.

On the other side, we may use `LinkedHashMap` for implementations that could be handled in a better way. I can imagine these would be exotic map implementations. It could be worth getting the knowledge about them and add proper tests.

Currently, I would take the risk, i.e., no catch for `IllegalStateException`.

src/main/java/org/assertj/core/util/Sets.java
```diff
@@ -22,9 +22,22 @@
...
+   * Creates a <em>mutable</em> {@link HashSet} containing the given elements.
+   *
+   * @param <T> the generic type of the {@code HashSet} to create.
+   * @param elements the elements to store in the {@code HashSet}.
+   * @return the created {@code HashSet}, or {@code null} if the given array of elements is {@code null}.
+   */
+  @SafeVarargs
+  public static <T> Set<T> set(T... elements) {
+    return newLinkedHashSet(elements);
+  }
```
scordio, line 39: Perhaps I should highlight this addition in the util class, which would be publicly available. Let me know if you have any problems with it.

src/main/java/org/assertj/core/api/AbstractMapAssert.java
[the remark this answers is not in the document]
joel-costigliola: sounds good to me

src/main/java/org/assertj/core/internal/Maps.java
joel-costigliola: fair enough, let's add a comment saying we don't use stream for performance reasons.

src/main/java/org/assertj/core/internal/Maps.java
joel-costigliola: fine by me

src/main/java/org/assertj/core/util/Sets.java
joel-costigliola: we can but for me these util classes are not meant to be API so we don't need to publicize them.

src/test/java/org/assertj/core/internal/maps/Maps_assertContainsOnlyKeys_Test.java
```diff
@@ -12,112 +12,216 @@
...
+
+  @SuppressWarnings("unchecked")
+  private static final Supplier<Map<String, String>>[] CASE_INSENSITIVE_MAPS = new Supplier[] {
+      // org.apache.commons.collections4.map.CaseInsensitiveMap not included due to slightly different behavior
+      LinkedCaseInsensitiveMap::new,
+      CASE_INSENSITIVE_TREE_MAP
+  };
+
+  @SuppressWarnings("unchecked")
+  private static final Supplier<Map<String, String>>[] MODIFIABLE_MAPS = ArrayUtils.addAll(CASE_INSENSITIVE_MAPS,
```
joel-costigliola, line 75: just curious why not using `new Supplier[]` as in L68

src/test/java/org/assertj/core/internal/maps/Maps_assertContainsOnlyKeys_Test.java
```diff
@@ -12,112 +12,216 @@
...
-  @Test
-  void should_pass_if_actual_contains_only_expected_keys() {
-    maps.assertContainsOnlyKeys(someInfo(), actual, array("color", "name"));
-    maps.assertContainsOnlyKeys(someInfo(), actual, array("name", "color"));
+  private static Stream<Arguments> unmodifiableMapsSuccessfulTestCases() {
+    return Stream.of(arguments(Collections.emptyMap(), emptyKeys()),
+                     arguments(Collections.singletonMap("name", "Yoda"), array("name")),
+                     arguments(new SingletonMap<>("name", "Yoda"), array("name")),
+                     arguments(Collections.unmodifiableMap(mapOf(entry("name", "Yoda"), entry("job", "Jedi"))),
+                               array("name", "job")),
```
joel-costigliola, line 130: minor: we don't loose information by static import `Collections.` and that will make the code a bit lighter.

src/test/java/org/assertj/core/internal/maps/Maps_assertContainsOnlyKeys_Test.java
```diff
@@ -12,112 +12,216 @@
...
+                     arguments(ImmutableMap.of("name", "Yoda", "job", "Jedi"), array("name", "job")));
```
joel-costigliola, line 131: keep the static import here, if we don't it's not clear a map is being built.

joel-costigliola, line 131: add `arguments(ImmutableMap.of("name", "Yoda", "job", "Jedi"), array("job", "name"))` as in `modifiableMapsSuccessfulTestCases`

src/test/java/org/assertj/core/internal/maps/Maps_assertContainsOnlyKeys_Test.java
```diff
@@ -12,112 +12,216 @@
...
-    // WHEN
-    expectAssertionError(() -> maps.assertContainsOnlyKeys(info, underTest, expectedKeys));
-    // THEN
-    verify(failures).failure(info, shouldContainOnlyKeys(underTest, expectedKeys, set("color"), emptySet()));
+  private static Stream<Arguments> caseInsensitiveMapsSuccessfulTestCases() {
+    return Stream.of(ArrayUtils.add(CASE_INSENSITIVE_MAPS, CaseInsensitiveMap::new))
+                 .flatMap(supplier -> Stream.of(arguments(mapOf(supplier, entry("NAME", "Yoda"), entry("Job", "Jedi")),
+                                                          array("name", "job")),
+                                                arguments(mapOf(supplier, entry("NAME", "Yoda"), entry("Job", "Jedi")),
+                                                          array("Name", "Job"))));
```
joel-costigliola, line 147: invert the order to show order is not important or add another `arguments` with the inverted order.

src/test/java/org/assertj/core/internal/maps/Maps_assertContainsOnlyKeys_Test.java
```diff
@@ -12,112 +12,216 @@
...
+    then(error).as(actual.getClass().getName())
+               .hasMessage(shouldContainOnlyKeys(actual, expected, notFound, notExpected).create());
   }
 
-  private static HashSet<String> set(String entry) {
-    HashSet<String> set = new HashSet<>();
-    set.add(entry);
-    return set;
+  private static Stream<Arguments> unmodifiableMapsFailureTestCases() {
+    return Stream.of(arguments(Collections.emptyMap(),
```
joel-costigliola, line 166: similar comment for static import.

src/test/java/org/assertj/core/internal/maps/Maps_assertContainsOnly_Test.java
```diff
@@ -12,131 +12,225 @@
...
+  void should_pass(Map<String, String> actual, Entry<String, String>[] expected) {
+    // WHEN/THEN
+    assertThatNoException().as(actual.getClass().getName())
+                           .isThrownBy(() -> maps.assertContainsOnly(info, actual, expected));
   }
 
-  @Test
-  void should_pass_if_actual_contains_only_expected_entries() {
-    maps.assertContainsOnly(someInfo(), actual, array(entry("name", "Yoda"), entry("color", "green")));
+  private static Stream<Arguments> unmodifiableMapsSuccessfulTestCases() {
```
joel-costigliola, line 122: similar comment on static import

src/test/java/org/assertj/core/internal/maps/Maps_assertDoesNotContainKeys_Test.java
```diff
@@ -71,7 +78,7 @@ void should_fail_if_actual_contains_key() {
     Throwable error = catchThrowable(() -> maps.assertDoesNotContainKeys(info, actual, new String[] { key }));
```
joel-costigliola, line 78: use `expectAssertionError` to remove `assertThat(error).isInstanceOf(AssertionError.class);` and `error` local variable

src/test/java/org/assertj/core/internal/maps/Maps_assertDoesNotContainKeys_Test.java
```diff
@@ -83,6 +90,6 @@ void should_fail_if_actual_contains_keys() {
     Throwable error = catchThrowable(() -> maps.assertDoesNotContainKeys(info, actual, new String[] { key1, key2 }));
```
joel-costigliola, line 90: same comment here

joel-costigliola (commented):
just some minor code cleanup and we are good to go!

src/test/java/org/assertj/core/internal/maps/Maps_assertContainsOnlyKeys_Test.java
scordio: Test cases added

src/test/java/org/assertj/core/internal/maps/Maps_assertContainsOnlyKeys_Test.java
scordio: Because I want to concatenate the existing `CASE_INSENSITIVE_MAPS` with the additional case-sensitive suppliers: if we add anything new to the former, all the test cases for `MODIFIABLE_MAPS` should run and pass out of the box.

src/test/java/org/assertj/core/internal/maps/Maps_assertDoesNotContainKeys_Test.java
joel-costigliola: not sure why you resolved this one @scordio ;-)

scordio: Becase I rewrote the entire `Maps_assertDoesNotContainKeys_Test` 🙂

joel-costigliola:
Thanks @filiphr and @scordio, this is now integrated.

Issue #971, opened 2017-04-11
New Feature: to have ability to ignore all assert when expected value is null


Sean-PAN2014:
#### Summary
to have ability to ignore all assert when expected value is null.

#### Example
Let's call the new method "ignoreAssertWhenExpectedIsNull"

```java
assertThat("Hello World!").ignoreAssertWhenExpectedIsNull().isEqualTo(null); // will pass
```

joel-costigliola:
Hmm, I fail to see how this is useful, do you have a real use case for that ?
I feel that if you can't tell whether the object under test is null or not then your unit test is not properly designed.

filiphr:
@joel-costigliola a use case for this is when you create parametrized tests and for some inputs you actually expect `null`. 

In order to achieve a correct test I do an if check and invoke the required method. For example:

```java
@Test
public void testParametrized(String input, String expected) {
    String result = someMethocCall(input);

    if (expected == null) {
        assertThat(result).isNull();
    } else {
        assertThat(result).isEqualTo(expected);
    }
}
```

@Sean-PAN2014 did I got your use case right?

Sean-PAN2014:
Well, I was thinking of this case.

```java
@Test
public void testParametrized(String input, String expected) {
    String result = someMethocCall(input);

    if (expected != null) {
        assertThat(result).isEqualTo(expected);
    }
}
```
When expected is null, assertThat is not needed, and actual may not be null.

tetv:
Also when there is a condition (boolean or function):

```javascript
@Test
public void my_test() {
    assertThat(obj1).ignoreAssertsWhen(x -> x < 0).isEqualTo(10); // using function condition
    assertThat(obj2).ignoreAssertsWhen(obj9 == 0.0).isEqualTo(10); // using boolean condition
    assertThat(obj3).ignoreAssertsWhenIsNull().isEqualTo(10); // using null condition
}
```

joel-costigliola:
so basically conditional assertions, right ? 
thinking out loud, we could passe a `Predicate` or a `Condition`, another name that could work is `skipping`
```java
assertThat(obj1).skippingAssertionsWhen(x -> x < 0).isEqualTo(10); 
```

tetv:
A better alternative would be implement the opposite: `onlyAssertWhen` and `onlyAssertWhenNotNull`

```javascript

// using function condition (ignore assert when obj1 is a negative numbers)
[...]

PR #2171, opened 2021-04-18
Draft for using conditional asserts by redirecting to a common method

filiphr:
This is an idea for a potential solution for #971.

I had an initial idea to use ByteBuddy, but then came up with this.

This allows to use `ignoreWhen` (the name of the method doesn't really matter) multiple times. Otherwise, we would be creating a new instance on every call to `ignoreWhen` and you can't reuse them. i.e. if you call once then the second time will be ignored.

e.g.

From one of the examples in the issue:
```
// more complex (if obj4 is negative should be -10, if positive should be 100, ignore zero)
assertThat(obj4)
    .ignoreWhen(i -> i < 0).isEqualTo(-10)
    .ignoreWhen(i -> i > 0).isEqualTo(100);
```

with the approach here it will work as excepted. Otherwise, it won't work because `ignoreWhen` will create a new object every time.

The approach in the PR allows doing that.

In addition to the implementation, I have also added some tests to see how this can be done. If you like the approach I can go ahead and implement it for the entire API and add the needed test methods to `BaseTestTemplate`.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -72,6 +72,8 @@ public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, 
 
   protected Objects objects = Objects.instance();
 
+  protected Predicate<ACTUAL> ignoreWhen = actual -> false;
+
   @VisibleForTesting
   Conditions conditions = Conditions.instance();
 
@@ -358,105 +360,112 @@ public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, 
   /** {@inheritDoc} */
   @Override
   public SELF isEqualTo(Object expected) {
-    objects.assertEqual(info, actual, expected);
+    //return doAssert(innerExpected -> objects.assertEqual(info, actual, innerExpected), expected);
+    return doAssert(() -> objects.assertEqual(info, actual, expected));
+  }
+
+  // Using this method means that we have to create a new object always, in order to use the input parameter
+  protected SELF doAssert(Runnable assertRunnable) {
+    if (ignoreWhen.test(actual)) {
+      return myself;
+    }
+
+    assertRunnable.run();
+
+    return myself;
+  }
+
+  // Using this method we can avoid not needed creation of objects, but passing everything we need to the consumer
+  protected <T> SELF doAssert(Consumer<T> consumer, T expected) {
+    if (ignoreWhen.test(actual)) {
+      return myself;
+    }
+    consumer.accept(expected);
     return myself;
   }
 
   /** {@inheritDoc} */
   @Override
   public SELF isNotEqualTo(Object other) {
-    objects.assertNotEqual(info, actual, other);
-    return myself;
+    return doAssert(inner -> objects.assertNotEqual(info, actual, inner), other);
   }
 
   /** {@inheritDoc} */
   @Override
   public void isNull() {
-    objects.assertNull(info, actual);
+    doAssert(() -> objects.assertNull(info, actual));
   }
 
   /** {@inheritDoc} */
   @Override
   public SELF isNotNull() {
-    objects.assertNotNull(info, actual);
-    return myself;
+    return doAssert(() -> objects.assertNotNull(info, actual));
   }
 
   /** {@inheritDoc} */
   @Override
   public SELF isSameAs(Object expected) {
-    objects.assertSame(info, actual, expected);
-    return myself;
+    return doAssert(inner -> objects.assertSame(info, actual, inner), expected);
   }
 
   /** {@inheritDoc} */
   @Override
   public SELF isNotSameAs(Object other) {
-    objects.assertNotSame(info, actual, other);
-    return myself;
+    return doAssert(inner -> objects.assertNotSame(info, actual, inner), other);
   }
 
   /** {@inheritDoc} */
   @Override
   public SELF isIn(Object... values) {
-    objects.assertIsIn(info, actual, values);
-    return myself;
+    return doAssert(inner -> objects.assertIsIn(info, actual, inner), values);
   }
 
   /** {@inheritDoc} */
   @Override
   public SELF isNotIn(Object... values) {
-    objects.assertIsNotIn(info, actual, values);
-    return myself;
+    return doAssert(inner -> objects.assertIsNotIn(info, actual, inner), values);
   }
 
   /** {@inheritDoc} */
   @Override
   public SELF isIn(Iterable<?> values) {
-    objects.assertIsIn(info, actual, values);
-    return myself;
+    return doAssert(inner -> objects.assertIsIn(info, actual, inner), values);
   }
 
   /** {@inheritDoc} */
   @Override
   public SELF isNotIn(Iterable<?> values) {
-    objects.assertIsNotIn(info, actual, values);
-    return myself;
+    return doAssert(inner -> objects.assertIsNotIn(info, actual, inner), values);
   }
 
   /** {@inheritDoc} */
   @Override
   public SELF is(Condition<? super ACTUAL> condition) {
-    conditions.assertIs(info, actual, condition);
-    return myself;
+    return doAssert(inner -> conditions.assertIs(info, actual, inner), condition);
   }
 
   /** {@inheritDoc} */
   @Override
   public SELF isNot(Condition<? super ACTUAL> condition) {
-    conditions.assertIsNot(info, actual, condition);
-    return myself;
+    return doAssert(inner -> conditions.assertIsNot(info, actual, inner), condition);
   }
 
   /** {@inheritDoc} */
   @Override
   public SELF has(Condition<? super ACTUAL> condition) {
-    conditions.assertHas(info, actual, condition);
-    return myself;
+    return doAssert(inner -> conditions.assertHas(info, actual, inner), condition);
   }
 
   /** {@inheritDoc} */
   @Override
   public SELF doesNotHave(Condition<? super ACTUAL> condition) {
-    conditions.assertDoesNotHave(info, actual, condition);
-    return myself;
+    return doAssert(inner -> conditions.assertDoesNotHave(info, actual, inner), condition);
   }
 
   /** {@inheritDoc} */
   @Override
   public SELF satisfies(Condition<? super ACTUAL> condition) {
-    conditions.assertSatisfies(info, actual, condition);
-    return myself;
+    return doAssert(inner -> conditions.assertSatisfies(info, actual, inner), condition);
   }
 
   /** {@inheritDoc} */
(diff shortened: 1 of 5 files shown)


filiphr:
I just thought of another idea that could reuse the `doAssert` methods. In theory we can simplify the `Assumptions` and `SoftAssertions` proxy to only intercept the `doAssert` methods and wrap the consumer and rethrow the appropriate exception.

In addition to that, we could even look into an option of providing a wrapper of the consumers in the current classes. Something similar to the [`TaskDecorator`](https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/core/task/TaskDecorator.java) that Spring has.

src/main/java/org/assertj/core/api/AbstractAssert.java
```diff
@@ -358,105 +360,112 @@ private void printDescriptionText() {
...
+    if (ignoreWhen.test(actual)) {
+      return myself;
+    }
+
+    assertRunnable.run();
+
+    return myself;
+  }
+
+  // Using this method we can avoid not needed creation of objects, but passing everything we need to the consumer
```
joel-costigliola, line 378: which object are we talking about?
we still need to build a consumer so it does not feel different that creating a Runnable.

src/main/java/org/assertj/core/api/AbstractAssert.java
```diff
@@ -72,6 +72,8 @@
 
   protected Objects objects = Objects.instance();
 
+  protected Predicate<ACTUAL> ignoreWhen = actual -> false;
```
joel-costigliola, line 75: I would simply set it to `null` by default so that we don't have to execute a predicate for nothing, just need a `null` check

src/main/java/org/assertj/core/api/AbstractAssert.java
```diff
@@ -358,105 +360,112 @@ private void printDescriptionText() {
   /** {@inheritDoc} */
   @Override
   public SELF isEqualTo(Object expected) {
-    objects.assertEqual(info, actual, expected);
+    //return doAssert(innerExpected -> objects.assertEqual(info, actual, innerExpected), expected);
+    return doAssert(() -> objects.assertEqual(info, actual, expected));
```
joel-costigliola, line 364: another option would be:
```java
 public SELF isEqualTo(Object expected) {
    if (ignoreWhen != null && ignoreWhen.test(actual)) return myself;
    return objects.assertEqual(info, actual, expected);
}
```
This is not much more work that  wrapping the call in `doAssert` and it has low performance impact for tests not using `ignoreWhen(Predicate<ACTUAL> predicate)` since a simple null check is performed vs a predicate call.

`ignoreWhen != null && ignoreWhen.test(actual)` can be refactored into a `ignoreAssertion(actual)` method:
```java
 public SELF isEqualTo(Object expected) {
    if (ignoreAssertion(actual)) return myself;
    return objects.assertEqual(info, actual, expected);
}
```

src/main/java/org/assertj/core/api/AbstractAssert.java
```diff
@@ -1094,4 +1091,11 @@ private void propagateAssertionInfoFrom(AbstractAssert<?, ?> assertInstance) {
     return (ASSERT) assertFactory.createAssert(extractedValue).withAssertionState(myself);
   }
 
+  @CheckReturnValue
+  public SELF ignoreWhen(Predicate<ACTUAL> predicate) {
+    requireNonNull(predicate, shouldNotBeNull("predicate")::create);
+    this.ignoreWhen = predicate;
+    return myself;
+  }
```
joel-costigliola, line 1099: it is the right place for `ignoreWhen` but it also means that any subclass will have to support it, for AssertJ it means quite a bit of work to adapt all assertions, it is worse for custom assertions that inherit `AbstractAssert` as they will provide `ignoreWhen` but it won't work until they change all their assertions.

I think `ignoreWhen` should be defined as a default method in the `Assert` interface (there's also an `ExtensionPoints` interface but since it is assertion related I prefer `Assert`), the default implementation would throw an exception saying it is not supported.

`AbstractAssert` can provide a `doIgnoreWhen` method similar to the `ignoreWhen` above and subclass that supports `ignoreWhen` will simply override it and call `doIgnoreWhen`.

I haven't tried to POC this idea, but I believe it should work, anyway I feel we need to try a proxied base solution to avoid writing so much code and have it supported for any new assertions we would add.

src/test/java/org/assertj/core/api/AbstractAssertBaseTest.java
```diff
@@ -22,4 +27,41 @@
   protected ConcreteAssert create_assertions() {
     return new ConcreteAssert(6L);
   }
+
+  protected ConcreteAssert applyIgnoreWithoutEffects() {
+    return assertions.ignoreWhen(v -> (Long) v < 8);
```
joel-costigliola, line 32: why not `v -> true`?

src/test/java/org/assertj/core/api/AbstractAssertBaseTest.java
```diff
@@ -22,4 +27,41 @@
...
+  protected ConcreteAssert applyIgnoreWithEffects() {
+    return assertions.ignoreWhen(v -> (Long) v > 8);
+  }
+
+  @Test
+  public void should_have_no_internal_effects_when_assertions_should_be_ignored() {
+    this.assertions = applyIgnoreWithoutEffects();
+    invoke_api_method();
+    verify_no_internal_effects();
+  }
```
joel-costigliola, line 44: that's really nice! :+1:  it avoids us writing a test per assertions (which would be a lot of work!)

src/test/java/org/assertj/core/api/abstract_/AbstractAssert_isInstanceOfSatisfying_Test.java
```diff
@@ -48,6 +48,16 @@ protected ConcreteAssert create_assertions() {
     return new ConcreteAssert(yoda);
   }
 
+  @Override
+  protected ConcreteAssert applyIgnoreWithoutEffects() {
+    return assertions.ignoreWhen(v -> ((Jedi) v).getName().equals("Yoda"));
+  }
```
joel-costigliola, line 54: I believe we don't need to override this if we used `v -> true` for `applyIgnoreWithoutEffects` implementation

joel-costigliola (commented):
@filiphr thanks for the POC, I have added my comments, I'm keen to hear what @scordio has to say ;-)

src/main/java/org/assertj/core/api/AbstractAssert.java
```diff
@@ -1094,4 +1091,11 @@ private void propagateAssertionInfoFrom(AbstractAssert<?, ?> assertInstance) {
     return (ASSERT) assertFactory.createAssert(extractedValue).withAssertionState(myself);
   }
 
+  @CheckReturnValue
+  public SELF ignoreWhen(Predicate<ACTUAL> predicate) {
+    requireNonNull(predicate, shouldNotBeNull("predicate")::create);
```
scordio, line 1096: `requireNonNull` with a `Supplier` is indeed what we should use in many other places! 👍

src/main/java/org/assertj/core/api/AbstractAssert.java
scordio: I have the same concern, `ignoreWhen` will require quite some work within AssertJ projects and there is no guarantee that custom assertions will use the `doAssert` pattern.

The default method throwing an exception might be possible protection, but probably not the best we could achieve.

I agree we should try a proxy-based solution.

src/main/java/org/assertj/core/api/AbstractAssert.java
filiphr: I think that the compiler can infer whether we need to create a new consumer every time or not. If we don't use parameters outside of the consumer, I think that the compiler can infer that. I might be wrong though

src/test/java/org/assertj/core/api/AbstractAssertBaseTest.java
filiphr: No particular reason. I can change it

src/test/java/org/assertj/core/api/AbstractAssertBaseTest.java
filiphr: Thanks. When I realised how many tests we would need to write, my head started spinning, so I was trying to come up with some solution 😄

src/test/java/org/assertj/core/api/abstract_/AbstractAssert_isInstanceOfSatisfying_Test.java
filiphr: You are completely right, it would make it way easier. I'll try with that

filiphr:
Thanks for the review @scordio and @joel-costigliola. I am going to also try to do a POC with proxies. One of the reasons why I didn't go down that path was the soft assertions. We need to be able to recreate the current Assert object, perhaps we could do something like the extracting methods.

-------------
I know that the approach I am proposing here will mean a lot of work to bring this to all assertions in AssertJ, but I think it will be worth it, and I can work on it as well.

An alternative approach to possibly having a default method throwing an exception would be to pass something in the constructor saying whether conditionals are supported or not, or some other object for configuration. If that is not set to `true` we can throw an exception. This will mean that 3rd party assertions can decide on their own whether they want to support this or not. We could off course add this to `Assert` (with default throwing exception).

src/main/java/org/assertj/core/api/AbstractAssert.java
joel-costigliola: I think we should run a perf tests before and after to see the real impact of any approach, otherwise it's just gesstimating the impact.

joel-costigliola (commented):
@filiphr thanks a lot for this POC. Let's explore the proxy option in another PR.

> I know that the approach I am proposing here will mean a lot of work to bring this to all assertions in AssertJ, but I think it will be worth it, and I can work on it as well.

If we settle on this approach we can create issues per Assert classes and label them as _good first issue_ for first time contributors since this is not complicated coding, simply a lot of work.

> An alternative approach to possibly having a default method throwing an exception would be to pass something in the constructor saying whether conditionals are supported or not, or some other object for configuration. If that is not set to true we can throw an exception. This will mean that 3rd party assertions can decide on their own whether they want to support this or not. We could off course add this to Assert (with default throwing exception).

The main problem for me is adding `ignoreWhen` to `AbstractAssert` as all inheriting classes will expose it. For example guava assertions classes will expose it as soon as it used the assertj-core version providing this feature but if we don't update the guava assert classes users will be tricked in thinking this would work:

```java
assertThat(multimap).ignoreWhen(v -> v == null)
                    .contains(...)
```
Think also of custom assertions type written for a project that will see `ignoreWhen` appearing after bumping assertj-core version.

So what I'm saying is defining `ignoreWhen` in `AbstractAssert` is not great, thinking out loud we could simply define it in a new interface (`IgnorableAssert` ?) and let the `AbstractXxxAssert` implement it. That way no API pollution.

An advantage of this approach is to be able to deliver ignorable assertions in multiple releases instead having to do all of them in one release.

An important part of this work would be to write a test that detects new assertions not supporting `ignoreWhen` otherwise we will end up adding assertions and find later on that they don't play well with `ignoreWhen`. Now I'm not completely sure this is technically possible to write such a test but we should give it a try.

filiphr:
>If we settle on this approach we can create issues per Assert classes and label them as good first issue for first time contributors since this is not complicated coding, simply a lot of work.

Indeed if we settle on this approach it will be trivial for anyone to do the work. I also saw that you are participating in hack commit push. In theory this can be a good candidate for that as well.

>The main problem for me is adding `ignoreWhen` to `AbstractAssert` as all inheriting classes will expose it. For example guava assertions classes will expose it as soon as it used the assertj-core version providing this feature but if we don't update the guava assert classes users will be tricked in thinking this would work:

I understand the problem. That's why I proposed adding another field indicating whether `ignoreWhen` is supported or not. The top level classes will need to set this. If it is not set to `true` then calling `ignoreWhen` will throw an exception. This way we can even gradually move the AssertJ classes as well.

>So what I'm saying is defining `ignoreWhen` in `AbstractAssert` is not great, thinking out loud we could simply define it in a new interface (`IgnorableAssert` ?) and let the `AbstractXxxAssert` implement it. That way no API pollution.

I would still like to at least have a `protected ignoreWhen` in `AbstractAssert` in order to make it easier for us to implement this. The `IgnorableAssert` with `AbstractXxxAssert` will then make that public and expose it. I like the idea.

>An important part of this work would be to write a test that detects new assertions not supporting `ignoreWhen` otherwise we will end up adding assertions and find later on that they don't play well with `ignoreWhen`. Now I'm not completely sure this is technically possible to write such a test but we should give it a try.

That was my idea with the changes in `AbstractAssertBaseTest`. I think that I can even pull this all the way up in `BaseTestTemplate` and then with some reflection verify that none of the mocked or spied fields are invoked.

src/main/java/org/assertj/core/api/AbstractAssert.java
filiphr: You are right. I'll try to do some perf tests as well. Do you by any chance have some repo where you do some benchmarks already?

joel-costigliola: nope :( we have [assertj-examples](https://github.com/assertj/assertj-examples/tree/main/assertions-examples/src/test/java/org/assertj/examples) or maybe it's time to create one.

joel-costigliola:
> I would still like to at least have a protected `ignoreWhen` in `AbstractAssert` in order to make it easier for us to implement this. The `IgnorableAssert` with `AbstractXxxAssert` will then make that public and expose it. I like the idea.

Sure, no problem adding protected method in `AbstractAssert`, it's not polluting what regular users see.

> That was my idea with the changes in AbstractAssertBaseTest. I think that I can even pull this all the way up in BaseTestTemplate and then with some reflection verify that none of the mocked or spied fields are invoked.

Good idea!

src/main/java/org/assertj/core/api/AbstractAssert.java
filiphr: I'll try to see if I can come up with a base example first for `isEqualTo` and `isNotEqualTo` we can then go from there :)

filiphr: Have a look at https://github.com/filiphr/assertj-benchmark. It is extremely work in progress, but I think it could be beneficial. I am more than willing to transfer the repo to the assertj organization.

filiphr: So first benchmark results are in

image

Left is with the changes in this PR and right is with 3.19.0. So there is some performance impact. Although I find it interesting the it is faster for `isNotEqualTo`. Although there is a difference for how the `isEqualTo` and `isNotEqualTo` are done.

joel-costigliola: thanks, I'm happy to move your repo to the assertj organization and give you commit rights to.

filiphr: There is also a slight difference between `isEqualTo` and `isNotEqualTo`. I will try to align them and run the benchmark again.

For the repo, let me know how you want to do it. In order for me to move it I need to be part of the assertj organization. Another option would be you to create a repo and I can push to that one. I will leave this up to you

filiphr: I got a bit sidetracked with the benchmarks 😄. I realized that the drop in the `isEqualTo` is not 100% due to my changes, there is a drop due to https://github.com/assertj/assertj-core/commit/f2acb419eade8ab6704b62e8bf9d86c0104a643d. I also changed the benchmark a bit in order not to use a blackhole. And the latest results are the following:

image


* resuls.png -> 3.19.0
* results_main.png -> The code from the main branch
* results_conditional.png -> The code from this PR rebased on the main branch

Looking at the latest numbers. The drop is extremely significant (50%). Therefore, I will try to see if we can do better.

scordio:
> So what I'm saying is defining `ignoreWhen` in `AbstractAssert` is not great, thinking out loud we could simply define it in a new interface (`IgnorableAssert` ?) and let the `AbstractXxxAssert` implement it. That way no API pollution.

We might give special attention to `AbstractObjectAssert`, it is probably a common superclass candidate for custom assertions. If we let it implement `IgnorableAssert` we might have a similar issue like in `AbstractAssert`.

PR #2172, opened 2021-04-18
Use varargs for satisfiesAnyOf

filiphr:
Remove the previous grouped methods since the solution done in 3d5841fa110c029fc2104fd7425e02dd1ed19384 can be used here as well


joel-costigliola:
Good catch @filiphr, it's integrated now.

PR #2173, opened 2021-04-18
Align return types across assertions / assumptions / soft assertions

filiphr:
With 3d5841fa110c029fc2104fd7425e02dd1ed19384 there is no longer the need for some assertions to have different return types. I think that going forward the same approach can be used for other places were `@SafeVarargs` is needed.

In this PR I've applied all the backwards compatible fixes (returning a more specific type, i.e. from `AbstractXXXAssert` to `XXXAssert`).  However, there are some inconsistencies between `SoftAssertions` and `Assertions`. The `SoftAssertions` are mostly returning `XXXAssert` whereas the `Assertions` are returning `AbstractXXXAssert`. I think that going forward we should return the most specific assertion, i.e. `XXXAssert`? What do you think?

If you don't agree with this PR and would like everything to stay the same, then feel free to close it.

### DIFF
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -282,7 +282,7 @@ public class Assertions implements InstanceOfAssertFactories {
    * @param actual the actual value.
    * @return the created assertion object.
    */
-  public static AbstractBigDecimalAssert<?> assertThat(BigDecimal actual) {
+  public static BigDecimalAssert assertThat(BigDecimal actual) {
     return AssertionsForClassTypes.assertThat(actual);
   }
 
@@ -293,7 +293,7 @@ public class Assertions implements InstanceOfAssertFactories {
    * @return the created assertion object.
    * @since 2.7.0 / 3.7.0
    */
-  public static AbstractBigIntegerAssert<?> assertThat(BigInteger actual) {
+  public static BigIntegerAssert assertThat(BigInteger actual) {
     return new BigIntegerAssert(actual);
   }
 
@@ -303,7 +303,7 @@ public class Assertions implements InstanceOfAssertFactories {
    * @param actual the actual value.
    * @return the created assertion object.
    */
-  public static AbstractUriAssert<?> assertThat(URI actual) {
+  public static UriAssert assertThat(URI actual) {
     return AssertionsForClassTypes.assertThat(actual);
   }
 
@@ -313,7 +313,7 @@ public class Assertions implements InstanceOfAssertFactories {
    * @param actual the actual value.
    * @return the created assertion object.
    */
-  public static AbstractUrlAssert<?> assertThat(URL actual) {
+  public static UrlAssert assertThat(URL actual) {
     return AssertionsForClassTypes.assertThat(actual);
   }
 
@@ -323,7 +323,7 @@ public class Assertions implements InstanceOfAssertFactories {
    * @param actual the actual value.
    * @return the created assertion object.
    */
-  public static AbstractBooleanAssert<?> assertThat(boolean actual) {
+  public static BooleanAssert assertThat(boolean actual) {
     return AssertionsForClassTypes.assertThat(actual);
   }
 
@@ -333,7 +333,7 @@ public class Assertions implements InstanceOfAssertFactories {
    * @param actual the actual value.
    * @return the created assertion object.
    */
-  public static AbstractBooleanAssert<?> assertThat(Boolean actual) {
+  public static BooleanAssert assertThat(Boolean actual) {
     return AssertionsForClassTypes.assertThat(actual);
   }
 
@@ -343,7 +343,7 @@ public class Assertions implements InstanceOfAssertFactories {
    * @param actual the actual value.
    * @return the created assertion object.
    */
-  public static AbstractBooleanArrayAssert<?> assertThat(boolean[] actual) {
+  public static BooleanArrayAssert assertThat(boolean[] actual) {
     return AssertionsForClassTypes.assertThat(actual);
   }
 
@@ -364,7 +364,7 @@ public class Assertions implements InstanceOfAssertFactories {
    * @param actual the actual value.
    * @return the created assertion object.
    */
-  public static AbstractByteAssert<?> assertThat(byte actual) {
+  public static ByteAssert assertThat(byte actual) {
     return AssertionsForClassTypes.assertThat(actual);
   }
 
@@ -374,7 +374,7 @@ public class Assertions implements InstanceOfAssertFactories {
    * @param actual the actual value.
    * @return the created assertion object.
    */
-  public static AbstractByteAssert<?> assertThat(Byte actual) {
+  public static ByteAssert assertThat(Byte actual) {
     return AssertionsForClassTypes.assertThat(actual);
   }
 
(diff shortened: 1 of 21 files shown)


joel-costigliola:
I remember we wanted to return abstract type to ease extension but I'm not sure if that was really used or useful so I'm ok returning concrete type, the tests fail as some methods have not been synchronized accross different assertions/soft assertions/assumptions entry point

filiphr:
>I remember we wanted to return abstract type to ease extension but I'm not sure if that was really used or useful

I don't know how extensions will work for this. Perhaps on the `WithAssertions` and the other interfaces?

In any case I syncrhonized it everywhere and I am using the specific return type instead of Abstract. Let me know what you think

joel-costigliola:
Thanks @filiphr, this is integrated now.

filiphr:
Thanks @joel-costigliola

mi-laf:
Hi,
I have been using AssertJ for several years now and now I wanted to update to `3.20.1` from `3.19.0`. My testing tools don't compile anymore though.

---

I have my own `Assertions` class:

```
package org.example;

public class Assertions extends org.assertj.core.api.Assertions {
    ...
}
```

It's mainly used to add `assertThat` methods for some of custom types. However, I did also extend some of the built-in `Assert`s:

```
public class ExtendedFileAssert extends AbstractFileAssert<ExtendedFileAssert>

    public ExtendedFileAssert hasAncestor(File ancestor) {
        ...
    }
    
    ...
}
```

My Assertions class overrides contains the respective `assertThat` method:

```
    public static ExtendedFileAssert assertThat(File actual) {
        return new ExtendedFileAssert(actual);
    }
```
	
---

Maybe I should not have extended AssertJ that way, but I had thought the idea of the `AbstractXAssert` classes was to make them extensible. After the recent change my extension approach doesn't work though anymore and for now I just postponed the AssertJ version update, since I don't know how to proceed.

What's the suggested way to extend built-in assertions and make them conveniently accessible, e.g. simply via `assertThat` methods?

joel-costigliola:
Sorry about that @mi-laf, I see the following alternatives here:

1. we revert the changes to return abstract types (and do that consistently everywhere) but we'll discuss with the team to see if that's a path we want to take
2. contribute your extended file assertions to assertj if they are common enough to be used by other people
3. change to `ExtendedFileAssert` to extend `FileAssert`, the drawback is assertions chained from a `FileAssert` assertion will return a `FileAssert` type which won't show the extended assertions. this can be mitigated by overriding `FileAssert` assertions method and returning `this`, it works but that's not elegant to say the least.

Ex:

```java
public class MyBigDecimalAssert extends BigDecimalAssert {

  public static MyBigDecimalAssert assertThat(BigDecimal actual) {
    return new MyBigDecimalAssert(actual);
  }

  public MyBigDecimalAssert(BigDecimal actual) {
    super(actual);
  }

  @Override
  public MyBigDecimalAssert isOne() {
    // check condition
    if (actual != null && actual.compareTo(BigDecimal.ONE) != 0) {
      failWithMessage("Expected BigDecimal to be one but was <%s>", actual);
    }
    // return the current assertion for method chaining
    return this;
  }

  public MyBigDecimalAssert isTen() {
    // check condition
    if (actual != null && actual.compareTo(BigDecimal.TEN) != 0) {
      failWithMessage("Expected BigDecimal to be 10 but was <%s>", actual);
    }
    // return the current assertion for method chaining
    return this;
  }
  
}
```

this can be used as follow:
```java
MyBigDecimalAssert.assertThat(BigDecimal.ONE)
                   .isTen()
                    // can chain any assertion from  MyBigDecimalAssert or BigDecimalAssert
                   .isOne()
                    // can chain any assertion from  MyBigDecimalAssert as isOne() is overridden
                   .isTen()
                   .isNotNull()
                   .isTen(); // does not compile as isNotNull() was not overridden
```

mi-laf:
Thanks for the quick response, @joel-costigliola!

ad 2) That's a valid point, of course. Some of the extensions might be common enough. However, I still think it'd be nice to have an easy and clean way to add special extensions when needed.

ad 3) Thanks for the suggestion. I was thinking about this, but as you said it's not really elegant. `Assert`s are supposed to "know" their type (via `<SELF>`). This works when extending the abstract classes, but one would lose it when extending a concrete `Assert` such as `FileAssert`.

ad 1) This would be the best option for me ;), but I am not an AssertJ developer (nor expert) and can't say whether a revert is a good option.
I can say though is that based on the the way I used AssertJ so far concrete `Assert`s such as `FileAssert` were mainly convenience classes for me. One can easily instantiate them directly when needed and one doesn't have to deal with (ugly) generics. On the other hand, the abstract classes make it possible to add extensions (keeping things clean using the `<SELF>` type). They are therefore very valuable and in my opinion it makes sense to have them and keep the actual code there.
When using `Assertions.assertThat` methods as a user I don't really care, whether the returned type is abstract or not. But as someone who wants to extend `Assertions`, the abstract return types made this easy so far.
Therefore this seemed to be a good choice, but maybe keeping the abstract types there has some downsides I am not aware of.

filiphr:
As I was the one that worked on this I'd like to offer my 2 cents.

>Maybe I should not have extended AssertJ that way, but I had thought the idea of the AbstractXAssert classes was to make them extensible. After the recent change my extension approach doesn't work though anymore and for now I just postponed the AssertJ version update, since I don't know how to proceed.
>
>What's the suggested way to extend built-in assertions and make them conveniently accessible, e.g. simply via assertThat methods?

They way you are extending `AssertJ` is completely fine and I think that it is still valid to use the `AbstraxtXXXAssert` for extending them. You can keep extending from `AbstractFileAssert`. 

What is confusing to me is why:

```
public class Assertions extends org.assertj.core.api.Assertions {
    public static ExtendedFileAssert assertThat(File actual) {
        return new ExtendedFileAssert(actual);
    }
}
```

wouldn't compile? `org.assertj.core.api.Assertions` contains static methods. Therefore if you do `org.example.Assertions.assertThat(file)` you should get the `ExtendedFileAssert` without any problem.

Can you be more specific about what exactly does not compile?

scordio:
We decided to revert this change for two reasons:
* To reduce the binary compatibility issues with libraries compiled against an older version of AssertJ (e.g., `assertj-guava` 3.4.0 was compiled against `assertj-core` 3.15.0, and fails if it is in the classpath together with version 3.20.1)
* To allow third party libraries extending with their own classes (the use case mentioned by @mi-laf)

Reverting still brings back an inconsistent situation where most of the entry point methods are returning abstract types, but some are returning concrete ones. I added this topic to the scope of AssertJ 4 (#2229).

mi-laf:
@scordio: I can confirm that updating to `3.20.2` worked for me without any issues. Thanks!

I agree that it would be nice to have consistent (abstract) return types.

@filiphr: yes, the methods are static, but it still not possible to overload with an incompatible return type. This (simplified) code demonstrates the issue:

```
package org.example;

import java.io.File;

class AbstractFileAssert {
}

class FileAssert extends AbstractFileAssert {
}

class Assertions {
	// changing the return type to AbstractFileAssert fixes the compilation issue below
	static FileAssert assertThat(File file) {
		return new FileAssert();
	}
}

class ExtendedFileAssert extends AbstractFileAssert {
}

class ExtendedAssertions extends Assertions {
	// does not compile, because "The return type is incompatible with Assertions.assertThat(File)"
	static ExtendedFileAssert assertThat(File file) {
		return new ExtendedFileAssert();
	}
}
```

joel-costigliola:
Thanks for the feedback @mi-laf, appreciated.

PR #2174, opened 2021-04-19
CompositeRepresentation and CompositeRepresentation_Test draft.

### LINKED ISSUE #2048: Fine-grained Representations
#### Summary

I have recently discovered the `Representation` API. I am looking at using it with osgi-test and with idempiere-test as part of their respective domain-specific assertion libraries, so that they can produce more readable diagnostics by overriding `toString()` implementations in the domain objects.

For maximum usability and code readability, It would be nice if these Representation implementations could be auto-registered. AssertJ already provides a mechanism for this.

However, ideally, different domain-specific libraries would be able to independently register `Representation` implementations for their specific settings  osgi-test and idempiere-test could independently register `Representation` implementations for their respective domains. For example, osgi-test could register a `Representation` implementation for the type `org.osgi.framework.Bundle` while idempiere-test could register one for `org.compiere.model.MInvoice`.

Unfortunately, with the current implementation, the default `Representation` implementation is a global singleton setting that is applied to all objects. 

#### Proposal

I'm open to ideas. I'm thinking something along these lines:

1. Un-deprecate the `ServiceLoader` implementation of `Representation`. `Configuration` is by nature a singleton instance and not well-suited to the modular, multi-implementation approach being offered.
2.  `Representation` will become parameterized. Thus:
```java
public interface Representation<T> {
  String toStringOf(T object);
  String unambiguousToStringOf(T object);
```
Specific implementations of `Representation` can add their own bounds, eg:

```java
public class BundleRepresentation implements Representation<Bundle> {
...
}
```
3. AssertJ Core will collate all of the implementations and build a lookup table from object type to representation. This will be used by the core in determining which representation to use (if not overridden by a call to `usingReperesentation()`).

I think it will be possible to build the lookup table with no further changes to `Representation`, using reflection to find out the bound on `T`. However, if it is not possible, then add a `supportedType()` method to `Representation` that the core can interrogate:

```java
public interface Representation<T> {
  String toStringOf(T object);
  String unambiguousToStringOf(T object);
  default Class<T> supportedType() {return (Class<T>)Object.class; }
```

Annette0127:
#### Check List:
* Fixes #2048
* Unit tests : YES 

Maybe the method `String unambiguousToStringOf(Object object)` from the ''org.assertj.core.presentation.Representation.java'' is no longer needed since multiple `Representation` can be implemented in the class `CompositeRepresentation`.

### DIFF
--- a/src/main/java/org/assertj/core/presentation/CompositeRepresentation.java
+++ b/src/main/java/org/assertj/core/presentation/CompositeRepresentation.java
@@ -1,0 +1,76 @@
+package org.assertj.core.presentation;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.assertj.core.presentation.StandardRepresentation.STANDARD_REPRESENTATION;
+
+public class CompositeRepresentation implements Representation {
+  protected Map<String, Representation> representations;
+
+  /*Initialize the CompositeRepresentation according to a certain correspondence of the "name of representation" and "representation" itself.*/
+  public CompositeRepresentation(Map<String, Representation> representations) {
+    this.representations = representations;
+  }
+
+  /*Initialize the CompositeRepresentation when the user does not have the correspondence of the "name of representation" and "representation" itself yet.*/
+  public CompositeRepresentation() {
+    representations = new HashMap<>();
+  }
+
+  /*Implementation of the `toString` method.*/
+  @Override
+  public String toStringOf(Object object) {
+    for (Map.Entry<String, Representation> entry : representations.entrySet()) {
+      String value = entry.getValue().toStringOf(object);
+      if (value != null)
+        return value;
+    }
+    return STANDARD_REPRESENTATION.toStringOf(object);
+  }
+
+  /*Implementation of the `toString` method when a specific representation is desired.*/
+  public String toStringOf(Object object, String representationName) {
+    if (representations.containsKey(representationName))
+      return representations.get(representationName).toStringOf(object);
+    return null;
+  }
+
+  /*Change the correspondence of the "name of representation" and "representation"*/
+  public void setRepresentations(Map<String, Representation> representations) {
+    this.representations = representations;
+  }
+
+  /*Get the "representation"*/
+  public Map<String, Representation> getRepresentations() {
+    return representations;
+  }
+
+  /*Add an entry of the correspondence of the "name of representation" and "representation"*/
+  public boolean addRepresentation(String representationName, Representation representation) {
+    if (representations.containsKey(representationName))
+      return false;
+    representations.put(representationName, representation);
+    return true;
+  }
+
+  /*Delete an entry of the correspondence of the "name of representation" and "representation"*/
+  public boolean deleteRepresentation(String representationName) {
+    if (representations.containsKey(representationName)) {
+      representations.remove(representationName);
+      return true;
+    }
+    return false;
+  }
+
+  /*Maybe the method is redundant here.*/
+  @Override
+  public String unambiguousToStringOf(Object object) {
+    for (Map.Entry<String, Representation> entry : representations.entrySet()) {
+      String value = entry.getValue().unambiguousToStringOf(object);
+      if (value != null)
+        return value;
+    }
+    return null;
+  }
+}
--- a/src/main/java/org/assertj/core/presentation/CompositeRepresentation.java
+++ b/src/main/java/org/assertj/core/presentation/CompositeRepresentation.java
@@ -1,72 +1,28 @@
 package org.assertj.core.presentation;
 
-import java.util.HashMap;
+import java.util.ArrayList;
 import java.util.Map;
 
 import static org.assertj.core.presentation.StandardRepresentation.STANDARD_REPRESENTATION;
 
 public class CompositeRepresentation implements Representation {
-  protected Map<String, Representation> representations;
+  protected final Map<String, Representation> representationsByName;
 
-  /*Initialize the CompositeRepresentation according to a certain correspondence of the "name of representation" and "representation" itself.*/
   public CompositeRepresentation(Map<String, Representation> representations) {
-    this.representations = representations;
+    if (representations == null) throw new IllegalArgumentException();
+    this.representationsByName = representations;
   }
 
-  /*Initialize the CompositeRepresentation when the user does not have the correspondence of the "name of representation" and "representation" itself yet.*/
-  public CompositeRepresentation() {
-    representations = new HashMap<>();
-  }
-
-  /*Implementation of the `toString` method.*/
   @Override
   public String toStringOf(Object object) {
-    for (Map.Entry<String, Representation> entry : representations.entrySet()) {
-      String value = entry.getValue().toStringOf(object);
-      if (value != null)
-        return value;
-    }
-    return STANDARD_REPRESENTATION.toStringOf(object);
-  }
-
-  /*Implementation of the `toString` method when a specific representation is desired.*/
-  public String toStringOf(Object object, String representationName) {
-    if (representations.containsKey(representationName))
-      return representations.get(representationName).toStringOf(object);
-    return null;
-  }
-
-  /*Change the correspondence of the "name of representation" and "representation"*/
-  public void setRepresentations(Map<String, Representation> representations) {
-    this.representations = representations;
-  }
-
-  /*Get the "representation"*/
-  public Map<String, Representation> getRepresentations() {
-    return representations;
-  }
-
-  /*Add an entry of the correspondence of the "name of representation" and "representation"*/
-  public boolean addRepresentation(String representationName, Representation representation) {
-    if (representations.containsKey(representationName))
-      return false;
-    representations.put(representationName, representation);
-    return true;
-  }
-
-  /*Delete an entry of the correspondence of the "name of representation" and "representation"*/
-  public boolean deleteRepresentation(String representationName) {
-    if (representations.containsKey(representationName)) {
-      representations.remove(representationName);
-      return true;
-    }
-    return false;
+    return (new ArrayList<Representation>(this.representationsByName.values()))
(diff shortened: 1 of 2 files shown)


src/main/java/org/assertj/core/presentation/CompositeRepresentation.java
```diff
@@ -0,0 +1,76 @@
...
+    for (Map.Entry<String, Representation> entry : representations.entrySet()) {
+      String value = entry.getValue().toStringOf(object);
+      if (value != null)
+        return value;
+    }
+    return STANDARD_REPRESENTATION.toStringOf(object);
+  }
+
+  /*Implementation of the `toString` method when a specific representation is desired.*/
+  public String toStringOf(Object object, String representationName) {
```
joel-costigliola, line 33: I'm not sure we have a use case for this one, can you show an example that would justify adding it?

src/main/java/org/assertj/core/presentation/CompositeRepresentation.java
```diff
@@ -0,0 +1,76 @@
...
+    if (representations.containsKey(representationName))
+      return representations.get(representationName).toStringOf(object);
+    return null;
+  }
+
+  /*Change the correspondence of the "name of representation" and "representation"*/
+  public void setRepresentations(Map<String, Representation> representations) {
+    this.representations = representations;
+  }
```
joel-costigliola, line 42: I would rather we initialize `representations` in the constructor only and make this class immutable, this means we don't need this method.

src/main/java/org/assertj/core/presentation/CompositeRepresentation.java
```diff
@@ -0,0 +1,76 @@
...
+
+  /*Get the "representation"*/
+  public Map<String, Representation> getRepresentations() {
+    return representations;
+  }
```
joel-costigliola, line 47: remove unless we have a use case for it

src/main/java/org/assertj/core/presentation/CompositeRepresentation.java
```diff
@@ -0,0 +1,76 @@
...
+
+  /*Add an entry of the correspondence of the "name of representation" and "representation"*/
+  public boolean addRepresentation(String representationName, Representation representation) {
+    if (representations.containsKey(representationName))
+      return false;
+    representations.put(representationName, representation);
+    return true;
+  }
```
joel-costigliola, line 55: remove to make this class immutable

src/main/java/org/assertj/core/presentation/CompositeRepresentation.java
```diff
@@ -0,0 +1,76 @@
...
+
+  /*Delete an entry of the correspondence of the "name of representation" and "representation"*/
+  public boolean deleteRepresentation(String representationName) {
```
joel-costigliola, line 58: remove to make this class immutable

src/main/java/org/assertj/core/presentation/CompositeRepresentation.java
```diff
@@ -0,0 +1,76 @@
...
+
+  /*Implementation of the `toString` method.*/
+  @Override
+  public String toStringOf(Object object) {
+    for (Map.Entry<String, Representation> entry : representations.entrySet()) {
+      String value = entry.getValue().toStringOf(object);
+      if (value != null)
+        return value;
+    }
+    return STANDARD_REPRESENTATION.toStringOf(object);
```
joel-costigliola, line 29: I think we could use a more functional approach:

- in the constructor, in addition to setup the representations map, also setup the representations in a List
- use this list here and stream + map to the get the `toStringOf(object)`
- use findAny
- chain with orElse to return STANDARD_REPRESENTATION.toStringOf(object)

this would only work if we make the class immutable and initialize the representations map in the constructor.

src/test/java/org/assertj/core/presentation/CompositeRepresentation_Test.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.presentation.BinaryRepresentation.BINARY_REPRESENTATION;
+import static org.assertj.core.presentation.HexadecimalRepresentation.HEXA_REPRESENTATION;
+import static org.assertj.core.presentation.UnicodeRepresentation.UNICODE_REPRESENTATION;
+
+public class CompositeRepresentation_Test extends AbstractBaseRepresentationTest {
```
joel-costigliola, line 13: add GIVEN WHEN THEN and use `then` as indicated in https://github.com/assertj/assertj-core/blob/main/CONTRIBUTING.md

joel-costigliola (commented):
First round of comments, thanks @Annette0127

src/main/java/org/assertj/core/presentation/CompositeRepresentation.java
Annette0127: Thank you very much for your suggestions! As for this function, if there are many representations for one object, for example, we want to represent a number in binary or hexadecimal way in different situations, we can use `key` of the `representationsByName` to indicate which representation we want to have. So this function was added.

Annette0127: If the example I imagine does not exist, I suppose `List` is better than `Map` in this situation, and `toStringOf(Object object, String representationName)` can be removed.

Annette0127:
Hi, I pushed a new version of the code to this draft according to your advice several days before. If there are some improvements to be made, please tell me.

src/main/java/org/assertj/core/presentation/CompositeRepresentation.java
```diff
@@ -0,0 +1,32 @@
...
+
+  @Override
+  public String toStringOf(Object object) {
+    return (new ArrayList<Representation>(this.representationsByName.values()))
+      .stream().map(x -> x.toStringOf(object)).findAny().orElse(STANDARD_REPRESENTATION.toStringOf(object));
+  }
+
+  /*Maybe the method is redundant here.*/
+  @Override
+  public String unambiguousToStringOf(Object object) {
```
joel-costigliola, line 24: for consistency, I think this should follow the same algorithm as `toStringOf` to decide which `unambiguousToStringOf` to return.

src/main/java/org/assertj/core/presentation/CompositeRepresentation.java
```diff
@@ -0,0 +1,32 @@
...
+
+  public CompositeRepresentation(Map<String, Representation> representations) {
+    if (representations == null) throw new IllegalArgumentException();
+    this.representationsByName = representations;
+  }
+
+  @Override
+  public String toStringOf(Object object) {
+    return (new ArrayList<Representation>(this.representationsByName.values()))
+      .stream().map(x -> x.toStringOf(object)).findAny().orElse(STANDARD_REPRESENTATION.toStringOf(object));
```
joel-costigliola, line 19: this is going to change when we use priority, also this was not really following https://github.com/assertj/assertj-core/issues/2048#issuecomment-733214647 since there is no check that the representation returned null (meaning that it did not represented the value).

joel-costigliola (changes requested):
second review @Annette0127, let us know if the solution suggested by @filiphr  is not clear for you as your implementation does not follow it completely.

src/main/java/org/assertj/core/presentation/CompositeRepresentation.java
```diff
@@ -0,0 +1,32 @@
...
+public class CompositeRepresentation implements Representation {
+  protected final Map<String, Representation> representationsByName;
+
+  public CompositeRepresentation(Map<String, Representation> representations) {
+    if (representations == null) throw new IllegalArgumentException();
+    this.representationsByName = representations;
+  }
+
+  @Override
+  public String toStringOf(Object object) {
```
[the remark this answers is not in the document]
Annette0127: @joel-costigliola In this part I thought priority could be determined by the insertion order directly, so I did not add a `priority` field in `Representation`.  So we actually need to add a field called `priority` in `Representatioin`?  And I do not fully understand when  should we print a message, in that case, saying that there were 2 representations with the same priority? Could you please explain more details? Thank you very much! 😊

src/main/java/org/assertj/core/presentation/CompositeRepresentation.java
Annette0127: @joel-costigliola I am not quite sure what should I return if the representation returns null. According to the comment in #2048, I saw the null will be returned if the representation returns null by default. So could you give me more details so that I can modify the code? Thanks!😁

joel-costigliola: In our case null means the representation did not have a specific representation for the object, this is true except for the standard representation (at some point we need to return a representation for null).
This means we need to return the first non null values from the custom representations, if all of them returns null then we use the standard representation.
Does that make sense?

Annette0127: Oh, yes! I understand. Thank you very much!

src/main/java/org/assertj/core/presentation/CompositeRepresentation.java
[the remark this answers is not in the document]
Annette0127: @joel-costigliola Hi! I am very sorry to bother you again because I am not sure whether you have seen this question since you answered another question related to this pr several hours before but this one remains silent. Looking forward to your reply! 😉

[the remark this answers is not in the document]
Annette0127: Since the `Representation` is an `Interface`, where I believe a `int priority` field can not be added to implement this mechanism, I wonder if it is I got a wrong idea or we have to think about another implementation of the `priority` part.☺️😳

joel-costigliola:
Integrated in de1b3597 thanks @Annette0127!

PR #2182, opened 2021-04-21
 fix issue 971:to have ability to ignore all assert when expected value is null

### LINKED ISSUE #971: New Feature: to have ability to ignore all assert when expected value is null

wuqihan837:
#### Check List:
* Fixes #971
* Unit tests : YES 
In assertions.java, I add two methods for most types. The conditionalAssertThat method adds a boolean type argument, when it's false, the assertThat will be skipped. The notNullAssertThat method ask the argument called expected to not be null, if it's null, the assertThat will be skipped. Also, I add some tests to test these two methods.
I'm a student, and my ability to develop a project is poor. I just tried making a pull request. Maybe my pull request is totally wrong and doesn't meet your requirements. If it happens, I'm sorry about that and you can ignore my pull request and don't merge it.

### DIFF
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -169,6 +169,8 @@ public class Assertions implements InstanceOfAssertFactories {
     return AssertionsForInterfaceTypes.assertThat(actual);
   }
 
+
+
   /**
    * Create assertion for {@link IntPredicate}.
    *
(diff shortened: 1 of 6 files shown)


joel-costigliola:
Sorry @wuqihan837 but this does not solve the issue, I'm thus closing this PR.
I recommend to look for issues with "Good first issue" that are easy to address for less experienced developers.

PR #2183, opened 2021-04-22
Fix possible heap pollution in AtomicReferenceArrayAssert

filiphr:
Seems like I forgot the AtomicReferenceArrayAssert in https://github.com/assertj/assertj-core/commit/3d5841fa110c029fc2104fd7425e02dd1ed19384. This PR fixes the possible heap pollution in that class and fixes some other heap pollution for the `ForProxy` methods by using arrays instead of varArgs.


joel-costigliola:
I missed it too, thanks.

PR #2186, opened 2021-04-24
Warning fixes

filiphr:
Warning fixes in AssertJ codebase and warnings in user code.

The first commit is for warnings in the AssertJ codebase and the second one are warnings that will be shown to users.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractObjectArrayAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractObjectArrayAssert.java
@@ -88,6 +88,7 @@ import org.assertj.core.util.introspection.IntrospectionError;
  * @author Mateusz Haligowski
  * @author Lovro Pandzic
  */
+// suppression of deprecation works in Eclipse to hide warning for the deprecated classes in the imports
 @SuppressWarnings("deprecation")
 public abstract class AbstractObjectArrayAssert<SELF extends AbstractObjectArrayAssert<SELF, ELEMENT>, ELEMENT> extends
     AbstractAssert<SELF, ELEMENT[]>
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -1302,6 +1302,7 @@ public class Assertions implements InstanceOfAssertFactories {
    * @return the created assertion object.
    * @since 3.20.0
    */
+  @CanIgnoreReturnValue
   public static <T> ObjectAssert<T> assertWith(T actual, Consumer<T> requirements) {
     return assertThat(actual).satisfies(requirements);
   }
--- a/src/main/java/org/assertj/core/internal/Classes.java
+++ b/src/main/java/org/assertj/core/internal/Classes.java
@@ -249,8 +249,10 @@ public class Classes {
    * @throws AssertionError if {@code actual} is {@code null}.
    * @throws AssertionError if the actual {@code Class} doesn't contains all of these annotations.
    */
+  // varargs are not used in order to avoid a "Possible heap pollution" warning.
+  // The method cannot be made final and annotated with @SafeVarargs because the class is mocked in tests
   public void assertContainsAnnotations(AssertionInfo info, Class<?> actual,
-                                        @SuppressWarnings("unchecked") Class<? extends Annotation>... annotations) {
+                                        Class<? extends Annotation>[] annotations) {
     assertNotNull(info, actual);
     Set<Class<? extends Annotation>> expected = newLinkedHashSet(annotations);
     Set<Class<? extends Annotation>> missing = new LinkedHashSet<>();
--- a/src/main/java/org/assertj/core/internal/Iterables.java
+++ b/src/main/java/org/assertj/core/internal/Iterables.java
@@ -1162,8 +1162,10 @@ public class Iterables {
     return Optional.empty();
   }
 
+  // varargs are not used in order to avoid a "Possible heap pollution" warning.
+  // The method cannot be made final and annotated with @SafeVarargs because the class is mocked in tests
   public <E> void assertSatisfiesExactly(AssertionInfo info, Iterable<? extends E> actual,
-                                         @SuppressWarnings("unchecked") Consumer<? super E>... allRequirements) {
+                                         Consumer<? super E>[] allRequirements) {
     assertNotNull(info, actual);
     assertHasSameSizeAs(info, actual, allRequirements); // TODO
     List<E> actualAsList = newArrayList(actual);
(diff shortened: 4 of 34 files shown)


src/main/java/org/assertj/core/api/AbstractObjectArrayAssert.java
```diff
@@ -88,7 +88,6 @@
  * @author Mateusz Haligowski
  * @author Lovro Pandzic
  */
-@SuppressWarnings("deprecation")
```
joel-costigliola, line 91: this one is needed because of the now deprecated but imported `FieldByFieldComparator` and `IgnoringFieldsComparator` (line 62-63)

src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -1302,6 +1302,7 @@ public static ThrowableAssert assertThatCode(ThrowingCallable shouldRaiseOrNotTh
    * @return the created assertion object.
    * @since 3.20.0
    */
+  @CanIgnoreReturnValue
```
joel-costigliola, line 1305: good catch!

src/main/java/org/assertj/core/internal/Classes.java
```diff
@@ -250,7 +250,7 @@ public void assertIsNotFinal(AssertionInfo info, Class<?> actual) {
    * @throws AssertionError if the actual {@code Class} doesn't contains all of these annotations.
    */
   public void assertContainsAnnotations(AssertionInfo info, Class<?> actual,
-                                        @SuppressWarnings("unchecked") Class<? extends Annotation>... annotations) {
+                                        Class<? extends Annotation>... annotations) {
```
joel-costigliola, line 253: removing this leads to a "Type safety: Potential heap pollution via varargs parameter annotations" warning, adding `@SafeVarargs` is possible here since this is not proxyed or subclassed.

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1163,7 +1163,7 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
   }
 
   public <E> void assertSatisfiesExactly(AssertionInfo info, Iterable<? extends E> actual,
-                                         @SuppressWarnings("unchecked") Consumer<? super E>... allRequirements) {
+                                         Consumer<? super E>... allRequirements) {
```
joel-costigliola, line 1166: removing this leads to a "Type safety: Potential heap pollution via varargs parameter annotations" warning, adding @SafeVarargs is possible here since this is not proxyed or subclassed.

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -572,7 +570,7 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
    * @throws AssertionError if the actual map not contains the given key.
    */
   public <K, V> void assertContainsKeys(AssertionInfo info, Map<K, V> actual,
-                                        @SuppressWarnings("unchecked") K... keys) {
+                                        K... keys) {
```
joel-costigliola, line 573: surprisingly eclipse does complain about the head pollution when I remove `@SuppressWarnings("unchecked")`

joel-costigliola (commented):
just some minor comments

src/main/java/org/assertj/core/api/AbstractObjectArrayAssert.java
filiphr: Interesting. When I compile the the class with Java I only get warnings on lines: the imports (line 62, 63 and 68) and also on line 426 for the `containsOnlyElementsOf`.

When I bring back the `SuppressWarnings` the warning on line 426 is gone, but the ones for the imports are still there.

src/main/java/org/assertj/core/internal/Classes.java
filiphr: For me the `SuppressWarnings` have no influence on the possible heap pollution (when compiling with `-Xlint:unchecked` in IntelliJ). 

We can't make it `final` because `Classes` is mocked in the tests. If we make it final tests start to fail (already tried that 😄 ). However, what we can do is to use an array instead of varargs. That's how I got rid of those errors in the `ForProxy` methods.

filiphr: Another solution would be to annotated the method with `@SuppressWarnings("unchecked")`. The heap pollution is gone for me with that.

src/main/java/org/assertj/core/internal/Iterables.java
filiphr: Same comment as for the `Classes`

src/main/java/org/assertj/core/internal/Maps.java
filiphr: Same comment as for the `Classes`. It seems that the ecj compiler is slightly different that the javac in this area.

filiphr:
Interesting findings @joel-costigliola. I rechecked and uses the `javac` compiler in IntelliJ to complile the classes and the warnings you are seeing are there for me with and without the `@SuppressWarnings`. Not sure how you want to proceed with this.

There are 2 possible sollutions:

* Use array instead of varargs in the internal classes
*  Add `@SuppressWarnings("unchecked")` on the methods

src/main/java/org/assertj/core/internal/Classes.java
joel-costigliola: Using an array is the best solution IMO, varargs are great for user facing APIs but for internal code arrays are fine.

src/main/java/org/assertj/core/api/AbstractObjectArrayAssert.java
joel-costigliola: if we could apply the SuppressWarnings at the import level that would be great but since it's not possible let's bring back this SuppressWarnings, it's not fine grained but I don't want to see deprecation warnings in my eclipse IDE ;-)

joel-costigliola:
let's use arrays to fix the warnings in non api code.

filiphr:
I'll bring back the `deprecation` warnings since they work correctly in Eclipse. I will also switch ti using arrays to fix the warnings in non API code.

filiphr:
@joel-costigliola I brought back the suppression (and added a comment why are there) and I also switched to using arrays.

joel-costigliola:
Integrated with a few minor changes thanks @filiphr!

filiphr:
Thanks a lot @joel-costigliola

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -637,8 +661,10 @@ public void assertHasSameSizeAs(AssertionInfo info, Map<?, ?> map, Map<?, ?> oth
    * @throws AssertionError if the given {@code Map} does not contain the given keys or if the given {@code Map}
    *           contains keys that are not in the given array.
    */
+  // varargs are not used in order to avoid a "Possible heap pollution" warning.
+  // The method cannot be made final and annotated with @SafeVarargs because the class is mocked in tests
```
scordio, line 665: I'm currently doing some cleanup in this class and I would like to remove these comments as the callers don't have vararg parameters (e.g., `ForProxy` methods).

Nothing suggests that we should switch to varargs in internal classes so I don't see much value in keeping them. If you don't mind, I'll remove them.

joel-costigliola: I agree, this comment is only relevant in the API layer.

filiphr: I am completely fine in removing the comments. I put them there for future reviews of the class in case someone wonders why `varargs` or not used

PR #2188, opened 2021-04-25
FIX #2011 wrap in method withAdditionalFailMessage

### LINKED ISSUE #2011: Add functionality to append supplied String (or Supplier<String>) to default error message
#### Summary
Currently we can call withFailMessage with either a String or Supplier<String>. This is nice but the default error message can be useful **in addition** to a custom error message. Please add this kind of method with both signatures.

Thanks for the consideration!

#### Example

```java
// logs "Inconceivable!" only
Assertions.assertThat("string").withFailMessage("Inconceivable!").contains("sub");
/** Would log the following:
Expecting:
 <"string">
to contain:
 <"sub">
Inconceivable!
 */
Assertions.assertThat("string").withAdditionalFailMessage("no").contains("sub");
```

WhiteLittleDuck:
#### Check List:
* Fixes #2011 
* Unit tests : NO 

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -652,6 +652,41 @@ public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, 
     return overridingErrorMessage(newErrorMessage, args);
   }
 
+  //fix issue 2011
+  /**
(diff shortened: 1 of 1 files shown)


joel-costigliola:
@WhiteLittleDuck I'm closing this PR we haven't even decided to address  #2011 yet and if we had this PR would not solve it, moreover this not following the contributing guidelines.

WhiteLittleDuck:
Sorry about my carelessness. If there is something else about this issue, I would like to help!

PR #2103, opened 2021-01-19
Mapped-Condition

stbischof:
**Mapped**: A Condition that describes the way of Mapping to an nested Condition.

PR #2193, opened 2021-04-30
Add Mapped Condition

joel-costigliola:
Fixes #2103

### DIFF
--- a/src/main/java/org/assertj/core/condition/MappedCondition.java
+++ b/src/main/java/org/assertj/core/condition/MappedCondition.java
@@ -15,7 +15,6 @@ package org.assertj.core.condition;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 
-import java.util.function.BiFunction;
 import java.util.function.Function;
 
 import org.assertj.core.annotations.Beta;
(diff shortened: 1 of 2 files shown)


joel-costigliola:
@stbischof I have worked on your PR and created this one, can you review it (your turn! :smile:) to check if I have properly captured your idea - please check the javadoc.

A few notable points:
- I have renamed `mapped` to `mappedCondition` which feels more expressive to me.
- I have revamped the javadoc
- refactor the tests.

stbischof:
I will have a look. But at first I would say thank you very much!

stbischof (changes requested):
just found  the old name in the JavaDoc

src/main/java/org/assertj/core/condition/MappedCondition.java
```diff
@@ -0,0 +1,121 @@
...
+   * @param <TO> the type of object the nested condition accept.
+   * @param mapping the Function that maps the value to test to the a value for the nested condition.
+   * @param mappingDescription describes the mapping verbal.
+   * @param condition the nested condition to evaluate.
+   * @return the created {@code Mapped}.
+   * @throws NullPointerException if the given condition is {@code null}.
+   * @throws NullPointerException if the given mapping is {@code null}.
+   */
+  public static <FROM, TO> MappedCondition<FROM, TO> mappedCondition(Function<FROM, TO> mapping, String mappingDescription,
+                                                                     Condition<TO> condition) {
```
nach-o-man, line 63: @joel-costigliola just curious if there is any reason **mappingDescription** is not last argument? That way a varargs method can be introduced like in Condition class.

joel-costigliola: good idea!

joel-costigliola:
@stbischof @nach-o-man I have further simplified the code by having only one method to set the mapping description and made it protected so that it can be overridden.
I also simplified a bit the description itself, check the latest commit https://github.com/assertj/assertj-core/pull/2193/commits/0a0af1a587a5767fa1188c63ca80d6f5ec0a2e96

stbischof:
Lgtm

nach-o-man:
> Lgtm

I sevond that.

joel-costigliola:
thanks for the review!

PR #2194, opened 2021-05-01
Fix 'toString' method of subclasses of AtomicReference isn't used for…

### LINKED ISSUE #2192: 'toString' method of subclasses of AtomicReference isn't used for presentation
#### Summary
Sometimes we extend AtomicReference in our classes to store some internal data. (Of course, usually it's better to use delegation instead of inheritance, but it may increase memory footprint and it may be undesirable in some cases.) `StandardRepresentation.toStringOf` returns special representation for instances of such classes, which doesn't include result of 'toString' call.

#### Example

Consider the following class
```java
class MyData extends AtomicReference<String> {
  private String description;

  MyData(String description) {
    this.description = description;
  }

  @Override
  public String toString() {
    return "MyData[" + description + "]";
  }
}
```
If you use its instances in assertions like 
```java
    assertThat(new MyData[] {new MyData("foo")}).contains(new MyData("bar"));
```
you'll get error messages like
> Expecting MyData[]:
  [AtomicReference[null]]
to contain:
  [AtomicReference[null]]
but could not find the following mydata(s):
  [AtomicReference[null]]

which aren't helpful.

Lanninger08:
#### Check List:
* Fixes #2192 
* Unit tests : YES

Thanks to the great help of all comments and my teammate @sustc11810424 

We modified `hasOverriddenToString` to check whether the declaring class of `toString` is the object's own class. Changes are applied to the three types of AtomicReferences mentioned,  with two tests for each. And two more tests are to make sure things still work for non-collection `Iterable`s.

### DIFF
--- a/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
+++ b/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
@@ -111,7 +111,7 @@ public class StandardRepresentation implements Representation {
 
   private static final Map<Class<?>, Function<?, String>> customFormatterByType = new HashMap<>();
   private static final Class<?>[] TYPE_WITH_UNAMBIGUOUS_REPRESENTATION = { Date.class, LocalDateTime.class, ZonedDateTime.class,
-      OffsetDateTime.class, Calendar.class };
+    OffsetDateTime.class, Calendar.class };
 
   protected enum GroupType {
     ITERABLE("iterable"), ARRAY("array");
@@ -213,9 +213,9 @@ public class StandardRepresentation implements Representation {
     if (object instanceof OffsetDateTime) return toStringOf((OffsetDateTime) object);
     if (object instanceof ZonedDateTime) return toStringOf((ZonedDateTime) object);
     if (object instanceof LongAdder) return toStringOf((LongAdder) object);
-    if (object instanceof AtomicReference) return toStringOf((AtomicReference<?>) object);
-    if (object instanceof AtomicMarkableReference) return toStringOf((AtomicMarkableReference<?>) object);
-    if (object instanceof AtomicStampedReference) return toStringOf((AtomicStampedReference<?>) object);
+    if (object instanceof AtomicReference && !hasOverriddenToString(object)) return toStringOf((AtomicReference<?>) object);
+    if (object instanceof AtomicMarkableReference && !hasOverriddenToString(object)) return toStringOf((AtomicMarkableReference<?>) object);
+    if (object instanceof AtomicStampedReference && !hasOverriddenToString(object)) return toStringOf((AtomicStampedReference<?>) object);
     if (object instanceof AtomicIntegerFieldUpdater) return AtomicIntegerFieldUpdater.class.getSimpleName();
     if (object instanceof AtomicLongFieldUpdater) return AtomicLongFieldUpdater.class.getSimpleName();
     if (object instanceof AtomicReferenceFieldUpdater) return AtomicReferenceFieldUpdater.class.getSimpleName();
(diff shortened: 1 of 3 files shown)


src/main/java/org/assertj/core/presentation/StandardRepresentation.java
```diff
@@ -630,19 +631,19 @@ private static String representGroup(List<String> representedElements, String st
 
   private String toStringOf(ChangeDelta<?> changeDelta) {
     return String.format("Changed content at line %s:%nexpecting:%n  %s%nbut was:%n  %s%n",
-                         changeDelta.lineNumber(),
-                         formatLines(changeDelta.getOriginal().getLines()),
-                         formatLines(changeDelta.getRevised().getLines()));
+      changeDelta.lineNumber(),
+      formatLines(changeDelta.getOriginal().getLines()),
+      formatLines(changeDelta.getRevised().getLines()));
```
joel-costigliola, line 636: use assertj formatting preferences to avoid these kind of changes, the formatting preferences are in https://github.com/assertj/assertj-core/tree/main/src/ide-support

src/main/java/org/assertj/core/presentation/StandardRepresentation.java
```diff
@@ -248,10 +248,11 @@ public String toStringOf(Object object) {
     return fallbackToStringOf(object);
   }
 
-  private static boolean hasOverriddenToString(Iterable<?> iterable) {
+  private static boolean hasOverriddenToString(Object object) {
     try {
-      Method method = iterable.getClass().getMethod("toString");
+      Method method = object.getClass().getMethod("toString");
       Class<?> declaringClass = method.getDeclaringClass();
+//      return object.getClass()==declaringClass;
```
joel-costigliola, line 255: remove this comment unless there is a reason too keep it

src/test/java/org/assertj/core/presentation/StandardRepresentation_toStringOf_AtomiceReferences_Test.java
```diff
@@ -0,0 +1,165 @@
...
+
+import org.junit.Test;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.*;
+import java.util.concurrent.atomic.AtomicMarkableReference;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.atomic.AtomicStampedReference;
+
+public class StandardRepresentation_toStringOf_AtomiceReferences_Test {
```
joel-costigliola, line 23: please follow the https://github.com/assertj/assertj-core/blob/main/CONTRIBUTING.md
- add GIVEN WHEN THEN
- use `then` instead of `assertThat`

src/test/java/org/assertj/core/presentation/StandardRepresentation_toStringOf_AtomiceReferences_Test.java
```diff
@@ -0,0 +1,165 @@
...
+  static Representation standardRepresentation = new StandardRepresentation();
```
joel-costigliola, line 24: use `StandardRepresentation.STANDARD_REPRESENTATION` instead, no need for a field, use statically imported `STANDARD_REPRESENTATION` directly

joel-costigliola:
Integrated thanks @Lanninger08! 
I made some changes so that toString redefined AtomicReference subclasses would be used (the code from your PR was simply checking if toString was overridden or not, it turns out that it is the case in `AtomicReference` so even if a sublcass would redefine it, it would use AssertJ representation instead of the overriddent toString).

PR #2195, opened 2021-05-04
Configure pitest

hcoles:
This PR adds a profile so pitest can be easily run against assertj core

```bash
mvn -Ppitest test-compile
```

The plugin has been bound to the test-compile (rather than test) as an easy way of preventing surefire from also running. As pitest runs the tests itself, additional running surefire just slows feedback.

A full analysis of the project takes around 45 minutes.

The PR also adds git integration, this allows just local changes to be analysed with

```bash
mvn -Ppitest -Dfeatures="+GIT" test-compile
```

Or for changes to be brought into scope from git history

```bash
mvn -Ppitest -Dfeatures="+GIT(from[HEAD~10])" test-compile
```

Once merged, this can be built on in a future PR to provide mutation testing feedback on all incoming PRs as described here -> https://blog.pitest.org/dont-let-your-code-dry/

The PR also modifies several tests to remove test order dependencies, and to exclude the pitest test runners from stack traces.

### DIFF
--- a/src/test/java/org/assertj/core/api/abstract_/AbstractAssert_equal_hashCode_Test.java
+++ b/src/test/java/org/assertj/core/api/abstract_/AbstractAssert_equal_hashCode_Test.java
@@ -39,7 +39,11 @@ class AbstractAssert_equal_hashCode_Test {
   @SuppressWarnings("deprecation")
   void should_not_fail_when_equals_exceptions_is_deactivated() {
     AbstractAssert.throwUnsupportedExceptionOnEquals = false;
-    assertions.equals("anotherString");
+    try {
+      assertions.equals("anotherString");
+    } finally {
+      AbstractAssert.throwUnsupportedExceptionOnEquals = true;
+    }
   }
 
   @Test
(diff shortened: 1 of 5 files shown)


src/test/java/org/assertj/core/api/abstract_/AbstractAssert_equal_hashCode_Test.java
```diff
@@ -39,7 +39,11 @@ void should_fail_because_not_supported_operation() {
   @SuppressWarnings("deprecation")
   void should_not_fail_when_equals_exceptions_is_deactivated() {
     AbstractAssert.throwUnsupportedExceptionOnEquals = false;
-    assertions.equals("anotherString");
+    try {
+      assertions.equals("anotherString");
+    } finally {
+      AbstractAssert.throwUnsupportedExceptionOnEquals = true;
+    }
```
joel-costigliola, line 46: good catch!

joel-costigliola:
Thanks a lot @hcoles!

PR #2200, opened 2021-05-06
Update ShouldBeEqual assertion message

### LINKED ISSUE #2199: Remove space in default error message for ShouldBeEqual to help IntelliJ diff detection
Referencing this StackOverflow ticket: https://stackoverflow.com/questions/10934743/formatting-output-so-that-intellij-idea-shows-diffs-for-two-texts

It seems that IntelliJ expects assertion errors to be in this format in order to have nicely rendered diffs:

`expected: xxx but was: yyy`

But AssertJ would format this with an extra space after the `was`:

`expected: xxx but was : yyy`

Which would cause it to fail the regex, and IntelliJ defaults to just printing the error string. I tried patching my AssertJ `ShouldBeEqual` locally without that extra space, and IntelliJ was then able to detect it properly.

Would it be possible to remove the extraneous space after the `was` so that the test output can automatically be detected as a diff?

Thanks!

Link to message: https://github.com/assertj/assertj-core/blob/main/src/main/java/org/assertj/core/error/ShouldBeEqual.java#L44

rpeng:
Move the trailing space after "was" to a leading space before "but" so that IntellIJ's diff regexes can pick it up

#### Check List:
* Fixes #2199 (ignore if not applicable)
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/Descriptable.java
+++ b/src/main/java/org/assertj/core/api/Descriptable.java
@@ -46,7 +46,7 @@ public interface Descriptable<SELF> {
    * } catch (AssertionError e) {
    *   assertThat(e).hasMessage(&quot;[check Frodo's age]\n
    *                             expected: 33\n
-   *                             but was : 50&quot;);
+   *                              but was: 50&quot;);
    * }</code></pre>
    *
    * @param description the new description to set.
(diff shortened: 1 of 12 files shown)


joel-costigliola:
Integrated thanks @rpeng for the quick fix!

Issue #2189, opened 2021-04-27
Refactoring: introduce IntrospectionStrategy in the recursive comparison


joel-costigliola:
#### Summary

The recursive comparison would benefit from a clear introspection strategy when gathering the field to compare and getting their values.

The recursive comparison is a bit inconsistent, it introspects fields to get the list of fields to compare but uses a different strategy to read their values by looking first to a getter then reading the field directly or looking for a map value (key being the field name).

The IntrospectionStrategy would define what to compare and how to read the values to compare.

This is needed for: 
* https://github.com/assertj/assertj-core/issues/2108 
* https://github.com/assertj/assertj-core/issues/2149

edit: other related issues
* https://github.com/assertj/assertj-core/issues/2554
* https://github.com/assertj/assertj/issues/2314

scordio:
@nach-o-man could you please comment here so I can assign the issue to you?

nach-o-man:
Comment

Issue #2149, opened 2021-03-16
[Feature request] Add recursive comparison support to do the value resolution only by field (reflection)


fegbers:
Hi,

#### Summary
If you have a getter that modifies the value or only returns a subset of the value, the [current resolving strategy](https://assertj.github.io/doc/#how-field-values-are-resolved) is error prone / unusable for recursive comparison tests (see example below).
Perhaps it's possible to add the support to do the value resolution only by field via reflection? Perhabs a boolean configuration could override the [current implementation](https://github.com/assertj/assertj-core/blob/a49869bf382f19a3dd4ada49e655c07f90fab05a/src/main/java/org/assertj/core/util/introspection/PropertyOrFieldSupport.java#L70)?
Currently, we use the recursive comparison for all our tests and love the fact that new properties of a class are always tested immediately. Thanks for the new kind of testing 👍🏼 

#### Example
Note: This is only an example to illustrate the problem. The easy solution would be the usage of "clean" getters. But if you have a public API you couldn't just change the internal behaviour…

```java
public class Example
{
    private Collection<String> value;

    public Example(Collection<String> value)
    {
        this.value = value;
    }

    public String getValue()
    {
        if (value == null)
        {
            return null;
        }
        return value.iterator().next();
    }
}

class ExampleTest
{
    @Test
    void testRecCom() // should fail
    {
        Assertions.assertThat(new Example(List.of("A", "B")))
                .usingRecursiveComparison()
                .ignoringAllOverriddenEquals()
                .withStrictTypeChecking()
                .isEqualTo(new Example(List.of("A")));
    }
}
```

Greetings!

joel-costigliola:
The best way to go would be to specify in the configuration to only use fields, something like: `comparingOnlyFields()` and change the implementation to introspect fields only.

sustc11810424:
Hello there. What about adding a bit logic to **introspection/PropertyOrFieldSupport/getSimpleValue** such that if **comparingOnlyFields** is set to true then we skip the property part and look straight into fields?

`public Object getSimpleValue(String name, Object input) {
    if (comparingOnlyFields){ //
        // try as a field, then map key
        ...
    }else{
        // try as a property, then field, then map key
        ...
  }`

I can implement this with tests if you'd like. 
[...]

PR #2201, opened 2021-05-06
POC introduction of introspectionStrategy

nach-o-man:
Here is a brief implementation of POC for #2189 and is far from being complete.
I am using a test example from #2149 to show that it is working. 
@joel-costigliola, @scordio  can you, please, have a glance and tell if I am going the right way with this.

After some thought it seems not really what was desired.
As I understand, by introducing strategies we introduce the way to find objects for comparison (all fields or all getters).
Is this the way?

### DIFF
--- a/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
+++ b/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
@@ -129,8 +129,8 @@ public class RecursiveComparisonDifferenceCalculator {
             for (String nonIgnoredActualFieldName : nonIgnoredActualFieldsNames) {
               IntrospectionStrategy introspectionStrategy = recursiveComparisonConfiguration.getIntrospectionStrategy();
               DualValue fieldDualValue = new DualValue(fieldLocation.field(nonIgnoredActualFieldName),
-                                                       introspectionStrategy.getValue(nonIgnoredActualFieldName, actual),
-                                                       introspectionStrategy.getValue(nonIgnoredActualFieldName, expected));
+                                                       introspectionStrategy.getMemberValue(nonIgnoredActualFieldName, actual),
+                                                       introspectionStrategy.getMemberValue(nonIgnoredActualFieldName, expected));
               dualValuesToCompare.addFirst(fieldDualValue);
             }
           } else {
@@ -294,9 +294,10 @@ public class RecursiveComparisonDifferenceCalculator {
         // - if actual has more fields than expected, the additional fields are ignored as expected is the reference
         for (String actualFieldName : actualNonIgnoredFieldsNames) {
           if (expectedFieldsNames.contains(actualFieldName)) {
+            IntrospectionStrategy introspectionStrategy = recursiveComparisonConfiguration.getIntrospectionStrategy();
             DualValue newDualValue = new DualValue(dualValue.fieldLocation.field(actualFieldName),
-                                                   COMPARISON.getSimpleValue(actualFieldName, actualFieldValue),
-                                                   COMPARISON.getSimpleValue(actualFieldName, expectedFieldValue));
+              introspectionStrategy.getMemberValue(actualFieldName, actualFieldValue),
+              introspectionStrategy.getMemberValue(actualFieldName, expectedFieldValue));
             comparisonState.registerForComparison(newDualValue);
           }
         }
--- a/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
+++ b/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
@@ -44,6 +44,7 @@ import java.util.stream.Stream;
 
 import org.assertj.core.internal.DeepDifference;
 import org.assertj.core.util.Objects;
+import org.assertj.core.util.introspection.IntrospectionStrategy;
 
 /**
  * Based on {@link DeepDifference} but takes a {@link RecursiveComparisonConfiguration}, {@link DeepDifference}
@@ -126,9 +127,10 @@ public class RecursiveComparisonDifferenceCalculator {
           if (expectedFieldsNames.containsAll(nonIgnoredActualFieldsNames)) {
             // we compare actual fields vs expected, ignoring expected additional fields
             for (String nonIgnoredActualFieldName : nonIgnoredActualFieldsNames) {
+              IntrospectionStrategy introspectionStrategy = recursiveComparisonConfiguration.getIntrospectionStrategy();
               DualValue fieldDualValue = new DualValue(fieldLocation.field(nonIgnoredActualFieldName),
-                                                       COMPARISON.getSimpleValue(nonIgnoredActualFieldName, actual),
-                                                       COMPARISON.getSimpleValue(nonIgnoredActualFieldName, expected));
+                                                       introspectionStrategy.getValue(nonIgnoredActualFieldName, actual),
+                                                       introspectionStrategy.getValue(nonIgnoredActualFieldName, expected));
               dualValuesToCompare.addFirst(fieldDualValue);
             }
           } else {
--- a/src/main/java/org/assertj/core/util/introspection/GetterIntrospectionStrategy.java
+++ b/src/main/java/org/assertj/core/util/introspection/GetterIntrospectionStrategy.java
@@ -12,13 +12,56 @@
  */
 package org.assertj.core.util.introspection;
 
+import static java.util.Arrays.stream;
+import static java.util.Locale.ENGLISH;
+import static java.util.Objects.requireNonNull;
+
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.Set;
+import java.util.stream.Collectors;
+
 public class GetterIntrospectionStrategy implements IntrospectionStrategy {
 
   private static final GetterIntrospectionStrategy INSTANCE = new GetterIntrospectionStrategy();
   private static final PropertySupport propertySupport = PropertySupport.instance();
 
+  private static final String GETTER_REGEX = "^(is|get)\\p{Upper}\\w*";
+
   public static GetterIntrospectionStrategy instance() {
     return INSTANCE;
+  }
+
+  @Override
+  public Set<String> getMemberNames(Class<?> clazz) {
+    requireNonNull(clazz, "expecting Class parameter not to be null");
+    Set<String> declaredFields = getDeclaredMethodsIgnoringSyntheticAndStatic(clazz);
+
+    Class<?> superclazz = clazz.getSuperclass();
+    while (superclazz != null && !superclazz.getName().startsWith("java.lang")) {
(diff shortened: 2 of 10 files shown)


nach-o-man:
I made a few adjustments how I understood the target.
If it is something we are looking for then I will tidy up the code.
Also I have a question if current behaviour should be an option?

src/main/java/org/assertj/core/util/introspection/IntrospectionStrategy.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package org.assertj.core.util.introspection;
+
+import java.util.Set;
+
+public interface IntrospectionStrategy {
+
+  Set<String> getMemberNames(Class<?> clazz);
+
+  Set<String> getMemberNamesAsFields(Class<?> clazz);
```
joel-costigliola, line 21: My intuition tells me we should not this one, I mean if member refers to a getter/property then why would need the fields names, we would get the values of member by calling its getter.

src/main/java/org/assertj/core/util/introspection/GetterIntrospectionStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  @Override
+  public Set<String> getMemberNamesAsFields(Class<?> clazz) {
+    return getMemberNames(clazz)
+                                .stream()
+                                .map(s -> s.startsWith("is") ? s.substring(2) : s.substring(3))
+                                .map(s -> s.substring(0, 1).toLowerCase(ENGLISH) + s.substring(1))
+                                .collect(Collectors.toSet());
+  }
```
joel-costigliola, line 56: Following up on my `IntrospectionStrategy` comment, I think this is `getMemberNames` and it should introspect all getter methods but not base this on fields.

After all if users want to compare classes according to their getters, we should be not look at fields.

@nach-o-man  I'm not sure I'm explaining it well, let me know if that's clear for you.

src/main/java/org/assertj/core/util/introspection/IntrospectionStrategy.java
```diff
@@ -0,0 +1,25 @@
...
+
+  Object getValue(String fieldName, Object target);
```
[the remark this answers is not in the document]
joel-costigliola: maybe rename the method to `getMemberValue`

src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
```diff
@@ -126,9 +127,10 @@ private void initDualValuesToCompare(Object actual, Object expected, FieldLocati
           if (expectedFieldsNames.containsAll(nonIgnoredActualFieldsNames)) {
             // we compare actual fields vs expected, ignoring expected additional fields
             for (String nonIgnoredActualFieldName : nonIgnoredActualFieldsNames) {
+              IntrospectionStrategy introspectionStrategy = recursiveComparisonConfiguration.getIntrospectionStrategy();
               DualValue fieldDualValue = new DualValue(fieldLocation.field(nonIgnoredActualFieldName),
-                                                       COMPARISON.getSimpleValue(nonIgnoredActualFieldName, actual),
-                                                       COMPARISON.getSimpleValue(nonIgnoredActualFieldName, expected));
+                                                       introspectionStrategy.getValue(nonIgnoredActualFieldName, actual),
+                                                       introspectionStrategy.getValue(nonIgnoredActualFieldName, expected));
```
joel-costigliola, line 133: That's a good first step, we need to go further, `RecursiveComparisonDifferenceCalculator` should not use at all `COMPARISON` nor `Objects.getDeclaredFieldsIncludingInherited` or `Objects.getFieldsNames` since these are ways to get members to use in the comparison (indirectly sometimes).

`RecursiveComparisonDifferenceCalculator` is a complex piece of code by itself, it should delegate member introspection to `IntrospectionStrategy`, that will move a bit of complexity out of it and make it clear when introspection is used.

joel-costigliola (changes requested):
Thanks @nach-o-man for the PR! This is a good starting point.

I have tried to clarify where this should be going, let me know if it makes sense for you.

src/main/java/org/assertj/core/util/introspection/IntrospectionStrategy.java
nach-o-man: The idea was to try keep existing approach of addressing fields as properties, if it makes sense.
Just cause we can have getters that have nothing to do with fields 
```
public String getFullName(){
  return firstName + " " + lastName;
}
```
there is a very high chance that the compartibility will be broken. For example at the moment I cant imagine ignoring/comparing fields via string chains `"one.two.three"` when using getters comparsion.

src/main/java/org/assertj/core/util/introspection/GetterIntrospectionStrategy.java
nach-o-man: If I understand correct you want really to ditinct the code rather that have some complex reusable logic. Is that it?

src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
nach-o-man: I see. Maybe new type of Assert should be introduced to existing methods naming `ignoringFieldsMatchingRegexes` etc. if we are bringing the `member` terminology. As for now I will try to work your comment and see how it turns out.

src/main/java/org/assertj/core/util/introspection/IntrospectionStrategy.java
joel-costigliola: fair point ... that would be interpreted as `getOne().getTwo().getThree()` but I need to think if it is worth it to break the API and introduce `ignoringMembers` or something alike.

src/main/java/org/assertj/core/util/introspection/GetterIntrospectionStrategy.java
joel-costigliola: I'm not sure to fully understand your question, what I had in mind was to extract the introspection logic in one place.

src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
joel-costigliola: I need to give some thoughts about it, I'm not sure I want to deprecate a good part of the API mentioning fields and introducing `ignoringMembersMatchingRegexes` might not be that clear ... let me think about it.

src/main/java/org/assertj/core/util/introspection/GetterIntrospectionStrategy.java
joel-costigliola: yes, we want the implementation to depend on the `IntrospectionStrategy` interface only.

we might indeed need `getMemberNamesAsFields` to translate the member location/path (at the moment the location is field based, ex: `person.address.street`), maybe we can rename it to `getMemberLocationNames` or something alike, I'm not completely sure (I'll take the PR branch and see if that makes, it's a bit hard to see from the PR diff)

nach-o-man: Did you get any insites? At the moment I am trying an approach with changing the  assertion class when switching strategy using 3 strategies - fieldsOnly, Getters, Properties (current). Atmo code is not looking well 😞

src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
joel-costigliola: did not have the time this week, hopefully in the week end but no promises

nach-o-man: Sure, I am doing some slow and steady research 😄

joel-costigliola:
@nach-o-man I have pushed two commits on https://github.com/assertj/assertj-core/tree/feature/introspection-strategy which is based on your branch (rebased on the latest `main` commits). I tried to push on your repo fork but I don't have the rights to :(.

The commits simply replace calls to get fields by the `introspectionStrategy.getMemberNamesAsFields` to be consistent with `RecursiveComparisonConfiguration.getNonIgnoredActualFieldNames` implementation.

I think the code in the branch is good as a refactoring for getting fields and their values, it is more cleanly encapsulated.
This is not a true refactoring as we break the way to get field values as it was done in `PropertyOrFieldSupport.COMPARISON`, I believe this is for the best though (I hope not to be wrong on this one but we can always provide an original introspectionStrategy to provide a non breaking change behavior). 

I would like to keep the code in this branch not to expose new API  methods, that would exclude `comparingUsingFields` and `comparingUsingGetters` (this would be done later on).

filiphr:
I'd like to chime in regarding this. I have some experience via MapStruct regarding bean properties / fields. We rely mostly on the [Java Beans Specification](https://download.oracle.com/otndocs/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/) when determining what is a property or not. Our main logic for determining whether a method is a property or not can be found [here](https://github.com/mapstruct/mapstruct/blob/5c22eee6c327aa893109f098d5a7cc2b5976cdc1/processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java#L72-L87). It doesn't use the reflection API, but I think you get the idea.

In a nutshell, I think that relying only on the method name is not enough. A getter cannot have parameters and should not be a void. In addition to that when extracting the property name from a getter it is not enough to decapitalize the first character always. e.g. for `getURL` the property name is `URL` and not `uRL`. 

What I would suggest for the introspection strategy (that might also improve the performance is the following:

```
public interface IntrospectionStrategy {

  Set<Property> getProperties(Class<?> clazz);

  interface Property {
  
    String getName();

    Class<?> getType(); // Not sure if this is needed for the nested property usage or not

    Object getValue(Object target);

    // Perhaps modifiers can be exposed as well
  }
}
```

With something like this you could even offer people to only use public properties and only check the public getters and public fields.

joel-costigliola:
I agree that getter/property should be rock solid, I would not put that on `IntrospectionStrategy` though as it is only relevant to `GetterIntrospectionStrategy`.

filiphr:
I was talking about the `GetterIntrospectionStrategy` as I saw things in there that I think might not work in some cases. The idea for the possible API is valid for both, although I am not sure what your end goal is with the strategy

nach-o-man:
@filiphr Thank you for the provided links and code. I totally think that getter/propperty defining logic should be straightforward.
As for provided example of decapitalisation, I think when we want to translate getter to property it is pretty much correct as camel-case is widely used and enforced in naming methods/members. In other words private property is much likely to be called `url` with `getUrl` getter.
I think we either should pick the one strategy and consider having an option for defining own property-matching strategy. Which I think is kinda over-design.

scordio:
@joel-costigliola I assume this superseded by ba6b5dfbd320e0149b6e6d6bf6452195288c73c8, right?

joel-costigliola:
@nach-o-man thanks for the initial PoC, it gave me good ideas to complete it.

PR #2209, opened 2021-05-13
Add stack trace information for Throwable when hasCauseInstanceOf and hasCauseExactlyInstanceOf fails

RGalways17:
Show the stack trace for the `Throwable` when `hasCauseInstanceOf` and `hasCauseExactlyInstanceOf` fails. I also noticed that there was only test for no cause, so I decided for myself to write a test on wrong cause. Please let me know if there is any problem.

#### Check List:
* Fixes #1355
* Unit tests : YES


src/test/java/org/assertj/core/error/ShouldHaveCauseExactlyInstance_create_Test.java
```diff
@@ -32,4 +35,21 @@ void should_create_error_message_for_no_cause() {
...
+    // WHEN
+    String message = shouldHaveCauseInstance(actual, expected.getClass()).create();
+    // THEN
+    then(message).isEqualTo(format("%nExpecting a throwable with cause being an instance of:%n" +
+      "  %s%n" +
+      "but was an instance of:%n" +
+      "  %s%n" +
+      "Throwable that failed the check:%n" +
+      "%n" + escapePercent(getStackTrace(actual)), expected, cause));
+  }
```
joel-costigliola, line 54: `isEqualTo` for String as an overload version that supports the `String.format` syntax, you can then remove the call to `format` here.

Also can you add a cause message containing a `%` to make sure it does not break the code?

see https://www.javadoc.io/doc/org.assertj/assertj-core/latest/org/assertj/core/api/AbstractStringAssert.html

this comment applies to the other tests too ;-)

joel-costigliola (changes requested):
Thanks @RGalways17, just some minor comments!

RGalways17:
@joel-costigliola I've finished the syntax correction. Not quite sure about what you mean by 'cause message containing a %'? Do you mean `then(message).isEqualTo("......" +
      "Cause message :%n" +
      "  %s%n" +
      "......" +
      "%n" + escapePercent(getStackTrace(actual)),...,..., causeMessage);`
The cause of the actual throwable should be already displayed in the message like this:
`Throwable that failed the check:

java.lang.RuntimeException: java.lang.IllegalAccessError
.....
Caused by: java.lang.IllegalAccessError
.....
`

RGalways17:
@joel-costigliola I've finished the syntax correction. Not quite sure about what you mean by 'cause message containing a %'? Do you mean `then(message).isEqualTo("......" +
      "Cause message :%n" +
      "  %s%n" +
      "......" +
      "%n" + escapePercent(getStackTrace(actual)),...,..., causeMessage);` ?
The cause of the actual throwable should be already displayed in the message like this:
`Throwable that failed the check:

java.lang.RuntimeException: java.lang.IllegalAccessError
.....
Caused by: java.lang.IllegalAccessError
.....
`

RGalways17:
@joel-costigliola I've finished the syntax correction. Not quite sure about what you mean by 'cause message containing a %'? Do you mean then(message).isEqualTo("......" + "Cause message :%n" + " %s%n" + "......" + "%n" + escapePercent(getStackTrace(actual)),...,..., causeMessage); ?
The cause of the actual throwable should be already displayed in the message like this:
`Throwable that failed the check:

java.lang.RuntimeException: java.lang.IllegalAccessError
.....
Caused by: java.lang.IllegalAccessError
.....`

RGalways17:
@joel-costigliola I've finished the syntax correction. Not quite sure about what you mean by 'cause message containing a %'? Do you mean 
`then(message).isEqualTo("......" + "Cause message :%n" + " %s%n" + "......" + "%n" + escapePercent(getStackTrace(actual)),...,..., causeMessage);` ?
The cause of the actual throwable should be already displayed in the message like this:
`Throwable that failed the check:

java.lang.RuntimeException: java.lang.IllegalAccessError
.....
Caused by: java.lang.IllegalAccessError
.....`

joel-costigliola:
I was saying to create the cause exception with a message having a `%`, ex: 
```java
Throwable cause = new IllegalAccessError("boom % %n");
```
I want to make sure that `these %` are not interpreted as `String.format` arguments.

RGalways17:
@joel-costigliola I've finished modifying the code. Please check :-).

joel-costigliola:
Integrated thanks @RGalways17!

PR #2212, opened 2021-05-15
Improve performance in StandardComparisonStrategy#areEqual

filiphr:
While working on PR #2171 I noticed that the performance degradation with the changes in that PR + this change are way lower than the changes in that PR without this change here.

Did some JFR profiling and most of the time is spend in this method here. With this PR the idea is to avoid doing `instanceof` checks if the actual is not an array. I know that `Objects#deepEquals` is similar, but I cannot explain why the performance degrades so much in the AssertJ code (perhaps because I am using `Long` for the testing and the reference is the same.

I can add a different object type in the benchmark if you want

### DIFF
--- a/src/main/java/org/assertj/core/internal/StandardComparisonStrategy.java
+++ b/src/main/java/org/assertj/core/internal/StandardComparisonStrategy.java
@@ -83,24 +83,33 @@ public class StandardComparisonStrategy extends AbstractComparisonStrategy {
   @Override
   public boolean areEqual(Object actual, Object other) {
     if (actual == null) return other == null;
-    if (actual instanceof Object[] && other instanceof Object[])
-      return java.util.Arrays.deepEquals((Object[]) actual, (Object[]) other);
-    if (actual instanceof byte[] && other instanceof byte[])
-      return java.util.Arrays.equals((byte[]) actual, (byte[]) other);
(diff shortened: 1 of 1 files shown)


scordio:
I definitely didn't see it coming and I also don't understand where the performance penalty is. Nice catch @filiphr!

Your changes look already good to me.

If you have time, you could try to see how the current implementation behaves with `Object[]` and `boolean[]`. You might put the `Object[]` check before the primitive types in your implementation to have the a similar flow like `main`.

filiphr:
I adapted the benchmark and I can now easily test with anything. I added tests for `Object[]` and `boolean[]` and the results are confusing to me.

image

Seems like the benchmark is slower with my changes from here. I really cannot explain it. It doesn't make much sense to me. I'll try running it with the other PR as well

filiphr:
OK I know what the problem was. I had typo in the setup of the benchmark so the `long` assert was actually the `Object[]` one. Running it now again 🤦

filiphr:
The new results are in and they are like I expected them:

image

filiphr:
The benchmark I used to get this is https://github.com/filiphr/assertj-benchmark/blob/main/src/main/java/org/assertj/benchmark/ObjectAssertBenchmark.java.

scordio:
Merged, thanks @filiphr!

PR #2213, opened 2021-05-15
add the feature mentioned in #2198

### LINKED ISSUE #2198: Feature Request: `assertThat(file).doesNotHaveContent("string")`
Currently it is possible to conveniently check a file's content with `hasContent`, but not the opposite.

Also I'm regularly bumping into the absence of the possibility to directly analyse files as it is possible for strings, so to have something analogous to `assertThat("String").contains` & `doesNotContain`. Of course one can do `assertThat(FileUtils.readFileToString(file))` but having syntactic sugar in assertJ to avoid this would be lovely as the brevity and clarity assertJ enables is one of it's main strengths IMHO.
These last two may be a separate feature request, however i wanted to gather general feedback about feasibility first.

wtd2:
#### Check List:
* Fixes #2198
* Unit tests : YES

Hi, the feature mentioned is implemented in the code. However, I am not sure whether the tests are sufficient.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractFileAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractFileAssert.java
@@ -16,14 +16,12 @@ import static java.util.Objects.requireNonNull;
 import static org.assertj.core.util.Preconditions.checkArgument;
 
 import java.io.File;
-import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.charset.Charset;
 import java.nio.file.FileSystem;
 import java.security.MessageDigest;
 import java.util.function.Predicate;
 
-import org.apache.commons.io.FileUtils;
 import org.assertj.core.internal.Files;
 import org.assertj.core.util.CheckReturnValue;
 import org.assertj.core.util.VisibleForTesting;
--- a/src/main/java/org/assertj/core/api/AbstractFileAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractFileAssert.java
@@ -12,10 +12,14 @@
  */
 package org.assertj.core.api;
 
+import static java.nio.file.Files.newBufferedReader;
+import static java.nio.file.Files.readAllBytes;
 import static java.util.Objects.requireNonNull;
 import static org.assertj.core.util.Preconditions.checkArgument;
 
+import java.io.BufferedReader;
 import java.io.File;
+import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.charset.Charset;
 import java.nio.file.FileSystem;
@@ -1237,10 +1241,16 @@ public abstract class AbstractFileAssert<SELF extends AbstractFileAssert<SELF>> 
 
   /**
    *
-   * @return
+   * @return a new StringAssert Object.
+   * @throws AssertionError if the actual {@code File} is not readable.
+   * @since 3.19.0
    */
-
   public StringAssert content() {
-    return new StringAssert(files.getFileContent(info, actual));
+    return content(Charset.defaultCharset());
+  }
+
+  public StringAssert content(Charset charset){
+    String fileContent = files.getFileContent(info, actual, charset);
+    return new StringAssert(fileContent);
   }
 }
--- a/src/main/java/org/assertj/core/api/AbstractFileAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractFileAssert.java
@@ -12,14 +12,10 @@
  */
 package org.assertj.core.api;
 
-import static java.nio.file.Files.newBufferedReader;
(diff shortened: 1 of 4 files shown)


src/main/java/org/assertj/core/api/AbstractFileAssert.java
```diff
@@ -1234,4 +1234,27 @@ public SELF isNotEmpty() {
...
+   * assertThat(file).content().contains('tet');</code></pre>
+   *
+   * @return a StringAssert object.
+   * @throws AssertionError if the actual {@code File} is not readable.
+   * @since 3.??
+   */
+  public StringAssert content() {
+    String fileContent = files.getFileContent(info, actual);
+    return new StringAssert(fileContent);
+  }
```
joel-costigliola, line 1258: we need to add `public StringAssert content(Charset actualCharset)` and make this one use the default charset.

src/main/java/org/assertj/core/internal/Files.java
```diff
@@ -507,6 +507,25 @@ public void assertIsDirectoryNotContaining(AssertionInfo info, File actual, Stri
...
+   * @throws AssertionError if the actual {@code File} can not be read.
+   * @return a String object with the content of the given {@code File}.
+   */
+
+  public String getFileContent(AssertionInfo info, File actual) {
+    assertCanRead(info, actual);
+    try {
+      return new String(readAllBytes(actual.toPath()));
+    } catch (IOException e) {
+      throw failures.failure(info, shouldBeReadable(actual));
```
joel-costigliola, line 525: rethrow an `UncheckedIOException` as the file is readable (in term of permissions) but the content led to an I/O error.

src/test/java/org/assertj/core/api/file/FileAssert_content_Test.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  @Test
+  void should_return_actual_content() throws IOException {
+    // WHEN
+    String fileContent = new String(java.nio.file.Files.readAllBytes(actual.toPath()));
+    // THEN
+    assertThat(actual).content()
+                      .isEqualTo(fileContent);
+  }
```
joel-costigliola, line 41: we need to test two things:
- that the returned type is `StringAssert`
- that the content is equal to `"actual"` (no need to read the file, we already know the content).

use `then` instead of `assertThat` as per the [contributing guidelines](https://github.com/assertj/assertj-core/blob/main/CONTRIBUTING.md).

src/test/java/org/assertj/core/api/file/FileAssert_content_Test.java
```diff
@@ -0,0 +1,53 @@
...
+  @Test
+  void should_pass_if_has_proper_functionality() {
+    // GIVEN
+    String prefix = "ac";
+    String subString = "ct";
+    // THEN
+    assertThat(actual).content()
+                      .startsWith(prefix)
+                      .contains(subString);
+  }
```
joel-costigliola, line 52: add a test to simulate an `IOException` by mocking actual to throw one when calling `actual.toPath()`.

src/test/java/org/assertj/core/internal/files/Files_getFileContent_Test.java
```diff
@@ -0,0 +1,25 @@
...
+import org.assertj.core.internal.FilesBaseTest;
+import org.junit.jupiter.api.Test;
+
+class Files_getFileContent_Test extends FilesBaseTest {
+  @Test
+  void should_fail_if_can_not_read() {
+    when(actual.canRead()).thenReturn(false);
+    AssertionInfo info = someInfo();
+
+    Throwable error = catchThrowable(() -> files.getFileContent(info, actual));
```
joel-costigliola, line 20: use GIVEN  WHEN THEN as per  the [contributing guidelines](https://github.com/assertj/assertj-core/blob/main/CONTRIBUTING.md)

joel-costigliola (changes requested):

src/main/java/org/assertj/core/api/AbstractFileAssert.java
```diff
@@ -1234,4 +1234,27 @@ public SELF isNotEmpty() {
...
+
+  /**
+   * Get a StringAssert object with the content of the actual {@code File}.
+   * <p>
+   * Example:
+   * <pre><code class='java'> File file = File.createTempFile(&quot;tmp&quot;, &quot;txt&quot;);
+   * Files.write(file.toPath(), new byte[]{'t', 'e', 's', 't'});
+   *
+   * // assertion will pass
+   * assertThat(file).content().contains('te');
```
joel-costigliola, line 1246: String takes `""`, not `''`

joel-costigliola:
Integrated thanks @wtd2!

PR #2215, opened 2021-05-16
fix issue 2210

### LINKED ISSUE #2210: `hasAllNullFieldsOrProperties` and primitive types
#### Summary

`hasAllNullFieldsOrProperties` check that the actual object has only null fields or properties.

All fields and properties are inspected but fields with primitive types always will have some value.

#### Example

```java
    class Book {
        String title;
        int pages;
    }

    @Test
    void testHasAllNullFieldsOrProperties() {
        Book book = new Book();
        assertThat(book).hasAllNullFieldsOrProperties();
    }
```

test will fail because of field `pages` is not null.

#### Proposition

Exclude by default all field with primitive types from assertion.

wuqihan837:
#### Check List:
* Fixes #2210 
* Unit tests : YES
I added hasAllNullFieldsOrProperties(boolean excludePrimitiveTypes)method and assertHasAllNullFieldsOrPropertiesExceptExcludePrimitives method. When boolean excludePrimitiveTypes is true, assertHasAllNullFieldsOrPropertiesExceptExcludePrimitives will be called. If excludePrimitiveTypes is false, the method is the same as hasAllNullFieldsOrProperties(). I also added two unit tests.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
@@ -355,9 +355,7 @@ public abstract class AbstractObjectAssert<SELF extends AbstractObjectAssert<SEL
     return myself;
   }
   /**
-   * The method is similar to hasAllNullFieldsOrProperties().
-   * The method adds one boolean argument called excludePrimitiveTypes,if the argument is true, the fields
-   * of primitive types are ignored and not considered.
+   * The method is similar to hasAllNullFieldsOrProperties() but ignore fields of primitive types.
    * Asserts that the actual object has only null fields or properties.
    * <p>
    * If an object has a field and a property with the same name, the property value will be used over the field.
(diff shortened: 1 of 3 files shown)


wuqihan837:
I added hasAllNullFieldsOrPropertiesExceptPrimitiveTypes() method in AbstractObjectAssert.java and assertHasAllNullFieldsOrPropertiesExceptExcludePrimitives method in Objects.java and two tests in ObjectAssert_hasAllNullFieldsOrPropertiesExcept_Test.java. In assertHasAllNullFieldsOrPropertiesExceptExcludePrimitives method, after getting Set<Field> declaredFields, I excluded the fields of primitive types in the set and get a new set called filtereddeclaredFields.

joel-costigliola:
superseded by the introduction of the recursive assertion as per https://github.com/assertj/assertj-core/issues/2210#issuecomment-1195316875

PR #2219, opened 2021-05-18
isHTTPReachable

### LINKED ISSUE #2196: Add `isReachable` / `isUnreachable` for URL assertions
#### Summary
Inspired by junit-pioneer/junit-pioneer#468, `isReachable` / `isUnreachable` assertions could be added to `AbstractUrlAssert`.

One of the most valuable use cases could be when the assertion is used as an assumption, allowing to skip a test if a remote dependency is not available.

The assertion name could be generic enough (i.e., not strictly bound to HTTP) and internally we could have different behaviors based on the protocol.

At first, the target should be only HTTP URLs.

[This gist](https://gist.github.com/nipafx/a21f1d1c5211617453e43a88c9df439c) might inspire the assertion implementation.

#### Example

```java
assertThat(new URL("http://google.com")).isReachable();
assertThat(new URL("http://invalid")).isUnreachable();
```

#### Points to discuss

- [ ] Naming: `isReachable` good enough, especially if we want to have it generic? What about `isAccessible` or `exists`?
- [ ] Testing: [MockServer](https://www.mock-server.com/)? Something else?
- [ ] Should we support HTTPS? We might need to [disable certificate chain and hostname validation](https://nakov.com/blog/2009/07/16/disable-certificate-validation-in-java-ssl-connections/)
- [ ] Method: [`GET` vs `HEAD`](https://www.baeldung.com/java-check-url-exists#using-a-head-request)
- [ ] What to do with [redirects](https://www.baeldung.com/java-check-url-exists#redirects)

jiashu-z:
#### Check List:
* Fixes #2196 
* Unit tests : YES

### DIFF
--- a/src/test/java/org/assertj/core/api/url/UrlAssert_isReachableHTTP_Test.java
+++ b/src/test/java/org/assertj/core/api/url/UrlAssert_isReachableHTTP_Test.java
@@ -1,0 +1,58 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2021 the original author or authors.
+ */
+package org.assertj.core.api.url;
+
+import static org.mockito.Mockito.verify;
+
+import com.sun.net.httpserver.HttpExchange;
+import com.sun.net.httpserver.HttpHandler;
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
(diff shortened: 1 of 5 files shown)


wateryloo:
@scordio Thanks for your review!

src/test/java/org/assertj/core/api/url/UrlAssert_isReachableHTTP_Test.java
```diff
@@ -0,0 +1,58 @@
...
+  private static class LocalHTTPHandler implements HttpHandler {
+
+    static final LocalHTTPHandler INSTANCE = new LocalHTTPHandler();
+
+    @Override
+    public void handle(HttpExchange exchange) throws IOException {
+      exchange.sendResponseHeaders(404, 0);
+      exchange.close();
+    }
+  }
```
scordio, line 38: This seems to be unused. Do you have something in mind for it?

scordio (commented):
Thanks @wateryloo, the direction is good. I'm keen to see the approach for the internal assertion testing, let me know if you need some input about it.

scordio:
Hi @jiashu-z, sorry for closing your PR without notice! It appears to have been a side effect of the maintenance we did on the target branch.

Nevertheless, due to https://github.com/assertj/assertj/issues/2196#issuecomment-3343877472, we decided not to proceed with this feature.

Thanks again for your interest!

PR #2220, opened 2021-05-18
GH-2134 avoid a copy in assertContains when actual is already a Collection

### LINKED ISSUE #2134: Avoid copy of Collection instances in Iterables#assertContains and Iterables#assertContainsAll
Following https://github.com/assertj/assertj-core/issues/2132#issuecomment-784523615 and https://github.com/assertj/assertj-core/issues/2132#issuecomment-785406008, the current implementation of `Iterables#assertContains` and `Iterables#assertContainsAll` could be improved avoiding the creation of a new array list when `actual` is a `Collection` instance.

mbaechler:
#### Check List:
* Fixes #2134
* Unit tests : NA (optimization)

### DIFF
--- a/src/main/java/org/assertj/core/internal/Iterables.java
+++ b/src/main/java/org/assertj/core/internal/Iterables.java
@@ -84,6 +84,7 @@ import static org.assertj.core.util.Streams.stream;
 
 import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Comparator;
 import java.util.Deque;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -339,10 +340,18 @@ public void assertHasSameSizeAs(AssertionInfo info, Iterable<?> actual, Iterable
...
-    assertIterableContainsGivenValues(actual.getClass(), actualAsList, values, info);
+    assertIterableContainsGivenValues(actual.getClass(), actualAsCollection, values, info);
+  }
+
+  private Collection<?> ensureActualCanBeReadMultipleTimes(Iterable<?> actual) {
+    if (actual instanceof Collection<?>) {
+      return (Collection<?>) actual; // Avoid copy when actual is already a Collection.
+    } else {
+      return newArrayList(actual); //ensure the assertion works for non-repeatable iterables
+    }
```
joel-costigliola, line 354: the else block is not necessary since we return in the if block.
could use the ? ternary operator too.

joel-costigliola:
thanks @mbaechler, just a minor comment.

scordio:
We actually have the same issue for all the `assertContains*` methods. Do we want to address all of them in one go?

joel-costigliola:
I would say all of them

scordio:
@mbaechler would it be fine for you to cover also the other methods?

mbaechler:
sorry, I won't be able to handle that for some time

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -339,10 +340,18 @@ public void assertHasSameSizeAs(AssertionInfo info, Iterable<?> actual, Iterable
...
-    final List<?> actualAsList = newArrayList(actual);
-    if (commonCheckThatIterableAssertionSucceeds(info, actualAsList, values)) return;
+    final Collection<?> actualAsCollection = ensureActualCanBeReadMultipleTimes(actual);
+    if (commonCheckThatIterableAssertionSucceeds(info, actualAsCollection, values)) return;
     // check for elements in values that are missing in actual.
-    assertIterableContainsGivenValues(actual.getClass(), actualAsList, values, info);
+    assertIterableContainsGivenValues(actual.getClass(), actualAsCollection, values, info);
+  }
+
+  private Collection<?> ensureActualCanBeReadMultipleTimes(Iterable<?> actual) {
```
vlsi, line 349: Should it be static to make it explicit it does not depend on the fields?

joel-costigliola: indeed

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -339,10 +340,18 @@ public void assertHasSameSizeAs(AssertionInfo info, Iterable<?> actual, Iterable
...
+    if (actual instanceof Collection<?>) {
+      return (Collection<?>) actual; // Avoid copy when actual is already a Collection.
```
vlsi, line 351: There still may be edge cases like "user-provided-collection that loads items lazily from the database and allows calling .iterator() just once"

WDYT of adding an extra check like `actual.getClass().getClassLoader() == null` (or `==Collection.class.getClassLoader` for brevity)?

joel-costigliola:
Integrated thanks @mbaechler!

PR #2221, opened 2021-05-21
Fix #2217

### LINKED ISSUE #2217: exclusion non existing fields, properties
#### Summary

We can exclude fields, properties from some of checking like in `ObjectAssert#hasNoNullFieldsOrPropertiesExcept`, `RecursiveComparisonAssert#ignoringFields` (maybe in other place also)

We can misspell in field name or filed can be removed from object in future development, so assertions in this situation should fail with message that excluded filed not exist.

#### Example

```java
    class Book {
        String title;
    }

   Book aBook = ...

   assertThat(aBook).hasNoNullFieldsOrPropertiesExcept("name");
```
Assertions should fail because class `Book` not contains field `name`

btx0424:
#### Check List:
* Fixes #2217 
* Unit tests : YES 

Implemented what's suggested in #2217. Check for field existence in `assertHasNoNullFieldsOrPropertiesExcept`. A `shouldHaveDeclaredFields` error is thrown if any of the fields does not exist.

### DIFF
--- a/src/main/java/org/assertj/core/internal/Objects.java
+++ b/src/main/java/org/assertj/core/internal/Objects.java
@@ -33,6 +33,7 @@ import static org.assertj.core.error.ShouldBeOfClassIn.shouldBeOfClassIn;
 import static org.assertj.core.error.ShouldBeSame.shouldBeSame;
 import static org.assertj.core.error.ShouldContainOnly.shouldContainOnly;
 import static org.assertj.core.error.ShouldHaveAllNullFields.shouldHaveAllNullFields;
+import static org.assertj.core.error.ShouldHaveFields.shouldHaveDeclaredFields;
 import static org.assertj.core.error.ShouldHaveNoNullFields.shouldHaveNoNullFieldsExcept;
 import static org.assertj.core.error.ShouldHavePropertyOrField.shouldHavePropertyOrField;
 import static org.assertj.core.error.ShouldHavePropertyOrFieldWithValue.shouldHavePropertyOrFieldWithValue;
@@ -721,6 +722,7 @@ public class Objects {
    * @param actual                     the given object.
    * @param propertiesOrFieldsToIgnore the fields to ignore in comparison.
    * @throws AssertionError if actual is {@code null}.
+   * @throws AssertionError if any of the properties or fields to ignore doesn't exist.
    * @throws AssertionError if some of the fields of the actual object are null.
    */
   public <A> void assertHasNoNullFieldsOrPropertiesExcept(AssertionInfo info, A actual,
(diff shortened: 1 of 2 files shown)


src/main/java/org/assertj/core/internal/Objects.java
```diff
@@ -732,10 +734,18 @@ static boolean propertyOrFieldValuesAreEqual(Object actualFieldValue, Object oth
     for (Field field : declaredFieldsIncludingInherited) {
       // ignore private field if user has decided not to use them in comparison
       String fieldName = field.getName();
-      if (ignoredFields.contains(fieldName) || !canReadFieldValue(field, actual)) continue;
+      if (ignoredFields.contains(fieldName)) {
+        ignoredFields.remove(fieldName);
+        continue;
+      }
```
slawekjaranowski, line 740: why not 
`if (ignoredFields.remove(fieldName)) continue`

src/test/java/org/assertj/core/api/object/ObjectAssert_hasNoNullFieldsOrProperties_Test.java
```diff
@@ -96,4 +96,17 @@ void should_pass_if_all_fields_or_properties_but_the_ignored_ones_are_set() {
...
+  public void should_fail_if_a_field_does_not_exist() {
+    Jedi jedi = new Jedi("Yoda", "Blue");
+    assertThatExceptionOfType(AssertionError.class).isThrownBy(() -> assertThat(jedi).hasNoNullFieldsOrPropertiesExcept("father"))
+                                                   .withMessage(format("%n"
+                                                                       + "Expecting%n"
+                                                                       + "  org.assertj.core.test.Jedi%n"
+                                                                       + "to have the following declared fields:%n"
+                                                                       + "  [\"father\"]%n"
+                                                                       + "but it doesn't have:%n"
+                                                                       + "  [\"father\"]"));
```
joel-costigliola, line 110: Use `expectAssertionError` as indicated in the [contributing guidelines](https://github.com/assertj/assertj-core/blob/main/CONTRIBUTING.md).

src/main/java/org/assertj/core/internal/Objects.java
```diff
@@ -721,6 +722,7 @@ static boolean propertyOrFieldValuesAreEqual(Object actualFieldValue, Object oth
    * @param actual                     the given object.
    * @param propertiesOrFieldsToIgnore the fields to ignore in comparison.
    * @throws AssertionError if actual is {@code null}.
+   * @throws AssertionError if any of the properties or fields to ignore doesn't exist.
```
joel-costigliola, line 725: this should be added to the `AbstractObjectAssert#hasNoNullFieldsOrPropertiesExcept` javadoc with a note saying it is a breaking change introduced in 3.20.0

joel-costigliola:
Integrated in cfb7b680f82eb5bf09f5b3bf41131adffb25fd6b thanks  @btx0424 !

PR #2247, opened 2021-06-05
[#2236] Add hasFileName as an alias of hasName for File assertions

### LINKED ISSUE #2236: Add `hasFileName` as an alias of `hasName` for `File` assertions
#### Example

```java
assertThat(new File("file.txt")).hasFileName("file.txt");
```

Related to #2205.

ShivakumarSwamy:
#### Check List:
* Fixes #2236
* Unit tests : YES 
* Javadoc with a code example (on API only) : Yes

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractFileAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractFileAssert.java
@@ -686,6 +686,7 @@ public abstract class AbstractFileAssert<SELF extends AbstractFileAssert<SELF>> 
    * @throws AssertionError if the actual {@code File} does not have the expected name.
    *
    * @see java.io.File#getName() name definition.
+   * @see #hasFileName(String)
    */
   public SELF hasName(String expected) {
     files.assertHasName(info, actual, expected);
@@ -693,7 +694,20 @@ public abstract class AbstractFileAssert<SELF extends AbstractFileAssert<SELF>> 
   }
 
   /**
-   * Alias for {@link #hasName(String)}.
+   * Verifies that the actual {@code File} has given name (alias of {@link #hasName(String)}).
+   *
+   * <p>
+   * Example:
+   * <pre><code class='java'> File xFile = new File(&quot;somewhere/xFile.java&quot;);
+   * File xDirectory = new File(&quot;somewhere/xDirectory&quot;);
+   *
+   * // assertion will pass
+   * assertThat(xFile).hasFileName(&quot;xFile.java&quot;);
+   * assertThat(xDirectory).hasFileName(&quot;xDirectory&quot;);
+   *
+   * // assertion will fail
+   * assertThat(xFile).hasFileName(&quot;xFile&quot;);
+   * assertThat(xDirectory).hasFileName(&quot;somewhere&quot;);</code></pre>
    *
    * @param expected the expected {@code File} name.
    * @return {@code this} assertion object.
@@ -702,6 +716,8 @@ public abstract class AbstractFileAssert<SELF extends AbstractFileAssert<SELF>> 
    * @throws AssertionError if the actual {@code File} does not have the expected name.
    *
    * @see java.io.File#getName() name definition.
+   * @see #hasName(String)
+   * @since 3.20.0
    */
   public SELF hasFileName(String expected) {
     return hasName(expected);
--- a/src/main/java/org/assertj/core/api/AbstractFileAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractFileAssert.java
@@ -19,6 +19,7 @@ import java.io.File;
 import java.io.UncheckedIOException;
 import java.nio.charset.Charset;
 import java.nio.file.FileSystem;
+import java.nio.file.Path;
 import java.security.MessageDigest;
 import java.util.function.Predicate;
 
@@ -689,6 +690,15 @@ public abstract class AbstractFileAssert<SELF extends AbstractFileAssert<SELF>> 
   public SELF hasName(String expected) {
     files.assertHasName(info, actual, expected);
     return myself;
+  }
+
+  /**
+   * Alias for {@link #hasName(String)}.
+   *
+   * This method can be used to easily switch between {@link File} and {@link Path} assertions.
+   */
+  public SELF hasFileName(String expected) {
+    return hasName(expected);
   }
 
   /**
--- a/src/main/java/org/assertj/core/api/AbstractFileAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractFileAssert.java
@@ -695,7 +695,13 @@ public abstract class AbstractFileAssert<SELF extends AbstractFileAssert<SELF>> 
   /**
    * Alias for {@link #hasName(String)}.
    *
-   * This method can be used to easily switch between {@link File} and {@link Path} assertions.
+   * @param expected the expected {@code File} name.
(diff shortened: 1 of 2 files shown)


src/main/java/org/assertj/core/api/AbstractFileAssert.java
```diff
@@ -691,6 +692,15 @@ public SELF hasName(String expected) {
...
 
+  /**
+   * Alias for {@link #hasName(String)}.
+   *
+   * This method can be used to easily switch between {@link File} and {@link Path} assertions.
+   */
+  public SELF hasFileName(String expected) {
+    return hasName(expected);
+  }
+
```
ShivakumarSwamy, line 703: - Have re-used `hasName` instead of adding a new method `assertHasFileName(info, actual, expected);` to `files` object.  Please let me know is this fine. 
- Please suggest a better javadoc if this does not explain it well 😉

src/main/java/org/assertj/core/api/AbstractFileAssert.java
```diff
@@ -691,6 +692,15 @@ public SELF hasName(String expected) {
     return myself;
   }
 
+  /**
+   * Alias for {@link #hasName(String)}.
+   *
+   * This method can be used to easily switch between {@link File} and {@link Path} assertions.
```
nach-o-man, line 698: I think the correct way (judging by surrounding code) is to use `@code` instead of `@link`.
Imho like 698 is not really needed - like 696 explains everything.

src/test/java/org/assertj/core/api/file/FileAssert_hasFileName_Test.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.assertj.core.api.file;
+
+import static org.mockito.Mockito.verify;
+
+import org.assertj.core.api.FileAssert;
+import org.assertj.core.api.FileAssertBaseTest;
+import org.junit.jupiter.api.DisplayName;
+
+@DisplayName("FileAssert hasFileName")
```
nach-o-man, line 21: This is not nessesary as the name generator is used
https://github.com/assertj/assertj-core/blob/b45e499446d3296dd11c480b0bd2087328989f05/src/test/resources/junit-platform.properties#L1

nach-o-man (changes requested):

src/test/java/org/assertj/core/api/file/FileAssert_hasFileName_Test.java
ShivakumarSwamy: Was not aware of that. Thank you 😄 . Then the display name will be `FileAssert hasFileName Test` instead of `FileAssert hasFileName`. 

Would that be fine? as the [example in contributing guide](https://github.com/assertj/assertj-core/blob/main/CONTRIBUTING.md) has display name without `Test`

src/main/java/org/assertj/core/api/AbstractFileAssert.java
ShivakumarSwamy: Tired to use the javadoc similar to [`assertThatNullPointerException` javadoc](https://github.com/assertj/assertj-core/blob/main/src/main/java/org/assertj/core/api/Assertions.java#L1416-L1422)

Removed line 698.

src/test/java/org/assertj/core/api/file/FileAssert_hasFileName_Test.java
nach-o-man: Ideed the guide says that your code is good. I will have to revisit it .... 😄

src/main/java/org/assertj/core/api/AbstractFileAssert.java
```diff
@@ -691,6 +692,21 @@ public SELF hasName(String expected) {
...
+   * Alias for {@link #hasName(String)}.
+   *
+   * @param expected the expected {@code File} name.
+   * @return {@code this} assertion object.
+   * @throws NullPointerException if the expected name is {@code null}.
+   * @throws AssertionError if the actual {@code File} is {@code null}.
+   * @throws AssertionError if the actual {@code File} does not have the expected name.
+   *
+   * @see java.io.File#getName() name definition.
+   */
```
scordio, line 705: I'd rather reuse the javadoc of `hasName` and introduce `@since`. Something like:

```suggestion
  /**
   * Verifies that the actual {@code File} has given name. Alias for {@link #hasName(String)}.
   *
   * <p>
   * Example:
   * <pre><code class='java'> File xFile = new File(&quot;somewhere/xFile.java&quot;);
   * File xDirectory = new File(&quot;somewhere/xDirectory&quot;);
   *
   * // assertion will pass
   * assertThat(xFile).hasFileName(&quot;xFile.java&quot;);
   * assertThat(xDirectory).hasFileName(&quot;xDirectory&quot;);
   *
   * // assertion will fail
   * assertThat(xFile).hasFileName(&quot;xFile&quot;);
   * assertThat(xDirectory).hasFileName(&quot;somewhere&quot;);</code></pre>
   *
   * @param expected the expected {@code File} name.
   * @return {@code this} assertion object.
   * @throws NullPointerException if the expected name is {@code null}.
   * @throws AssertionError if the actual {@code File} is {@code null}.
   * @throws AssertionError if the actual {@code File} does not have the expected name.
   *
   * @see java.io.File#getName() name definition.
   * @see #hasName(String)
   * @since 3.20.0
   */
```

Please also add:

```java
   * @see #hasFileName(String)
```

to the javadoc of `hasName`.

src/test/java/org/assertj/core/api/file/FileAssert_hasFileName_Test.java
scordio: The correct display name is indeed `FileAssert hasFileName`.

@joel-costigliola thinking aloud, what if we had our own custom implementation of `DisplayNameGenerator`, which removes underscores and also the suffix `Test`? This way we might avoid the repetitive `@DisplayName` on every test.

src/main/java/org/assertj/core/api/AbstractFileAssert.java
scordio: > * Have re-used `hasName` instead of adding a new method `assertHasFileName(info, actual, expected);` to `files` object.  Please let me know is this fine.

This is fine, thanks.

>     * Please suggest a better javadoc if this does not explain it well 😉

Answered in the other comment.

src/main/java/org/assertj/core/api/AbstractFileAssert.java
ShivakumarSwamy: Thank for the inputs, javadoc updated 👍

src/test/java/org/assertj/core/api/file/FileAssert_hasFileName_Test.java
scordio: With e8e221dafb69c88b64bddb8accd497e26c1635ee it is no longer necessary to add a `@DisplayName` on the tests. @ShivakumarSwamy feel free to rebase this PR.

ShivakumarSwamy: Cool 🎉 
Removed display name, updated minor version in since tag and rebased.

scordio (approved):

scordio:
Thank you, @ShivakumarSwamy!

Issue #2233, opened 2021-05-29
Add `hasExtension` to `Path` assertions


scordio:
#### Example

```java
assertThat(Paths.get("file.txt")).hasExtension("txt");
```

Related to #2205.

Syemon:
Hi, I would like to work on this issue.

scordio:
Thanks @Syemon, let us know if you need anything.

PR #2255, opened 2021-06-13
#2233 Add hasExtension to Path assertions

Syemon:
#### Check List:
* Fixes : #2233 
* Unit tests : YES

I am not sure if moving getFileExtension from internal/Files to util/Files is a correct approach. I moved it because i wanted to use it in internal/Paths.


joel-costigliola:
Thanks, we will review your PR shortly!

scordio:
> I am not sure if moving getFileExtension from internal/Files to util/Files is a correct approach. I moved it because i wanted to use it in internal/Paths.

As the complexity is not so high, I prefer to duplicate the implementation and keep it private instead of exposing a new public method in `org.assertj.core.util.Files`.

scordio:
@Syemon I updated your PR before merging as there have been some changes on how we test `Path` assertions, plus I did some small refactoring.

Thanks for your first contribution!

PR #2265, opened 2021-06-17
Map entry values handle deep equality comparison again

schlosna:
This appears to have broken in PR #2167 specifically with the use of
`Map.remove(Object, Object)` that does not perform deep equality
checking on the map entry value (e.g. arrays).

#### Check List:
* Fixes https://github.com/assertj/assertj-core/issues/2264
* Unit tests : YES


scordio:
@schlosna thank you so much for the PR. I moved the test closer to the internal implementation and added some minor changes.

schlosna:
Excellent, thanks for the quick review and merge @scordio ! Do you have an idea when you might cut the next release? Thanks!

scordio:
We would like to address https://github.com/assertj/assertj-core/pull/2173#issuecomment-863089904 and then we will deliver 3.20.2.

PR #2274, opened 2021-06-29
improve javadoc

GolanLevy:
* Unit tests : NA

Just a minor fix for the docs.


scordio:
Good catch, thanks @GolanLevy!

PR #2282, opened 2021-07-08
[MappedCondition] fix handling of mappings to null

### LINKED ISSUE #2281: MappedCondition must prevent NPE
when calling `to.getClass()`

https://github.com/assertj/assertj-core/blob/c964ea6d3f94286ac33fce2de6180288a4794313/src/main/java/org/assertj/core/condition/MappedCondition.java#L137

stbischof:
#### Check List:
* Fixes  #2281
* Unit tests :  NO 
* Javadoc with a code example (on API only) : NO  (private method)

### DIFF
--- a/src/main/java/org/assertj/core/condition/MappedCondition.java
+++ b/src/main/java/org/assertj/core/condition/MappedCondition.java
@@ -133,11 +133,16 @@ public class MappedCondition<FROM, TO> extends Condition<FROM> {
   protected String buildMappingDescription(FROM from, TO to) {
     StringBuilder sb = new StringBuilder("mapped");
     if (!mappingDescription.isEmpty()) sb.append(format("%n   using: %s", mappingDescription));
-    sb.append(format("%n   from: <%s> %s%n", from.getClass().getSimpleName(), from));
-    sb.append(format("   to:   <%s> %s%n", to.getClass().getSimpleName(), from, to));
+    sb.append(format("%n   from: <%s> %s%n", classNameOrNull(from), from));
+    sb.append(format("   to:   <%s> %s%n", classNameOrNull(to), to));
     sb.append("   then checked:");
     sb.append(format("%n      %-10s", condition));
     return sb.toString();
+  }
+
+  private static String classNameOrNull(Object object) {
+
+    return object == null?"null":object.getClass().getSimpleName();
   }
 
 }
--- a/src/test/java/org/assertj/core/condition/MappedConditionTest.java
+++ b/src/test/java/org/assertj/core/condition/MappedConditionTest.java
@@ -14,11 +14,13 @@ package org.assertj.core.condition;
 
 import static java.lang.String.format;
 import static java.lang.System.lineSeparator;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.BDDAssertions.then;
 import static org.assertj.core.api.BDDAssertions.thenNullPointerException;
 import static org.assertj.core.condition.MappedCondition.mappedCondition;
 
 import java.util.Optional;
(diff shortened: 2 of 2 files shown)


src/main/java/org/assertj/core/condition/MappedCondition.java
```diff
@@ -133,11 +133,19 @@ public boolean matches(FROM value) {
...
+    sb.append(format("   to:   <%s> %s%n", classNameOrNull(to), from, to));
     sb.append("   then checked:");
     sb.append(format("%n      %-10s", condition));
     return sb.toString();
   }
 
+  private static String classNameOrNull(Object object) {
+
+    if (object == null) {
+      return "<null>";
```
joel-costigliola, line 146: since the message L136 already specifies `<%s>`, no need to double down with additional `<>`

src/main/java/org/assertj/core/condition/MappedCondition.java
```diff
@@ -133,11 +133,19 @@ public boolean matches(FROM value) {
...
+    }
+    return object.getClass().getSimpleName();
```
joel-costigliola, line 148: use ternary operator `?` to get a one liner for `classNameOrNull`

src/main/java/org/assertj/core/condition/MappedCondition.java
```diff
@@ -133,11 +133,19 @@ public boolean matches(FROM value) {
   protected String buildMappingDescription(FROM from, TO to) {
     StringBuilder sb = new StringBuilder("mapped");
     if (!mappingDescription.isEmpty()) sb.append(format("%n   using: %s", mappingDescription));
-    sb.append(format("%n   from: <%s> %s%n", from.getClass().getSimpleName(), from));
-    sb.append(format("   to:   <%s> %s%n", to.getClass().getSimpleName(), from, to));
+    sb.append(format("%n   from: <%s> %s%n", classNameOrNull(from), from));
+    sb.append(format("   to:   <%s> %s%n", classNameOrNull(to), from, to));
```
joel-costigliola, line 137: I did not notice that before but `" to: <%s> %s%n"` has two placeholders but 3 arguments are given!

src/main/java/org/assertj/core/condition/MappedCondition.java
```diff
@@ -133,11 +133,19 @@ public boolean matches(FROM value) {
   protected String buildMappingDescription(FROM from, TO to) {
```
joel-costigliola, line 133: please add a test showing the new behavior

joel-costigliola (changes requested):
Thanks @stbischof, PR commented.

src/main/java/org/assertj/core/condition/MappedCondition.java
stbischof: ok

src/main/java/org/assertj/core/condition/MappedCondition.java
stbischof: changed

src/main/java/org/assertj/core/condition/MappedCondition.java
stbischof: removed the from

stbischof:
did all requested changes of Review1

joel-costigliola:
I took the liberty to change the description when `from` or `to` are null, in that case no need to print the class since we don't know it so I simply print the null value.

ex:
```
mapped
   using: identity
   from: null
   to:   null
   then checked:
      is null   
```

joel-costigliola:
Integrated thanks @stbischof!

stbischof:
Maybe you would add the Version label

PR #2283, opened 2021-07-09
use conditionDescriptionWithStatus on Join

stbischof:
at the moment it is only on AllOf but not AnyOf

#### Check List:
* Fixes: none
* Unit tests : NO - not new but fixes existing one
* Javadoc with a code example (on API only) : NO - not new


joel-costigliola:
Integrated thanks @stbischof!

PR #2292, opened 2021-07-15
Remove `MockPathsBaseTest`

scordio:
Notable changes:
* `MockPathsBaseTest` is gone, let's test reality!
* The scope of `NioFilesWrapper` is now reduced to the methods from `java.nio.file.Files` that might throw `IOException`
* `PathsException` is gone in favor of `UncheckedIOException`
* All tests originally inheriting from `MockPathsBaseTest` are rewritten with BDD style
* Some negative cases are failing on Windows (GitHub Actions) --> currently disabled and investigation will be covered by #2312


src/test/java/org/assertj/core/internal/paths/Paths_assertIsDirectoryContaining_with_Predicate_Test.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  @Test
+  @DisabledOnOs(value = WINDOWS, disabledReason = "gh-FIXME")
+  void should_rethrow_IOException_as_UncheckedIOException() throws Exception {
+    // GIVEN
+    Path actual = Files.createDirectory(tempDir.resolve("actual"));
+    Predicate<Path> filter = path -> true;
+    IOException cause = new IOException("boom!");
+    given(nioFilesWrapper.newDirectoryStream(actual, filter)).willThrow(cause);
```
joel-costigliola, line 96: `Paths_assertIsDirectoryContaining_with_String_Test#should_rethrow_IOException_as_UncheckedIOException` uses     `willThrow(cause).given(nioFilesWrapper).newDirectoryStream(any(), any());`  and is not disabled on windows, I have not idea why though.

src/test/java/org/assertj/core/internal/paths/Paths_assertIsExecutable_Test.java
```diff
@@ -12,50 +12,58 @@
...
-    when(nioFilesWrapper.isExecutable(actual)).thenReturn(false);
-
-    Throwable error = catchThrowable(() -> paths.assertIsExecutable(info, actual));
-
-    assertThat(error).isInstanceOf(AssertionError.class);
-    verify(failures).failure(info, shouldBeExecutable(actual));
+  @DisabledOnOs(value = WINDOWS, disabledReason = "gh-FIXME")
+  void should_fail_if_actual_is_not_executable() throws IOException {
+    // GIVEN
+    Path actual = Files.createFile(tempDir.resolve("actual"));
```
joel-costigliola, line 54: I wonder if we could explicitly make the file non executable (in case it is by default on windows which would be weird)

joel-costigliola (commented):
thanks @scordio, looks good to me, I just have some minor comments.

src/test/java/org/assertj/core/internal/paths/Paths_assertEndsWithRaw_Test.java
```diff
@@ -12,46 +12,84 @@
...
+    then(error).hasMessage(actualIsNull());
   }
 
   @Test
-  void should_fail_if_other_is_null() {
-    assertThatNullPointerException().isThrownBy(() -> paths.assertEndsWithRaw(info, actual, null))
-                                    .withMessage("the expected end path should not be null");
+  void should_fail_if_other_is_null() throws IOException {
+    // GIVEN
+    Path actual = Files.createFile(tempDir.resolve("actual"));
```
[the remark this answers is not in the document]
scordio: What about `Files.write(Path, byte[])`? E.g.:

```java
Path actual = write(tempDir.resolve("actual"), "Content".getBytes());
```

src/test/java/org/assertj/core/internal/paths/Paths_assertIsExecutable_Test.java
scordio: It seems that `actual.toFile().setExecutable(false)` [has no impact on Windows](https://github.com/assertj/assertj-core/pull/2292/checks?check_run_id=3332458561#step:5:6491).

Anyway, I'm going to rework this part (in a separate PR), removing the `File` API and using only the NIO2. With that, there should be two different ways to properly set those flags on both Unix and Windows.

src/test/java/org/assertj/core/internal/paths/Paths_assertIsDirectoryContaining_with_Predicate_Test.java
scordio: It seems we get failures on Windows only when Files.newInputStream is used.

There are also other tests like `Paths_assertIsCanonical_Test#should_rethrow_IOException_as_UncheckedIOException` or `Paths_assertHasSize_Test#should_rethrow_IOException_as_UncheckedIOException` which are working fine on Windows.

scordio: I overlooked your comment, and there were tests failing and disabled using `newDirectoryStream`. It seems that inverting the stubbing procedure fixed the issue: 57680333d70827eba86c443a10ec3f840b088414

I'll try the same with the other strange failures.

src/test/java/org/assertj/core/internal/paths/Paths_assertEndsWithRaw_Test.java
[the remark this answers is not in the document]
joel-costigliola: I would keep `Files` because `write` does not always mean writing to a file but I'm not strongly opinionated about it.

PR #2295, opened 2021-07-20
Improve assertion error for hasSameElementsAs(emptyList())

### LINKED ISSUE #2294: nondescript assertion error for hasSameElementsAs(emptyList())
#### Summary
Compared to other asserts, `hasSameElementsAs` has only a brief and generic assertion error given an empty expected collection.

#### Example

```java
    List<String> actual = List.of("element");
    Assertions.assertThat(actual).as("helpful message").hasSameElementsAs(Collections.emptyList());
```
fails with
```
java.lang.AssertionError: actual is not empty
```

In comparision
```java
    List<String> actual = List.of("element");
    Assertions.assertThat(actual).as("helpful message").containsExactlyElementsOf(Collections.emptyList());
```
has a much richer message as I would expect from AssertJ:
```
java.lang.AssertionError: [helpful message] 
Expecting actual:
  ["element"]
  []
but some elements were not expected:
  ["element"]
```

jhonnen:
Inlined commonCheckThatIterableAssertionSucceeds in assertContainsOnly and removed failIfEmptySinceActualIsNotEmpty call because expectedValues is not a subsequence.

#### Check List:
* Fixes #2294
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/internal/Iterables.java
+++ b/src/main/java/org/assertj/core/internal/Iterables.java
@@ -380,7 +380,9 @@ public class Iterables {
    */
   public void assertContainsOnly(AssertionInfo info, Iterable<?> actual, Object[] expectedValues) {
     final List<?> actualAsList = newArrayList(actual);
-    if (commonCheckThatIterableAssertionSucceeds(info, actualAsList, expectedValues)) return;
+    checkNotNullIterables(info, actualAsList, expectedValues);
+    // if both actual and values are empty, then assertion passes.
+    if (actualAsList.isEmpty() && expectedValues.length == 0) return;
 
     // after the for loop, unexpected = expectedValues - actual
     List<Object> unexpectedValues = newArrayList(actualAsList);
--- a/src/test/java/org/assertj/core/api/iterable/IterableAssert_hasSameElementsAs_Test.java
+++ b/src/test/java/org/assertj/core/api/iterable/IterableAssert_hasSameElementsAs_Test.java
@@ -19,7 +19,7 @@ import static org.assertj.core.util.Lists.newArrayList;
 import static org.mockito.Mockito.verify;
 
 /**
- * Tests for <code>{@link org.assertj.core.api.AbstractIterableAssert#containsOnlyElementsOf(Iterable)} </code>.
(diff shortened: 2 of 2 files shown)


src/test/java/org/assertj/core/api/iterable/IterableAssert_hasSameElementsAs_Test.java
```diff
@@ -19,7 +19,7 @@
 import static org.mockito.Mockito.verify;
 
 /**
- * Tests for <code>{@link org.assertj.core.api.AbstractIterableAssert#containsOnlyElementsOf(Iterable)} </code>.
+ * Tests for <code>{@link org.assertj.core.api.AbstractIterableAssert#hasSameElementsAs(Iterable)} </code>.
```
joel-costigliola, line 22: good catch

src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -380,7 +380,9 @@ private void iterablesRemove(Iterable<?> actual, Object value) {
    */
   public void assertContainsOnly(AssertionInfo info, Iterable<?> actual, Object[] expectedValues) {
     final List<?> actualAsList = newArrayList(actual);
-    if (commonCheckThatIterableAssertionSucceeds(info, actualAsList, expectedValues)) return;
+    checkNotNullIterables(info, actualAsList, expectedValues);
+    // if both actual and values are empty, then assertion passes.
+    if (actualAsList.isEmpty() && expectedValues.length == 0) return;
```
joel-costigliola, line 385: @jhonnen how does that address the issue?
`commonCheckThatIterableAssertionSucceeds` was doing the same thing with:
```
if (!actual.iterator().hasNext() && sequence.length == 0) return true;
```
You would need to add a test showing the improved behavior.

joel-costigliola (commented):
@jhonnen PR commented!

src/main/java/org/assertj/core/internal/Iterables.java
jhonnen: The issue was not caused by this check, but by `failIfEmptySinceActualIsNotEmpty(sequence)` from `commonCheckThatIterableAssertionSucceeds`.

joel-costigliola:
Integrated thanks @jhonnen!

PR #2297, opened 2021-07-26
Add satisfies with ThrowingConsumer that allow consumers to throw che…

### LINKED ISSUE #731: Add satisfies with checked exceptions consumer.
Hello,

Currently it is not possible to write code like that:

``` java

assertThat(item).satisfies(i -> {
   assertThat(i.getContent()).isEqualTo("data"); // i.getContent() throws IOException
});
```

Would it be possible to add satisfies method which can handle consumer throwing checked exception? If exception would be thrown then it would be assumed that condition failed.

Thanks,
Michał

joel-costigliola:
…cked exception

* Fixes  #731

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -817,6 +817,37 @@ public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, 
    * @throws NullPointerException if given Consumer is null
    */
   public SELF satisfies(Consumer<ACTUAL> requirements) {
+    return internalSatisfies(requirements);
+  }
+
+  /**
+   * Verifies that the actual object satisfied the given requirements expressed as a {@link ThrowingConsumer}.
+   * <p>
+   * This is the same assertion as {@link #satisfies(java.util.function.Consumer)} except that a {@link ThrowingConsumer} rethrows checked exceptions as {@link RuntimeException}.
+   * More precisely, {@link RuntimeException} and {@link AssertionError} are rethrown as they are while any other {@link Throwable} are rethrown as {@link RuntimeException}. 
+   * <p>
+   * Example:
+   * <pre><code class='java'> // read() throws IOException
+   * ThrowingConsumer&lt;Reader&gt; hasReachedEOF = reader -&gt; assertThat(reader.read()).isEqualTo(-1);
+   *
+   * // assertion succeeds as the file is empty (note that if hasReachedEOF was declared as Consumer&lt;Reader&gt; the following line would not compile): 
+   * assertThat(new FileReader("empty.txt")).satisfies(hasReachedEOF);
+   *
+   * // assertion fails as the file is not empty:
+   * assertThat(new FileReader("nonEmpty.txt")).satisfies(hasReachedEOF);</code></pre>
+   *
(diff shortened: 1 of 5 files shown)


src/main/java/org/assertj/core/api/AbstractAssert.java
```diff
@@ -817,6 +817,37 @@ public SELF matches(Predicate<? super ACTUAL> predicate, String predicateDescrip
...
+   * assertThat(new FileReader("nonEmpty.txt")).satisfies(hasReachedEOF);</code></pre>
+   *
+   * @param throwingConsumer requirements to assert on the actual object - must not be null..
+   * @return this assertion object.
+   *
+   * @throws NullPointerException if given {@link ThrowingConsumer} is null
+   * @throws RuntimeException rethrown as is by given {@link ThrowingConsumer} or wrapping any {@link Throwable}.    
+   * @throws AssertionError rethrown as is by given {@link ThrowingConsumer}  
+   */
+  public SELF satisfies(ThrowingConsumer<ACTUAL> throwingConsumer) {
```
scordio, line 846: The assertion under test below in `AbstractAssert_satisfies_with_Consumer_Test` is now the one with the `ThrowingConsumer` parameter. 

https://github.com/assertj/assertj-core/blob/45e8903129f5a45eff9cec17dc075051020b975e/src/test/java/org/assertj/core/api/abstract_/AbstractAssert_satisfies_with_Consumer_Test.java#L43-L46

Perhaps it would make sense to extract that lambda into a `Consumer` variable so that the test continues to behave as before.

src/main/java/org/assertj/core/api/ThrowingConsumer.java
```diff
@@ -0,0 +1,39 @@
...
+    try {
+      consumeThrows(input);
+    } catch (final RuntimeException | AssertionError e) {
+      throw e;
+    } catch (Throwable e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  void consumeThrows(T input) throws Throwable;
```
scordio, line 38: Minor, just to keep naming aligned:
```suggestion
  void acceptThrows(T input) throws Throwable;
```

src/test/java/org/assertj/core/api/ThrowingConsumer_accept_Test.java
```diff
@@ -0,0 +1,53 @@
...
+    ThrowingConsumer<?> throwingConsumer = throwingConsumer(ioException);
+    // WHEN
+    Throwable throwable = catchThrowable(() -> throwingConsumer.accept(null));
+    // THEN
+    then(throwable).isExactlyInstanceOf(RuntimeException.class)
+                   .hasCause(ioException);
+  }
+
+  @Test
+  void should_rethrow_runtime_exception_as_is() {
```
scordio, line 37: I'd also add a similar test for `AssertionError`

scordio (commented):
LGTM, just a few comments.

Do you also plan to add the `assertWith` / `thenWith` variants with `ThrowingConsumer`?

src/main/java/org/assertj/core/api/AbstractAssert.java
joel-costigliola: yes, good catch!

joel-costigliola:
Yes, I think we need to provide the throwing version of other assertions taking a `Consumer` (like `satisfiesAnyOf`) but I'll create issues for these

Issue #2102, opened 2021-01-15
`isUnmodifiable` assertion for Collections


scordio:
#### Summary
Considering the `java.util.Collections.unmodifiableXXX()` methods, we could add a `isUnmodifiable` method to the assert classes for collections.

#### Example

```java
List<Integer> list = Collections.unmodifiableList(Arrays.asList(1, 2, 3));
assertThat(list).isUnmodifiable();
```

HaydenMeloche:
That would be cool! How would we assert the list is unmodifiable under the hood?
We could do:
```java
list.getClass().isInstance(Collections.unmodifiableList(Arrays.asList("any value")))
```
This should work if `UnmodifiableList` or `UnmodifiableRandomAccessList` is returned.

joel-costigliola:
@HaydenMeloche that would cover the JDK types but if I use a Guava ImmutableCollection that won't work so the best implementation IMO is following the possible workaround by calling clear and catching the UnsupportedOperationException, if no exceptin is caught then fail the assertion

HaydenMeloche:
@joel-costigliola Fair point. I'd be willing to tackle this if you think it's worth working on :)

joel-costigliola:
Sure, go ahead, thanks!

scordio:
@HaydenMeloche thank you! Before start working on it please let me put together a more detailed description of the approach I have in mind, so we can blame it also considering Guava. I should be able to do it today or tomorrow max.

scordio:
I added some more details and I'll complete them during the weekend. However, this should give already a feeling so please let me know your thoughts.

I'd suggest starting the implementation of `AbstractCollectionAssert` which would inherit from `AbstractIterableAssert` and provide `isUnmodifiable`. The rest are extension points and could be done separately.

HaydenMeloche:
@scordio Thanks for the detailed explanation. I'll get started when I have some time this week. I might create a draft PR when I have something concrete so I can get feedback

HaydenMeloche:
hey @scordio got a chance to working on this over the weekend. I setup my assertion class as such
```java
public abstract class AbstractCollectionAssert <SELF extends AbstractCollectionAssert<SELF, ACTUAL, ELEMENT, ELEMENT_ASSERT>,
  ACTUAL extends java.util.Collection<? extends ELEMENT>,
  ELEMENT,
  ELEMENT_ASSERT extends AbstractAssert<ELEMENT_ASSERT, ELEMENT>>
[...]

PR #2315, opened 2021-08-17
Add `CollectionAssert` hierarchy

scordio:
Prerequisite for #2102.


scordio:
@joel-costigliola could you please check it to make sure I haven't missed anything?

src/main/java/org/assertj/core/api/ClassBasedNavigableCollectionAssert.java
```diff
@@ -0,0 +1,70 @@
...
+import java.lang.reflect.Constructor;
+import java.util.Collection;
+
+/**
+ * Build the Assert instance by reflection.
+ * 
+ * @since 3.21.0
+ */
+//@format:off
+public class ClassBasedNavigableCollectionAssert<SELF extends ClassBasedNavigableCollectionAssert<SELF, ACTUAL, ELEMENT, ELEMENT_ASSERT>,
```
scordio, line 26: I'm not sure about the concrete use cases for `ClassBasedNavigableCollectionAssert` and `FactoryBasedNavigableCollectionAssert`, and if it makes sense at all to have them. For now, the structure is similar to `ListAssert` and `IterableAssert`.

joel-costigliola: it's to get narrowed type assertions when navigating to elements,ex:
```java
Iterable<String> hobbits = newHashSet("frodo", "sam", "pippin");
assertThat(hobbits, StringAssert.class).first()
                                        .startsWith("fro")
                                        .endsWith("do");
```
This was introduced before instanceof factories which fills a similar purpose.
I don't mind leaving them out if that's possible.

src/test/java/org/assertj/core/api/InstanceOfAssertFactoriesTest.java
```diff
@@ -1089,6 +1092,26 @@ void typed_iterator_factory_should_allow_typed_iterator_assertions() {
     result.hasNext();
   }
 
+  @Test
+  void collection_factory_should_allow_collection_assertions() {
+    // GIVEN
+    Object value = asList("Homer", "Marge", "Bart", "Lisa", "Maggie");
```
joel-costigliola, line 1098: minor: I usually use `list()` from `org.assertj.core.util.util.Lists`

src/main/java/org/assertj/core/api/ClassBasedNavigableCollectionAssert.java
scordio: Ok, so I'll keep it simple and remove them. We can always add them later if users request them.

PR #2321, opened 2021-08-23
Add new assertion method for BigDecimal class

talentedasian:
This PR was really supposed to be an issue but I just wanted to play around the 
assertJ codebase. Even if this PR wouldn't get merged, it would still be wonderful to have assertion methods for scales of BigDecimals values.
#### Check List:
* Unit tests : YES 


src/main/java/org/assertj/core/api/AbstractBigDecimalAssert.java
```diff
@@ -263,6 +264,25 @@ public SELF isNotEqualByComparingTo(String notExpected) {
...
+   * assertThat(new BigDecimal(&quot;8.00&quot;)).isScaleEqualTo(2);
+   * assertThat(new BigDecimal(&quot;8.00&quot;).setScale(4)).isScaleEqualTo(4);
+   *
+   * // assertion will fail
+   * assertThat(new BigDecimal(&quot;8.00&quot;)).isScaleEqualTo(3);
+   * assertThat(new BigDecimal(&quot;8.00&quot;).setScale(4)).isScaleEqualTo(2);</code></pre>
+   * @param expected the expected scale value..
+   * @return {@code this} assertion object.
+   */
+  public SELF isScaleEqualTo(int expected) {
```
joel-costigliola, line 281: would `hasScale` be nicer?

talentedasian: Yes. Sorry for being pedantic but how about `hasScaleOf`?

joel-costigliola: sounds better :+1:

talentedasian:
On a side note, would it be cool to also add assertion methods like `isBetween` `isStrictlyBetween` for scales on BigDecimals too?

src/main/java/org/assertj/core/api/AbstractBigDecimalAssert.java
```diff
@@ -263,6 +265,26 @@ public SELF isNotEqualByComparingTo(String notExpected) {
...
+   *
+   * // assertion will fail
+   * assertThat(new BigDecimal(&quot;8.00&quot;)).hasScaleOf(3);
+   * assertThat(new BigDecimal(&quot;8.00&quot;).setScale(4)).hasScaleOf(2);</code></pre>
+   * @param expected the expected scale value..
+   * @return {@code this} assertion object.
+   */
+  public SELF hasScaleOf(int expected) {
+    isNotNull();
+    bigDecimals.assertEqual(info, actual.scale(), expected);
```
joel-costigliola, line 284: introduce `BigDecimals.asserthasScale` method so that we can throw an assertion error that mentions scale instead of a basic equal error.
This means creating a `ShouldHaveScale` error and throwing it (see https://github.com/assertj/assertj-core/blob/main/src/main/java/org/assertj/core/error/ShouldHaveMessage.java for an example)
You would have to test `ShouldHaveScale` too ;-)

joel-costigliola (changes requested):
a bit more work for you @talentedasian ;-)

joel-costigliola:
> On a side note, would it be cool to also add assertion methods like isBetween isStrictlyBetween for scales on BigDecimals too?

I would prefer to have a `scale` navigation method to be able to chain any number assertions to avoid duplcating them, ex:
```java
assertThat(bigDecimal).scale().isBetween(1, 5);
```

src/main/java/org/assertj/core/api/AbstractBigDecimalAssert.java
talentedasian: > > On a side note, would it be cool to also add assertion methods like isBetween isStrictlyBetween for scales on BigDecimals too?
> 
> I would prefer to have a `scale` navigation method to be able to chain any number assertions to avoid duplcating them, ex:
> 
> ```java
> assertThat(bigDecimal).scale().isBetween(1, 5);
> ```
> 
> We won't have `scale` mentioned in the error message though but it's ok in this case.


Makes sense. Maybe I'll put up a separate PR for that.

talentedasian:
Makes sense. Maybe I'll put up a separate PR for that.

src/main/java/org/assertj/core/internal/StandardComparisonStrategy.java
```diff
@@ -82,6 +82,7 @@ public String asText() {
    */
   @Override
   public boolean areEqual(Object actual, Object other) {
+    System.out.println(actual.equals(other) + " tite");
```
talentedasian: Please don't mind this and don't bother googling what it means either and if you do, I apologize.

src/test/java/org/assertj/core/internal/bigdecimals/BigDecimals_assertHasScale_Test.java
```diff
@@ -0,0 +1,50 @@
...
+import java.math.BigDecimal;
+
+import static java.lang.String.format;
+import static java.math.BigDecimal.ONE;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.assertj.core.api.BDDAssertions.then;
+import static org.assertj.core.test.TestData.someInfo;
+import static org.assertj.core.util.FailureMessages.actualIsNull;
+
+class BigDecimals_assertHasScale_Test extends BigDecimalsBaseTest {
```
joel-costigliola, line 16: follow the GIVEN WHEN THEN pattern as in `Iterables_assertContainsOnly_Test`

src/main/java/org/assertj/core/error/ShouldHaveScale.java
```diff
@@ -0,0 +1,12 @@
+package org.assertj.core.error;
+
+public class ShouldHaveScale extends BasicErrorMessageFactory {
+
+  public static ErrorMessageFactory shouldHaveScale(int actual, int expected) {
+    return new ShouldHaveScale(actual, expected);
+  }
+
+  private ShouldHaveScale(int actual, int expected) {
+    super("%nexpecting actual to have a scale of: %s%n  but has a scale of: %s", expected, actual);
```
joel-costigliola, line 10: let's pass actual so that we can display in the error message as in:
```
"%nexpecting actual %s to have a scale of:%n  %s%nbut had a scale of:%n  %s", actual, expected, actual.getScale()
```
with this error message (if I got it correctly) scales are aligned in the output

src/test/java/org/assertj/core/error/ShouldHaveScale_create_Test.java
```diff
@@ -0,0 +1,23 @@
...
+import org.assertj.core.presentation.StandardRepresentation;
+import org.junit.jupiter.api.Test;
+
+import static java.lang.String.format;
+import static org.assertj.core.api.BDDAssertions.then;
+
+public class ShouldHaveScale_create_Test {
+
+  @Test
+  public void should_create_error_message() {
```
joel-costigliola, line 14: add GIVEN WHEN THEN

src/test/java/org/assertj/core/internal/bigdecimals/BigDecimals_assertHasScale_Test.java
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  @Test
+  void should_pass_if_scales_are_equal_whatever_custom_strategy_is_used() {
+    numbersWithAbsValueComparisonStrategy.assertHasScale(someInfo(), ONE_WITH_3_DECIMALS, 3);
+  }
+
+  @Test
+  void should_fail_if_scales_are_not_equal() {
+    AssertionError assertionError = AssertionsUtil.expectAssertionError(() -> numbers.assertHasScale(someInfo(), ONE, 3));
```
joel-costigliola, line 38: import static `expectAssertionError` for readability

joel-costigliola (changes requested):
a few more changes and we are good thanks @talentedasian

src/test/java/org/assertj/core/internal/bigdecimals/BigDecimals_assertHasScale_Test.java
talentedasian: I was kind of confused as to what to do because of some tests that I looked at. If what I did was incorrect just let me know :)

joel-costigliola:
Integrated thanks @talentedasian!

PR #2333, opened 2021-09-02
Apply pattern to file name only with `isDirectoryContaining` / `isDirectoryNotContaining`

### LINKED ISSUE #2329: isDirectoryContaining fails for expected patterns
Let's assume that there is a directory `C:\Temp\junit12854012982575283912` which contains a file `test-6e672c4ba0be4f0886d434125a0d062c.txt`.

The assertion of `assertThat(dir).isDirectoryContaining("glob:test-[0-9a-f]*.txt")` and `assertThat(dir).isDirectoryContaining("regex:test-[0-9a-f]+\\.txt")` fails although it is not expected do so.

The error messages is confusing and does not give a hint why this pattern would be wrong.

```
Expecting directory:
  C:\Temp\junit12854012982575283912\out
to contain at least one file matching the 'glob:test-[0-9a-f]*.txt' pattern but there was none.
The directory content was:
  [test-6e672c4ba0be4f0886d434125a0d062c.txt]
```

Using the same with a `PathMatcher` succeeds.

```
assertThat(dir.getFileSystem().getPathMatcher("glob:test-[0-9a-f]*.txt").matches(Path.of("test-6e672c4ba0be4f0886d434125a0d062c.txt"))).isTrue();
assertThat(dir.getFileSystem().getPathMatcher("regex:test-[0-9a-f]+\\.txt").matches(Path.of("test-6e672c4ba0be4f0886d434125a0d062c.txt"))).isTrue();
```

There is also a difference between using `Path` instead of `File`. If the directory variable is of type `Path`, the error outputs the absolute paths of the file in the directory. That gives a hint about what might be wrong.

```
Expecting directory:
  C:\Temp\junit12854012982575283912\out
to contain at least one file matching the 'glob:test-[0-9a-f]*.txt' pattern but there was none.
The directory content was:
  [C:\Temp\junit12854012982575283912\out\test-6e672c4ba0be4f0886d434125a0d062c.txt]
```

`isDirectoryContaining` tests against the absolute path of the files although it should test against their relative paths. The patterns from my example will of course never match against the absolute path! Using `"glob:**test-[0-9a-f]*.txt"` works but the expectation when using the API is that the paths are relative. If it is really the intention to test against the absolute path, the docs should be very clear about this.

p.s.
This example in the javadoc `.isDirectoryContaining("glob:**.{ext,bin")` is missing a closing "}". It should be `.isDirectoryContaining("glob:**.{ext,bin}")`.

scordio:
Fixes #2329.

- [x] Update `Path` assertions
- [x] Update `File` assertions
- [x] Align error message format between `File` and `Path` assertions: absolute or relative paths?
- [x] Evaluate `isDirectoryRecursivelyContaining` behavior change for both `File` and `Path` assertions
- [x] Update javadoc

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractFileAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractFileAssert.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.charset.Charset;
 import java.nio.file.FileSystem;
+import java.nio.file.PathMatcher;
 import java.security.MessageDigest;
 import java.util.function.Predicate;
 
@@ -619,7 +620,7 @@ public abstract class AbstractFileAssert<SELF extends AbstractFileAssert<SELF>> 
    * @since 3.21.0
    */
   public SELF isWritable() {
-	  return canWrite();
+    return canWrite();
   }
 
   /**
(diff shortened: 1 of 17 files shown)


scordio:
> Evaluate `isDirectoryRecursivelyContaining` behavior change for both `File` and `Path` assertions

The behavior will be changed in a separate PR, most likely for version 3.22.0.

scordio:
> Align error message format between `File` and `Path` assertions: absolute or relative paths?

As per internal discussion, all the error messages should display the filenames only.

scordio:
Closing as the fix will take a different direction.

PR #2373, opened 2021-10-17
Fix #2366: Multiline object formatting

### LINKED ISSUE #2366: Create beautiful error messages for multi-line values and nested asserts
#### Summary

AssertJ error messages currently do not take into account that `toString` of asserted values may contain new lines. 

The resulting error messages can at times look awkward due to missing indentation. 

#### Example

For example let's consider a class where `toString` returns multi-line XML:

```xml
<xml>
  <value>A</value>
</xml>
```

For simplicity I will use the following:

```java
  public static class Xml {
    private final String value;
    public Xml(String value) {
      this.value = value;
    }
    @Override
      if (this == o) return true;
      Xml xml = (Xml) o;
      return Objects.equals(value, xml.value);
    }
    @Override
    public int hashCode() {
      return Objects.hash(value);
    }
    @Override
    public String toString() {
      return "<xml>\n  <value>" + value + "</value>\n</xml>";
    }
  }

  private static Xml xml(String value) {
    return new Xml(value);
  }
```

```java
assertThat(xml("A")).isNull();
```

This will produce the following error message:

```
org.opentest4j.AssertionFailedError: 
expected: null
 but was: <xml>
  <value>A</value>
</xml>
```

Notice that the first line is awkwardly placed directly after "but was". Also notice that last line has same indentation as "expected".

This message could be improved to e.g. 

```
org.opentest4j.AssertionFailedError: 
expected: 
  null
but was: 
  <xml>
    <value>A</value>
  </xml>
```

Similarly:

```java
assertThat(xml("A")).isEqualTo(xml("B"));
```

currently produces the following error:

```
org.opentest4j.AssertionFailedError: 
expected: <xml>
  <value>B</value>
</xml>
 but was: <xml>
  <value>A</value>
</xml>
```

This could be improved to e.g.:

```
org.opentest4j.AssertionFailedError: 
expected: 
  <xml>
    <value>B</value>
  </xml>
but was: 
  <xml>
    <value>A</value>
  </xml>
```

Indentation becomes even more useful to decipher error messages when we use `anySatisfy`:

```java
assertThat(Arrays.asList(xml("A"), xml("B"))).anySatisfy(v1 -> {
  assertThat(v1).isEqualTo(xml("C"));
});
```

Currently AssertJ produces:

```
java.lang.AssertionError: 
Expecting any element of:
  [<xml>
  <value>A</value>
</xml>, <xml>
  <value>B</value>
</xml>]
to satisfy the given assertions requirements but none did:

<xml>
  <value>A</value>
</xml>
error: 
expected: <xml>
  <value>C</value>
</xml>
 but was: <xml>
  <value>A</value>
</xml>

<xml>
  <value>B</value>
</xml>
error: 
expected: <xml>
  <value>C</value>
</xml>
 but was: <xml>
  <value>B</value>
</xml>
```

This could be improved to e.g.:

```
java.lang.AssertionError: 
Expecting any element of:
  [
    <xml>
      <value>A</value>
    </xml>
  , 
    <xml>
      <value>B</value>
    </xml>
  ]
to satisfy the given assertions requirements but none did:

  Element
    <xml>
      <value>A</value>
    </xml>
  : 
    expected: 
      <xml>
        <value>C</value>
      </xml>
    but was: 
      <xml>
        <value>A</value>
      </xml>
  
  Element
    <xml>
      <value>B</value>
    </xml>
  : 
    expected: 
      <xml>
        <value>C</value>
      </xml>
    but was: 
      <xml>
        <value>B</value>
      </xml>
```

The following extreme case produces very long and hard to decipher error messages. 

```java
assertThat(Arrays.asList(xml("A"), xml("B"))).anySatisfy(v1 -> {
  assertThat(Arrays.asList(xml("C"), xml("D"))).anySatisfy(v2 -> {
    assertThat(v1).isEqualTo(v2);
  });
});
```

I will not include the long error here but I think indentation could also help a lot here to make errors parsable by humans.

temp-droid:
WIP - Some tests are broken, to be investigated.

#### Check List:
* Fixes #2366
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
+++ b/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
@@ -330,7 +330,12 @@ public class StandardRepresentation implements Representation {
    * @return to {@code toString} representation for the given object
    */
   protected String fallbackToStringOf(Object object) {
-    return object.toString();
+    String str = object.toString();
+    if (str == null) return null;
+    if (str.contains("\r\n")) return String.format("\r\n%s", str).replace("\r\n", "\r\n  ");
+    if (str.contains("\n")) return String.format("\n%s", str).replace("\n", "\n  ");
+    if (str.contains("\r")) return String.format("\r%s", str).replace("\r", "\r  ");
+    return str;
   }
 
   protected String toStringOf(Number number) {
@@ -368,7 +373,7 @@ public class StandardRepresentation implements Representation {
   protected String toStringOf(ComparatorBasedComparisonStrategy comparatorBasedComparisonStrategy) {
     String comparatorDescription = comparatorBasedComparisonStrategy.getComparatorDescription();
     return comparatorDescription == null ? toStringOf(comparatorBasedComparisonStrategy.getComparator())
-        : quote(comparatorDescription);
+      : quote(comparatorDescription);
   }
 
   protected String toStringOf(Calendar calendar) {
@@ -470,7 +475,7 @@ public class StandardRepresentation implements Representation {
     if (!entriesIterator.hasNext()) return "{}";
     StringBuilder builder = new StringBuilder("{");
     int printedElements = 0;
-    for (;;) {
+    for (; ; ) {
       Entry<?, ?> entry = (Entry<?, ?>) entriesIterator.next();
       if (printedElements == maxElementsForPrinting) {
         builder.append(DEFAULT_MAX_ELEMENTS_EXCEEDED);
--- a/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
+++ b/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
@@ -65,6 +65,7 @@ import java.util.concurrent.atomic.LongAdder;
 import java.util.function.Function;
 import java.util.stream.Stream;
 
+import org.assertj.core.api.recursive.comparison.RecursiveComparisonConfiguration;
 import org.assertj.core.configuration.Configuration;
 import org.assertj.core.configuration.ConfigurationProvider;
 import org.assertj.core.data.MapEntry;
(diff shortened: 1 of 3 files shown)


temp-droid:
The impacted tests are in the classes `ShouldBeEqualByComparingFieldByFieldRecursively_create_Test` and `ShouldNotBeEqualComparingFieldByFieldRecursively_create_Test`.

They are failing because the formatting has changed while displaying the `configuration` used, for example (issue occurs after `The recursive comparison was performed with this configuration`):
```
org.opentest4j.AssertionFailedError: 
expected: "[Test] 
Expecting actual:
  "Yoda"
not to be equal to:
  "Luke"
when recursively comparing field by field
The recursive comparison was performed with this configuration:

  - no overridden equals methods were used in the comparison (except for java types)
  - these types were compared with the following comparators:
    - java.lang.Double -> DoubleComparator[precision=1.0E-15]
    - java.lang.Float -> FloatComparator[precision=1.0E-6]
    - java.nio.file.Path -> lexicographic comparator (Path natural order)
  - actual and expected objects and their fields were compared field by field recursively even if they were not of the same type, this allows for example to compare a Person to a PersonDto (call strictTypeChecking(true) to change that behavior).
  "
 but was: "[Test] 
Expecting actual:
  "Yoda"
not to be equal to:
  "Luke"
when recursively comparing field by field
The recursive comparison was performed with this configuration:
- no overridden equals methods were used in the comparison (except for java types)
- these types were compared with the following comparators:
  - java.lang.Double -> DoubleComparator[precision=1.0E-15]
  - java.lang.Float -> FloatComparator[precision=1.0E-6]
  - java.nio.file.Path -> lexicographic comparator (Path natural order)
- actual and expected objects and their fields were compared field by field recursively even if they were not of the same type, this allows for example to compare a Person to a PersonDto (call strictTypeChecking(true) to change that behavior).
"
```

Should we apply the same formatting to those messages or find a way to keep them with the current formatting?

joel-costigliola:
I prefer the current formatting that does not add a new line after `The recursive comparison was performed with this configuration:`

joel-costigliola:
Integrated thanks @temp-droid!
the issue was more complex than what I imagined, the merged code is quite different from the initial PR but your PR was valuable, I kept the idea of detecting newline and indenting the representation but limited it to the `isEqualTo` error message instead of baking it in the standard representation

PR #2382, opened 2021-10-21
Issue 2379 map assert throws npe on null key and value

### LINKED ISSUE #2379: MapAssert throws NPE on null key and value
#### Summary
`MapAssert` throws NPE on `null` key and value, if `Map` implementation does not accept `null` values like `java.util.Map#of(*)`.
For `List` there ist a special method  `ListAssert#doesNotContainNull()`.
Maybe `MapAssert` should also have special treatment for `null` keys and values.
At least the negative verification methods shouldn't throw an exception.

#### Example

```java
assertThat(List.of("a", "b")).doesNotContainNull(); // OK

assertThat(Map.of("a", "b")).containsValue(null); // NPE
assertThat(Map.of("a", "b")).containsKey(null); // NPE

assertThat(Map.of("a", "b")).doesNotContainKey(null); // NPE
assertThat(Map.of("a", "b")).doesNotContainValue(null); // NPE
```

run with org.assertj:assertj-core:3.21.0

BenTels:
#### Check List:
* Fixes #2379 (partially)
* Unit tests : YES 
* PR meets the [contributing guidelines](https://github.com/assertj/assertj-core/blob/main/CONTRIBUTING.md) : Think so...

This is a draft PR as requested by @scordio .

This fixes the NPE thrown when asserting containsKey(null) and doesNotContainKey(null) on certain map implementations that do not allow null keys.

Note: I am hoping that in fixing the *Value(null) side of it a pattern will emerge that will allow me to refactor the try/catch constructions into something neater...


src/test/java/org/assertj/core/api/map/MapAssert_containsKey_Test.java
```diff
@@ -35,4 +44,25 @@
   protected void verify_internal_effects() {
     verify(maps).assertContainsKeys(getInfo(assertions), getActual(assertions), array("key1"));
   }
+
+  @ParameterizedTest
+  @MethodSource("maps_without_null_keys_or_values")
+  protected void should_fail_when_asserting_for_null_on_map_that_does_not_contain_null(final Map<?, ?> actual) {
```
scordio, line 50: With the current architecture, the best would be adding cases to the corresponding test class in `org.assertj.core.internal.maps`.

src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -614,4 +625,44 @@ private void assertNotNull(AssertionInfo info, Map<?, ?> actual) {
...
+       * containsValue, for example). If we catch an NPE here and the argument to the predicate
+       * is null, we will assume we are dealing with a predicate that uses such a method and ignore the 
+       * exception as the map will definitely not contain null.
+       */
+      if (predicateOperand != null) {
+        // The NPE was thrown for a different reason than described above. Rethrow it.
+        throw npe;
+      }
+    }
+  }
```
scordio, line 666: I have the feeling that these two methods are a bit over the threshold of necessary complexity, probably two "smart" wrappers for `containsKey` and `containsValue` could also be enough. I'll experiment in this direction.

scordio (commented):
@BenTels thanks for your patience and sorry if I made you wait for this change, I finally have some time to focus on it.

I have some comments and proposed changes, but no action needs to be taken at the moment. I need a little more time to think about it and will take care of the changes if necessary (but the credits are yours, of course).

src/test/java/org/assertj/core/api/MapAssertBaseTest.java
```diff
@@ -56,4 +62,49 @@ protected void inject_internal_objects() {
...
+  }
+
+  private static Map<String, String> concurrentMap() {
+    Map<String, String> concurrentMap = new ConcurrentHashMap<>();
+    concurrentMap.put("Whatever", "Don't care");
+    return concurrentMap;
+  }
+
+  private static Map<String, String> java_util_Map_of_simulation() {
+    class SingletonNoNullKeysMap<K, V> extends HashMap<K, V> {
```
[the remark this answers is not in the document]
scordio: Fair point. I think we could add a dedicated test to protect against the behavioral changes of that implementation, making sure that's what we want to have for the assertion test case.

src/test/java/org/assertj/core/api/map/MapAssert_containsValue_Test.java
```diff
@@ -34,4 +44,25 @@
...
+    // actual is a Map that has no null keys
+    // WHEN
+    AssertionError assertionError = expectAssertionError(() -> assertThat(actual).containsValue(null));
+    // THEN
+    /*
+     * This is not ideal. It feels to me like this should be a call to
+     * ShouldContainKey.shouldContainKey(). However, internally the assertion
+     * error is generated by ShouldContainKeys -- and that has a different
+     * error message.
+     */
```
scordio, line 64: I agree that this could be improved. I raised #2428 about it.

scordio:
Thanks for all your work, @BenTels! The fix took a slightly different direction, but I used your changes as input and kept you as a co-author.

PR #2389, opened 2021-10-30
#2374 Accept vararg for AbstractAssert.satisfies

### LINKED ISSUE #2374: Add `satisfiesAllOf` to `AbstractAssert`
#### Summary
AssertJ already has support for soft assertions. However, that feature might be difficult to discover as it's not part of the main API and also a bit error-prone considering that you have to remember to call `assertAll` in the end.

`AbstractAssert` already contains a `satisfiesAnyOf` method, so this proposed method would be a nice complement to that. It would be easy to discover and would be less error-prone.

#### Example

```java
TolkienCharacter frodo = new TolkienCharacter("Frodo", HOBBIT);
    
Consumer<TolkienCharacter> isHobbit = tolkienCharacter -> assertThat(tolkienCharacter.getRace()).isEqualTo(HOBBIT);
Consumer<TolkienCharacter> isElf = tolkienCharacter -> assertThat(tolkienCharacter.getRace()).isEqualTo(ELF);
Consumer<TolkienCharacter> isFrodo = tolkienCharacter -> assertThat(tolkienCharacter.getName()).isEqualTo("Frodo");
Consumer<TolkienCharacter> isBoromir = tolkienCharacter -> assertThat(tolkienCharacter.getRace()).isEqualTo("Boromir");
    
// assertion succeeds:
assertThat(frodo).satisfiesAllOf(isHobbit, isFrodo);
    
// assertion fails reporting all failures:
assertThat(boromir).satisfiesAllOf(isElf, isBoromir);
```

If the proposal is accepted I would be happy to contribute.

martin-tarjanyi:
#### Check List:
* Fixes #2374 
* Unit tests : YES


src/main/java/org/assertj/core/api/WithAssertions.java
```diff
@@ -2601,7 +2601,7 @@ default OptionalLongAssert assertThat(final OptionalLong optional) {
    *               assertThat(stats.reboundsPerGame).isBetween(9, 12);
    *            });</code></pre>
    * <p>
-   * {@code assertWith} is variation of {@link AbstractAssert#satisfies(Consumer)} hopefully easier to find for some users.
+   * {@code assertWith} is variation of {@link AbstractAssert#satisfies(Consumer[])} hopefully easier to find for some users.
```
martin-tarjanyi, line 2604: I was not sure whether I should add a vararg version to the related method here, since it is an interface:

https://github.com/assertj/assertj-core/blob/016ebe460da0a144f7d24042db4672ee7bc31923/src/main/java/org/assertj/core/api/WithAssertions.java#L2612-L2615

src/test/java/org/assertj/core/api/abstract_/AbstractAssert_satisfies_with_Consumers_Test.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  @Test
+  void should_fail_not_satisfying_single_requirement() {
+    // GIVEN
+    Consumer<Jedi> isNamedVader = jedi -> assertThat(jedi.getName()).as("check vader").isEqualTo("Darth Vader");
+    // WHEN
+    AssertionError assertionError = expectAssertionError(() -> assertThat(yoda).satisfies(isNamedVader));
+    // THEN
+    then(assertionError.getMessage()).contains("check vader");
```
martin-tarjanyi, line 52: I noticed that even if there is only a single failing requirement, the error report contains message "Multiple failures...". Would it make sense to create a shortcut for the case when there is only one failure, so the error message would be more simple?

martin-tarjanyi:
I see binary compatibility check fails. Should I keep the original version of `satisfies`?

src/main/java/org/assertj/core/api/WithAssertions.java
joel-costigliola: I would say yes but it might clash with this one, in this case modify add vararg to this instead of adding a new `assertWith`

src/test/java/org/assertj/core/api/abstract_/AbstractAssert_satisfies_with_Consumers_Test.java
```diff
@@ -0,0 +1,95 @@
...
+    then(yoda).satisfies(notNullObjectConsumer);
+  }
+
+  @Test
+  void should_fail_if_consumer_is_null() {
+    // GIVEN
+    Consumer<Jedi> nullRequirements = null;
+    // WHEN/THEN
+    thenIllegalArgumentException().isThrownBy(() -> assertThat(yoda).satisfies(nullRequirements))
+                                  .withMessage("No assertions group should be null");
```
joel-costigliola, line 93: add a test with one null and one non null consumer.

src/test/java/org/assertj/core/api/abstract_/AbstractAssert_satisfies_with_ThrowingConsumers_Test.java
```diff
@@ -97,23 +114,12 @@ void should_propagate_RuntimeException_as_is() {
...
-
   @Test
   void should_fail_if_throwing_consumer_is_null() {
     // GIVEN
     ThrowingConsumer<String> nullRequirements = null;
     // WHEN/THEN
-    thenNullPointerException().isThrownBy(() -> assertThat("foo").satisfies(nullRequirements))
-                              .withMessage("The Consumer<T> expressing the assertions requirements must not be null");
+    thenIllegalArgumentException().isThrownBy(() -> assertThat("foo").satisfies(nullRequirements))
+                                  .withMessage("No assertions group should be null");
```
joel-costigliola, line 123: add a test with one null and one non null consumer.

src/test/java/org/assertj/core/api/abstract_/AbstractAssert_satisfies_with_ThrowingConsumers_Test.java
```diff
@@ -17,32 +17,20 @@
...
-
-  @BeforeEach
-  void setup() {
-    isEOF = path -> {
-      assertThat(isReadable(path)).isTrue();
-      // this would not compile if isEOF was declared as Consumer<Path> since it can throw an IOException
-      assertThat(readAllLines(path)).isEmpty();
-    };
-  }
+class AbstractAssert_satisfies_with_ThrowingConsumers_Test {
```
joel-costigliola, line 30: use `hasMessageNotContaining​` and `hasMessageContaining​` whenever possible in these tests

src/test/java/org/assertj/core/api/abstract_/AbstractAssert_satisfies_with_Consumers_Test.java
```diff
@@ -0,0 +1,95 @@
...
+import static org.assertj.core.api.BDDAssertions.then;
+import static org.assertj.core.api.BDDAssertions.thenIllegalArgumentException;
+import static org.assertj.core.util.AssertionsUtil.expectAssertionError;
+
+import java.util.function.Consumer;
+
+import org.assertj.core.test.Jedi;
+import org.junit.jupiter.api.Test;
+
+class AbstractAssert_satisfies_with_Consumers_Test {
```
joel-costigliola, line 25: use `hasMessageNotContaining​` and `hasMessageContaining​` whenever possible in these tests

joel-costigliola:
Thanks , first round of code review done, thanks @martin-tarjanyi.

src/main/java/org/assertj/core/api/WithAssertions.java
martin-tarjanyi: For consistency I modfied the existing one.

joel-costigliola:
Integrated thanks @martin-tarjanyi!

PR #2393, opened 2021-11-06
Issue2211 has no null fields or properties recurrence

### LINKED ISSUE #2211: `hasNoNullFieldsOrProperties` - recurrence
#### Summary

When we have more complicated objects structure it will be useful to have possibility to check that all fields for actual object and all fields in included object are also not null. 

#### Example

```java
    class Author {
        String name;
        String email;

        Author(String name, String email) {
            this.name = name;
            this.email = email;
        }
    }

    class Book {
        String title;
        Author author;

        Book(String title, Author author) {
            this.title = title;
            this.author = author;
        }
    }

    @Test
    void testHasNoNullFieldsOrProperties() {
        Book book = new Book("title", new Author(null, null));
        assertThat(book).hasNoNullFieldsOrProperties();
        assertThat(book.author).hasNoNullFieldsOrProperties();
    }

```

assertion `assertThat(book).hasNoNullFieldsOrProperties()` will pass and it is true
second assertion `assertThat(book.author).hasNoNullFieldsOrProperties()` will fail and it is also ok.

When we have object with many other object we must write many assertions for each fields.
When we add new field to object we must also remember to add new assertion.

#### Proposition
 
assertion `assertThat(book).hasNoNullFieldsOrProperties()` should fail because object `book.author` has fields with `null` values.

BenTels:
#### Check List:
* Fixes #2211
* Unit tests : YES

My solution is a little different than what was proposed in the original issue. I added a `usingRecursiveAssertion()` rather than adding onto `usingRecursiveComparison()` -- mostly because recursive assertion really isn't comparison at all.

I reused code from recursive comparison where possible, so recursive assertion supports some of the same ignore rules. In addition recursive assertion can be configured to ignore primitives, to include JCL types and can be configured for different ways of handling collections and maps.

I also added a convenience method for asserting not null, because I expect that to be used most frequently.

Interested to hear what you guys think. Please let me know whether the documentation is up to your standard.


scordio:
@BenTels thanks for the PR. We'll discuss the solution in the team and get back to you.

joel-costigliola:
Sorry for the delayed review, we'll try to provide feedback by the end of the year.

joel-costigliola:
@BenTels I have finally taken the time to work on this, I already rebased your work on main and did some refactoring (I'm keeping you at the commits author to give you credit), I haven't changed the approach which I found quite good, thanks!

I have a few questions but let's start with the optional handling, I'm not sure to see a use case for `ignoreAllEmptyOptionalFields`, did you have an example in mind where it would be valuable?
On the same topic, we could have a config to define how optional are handled a bit like collection (after all an optional is simply a collection of 0 or 1 element) so something like ELEMENT_ONLY, OPTIONAL_OBJECT_ONLY, OPTIONAL_OBJECT_AND_VALUE.

WDYT?

joel-costigliola:
Integrated thanks @BenTels!

Issue #2356, opened 2021-10-01
InjectSoftAssertions doesn't work as expected when parallel mode is enabled.


Rusgallaz:
#### Summary
Framework: Junit 5
Using InjectSoftAssertions and running tests in concurrent mode are leading to incorrect test results. Tests that should have failed are eventually passed.

#### Example

```java
@ExtendWith(SoftAssertionsExtension.class)
public class SoftAssertionTest {

    @InjectSoftAssertions private SoftAssertions softAssertions;

    @Test
    @DisplayName("Test 1")
    void test1() {
        softAssertions.assertThat("Test 1").as("Test 1").isEqualTo("Fail");
    }

    @Test
    @DisplayName("Test 2")
    void test2() {
        softAssertions.assertThat("Test 2").as("Test 2").isEqualTo("Fail");
    }
}
```
Both tests should fail, but one of them passes and the other fails with two errors(from test one and test two).

Screenshot from IDE
![image](https://user-images.githubusercontent.com/39244350/135622106-aa2564fa-d18d-4fe5-b9a3-c6f09c35a539.png)

If i disable parallel mode in the junit-platfrom.properties or set these two annotations 
```java
    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
    @Execution(ExecutionMode.CONCURRENT)
```
it will help. But this is not a solution for me.

Repository to reproduce the issue. You can run tests by maven or from IDE. The result will be the same.

#### Version 
assertj-core  3.21.0

joel-costigliola:
Thanks for reporting this issue with great details!

kriegfrj:
I'll try and have a look at this one next week.

JamesPeters98:
This is still an issue as of 3.23.1, any chance it can be looked at as it seems like this issue got missed :)

scordio:
Hi @JamesPeters98, we'll do our best to keep this issue in scope for 3.24.0.

kriegfrj:
> I'll try and have a look at this one next week.

Sorry, that has been an extremely long week. Hopefully it will get shorter in the coming... errmm... weeks...

etrandafir93:
This is similar to #3550. It appears that a single test thread entering afterTestExecution collects all assertion errors and reports them together, while other tests with no errors are shown as passing (green).

If handled correctly, this could resolve #3550, though I’m not sure of the exact approach. 

[...]

PR #2403, opened 2021-11-11
[soft] Meta-annotate using @ExtendWith

kriegfrj:
From Jupiter 5.8 and on, this allows Jupiter soft assertion tests to automatically register the `SoftAssertionsExtension` simply by use of the `@InjectSoftAssertions` field.

At least, that's the theory. The unit test I put in place to test it doesn't seem to be working. Hailing @bjhargrave (who added this feature to [osgi-test](https://github.com/osgi/osgi-test)) to see if he can shed any light on this issue.

### DIFF
--- a/src/main/java/org/assertj/core/api/junit/jupiter/InjectSoftAssertions.java
+++ b/src/main/java/org/assertj/core/api/junit/jupiter/InjectSoftAssertions.java
@@ -21,13 +21,15 @@ import java.lang.annotation.Target;
 import org.assertj.core.api.BDDSoftAssertions;
 import org.assertj.core.api.SoftAssertions;
 import org.assertj.core.api.SoftAssertionsProvider;
+import org.junit.jupiter.api.extension.ExtendWith;
 
 /**
- * Annotation used with {@link SoftAssertionsExtension} for specify wich test instance fields should be initialised with
+ * Annotation used with {@link SoftAssertionsExtension} for specify which test instance fields should be initialised with
  * a {@link SoftAssertionsProvider} concrete implementation, for example {@link SoftAssertions}, {@link BDDSoftAssertions} or any
  * custom soft assertions class.
  */
 @Retention(RUNTIME)
 @Target(FIELD)
+@ExtendWith(SoftAssertionsExtension.class)
 public @interface InjectSoftAssertions {
 }
--- a/src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
+++ b/src/main/java/org/assertj/core/api/junit/jupiter/SoftAssertionsExtension.java
@@ -45,7 +45,7 @@ import org.junit.jupiter.api.extension.ExtensionContext.Store;
 import org.junit.jupiter.api.extension.ParameterContext;
 import org.junit.jupiter.api.extension.ParameterResolutionException;
 import org.junit.jupiter.api.extension.ParameterResolver;
-import org.junit.jupiter.api.extension.TestInstancePostProcessor;
+import org.junit.jupiter.api.extension.TestInstances;
 import org.junit.jupiter.api.parallel.Execution;
 import org.junit.jupiter.api.parallel.ExecutionMode;
 import org.junit.platform.commons.annotation.Testable;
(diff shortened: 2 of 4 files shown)


bjhargrave:
I am not sure what is going on here. I built the assertj-core snapshot with the change to `InjectSoftAssertions` and tried to use it in the osgi-test build after removing the `@ExtendWith(SoftAssertionsExtension.class)` in `PromiseAssertTest`. It fails in the same way with the field injection not happening.

I then looked at how the two extensions, `SoftAssertionsExtension` and `BundleContextExtension`, do the field injection. In `BundleContextExtension`, we do the field injection during the `BeforeEachCallback` (and `BeforeAllCallback`) callback. In `SoftAssertionsExtension`, it appears to do the field injection during the `TestInstancePostProcessor` callback. So I am not sure if this is something about how JUnit's support of `@ExtendWith` on the annotation is working or not.

kriegfrj:
Thanks for the insight, @bjhargrave. I'll try digging further. Just wanted to make sure I hadn't missed something obvious.

kriegfrj:
> I then looked at how the two extensions, `SoftAssertionsExtension` and `BundleContextExtension`, do the field injection. In `BundleContextExtension`, we do the field injection during the `BeforeEachCallback` (and `BeforeAllCallback`) callback. In `SoftAssertionsExtension`, it appears to do the field injection during the `TestInstancePostProcessor` callback. So I am not sure if this is something about how JUnit's support of `@ExtendWith` on the annotation is working or not.

This is the key. `SoftAssertionsExtension` uses the `TestInstancePostProcessor` callback so that it can transparently handle `PER_CLASS` + `CONCURRENT` (something we have not concerned ourselves with in osgi-test) using `ThreadLocal` under-the-hood. The injection is done in two phases - the TIPP callback initialises & injects the `SoftAssertionsProvider` instances, and `beforeEach()` sets the delegate of each `SoftAssertionsProvider` to the common collector.

When you use implicit registration, it seems that the TIPP callback is not being called.

I'm trying to find a workaround but it's not easy...

bjhargrave:
> When you use implicit registration, it seems that the TIPP callback is not being called.

Perhaps this is a bug in JUnit? Or an undocumented restriction?

kriegfrj:
> > When you use implicit registration, it seems that the TIPP callback is not being called.
> 
> Perhaps this is a bug in JUnit? Or an undocumented restriction?

I think definitely it is one of those two...

kriegfrj:
> > > When you use implicit registration, it seems that the TIPP callback is not being called.
> > 
> > 
> > Perhaps this is a bug in JUnit? Or an undocumented restriction?
> 
> I think definitely it is one of those two...

Or perhaps a third option: it is a *documented* restriction:

From [section 5.2.1](https://junit.org/junit5/docs/current/user-guide/#extensions-registration-declarative):
> @ExtendWith fields may be either static or non-static. The documentation on Static Fields and Instance Fields for @RegisterExtension fields also applies to @ExtendWith fields.

The relevant documentation for `@RegisterExtension` [here](https://junit.org/junit5/docs/current/user-guide/#extensions-registration-programmatic-instance-fields): 
>If a @RegisterExtension field is non-static (i.e., an instance field), the extension will be registered after the test class has been instantiated and after each registered TestInstancePostProcessor has been given a chance to post-process the test instance (potentially injecting the instance of the extension to be used into the annotated field). Thus, if such an instance extension implements class-level or instance-level extension APIs such as BeforeAllCallback, AfterAllCallback, or TestInstancePostProcessor, those APIs will not be honored. 

So a workaround is required.

At this point, I'm extremely thankful we took the time to implement thorough tests - the rewrite required is substantial and the tests make sure that I haven't broken any of the (many) weird corner cases.  I think I have my workaround nearly ready.

kriegfrj:
This latest push passes all the tests (including the new one I added to check that the meta-annotation is working). There might be some obscure problems with nested tests that mix PER_CLASS and PER_METHOD, but other than that it should be fine.

scordio:
Thanks a lot to both of you for your effort!

If you have time, would you mind expressing your opinion on #2356?

PR #2406, opened 2021-11-12
Added possibility to override error messages for the recursive comparison

groov1kk:
Added possibility to override a error message for a specific fields/types for the recursive comparison.

This functionality especially useful for custom equals/comparators. For example:
```java
class Foo {

    private final Integer bar;

     private Foo(Integer bar) {
             this.bar = bar;
      }
}

Foo actual = new Foo(1);
Foo expected = new Foo(1);

BiPredicate<Integer, Integer> greaterThan = (a, e) -> Objects.equals(a, e + 1);

Assertions.assertThat(actual)
                .usingRecursiveComparison()
                .withEqualsForFields(greaterThan, "bar")
                .isEqualTo(expected);
```
This code snippet would throw a error with a description:

```
when recursively comparing field by field, but found the following difference:

field/property 'bar' differ:
- actual value  : 1 (Integer@3f3c7bdb)
- expected value: 1 (Integer@3f3c7bdb)
```

Looks silly, doesn't it?

This pull request provides to override the description:
```java
class Foo {

    private final Integer bar;

     private Foo(Integer bar) {
             this.bar = bar;
      }
}

Foo actual = new Foo(1);
Foo expected = new Foo(1);

BiPredicate<Integer, Integer> greaterThan = (a, e) -> Objects.equals(a, e + 1);
String message = "Property bar of expected object must be incremented by 1";

Assertions.assertThat(actual)
                .usingRecursiveComparison()
                .withEqualsForFields(greaterThan, "bar")
                .withErrorMessageForFields(message, "bar")
                .isEqualTo(expected);
```
And the result message for this field will be overridden.

```
when recursively comparing field by field, but found the following difference:

Property bar of expected object must be incremented by 1
```

#### Check List:
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
@@ -2245,7 +2245,7 @@ public abstract class AbstractIterableAssert<SELF extends AbstractIterableAssert
   @Deprecated
   @CheckReturnValue
   public <T> SELF usingComparatorForElementFieldsWithType(Comparator<T> comparator, Class<T> type) {
-    getComparatorsForElementPropertyOrFieldTypes().put(type, comparator);
+    getComparatorsForElementPropertyOrFieldTypes().registerComparator(type, comparator);
     return myself;
   }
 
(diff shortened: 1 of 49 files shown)


joel-costigliola:
Good idea thanks @groov1kk, I'll put that for our next release.

joel-costigliola:
Integrated thanks @groov1kk, very nice work!
When integrating this PR I did change one thing: I removed the overloaded `withErrorMessageForFields` and  `withErrorMessageForType` that took a list of args, the rationale being that the difference with `withErrorMessageForFields(String message, String... fieldLocations)` / `withErrorMessageForType(String message, Class<?> type)` was not super clear and the value not compelling as one can simply build  the same message with `String.format` before passing it to `withErrorMessageForFields` /  `withErrorMessageForType` (it's even better since you don't have to create a list to use `String.format`).
Thanks again for the good work!

PR #2410, opened 2021-11-14
issue-2397 catchThrowableOfType enriched alternatives

### LINKED ISSUE #2397: `catchThrowableOfType` enriched alternatives
#### Summary
Similarly to the enriched alternatives of [`assertThatExceptionOfType`](https://assertj.github.io/doc/#assertj-core-exception-assertions-assertThatExceptionOfType), we could introduce enriched versions of `catchThrowableOfType`.

#### Example

```java
NullPointerException npe = catchNullPointerException( () -> ... );
IllegalArgumentException iae = catchIllegalArgumentException( () -> ... );
IllegalStateException ise = catchIllegalStateException( () -> ... );
IOException ioe = catchIOException( () -> ... );
```

Spacca:
#### Check List:
* Fixes #2397
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -1381,6 +1381,222 @@ public class Assertions implements InstanceOfAssertFactories {
   }
 
   /**
+   * Allows catching an instance of {@link Exception}.
+   * <p>
+   * A call is made to {@code catchThrowable(ThrowingCallable)}, if no exception is thrown it returns null
+   * otherwise it checks that the caught {@link Throwable} is of type {@link Exception} and casts it making it convenient to perform subtype-specific assertions on it.
+   * <p>
+   * Example:
+   * <pre><code class='java'>
+   * Exception exception = catchException(() -&gt; {throw new Exception("boom!");});
+   * // assertions succeed
+   * assertThat(exception).hasMessage("boom!");
+   *
+   * // succeeds as catchException returns null when the code does not throw any exceptions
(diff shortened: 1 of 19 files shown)


scordio:
Thanks for the PR @Spacca. I see you added `catch` versions for additional exception types, which could be totally fine from my perspective. We may want to align the corresponding `assertThat...Exception` methods, probably in a separate PR.

What do you think @joel-costigliola?

src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -1380,6 +1380,222 @@ public static Throwable catchThrowable(ThrowingCallable shouldRaiseThrowable) {
...
+   * // succeeds as catchException returns null when the code does not throw any exceptions
+   * assertThat(catchException(() -&gt; {})).isNull();
+   *
+   * // fails as the thrown instance is not an Exception
+   * catchException(() -&gt; {throw new Throwable("boom!");});</code></pre>
+   *
+   * @param shouldRaiseException The lambda with the code that should raise the exception.
+   * @return The captured exception or <code>null</code> if none was raised by the callable.
+   * @see #catchThrowable(ThrowableAssert.ThrowingCallable)
+   * @since 3.21.0
```
joel-costigliola, line 1404: should be 3.22.0 (same of the other ones too)

joel-costigliola:
We do want to align with `assertThat...Exception`, probably  in a different issue, in the new issue we could also add a test to verify that we have matching `assertThat...Exception` with `catch...Exception`.

scordio:
Thanks for your first contribution, @Spacca!

PR #2413, opened 2021-11-17
Is directory recursively containing string

### LINKED ISSUE #2338: `isDirectoryRecursivelyContaining(String)` should apply filter pattern to filenames only
Similar to #2329, `isDirectoryRecursivelyContaining(String)` should apply the filter pattern to filenames only and not to the entire path.

Together with this change, the existing assertion tests for both variants with `String` and `Predicate` should be refactored to use the same style introduced with #2292 and to inherit from `PathsBaseTest` instead of `PathsSimpleBaseTest`, which should be deleted eventually.

spring1627:
#### Check List:
* Fixes #2338
* Unit tests : YES


scordio:
Thanks for the PR, @spring1627. We still have some points to clarify as mentioned in https://github.com/assertj/assertj-core/issues/2329#issuecomment-919539181 and the direction we want to take with these assertions is not clear yet. Unfortunately, it was not mentioned in #2338, sorry about that.

PR #2422, opened 2021-11-24
modified all subclasses of FilesBaseTest to use real filesystem file

### LINKED ISSUE #2337: Avoid mocks for testing `File` assertions
Similar to #2292, all the test cases in subclasses of `FilesBaseTest` should use real filesystem elements and not mocks.

We might still favor the NIO2 File API to initialize the test prerequisites, so `actual` would be obtained via `Path::toFile`.

ahmadsadeed:
#### Check List:
* Fixes #2337 
* Unit tests : NA

This is my first PR on Assertj. I will appreciate if I can get any suggestions on how to improve my work for this and future issues.

### DIFF
--- a/src/test/java/org/assertj/core/internal/files/File_assertHasSize_Test.java
+++ b/src/test/java/org/assertj/core/internal/files/File_assertHasSize_Test.java
@@ -26,7 +26,6 @@ import org.assertj.core.api.AssertionInfo;
 import org.assertj.core.internal.Files;
 import org.assertj.core.internal.FilesBaseTest;
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 
 /**
@@ -34,7 +33,6 @@ import org.junit.jupiter.api.Test;
  *
  * @author Krishna Chaithanya Ganta
  */
-@DisplayName("Files.assertHasSizeInBytes:")
 class File_assertHasSize_Test extends FilesBaseTest {
 
   private static File actual;
@@ -57,9 +55,8 @@ class File_assertHasSize_Test extends FilesBaseTest {
     // GIVEN
     AssertionInfo info = someInfo();
     // WHEN
-    AssertionError error = expectAssertionError(() -> files.assertHasSizeInBytes(info, actual, 36L));
+    expectAssertionError(() -> files.assertHasSizeInBytes(info, actual, 36L));
     // THEN
-    assertThat(error).isInstanceOf(AssertionError.class);
     verify(failures).failure(info, shouldHaveSize(actual, 36L));
   }
 
@@ -69,9 +66,8 @@ class File_assertHasSize_Test extends FilesBaseTest {
     AssertionInfo info = someInfo();
     File notAFile = new File("xyz");
     // WHEN
-    AssertionError error = expectAssertionError(() -> files.assertHasSizeInBytes(info, notAFile, 36L));
+    expectAssertionError(() -> files.assertHasSizeInBytes(info, notAFile, 36L));
     // THEN
-    assertThat(error).isInstanceOf(AssertionError.class);
     verify(failures).failure(info, shouldBeFile(notAFile));
   }
 
--- a/src/test/java/org/assertj/core/internal/files/File_assertHasSize_Test.java
+++ b/src/test/java/org/assertj/core/internal/files/File_assertHasSize_Test.java
@@ -45,34 +45,31 @@ class File_assertHasSize_Test extends FilesBaseTest {
   @Test
   void should_throw_error_if_actual_is_null() {
     // WHEN
-    AssertionError assertionError = expectAssertionError(() -> files.assertHasSizeInBytes(someInfo(), null, actual.length()));
+    AssertionError assertionError = expectAssertionError(() -> files.assertHasSizeInBytes(INFO, null, actual.length()));
     // THEN
     assertThat(assertionError).hasMessage(actualIsNull());
   }
 
   @Test
   void should_throw_error_if_actual_file_does_not_have_the_expected_size() {
-    // GIVEN
-    AssertionInfo info = someInfo();
-    // WHEN
-    expectAssertionError(() -> files.assertHasSizeInBytes(info, actual, 36L));
+    // GIVEN / WHEN
+    expectAssertionError(() -> files.assertHasSizeInBytes(INFO, actual, 36L));
     // THEN
-    verify(failures).failure(info, shouldHaveSize(actual, 36L));
+    verify(failures).failure(INFO, shouldHaveSize(actual, 36L));
   }
 
   @Test
   void should_fail_if_actual_is_not_a_file() {
     // GIVEN
-    AssertionInfo info = someInfo();
     File notAFile = new File("xyz");
     // WHEN
-    expectAssertionError(() -> files.assertHasSizeInBytes(info, notAFile, 36L));
+    expectAssertionError(() -> files.assertHasSizeInBytes(INFO, notAFile, 36L));
     // THEN
-    verify(failures).failure(info, shouldBeFile(notAFile));
+    verify(failures).failure(INFO, shouldBeFile(notAFile));
   }
 
   @Test
   void should_pass_if_actual_has_expected_size() {
-    files.assertHasSizeInBytes(someInfo(), actual, actual.length());
+    files.assertHasSizeInBytes(INFO, actual, actual.length());
   }
 }
--- a/src/test/java/org/assertj/core/internal/files/File_assertHasSize_Test.java
+++ b/src/test/java/org/assertj/core/internal/files/File_assertHasSize_Test.java
@@ -34,7 +34,7 @@ import org.junit.jupiter.api.Test;
  *
  * @author Krishna Chaithanya Ganta
  */
-@DisplayName("Files.assertHasSize:")
+@DisplayName("Files.assertHasSizeInBytes:")
 class File_assertHasSize_Test extends FilesBaseTest {
 
   private static File actual;
@@ -57,8 +57,9 @@ class File_assertHasSize_Test extends FilesBaseTest {
     // GIVEN
     AssertionInfo info = someInfo();
     // WHEN
-    expectAssertionError(() -> files.assertHasSizeInBytes(info, actual, 36L));
+    AssertionError error = expectAssertionError(() -> files.assertHasSizeInBytes(info, actual, 36L));
     // THEN
+    assertThat(error).isInstanceOf(AssertionError.class);
     verify(failures).failure(info, shouldHaveSize(actual, 36L));
   }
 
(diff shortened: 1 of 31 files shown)


src/test/java/org/assertj/core/internal/files/File_assertHasSize_Test.java
```diff
@@ -68,8 +69,9 @@ void should_fail_if_actual_is_not_a_file() {
     AssertionInfo info = someInfo();
     File notAFile = new File("xyz");
     // WHEN
-    expectAssertionError(() -> files.assertHasSizeInBytes(info, notAFile, 36L));
+    AssertionError error = expectAssertionError(() -> files.assertHasSizeInBytes(info, notAFile, 36L));
```
joel-costigliola, line 72: `expectAssertionError` already makes sure the error is an `AssertionError`, please revert all the changes in this file.
The only thing to do is remove @DisplayName("Files.assertHasSizeInBytes:") as it is computed automatically

joel-costigliola: this comment applies for the other tests too ;-)

src/test/java/org/assertj/core/internal/files/Files_assertHasBinaryContent_Test.java
```diff
@@ -94,13 +108,20 @@ void should_throw_error_wrapping_caught_IOException() throws IOException {
...
+    byte[] expected = "fake".getBytes();
+
+    BinaryDiff binaryDiff = new BinaryDiff();
+    BinaryDiffResult diff = binaryDiff.diff(actual, expected);
     AssertionInfo info = someInfo();
 
-    Throwable error = catchThrowable(() -> files.assertHasBinaryContent(info, actual, expected));
+    AssertionError error = expectAssertionError(() -> unMockedFiles.assertHasBinaryContent(info, actual, expected));
 
     assertThat(error).isInstanceOf(AssertionError.class);
```
joel-costigliola, line 124: can be removed as you are using `expectAssertionError`

src/test/java/org/assertj/core/internal/files/Files_assertHasBinaryContent_Test.java
```diff
@@ -94,13 +108,20 @@ void should_throw_error_wrapping_caught_IOException() throws IOException {
...
-    verify(failures).failure(info, shouldHaveBinaryContent(actual, diff));
+    then(error).hasMessage(shouldHaveBinaryContent(actual, diff).create(info.description(), info.representation()));
```
joel-costigliola, line 125: Revert this change as the previous code basically does the same thing but in a more concise way

joel-costigliola: this comment applies to other tests too

src/test/java/org/assertj/core/internal/files/Files_assertHasBinaryContent_Test.java
```diff
@@ -94,13 +108,20 @@ void should_throw_error_wrapping_caught_IOException() throws IOException {
 
   @Test
   void should_fail_if_file_does_not_have_expected_binary_content() throws IOException {
```
joel-costigliola, line 110: add GIVEN WHEN THEN to the tests you are updating, thanks!

joel-costigliola: this is a general comment that applies to other tests

joel-costigliola (changes requested):
Thanks , first round of code review done. Thanks @ahmadsadeed!

ahmadsadeed:
Thanks @joel-costigliola, I made the requested changes. I added `protected Failures unMockedFailures;` to FilesBaseTest.java since I was using `unMockedFiles` instead of `files` in some tests. Then used `verify(unMockedFailures)` instead of `verify(failures)`. I hope I did this part right.

ahmadsadeed:
By the way, I noticed that the CI workflow was failing due to an issue in Files_assertCanRead_Test. I fixed this issue so it shouldn't be a problem.

src/test/java/org/assertj/core/internal/files/Files_assertCanRead_Test.java
```diff
@@ -46,18 +42,19 @@ void should_fail_if_actual_is_null() {
 
   @Test
   void should_fail_if_can_not_read() {
-    when(actual.canRead()).thenReturn(false);
+    // GIVEN
+    File notAFile = new File("xyz");
```
joel-costigliola, line 46: minor: rename to `nonExistentFile` which is more accurate as `notAFile` is a `File` but simply not an existing one

src/test/java/org/assertj/core/internal/files/Files_assertCanWrite_Test.java
```diff
@@ -46,18 +43,19 @@ void should_fail_if_actual_is_null() {
...
-
+    // THEN
     assertThat(error).isInstanceOf(AssertionError.class);
     verify(failures).failure(info, shouldBeWritable(actual));
   }
 
   @Test
   void should_pass_if_actual_can_write() {
-    when(actual.canWrite()).thenReturn(true);
+    File actual = newFile(tempDir.getAbsolutePath() + "to_write.txt");
```
joel-costigliola, line 58: Even if it's experimental let's give a try to JUnit5 [@TempDir](https://junit.org/junit5/docs/current/user-guide/#writing-tests-built-in-extensions-TempDirectory) to avoid leaving temp files.

Have a look at `PathsBaseTest` where it is already used.

joel-costigliola: Comment applies to other tests using temp files

src/test/java/org/assertj/core/internal/files/Files_assertHasBinaryContent_Test.java
```diff
@@ -66,19 +67,25 @@ void should_fail_if_actual_is_null() {
...
   void should_pass_if_file_has_expected_binary_content() throws IOException {
-    when(binaryDiff.diff(actual, expected)).thenReturn(BinaryDiffResult.noDiff());
-    files.assertHasBinaryContent(someInfo(), actual, expected);
+    File actual = newFile(tempDir.getAbsolutePath() + "/tmp.txt");
+    byte[] data = "actual".getBytes();
+    try (FileOutputStream myWriter = new FileOutputStream(actual)) {
+      myWriter.write(data, 0, data.length);
+    }
+    byte[] expected = "actual".getBytes();
+    unMockedFiles.assertHasBinaryContent(someInfo(), actual, expected);
```
joel-costigliola, line 88: Take an approach similar to `Paths_assertHasBinaryContent_Test`, it's using a more concise way of writing files:
```
// call .toFile() other it returns a Path
Files.write(tempDir.resolve("actual"), "Content".getBytes()).toFile();
```

joel-costigliola: This comment applies to other tests too (I mean looking at the equivalent test for Path)

joel-costigliola (commented):
thanks for the PR @ahmadsadeed, this goes in the right direction.
I haven't commented everything to avoid repeating myself too much, the comments I made apply to many tests, if can you address them  to all the tests where they make sense that would be fantastic.

src/test/java/org/assertj/core/internal/files/Files_assertHasBinaryContent_Test.java
ahmadsadeed: Since the two unresolved comments are on the same topic, I will ask my question here:

- We already have `import org.assertj.core.internal.Files;` so I have to use `java.nio.file.Files.write`. Any suggestions or comments?
- `tmpDir` is declared as type `protected File tempDir;` in FilesBaseTest. I could use `tempDir.toPath()` or change it to `protected Path tempDir;`. Or keep both types:
```
  @TempDir
  protected File tempDir;
  @TempDir
  protected Path tempDirPath;
```
Final:
```
  @Test
  void should_throw_error_if_actual_file_does_not_have_the_expected_size() throws IOException {
    // GIVEN
    File actual = java.nio.file.Files.write(tempDirPath.resolve("actual"), "content".getBytes()).toFile();
    // WHEN
    expectAssertionError(() -> files.assertHasSizeInBytes(INFO, actual, 36L));
    // THEN
    verify(failures).failure(INFO, shouldHaveSize(actual, 36L));
  }
```

ahmadsadeed: Also, should I replace usage of files from resources folder with the suggestion above?
```@BeforeAll
  static void setUpOnce() {
    actual = new File("src/test/resources/actual_file.txt");
  }
```

```@Test
  void should_pass_if_actual_has_expected_size() {
    files.assertHasSizeInBytes(INFO, actual, actual.length());
  }
```

joel-costigliola: For files test let's use:
```
  @TempDir
  protected File tempDir;
```
unless I'm missing something we don't need a `Path tempDir` .

joel-costigliola: we could add a helper method in `FilesBaseTest`  to wrap:
```java
java.nio.file.Files.write(tempDirPath.resolve("actual"), "content".getBytes()).toFile();
```
to something like:
```
writeFile("actual", "content");
```
That's a suggestion, maybe try it and see for yourself it it makes it easier otherwise don't worry too much ;-)

src/test/java/org/assertj/core/internal/files/Files_assertHasBinaryContent_Test.java
```diff
@@ -54,53 +55,64 @@ static void setUpOnce() {
...
 
   @Test
   void should_pass_if_file_has_expected_binary_content() throws IOException {
-    when(binaryDiff.diff(actual, expected)).thenReturn(BinaryDiffResult.noDiff());
-    files.assertHasBinaryContent(someInfo(), actual, expected);
+    File actual = newFile(tempDir.getAbsolutePath() + "/tmp.txt");
+    byte[] data = "actual".getBytes();
+    try (FileOutputStream myWriter = new FileOutputStream(actual)) {
+      myWriter.write(data, 0, data.length);
+    }
```
joel-costigliola, line 84: this comment applies broadly: use [commons io FileUtils](https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/FileUtils.html) to write file content, there are many ways, ex `writeByteArrayToFileFile file, byte[] data)` or `write(File file, CharSequence data, String charsetName)`. These methods throw `IOException`, don't catch them just add them in the test method throw declaration.

joel-costigliola (changes requested):
@ahmadsadeed thanks this looks good!
I added a comment to use commons io to simplify writing files content.

The other one is optional (I'll do it if you don't have time) is to refactor the `should_fail_if_actual_is_null()` tests to be like:

```java
  @Test
  void should_fail_if_actual_is_null() {
    // GIVEN
    File actual = null;
    // WHEN
    AssertionError error = expectAssertionError(() -> files.assertSameBinaryContentAs(INFO, actual, expected));
    // THEN
    then(error).hasMessage(actualIsNull());
  } 
```
After that we are good to go!

joel-costigliola:
Integrated thanks @ahmadsadeed!

PR #2423, opened 2021-11-26
Add AbstractThrowableAssert#causesChain to simplify performing assertions on a chain of throwables

### LINKED ISSUE #1147: Provide semantic ways to test n-level (n>1) causes of an Exception, not only direct cause and root cause
Assertj currently provides methods to test a thrown exception, its direct cause, or its root cause. It does not offer a semantic way to test the exceptions/causes in between the first cause and the root cause.

Exceptions tend to get wrapped and wrapped by frameworks and libraries, and by project code. Sometimes the most meaningful exception is neither the one thrown nor its cause nor the root cause, but one somewhere in the middle.

For example, BusinessLogicEx caused by SqlRollbackEx caused by UnableToExecuteSqlEx in a test where I just want to assert that my sql will be rolled back by finding the SqlRollbackEx in the stack.

```java
assertThatThrownBy(() -> {}).hasAnyLevelCauseInstanceOf(type);
assertThatThrownBy(() -> {}).hasAnyLevelCauseWithMessageMatching(condition);
// etc
```

slinkydeveloper:
#### Check List:
* Fixes #1147 (ignore if not applicable)
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
@@ -204,6 +204,31 @@ public abstract class AbstractThrowableAssert<SELF extends AbstractThrowableAsse
   }
 
   /**
+   * Returns a new list assertion of every cause in the chain of causes.
+   * The list is ordered from the current {@link Throwable} up to the root cause.
+   * <p>
+   * Examples:
+   * <pre><code class='java'> Throwable rootCause = new JdbcException("invalid query");
+   * Throwable cause =  new RuntimeException("some failure", rootCause);
+   * Throwable throwable = new Exception("boom!", cause);
+   *
+   * // typical use:
+   * assertThat(throwable).asCausesChain()
+   *                      .extracting(Throwable::getMessage)
+   *                      .anyMatch("some failure");</code></pre>
+   *
+   * @return a new assertion object
+   * @throws AssertionError if the actual {@code Throwable} is {@code null}.
+   * @throws AssertionError if the actual {@code Throwable} does not have a cause.
+   *
+   * @since 3.16.0
+   */
+  public ListAssert<Throwable> asCausesChain() {
+    throwables.assertHasCause(info, actual);
+    return AssertionsForInterfaceTypes.assertThat(org.assertj.core.util.Throwables.getCausesChain(actual));
+  }
+
+  /**
    * Verifies that the message of the actual {@code Throwable} starts with the given description.
    * <p>
    * Examples:
--- a/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
@@ -15,7 +15,9 @@ package org.assertj.core.api;
 import static java.lang.String.format;
 import static org.assertj.core.error.ShouldNotHaveThrown.shouldNotHaveThrown;
 
+import java.util.ArrayList;
 import java.util.IllegalFormatException;
+import java.util.List;
 import java.util.regex.Pattern;
 
 import org.assertj.core.error.BasicErrorMessageFactory;
@@ -204,8 +206,8 @@ public abstract class AbstractThrowableAssert<SELF extends AbstractThrowableAsse
   }
 
   /**
-   * Returns a new list assertion of every cause in the chain of causes.
-   * The list is ordered from the current {@link Throwable} up to the root cause.
+   * Returns a new list assertion of every cause from the current {@link Throwable} down to the root cause.
+   * The list is empty if there is no cause.
    * <p>
    * Examples:
    * <pre><code class='java'> Throwable rootCause = new JdbcException("invalid query");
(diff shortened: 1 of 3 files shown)


slinkydeveloper:
Hi @joel-costigliola I stumbled on the issue #1147 and I created this PR to propose a solution, lmk what do you think. I'm a first time contributor here, so I hope I haven't messed it up badly :smile:

src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
```diff
@@ -203,6 +203,31 @@ public SELF hasNoCause() {
     return new ThrowableAssert<>(org.assertj.core.util.Throwables.getRootCause(actual));
   }
 
+  /**
+   * Returns a new list assertion of every cause in the chain of causes.
+   * The list is ordered from the current {@link Throwable} up to the root cause.
```
matriv, line 208: Myself, I kinda prefer the term `down to the root cause`.

src/main/java/org/assertj/core/util/Throwables.java
```diff
@@ -148,6 +150,21 @@ public static Throwable getRootCause(Throwable throwable) {
     return throwable;
   }
 
+  /**
+   * Get the causes chain of a {@link Throwable}.
+   *
+   * @param throwable the {@code Throwable} to get causes chain from.
+   * @return the causes chain, including the {@link Throwable} provided as parameter.
+   */
+  public static List<Throwable> getCausesChain(Throwable throwable) {
```
matriv, line 159: I'm kind of unsettled if the `throwable` in test should be included or not in the list.
On one hand it's nice to have the complete list and assert any of the throwables in the chain, on the other hand, one might say that `getCausesChain` only contains causes down from the "current" level, excluding the current level. Moreover, it's easier to assert the `throwable` in question separately rather than excluding it from the list manually if the user doesn't want to assert it.

I have a slight preference to keep it as is (including the `throwable` in test), but I'd say to specifically mention the inclusive behaviour in the javadocs, here and in `asCausesChain()` method.

matriv (commented):
Thank you @slinkydeveloper! That's a very useful functionality for asserting exceptions!

src/main/java/org/assertj/core/util/Throwables.java
scordio, line 159: I don't see the need of exposing this as a public API, can be a private method of `AbstractThrowableAssert`.

src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
```diff
@@ -203,6 +203,31 @@ public SELF hasNoCause() {
...
+   * Throwable throwable = new Exception("boom!", cause);
+   *
+   * // typical use:
+   * assertThat(throwable).asCausesChain()
+   *                      .extracting(Throwable::getMessage)
+   *                      .anyMatch("some failure");</code></pre>
+   *
+   * @return a new assertion object
+   * @throws AssertionError if the actual {@code Throwable} is {@code null}.
+   * @throws AssertionError if the actual {@code Throwable} does not have a cause.
```
scordio, line 222: I'm not sure if we should relax this and return an empty list instead.

src/main/java/org/assertj/core/util/Throwables.java
scordio: I tend to favor not having `actual` in the resulting list, starting with its direct cause instead.

scordio (commented):
Thanks @slinkydeveloper, some initial thoughts from my side. We would need some additional work to make sure everything works fine with soft assertions, but let's have first feedback from @joel-costigliola about naming and general direction.

src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
joel-costigliola: It's not an obvious call  but I think I'm in favor of returning an empty list.

pros: 
- you can write `assertThat(throwable).causesChain().isEmpty();` which is reasonable
- in API I tend to prefer returning empty collection over null/failure

cons: 
- if we return an empty list then some assertions like `allMatch` will pass (it succeeds for any empty list)
- users probably expect to have causes some failing if there are none might be better

The last point is not a real issue IMO as the assertions applied on the list of causes will most probably fail if we return an empty list (except for a few like `allMatch`).

TLDR: let's go with returning an empty list ;-)

src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
joel-costigliola: `Returns a new list assertion of every cause down to the root cause. The list is empty if there is no cause.`  sounds good to me

src/main/java/org/assertj/core/util/Throwables.java
joel-costigliola: When I first read `causeChain` I thought it would only contain causes and not the initial exception so I would leave it out.
We can always add a `causeChainWithInitialThrowable` later on if users ask for it.

joel-costigliola: Agreed, we'll move it here if we reuse that piece of code elsewhere.

src/test/java/org/assertj/core/api/throwable/ThrowableAssert_asCausesChain_Test.java
```diff
@@ -0,0 +1,56 @@
...
+
+  private final Throwable rootCause = new NullPointerException("root");
+  private final Throwable intermediateCause = new IllegalStateException("intermediate", rootCause);
+
+  @Test
+  void should_return_throwable_chain_assertion() {
+    // GIVEN
+    Throwable throwable = new Throwable("initial", intermediateCause);
+    // WHEN/THEN
+    assertThat(throwable).asCausesChain()
```
joel-costigliola, line 32: minor: we prefer `then` here (from `BDDAssertions.then` static import)

joel-costigliola:
@slinkydeveloper we have reviewed  your PR, thanks for the good work!

src/main/java/org/assertj/core/util/Throwables.java
slinkydeveloper: May I propose to introduce both in the same PR? Perhaps one could be `causesChain` and the other one `throwableChain` or `throwableTree`, which includes the initial one.

src/test/java/org/assertj/core/api/throwable/ThrowableAssert_asCausesChain_Test.java
slinkydeveloper: WDYM? I honestly don't get how you want me to refactor this particular line of code :smile: I modified the code in the tests below to use `BDDAssertions.then`

src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
slinkydeveloper: Modified with the empty list

src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
```diff
@@ -203,6 +203,31 @@ public SELF hasNoCause() {
...
+   *
+   * @since 3.16.0
+   */
+  public ListAssert<Throwable> asCausesChain() {
```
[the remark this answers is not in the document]
slinkydeveloper: How would you suggest to change the return type? I changed it to `AbstractListAssert<?, List<Throwable>, Throwable, ? extends AbstractThrowableAssert<?, Throwable>>`, although i'm not completely sure it's the correct one

joel-costigliola:
This fell under our radar, let's schedule it for the 4.0 release.

joel-costigliola:
Integrated thanks @slinkydeveloper, I changed a bit the naming to `throwablesChain` to make it clear it's not only the causes but also the root exception.

PR #2436, opened 2021-12-12
Added hasNoHost to AbstractUriAssert and AbstractUrlAssert

ascopes:
Both java.net.URI and java.net.URL have the ability to have null
hosts set (for example in URIs that point to file-system objects,
or custom URI schemes that may be non-standard).

To match the rest of this API where methods exist to provide checks
for no fragments, queries, parameters, etc being set, this provides
host visibility checking.

### DIFF
--- a/src/main/java/org/assertj/core/internal/Uris.java
+++ b/src/main/java/org/assertj/core/internal/Uris.java
@@ -75,14 +75,13 @@ public class Uris {
 
   public void assertHasHost(AssertionInfo info, URI actual, String expected) {
     assertNotNull(info, actual);
-    requireNonNull(expected, "The expected host should not be null.");
+    requireNonNull(expected, "The expected host should not be null");
     if (!Objects.equals(actual.getHost(), expected)) throw failures.failure(info, shouldHaveHost(actual, expected));
   }
 
   public void assertHasNoHost(AssertionInfo info, URI actual) {
     assertNotNull(info, actual);
-    if (actual.getHost() != null && !actual.getHost().isEmpty())
-      throw failures.failure(info, shouldHaveNoHost(actual));
+    if (actual.getHost() != null) throw failures.failure(info, shouldHaveNoHost(actual));
   }
 
   public void assertHasAuthority(AssertionInfo info, URI actual, String expected) {
--- a/src/main/java/org/assertj/core/internal/Uris.java
+++ b/src/main/java/org/assertj/core/internal/Uris.java
@@ -12,9 +12,11 @@
  */
 package org.assertj.core.internal;
 
+import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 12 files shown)


scordio:
Thanks for the PR, @ascopes!

We discussed the change in the team and we would prefer to have a dedicated error message instead of reusing the one from `assertHasHost`. Also, we would enforce that `assertHasHost` expects a non-null value.

The pattern to adopt would be similar to `assertHasExtension` / `assertHasNoExtension` for `Path` assertions.

Would it be okay for you to update your PR accordingly?

ascopes:
Sure, will take a look

ascopes:
Should be updated now.

I made the assumption that when you say "assertHasHost" expects a non-null value, that it is fine to leave it covered by the nonNull check within Uris.java and Urls.java as opposed to having a check on that method as well; if that isn't right or there is any other feedback, let me know!

Thanks

src/main/java/org/assertj/core/internal/Uris.java
```diff
@@ -76,6 +77,12 @@ public void assertHasHost(AssertionInfo info, URI actual, String expected) {
     if (!Objects.equals(actual.getHost(), expected)) throw failures.failure(info, shouldHaveHost(actual, expected));
   }
 
+  public void assertHasNoHost(AssertionInfo info, URI actual) {
+    assertNotNull(info, actual);
+    if (actual.getHost() != null && !actual.getHost().isEmpty())
```
ascopes, line 82: The empty check has been added as it appears that the URL can return either-or. Seems that you can specify the host as null manually during initialisation but the constructor that parses those expressions will always use an emptystring.

scordio:
> Also, we would enforce that `assertHasHost` expects a non-null value.

I mean that the `expected` parameter of `assertHasHost` should always be non-null and we can enforce it adding a `requireNonNull` check similar to:

https://github.com/assertj/assertj-core/blob/56c77698ea5a6b829a257ad9c341dd06904814f0/src/main/java/org/assertj/core/internal/Paths.java#L466-L467

src/main/java/org/assertj/core/internal/Uris.java
ascopes: Amended tests to explicitly check both of these cases in addition to the parsing case.

ascopes:
Should be good now, hopefully!

src/main/java/org/assertj/core/internal/Uris.java
scordio: I did some tries and I'm actually getting the opposite result: when `host` is either `null` or `""` in the constructor, `getHost()` always returns `null`.

scordio:
Thanks for your first contribution, @ascopes!

PR #2452, opened 2021-12-30
[mvn] Update maven wrapper

hazendaz:
Build change only to use official maven wrapper as it migrated to maven ownership.


scordio:
For reproducibility, the update is achieved by running:

```
mvn -N org.apache.maven.plugins:maven-wrapper-plugin:wrapper -Dtype=source -Dmaven=3.8.4
```

in the root of the project.

PR #2458, opened 2022-01-05
Add `isUnmodifiable` to `Map` assertions

### LINKED ISSUE #2381: `isUnmodifiable` assertion for Maps
#### Summary
Similarly to #2102, we could add a `isUnmodifiable()` method to the assertion methods for `Map` types.

#### Implementation Proposal

Under the hood, the assertion should try to call all the map APIs which are expected to throw `UnsupportedOperationException`. The assertion succeeds if all of them are throwing the expected exception, otherwise it fails.

Below is the mapping between the map type and the APIs to test. Each subtype would cover also the methods of the parent type. E.g., if we have a `NavigableMap`, the assertion should also verify `Map` methods. 

| Instance Of    | APIs to test                                                                                                                                                                                                                                                                            |
|----------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `Map`          | `clear()`, `compute(K, BiFunction)`, `computeIfAbsent(K, Function)`, `computeIfPresent(K, BiFunction)`, `entrySet().iterator().remove()`, `keySet().iterator().remove()`, `merge(K, V, BiFunction)`, `put(K, V)`, `putAll(Map)`, `putIfAbsent(K, V)`, `replace(K, V)`, `replace(K, V, V)`, `remove(Object)`, `remove(Object, Object)`, `replaceAll(BiFunction)` |
| `NavigableMap` |  `pollFirstEntry()`, `pollLastEntry() `                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
| `SortedMap`    | -                                                                                                                                                                                                                                                                                        |

#### Example

```java
Map<String, String> map = new HashMap<>();
map.put("key", "value");

assertThat(map).isUnmodifiable(); // should fail

Map<String, String> unmodifiableMap = Collections.unmodifiableMap(map);

assertThat(unmodifiableMap).isUnmodifiable(); // should succeed
```

WojciechZankowski:
#### Check List:
* Fixes #2381 
* Unit tests : YES

1. Check if you like moving out common code to a new class ```Immutables```. Maybe there is a better place for that code.
2. ~~Seems like not all libraries are creating fully unmodifiable maps?~~


assertj-core/src/test/java/org/assertj/core/api/map/MapAssert_isUnmodifiable_Test.java
```diff
@@ -0,0 +1,122 @@
...
+  // https://issues.apache.org/jira/browse/COLLECTIONS-xxx FIXME
+  @Test
+  void should_fail_with_commons_collections_UnmodifiableMap() {
+    // GIVEN
+    Map<?, ?> actual = UnmodifiableMap.unmodifiableMap(new HashMap<>());
+    // WHEN
+    AssertionError assertionError = expectAssertionError(() -> assertThat(actual).isUnmodifiable());
+    // THEN
+    then(assertionError).hasMessage(shouldBeUnmodifiable("Map.compute(null, (k, v) -> v)").create());
+  }
```
scordio, line 85: I think this is not a Commons Collections issue but rather caused by the way we test the `compute` method.

The following fails properly:

```java
Map<Object, Object> map = new HashMap<>();
map.put(null, null);
Map<Object, Object> unmodifiableMap = UnmodifiableMap.unmodifiableMap(map);
unmodifiableMap.compute(null, (k, v) -> v); // fails with UOE as it tries to remove (null -> null)
```

but the following doesn't because there is no modification caused to the map:

```java
Map<Object, Object> emptyMap = new HashMap<>();
Map<Object, Object> unmodifiableEmptyMap = UnmodifiableMap.unmodifiableMap(emptyMap);
unmodifiableEmptyMap.compute(null, (k, v) -> v); // does not fail as there is no modification
```

The behavior is also in line with the [`compute`](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#compute-K-java.util.function.BiFunction-) Javadoc:

> If the remapping function returns null, the mapping is removed (or remains absent if initially absent).

The same is also valid for `unmodifiableSortedMap`.

I'll check how we can improve the assertion for such cases.

scordio: In retrospect, I don't think we should invest more just to cover two implementations without fail-fast checks. Let's go ahead with the merge and we can always react if concrete use cases are reported.

scordio:
This is finally merged. Thanks for your contribution, @WojciechZankowski!

PR #2460, opened 2022-01-09
Fixes #2459: Avoid ValueNode being considered as iterable while using recursive comparison

### LINKED ISSUE #2459: Different behavior between recursive comparison and JsonUnit with Jackson `JsonNode`
Given:

```java
ObjectMapper om = new ObjectMapper();

JsonNode objectNode0 = om.readTree("{\"someNotImportantValue\":1,\"importantValue\":\"10\"}");
JsonNode objectNode1 = om.readTree("{\"someNotImportantValue\":15,\"importantValue\":\"1\"}");
```

with JsonUnit:
```java
JsonAssertions.assertThatJson(objectNode0).isEqualTo(objectNode1);
```

we get a failure with:
```
net.javacrumbs.jsonunit.core.internal.Opentest4jExceptionFactory$JsonAssertError: JSON documents are different:
Different value found in node "importantValue", expected: <"1"> but was: <"10">.
Different value found in node "someNotImportantValue", expected: <15> but was: <1>.
```

but with AssertJ and recursive comparison:

```
assertThat(objectNode0).usingRecursiveComparison().isEqualTo(objectNode1);
```

the assertion passes.

Being `JsonNode` an `Iterable`, it seems that something is not working properly with AssertJ.

Source: https://stackoverflow.com/questions/70602371/assertj-not-working-with-jackson-jsonnode

temp-droid:
#### Check List:
* Fixes #2459
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/recursive/comparison/DualValue.java
+++ b/src/main/java/org/assertj/core/api/recursive/comparison/DualValue.java
@@ -170,8 +170,8 @@ public final class DualValue {
   }
 
   public boolean isActualFieldAnIterable() {
-    // ignore Path to be consistent with isExpectedFieldAnIterable
-    return actual instanceof Iterable && !(actual instanceof Path);
+    // ignore Path and ValueNode to be consistent with isExpectedFieldAnIterable
+    return actual instanceof Iterable && !(actual instanceof Path || isActualFieldAJsonValueNode());
   }
 
   public boolean isExpectedFieldAnIterable() {
@@ -179,7 +179,26 @@ public final class DualValue {
     // Iterable are compared element by element recursively
     // Ex: /tmp/foo.txt path has /tmp as its first element
     // so /tmp is going to be compared recursively but /tmp first element is itself leading to an infinite recursion
-    return expected instanceof Iterable && !(expected instanceof Path);
+    // Don't consider ValueNode as an Iterable as they only contain one value and iterating them does not make sense.
+    return expected instanceof Iterable && !(expected instanceof Path || isExpectedFieldAJsonValueNode());
+  }
+
+  private boolean isActualFieldAJsonValueNode() {
+    return isAJsonValueNode(actual);
+  }
+
+  private boolean isExpectedFieldAJsonValueNode() {
+    return isAJsonValueNode(expected);
+  }
+
+  private static boolean isAJsonValueNode(Object value) {
+    try {
+      Class<?> valueNodeClass = Class.forName("com.fasterxml.jackson.databind.node.ValueNode");
+      return valueNodeClass.isInstance(value);
+    } catch (ClassNotFoundException e) {
+      // value cannot be a ValueNode because the class couldn't be located
+      return false;
+    }
   }
 
   private static boolean isAnOrderedCollection(Object value) {
--- a/src/main/java/org/assertj/core/api/recursive/comparison/DualValue.java
+++ b/src/main/java/org/assertj/core/api/recursive/comparison/DualValue.java
@@ -180,7 +180,7 @@ public final class DualValue {
     // Ex: /tmp/foo.txt path has /tmp as its first element
     // so /tmp is going to be compared recursively but /tmp first element is itself leading to an infinite recursion
     // Don't consider ValueNode as an Iterable as they only contain one value and iterating them does not make sense.
-    return expected instanceof Iterable && !(expected instanceof Path || isExpectedFieldAJsonValueNode());
+    return expected instanceof Iterable && !(expected instanceof Path || isAJsonValueNode(expected));
   }
 
   private boolean isActualFieldAJsonValueNode() {
--- a/src/main/java/org/assertj/core/api/recursive/comparison/DualValue.java
+++ b/src/main/java/org/assertj/core/api/recursive/comparison/DualValue.java
@@ -171,7 +171,7 @@ public final class DualValue {
 
   public boolean isActualFieldAnIterable() {
     // ignore Path and ValueNode to be consistent with isExpectedFieldAnIterable
-    return actual instanceof Iterable && !(actual instanceof Path || isActualFieldAJsonValueNode());
+    return actual instanceof Iterable && !(actual instanceof Path || isAJsonValueNode(actual)());
   }
 
   public boolean isExpectedFieldAnIterable() {
--- a/src/test/java/org/assertj/core/api/recursive/comparison/DualValue_iterableValues_Test.java
+++ b/src/test/java/org/assertj/core/api/recursive/comparison/DualValue_iterableValues_Test.java
@@ -26,6 +26,16 @@ import java.util.stream.Stream;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
 
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.BinaryNode;
+import com.fasterxml.jackson.databind.node.BooleanNode;
+import com.fasterxml.jackson.databind.node.IntNode;
+import com.fasterxml.jackson.databind.node.MissingNode;
+import com.fasterxml.jackson.databind.node.NullNode;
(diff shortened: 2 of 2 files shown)


temp-droid:
@scordio , @joel-costigliola , your feedback is welcome!

I feel like the test provided in the issue could also be included, but not sure where (if needed) to add it.

src/main/java/org/assertj/core/api/recursive/comparison/DualValue.java
```diff
@@ -170,16 +170,35 @@ public boolean isExpectedFieldAnOrderedCollection() {
   }
 
   public boolean isActualFieldAnIterable() {
-    // ignore Path to be consistent with isExpectedFieldAnIterable
-    return actual instanceof Iterable && !(actual instanceof Path);
+    // ignore Path and ValueNode to be consistent with isExpectedFieldAnIterable
+    return actual instanceof Iterable && !(actual instanceof Path || isActualFieldAJsonValueNode());
```
scordio, line 174:
```suggestion
    return actual instanceof Iterable && !(actual instanceof Path || isAJsonValueNode(actual)());
```

src/main/java/org/assertj/core/api/recursive/comparison/DualValue.java
```diff
@@ -170,16 +170,35 @@ public boolean isExpectedFieldAnOrderedCollection() {
...
   }
 
   public boolean isExpectedFieldAnIterable() {
     // Don't consider Path as an Iterable as recursively comparing them leads to a stack overflow, here's why:
     // Iterable are compared element by element recursively
     // Ex: /tmp/foo.txt path has /tmp as its first element
     // so /tmp is going to be compared recursively but /tmp first element is itself leading to an infinite recursion
-    return expected instanceof Iterable && !(expected instanceof Path);
+    // Don't consider ValueNode as an Iterable as they only contain one value and iterating them does not make sense.
+    return expected instanceof Iterable && !(expected instanceof Path || isExpectedFieldAJsonValueNode());
```
scordio, line 183:
```suggestion
    return expected instanceof Iterable && !(expected instanceof Path || isAJsonValueNode(expected));
```

src/main/java/org/assertj/core/api/recursive/comparison/DualValue.java
```diff
@@ -170,16 +170,35 @@ public boolean isExpectedFieldAnOrderedCollection() {
...
+  }
+
+  private boolean isActualFieldAJsonValueNode() {
+    return isAJsonValueNode(actual);
+  }
+
+  private boolean isExpectedFieldAJsonValueNode() {
+    return isAJsonValueNode(expected);
+  }
```
scordio, line 192: I'd suggest dropping these intermediate methods.

joel-costigliola:
> I feel like the test provided in the issue could also be included, but not sure where (if needed) to add it.

I agree, we usually add the tests reported in the issues to be user we have fixed the issue.

src/test/java/org/assertj/core/api/recursive/comparison/DualValue_iterableValues_Test.java
```diff
@@ -131,8 +142,9 @@ void isExpectedFieldAnIterable_should_return_false_when_expected_is_not_an_itera
   }
 
   static Stream<Object> nonIterables() {
-    // even though Path is an iterable, it must not be considered as such
-    return Stream.of(123, "abc", array("a", "b"), Paths.get("/tmp"));
+    // even though Path and ValueNode are iterables, they must not be considered as such
+    return Stream.of(123, "abc", array("a", "b"), Paths.get("/tmp"), new TextNode("foo"), new IntNode(42), BooleanNode.TRUE,
+                     new POJONode(new Object()), MissingNode.getInstance(), NullNode.getInstance(), new BinaryNode(new byte[0]));
```
scordio, line 147: Instantiating directly Jackson types together with other iterables/non-iterables seems fragile to me.

I would rather try to replicate what end users would do, i.e., `new ObjectMapper().readTree("...")`, and compose different test cases as input of `readTree` in a separate parameterized test(s).

However, I'm not sure yet how it would look like. I can try to play with it and come up with a proposal, but don't wait for me in case you have already an idea 🙂

src/main/java/org/assertj/core/api/recursive/comparison/DualValue.java
temp-droid: I replicated what was done for `isAnOrderedCollection` but totally fine for me to remove them.

src/test/java/org/assertj/core/api/recursive/comparison/DualValue_iterableValues_Test.java
temp-droid: Maybe something in this spirit?

```java
  @ParameterizedTest
  @MethodSource("nonIterableJsonNodes")
  void isExpectedFieldAnIterable_should_return_false_when_expected_is_a_non_iterable_json_node(JsonNode expected) {
    // GIVEN
    DualValue dualValue = new DualValue(PATH, "", expected.findValue("value"));
    // WHEN
    boolean isExpectedFieldAnIterable = dualValue.isExpectedFieldAnIterable();
    // THEN
    assertThat(isExpectedFieldAnIterable).isFalse();
  }

  static Stream<JsonNode> nonIterableJsonNodes() {
    // even though ValueNode is an iterable, it must not be considered as such
    return Stream.of("{\"value\": \"foo\"}", "{\"value\": 42}", "{\"value\": true}")
                 .map(json -> toJsonNode(json));
  }

  private static JsonNode toJsonNode(String value) {
    ObjectMapper objectMapper = new ObjectMapper();
    try {
      return objectMapper.readTree(value);
    } catch (JsonProcessingException e) {
      throw new RuntimeException(e);
    }
  }
```
and another test for `"{\"value\": []}"` and `"{\"value\": {}}"`

joel-costigliola:
I applied @scordio suggestion directly on the PR and broke the PR build, I was fixing locally and verifying that the issue example was fixed, I realized that the example failed but not exactly as expected, the error message being:
```
java.lang.AssertionError: 
Expecting actual:
  {"someNotImportantValue":1,"importantValue":"10"}
to be equal to:
  {"someNotImportantValue":10,"importantValue":"1"}
when recursively comparing field by field, but found the following difference:

Top level actual and expected objects differ:
- actual value  : {"someNotImportantValue":1,"importantValue":"10"}
- expected value: {"someNotImportantValue":10,"importantValue":"1"}
The following actual elements were not matched in the expected ObjectNode:
  [1, "10"]
```
That's not exactly what we want.
Why do we get this error message and not one mentioning `"importantValue"`and  `"someNotImportantValue"`?
The reason is actual and expected are `ObjectNode` and are treated as iterables and the ObjectNode's iterator on loop on values :(.

This means this test succeeds while it really should not:
```java
@Test
void test() throws IOException {
  // GIVEN
  ObjectMapper om = new ObjectMapper();
  JsonNode actual = om.readTree("{\"someNotImportantValue\":1,\"importantValue\":\"10\"}");
  JsonNode expected = om.readTree("{\"foo\":1,\"bar\":\"10\"}");
  // WHEN
  then(actual).usingRecursiveComparison().isEqualTo(expected);
}
```
I think the way forward is to treat ObjectNode as map (it has a field `_children = new LinkedHashMap<String, JsonNode>();`).

I also feel JSON comparison should really be done by a json assertions library, the recursive is a poor substitute for it.

joel-costigliola:
Not considering ObjectNode as an iterable makes thing better, the test now fails with:
```
java.lang.AssertionError: 
Expecting actual:
  {"someNotImportantValue":1,"importantValue":"10"}
to be equal to:
  {"foo":1,"bar":"10"}
when recursively comparing field by field, but found the following difference:

field/property '_children' differ:
- actual value  : {"importantValue"="10", "someNotImportantValue"=1}
- expected value: {"bar"="10", "foo"=1}
The following actual map entries were not found in the expected map:
  {importantValue="10", someNotImportantValue=1}
```

joel-costigliola:
Integrated thanks @temp-droid!

PR #2465, opened 2022-01-12
refactor: change visibility of constructor in final util class to private

MartinWitt:
#### Check List:
* Fixes: no issue
* Unit tests : No

The Fail class is final, but the constructor is protected. This seems like a little bad smell. Because `Fail` is pure static class, I changed the visibility to private.

### DIFF
--- a/src/main/java/org/assertj/core/api/Fail.java
+++ b/src/main/java/org/assertj/core/api/Fail.java
@@ -109,8 +109,7 @@ public final class Fail {
   }
 
   /**
-   * This constructor is protected to make it possible to subclass this class. Since all its methods are static, there is no point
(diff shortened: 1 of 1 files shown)


joel-costigliola:
@MartinWitt I see what the intention is here but it is a breaking change in case someone had created a subclass (quite unlikely though but you never know), we probably should have made the constructor private initially but it's not harming anybody having it protected so I'd rather keep it that to avoid a breaking change unless you can give me a compelling use case ;-)

MartinWitt:
I would be surprised if someone creates a subclass of a final class, but sure it's your code base.

scordio:
Raised siom79/japicmp#310 about it.

Issue #2412, opened 2021-11-15
flatExtracting/flatMap for iterable/array should throw a proper assertion error if any element is null


joel-costigliola:
#### Summary

Similar to  https://github.com/assertj/assertj-core/issues/2401 but for `flatExtracting`/`flatMap`.

There are quite a few variations to cover and tests;
- [ ] `flatExtracting(Function)`, `flatMap(Function) `, `flatExtracting(ThrowingExtractor)` and `flatMap(ThrowingExtractor)`, will be fixed by changing `doFlatExtracting`, tests needed for the 4 methods exposed in the API.
- [ ] `flatExtracting(Function...)`, flatExtracting(ThrowingExtractor...), `flatMap(Function...)` and `flatMap(ThrowingExtractor...)`, will be fixed by changing `flatExtractingForProxy`, tests needed for  4 methods exposed in the API.

zihanxu3:
Hi can I take this issue? Thanks!

joel-costigliola:
Sure, go ahead @zihanxu3! thanks!

Issue #2411, opened 2021-11-15
extracting/map for iterable/array should throw a proper assertion error if any element is null


joel-costigliola:
#### Summary

Similar to  https://github.com/assertj/assertj-core/issues/2401 but for iterable/arrays

There are quite a few variations to cover and tests;
- [ ] `extracting(Function)`, `map(Function) `, `extracting(ThrowingExtractor)` and `map(ThrowingExtractor)`, will be fixed by changing `internalExtracting`, tests needed for the 4 methods exposed in the API.
- [ ] `extracting(Function...)` and `map(Function...)`, will be fixed by changing `extractingForProxy`, tests needed for both

Issue #2401, opened 2021-11-09
`extracting` should throw a proper assertion error if `actual` is null


scordio:
#### Summary
Based on https://github.com/assertj/assertj-core/issues/2392#issuecomment-961969975 and further comments, the implementation of `extracting(Function, AssertFactory)` and `extracting(String, AssertFactory)` in `AbstractAssert` should be improved to ensure that a preliminary `isNotNull` check is performed.

#### Example

The following assertions:

```java
assertThatObject(null).extracting(Object::getClass);
assertThatObject(null).extracting("class");
```

should fail with:

```
java.lang.AssertionError: 
Expecting actual not to be null
```

instead of failing with:

```
java.lang.NullPointerException
	at org.assertj.core.api.AbstractAssert.extracting(AbstractAssert.java:1016)
	at org.assertj.core.api.AbstractObjectAssert.extracting(AbstractObjectAssert.java:953)
	...
```

and:

```
java.lang.IllegalArgumentException: The object to extract property/field from should not be null
	at org.assertj.core.util.Preconditions.checkArgument(Preconditions.java:129)
	at org.assertj.core.util.introspection.PropertyOrFieldSupport.getValueOf(PropertyOrFieldSupport.java:49)
	at org.assertj.core.extractor.ByNameSingleExtractor.apply(ByNameSingleExtractor.java:29)
	at org.assertj.core.api.AbstractAssert.extracting(AbstractAssert.java:990)
	at org.assertj.core.api.AbstractObjectAssert.extracting(AbstractObjectAssert.java:834)
	...
```

respectively.

scordio:
@bjhargrave as you used `extracting` in the past for custom assertions (https://github.com/assertj/assertj-core/pull/1781#issuecomment-588377368), do you see any issues if we add an `isNotNull()` call to the `extracting` implementations of `AbstractAssert`?

bjhargrave:
> do you see any issues if we add an `isNotNull()` call to the `extracting` methods of `AbstractAssert`?

No. In all the cases were we use `extracting` in [osgi-test](https://github.com/osgi/osgi-test/), we have already called `isNotNull` directly or indirectly first.

For example: https://github.com/osgi/osgi-test/blob/659f5c7fe6c57cb194e3a511f1f4e726eceed166/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/bundle/AbstractBundleAssert.java#L93-L96

cc: @kriegfrj

[...]

PR #2476, opened 2022-01-29
Breaking change: Extracting within SoftAssertions now throws an assertion error if actual is null

KENNYSOFT:
#### Check List:
* Unit tests : YES
* PR meets the [contributing guidelines](https://github.com/assertj/assertj-core/blob/main/CONTRIBUTING.md) : Should be, but feel free to notify me if not!

#2401 introduced `extracting` to throw an assertion error if `actual` is `null`, but it's not effective when using `SoftAssertions`. So I've fixed it.

Actually, I think I should see related #2411 and #2412. Will take a look if my approach is right.

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -1110,8 +1110,9 @@ public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, 
                                                                        AssertFactory<T, ASSERT> assertFactory) {
     requireNonNull(extractor, shouldNotBeNull("extractor")::create);
     requireNonNull(assertFactory, shouldNotBeNull("assertFactory")::create);
+    T extractedValue = null;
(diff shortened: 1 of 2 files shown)


KENNYSOFT:
Weirdly, the Pull request list page doesn't show this. @scordio Can I get some review?

scordio:
Sure @KENNYSOFT, I had already a first look.

I would like to check our Byte Buddy configuration, it might be that we can solve this issue directly there.

I'll play with BB during the weekend and get back to you.

KENNYSOFT:
Wow, sounds great! Looking forward to see the issue resolved in any form.

scordio:
@KENNYSOFT I'm still working on this topic and I might need a few more days for it.

KENNYSOFT:
Thanks for the update. Hope you finish well.

By the way, I now learned JUnit's `@ParameterizedTest` and it resolves my actual issue in some different way; I used `SoftAssertions` in the loop to check all violations within the given list, but now the loop is iterated by JUnit and I just use a simple `Assertion`. 😉

PR #2477, opened 2022-01-29
Remove Byte Buddy shading

### LINKED ISSUE #2470: Do not shade Byte Buddy
Byte Buddy is currently shaded. As Byte Buddy retains compatibility, not shading the library would allow running recent JVMs without an update of assertj. Other libraries like Mockito or Hibernate do not shade BB and there are no known issues with this approach. In case of an error, stack traces would also be easier to process as break points remain intact.

scordio:
Fixes #2470.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,6 @@
 
   <artifactId>assertj-core</artifactId>
   <version>3.23.0-SNAPSHOT</version>
-  <packaging>jar</packaging>
 
   <name>AssertJ fluent assertions</name>
   <description>Rich and fluent assertions for testing for Java</description>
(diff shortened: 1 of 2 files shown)


scordio:
@bjhargrave @kriegfrj would you mind having a look at this PR?

Removing the shading of byte buddy impacted the Maven configuration for OSGi. I inspected the generated manifest and it looks good to me but would be great to have a check from you as well.

bjhargrave:
I am out of the office until Feb 3. I can look then.

scordio:
Sure and thanks, @bjhargrave.

pom.xml
```diff
@@ -357,81 +356,20 @@
...
-            <id>bundle-after-unpack-shaded</id>
-            <phase>package</phase>
+            <id>bnd-process</id>
             <goals>
               <goal>bnd-process</goal>
             </goals>
             <configuration>
               <bnd><![CDATA[
                 Import-Package:\
+                  net.bytebuddy.*,\
```
kriegfrj, line 372: You *could* leave this line out - then the import of `net.bytebuddy.*` would be covered by the default catch-all below and become optional. As I understand it, BB is only needed for a subset of AssertJ functionality (assumptions, soft assertions, and...?), so one could make a case for possibly making it optional as then those who aren't using those features wouldn't need to install BB.

In that case, something that would need to be considered is how to handle `ClassNotFoundException`/`NoClassDefFoundError` at runtime in a user-friendly manner if BB is not installed. Actually, this is handling that we might want to consider having in-place for non-OSGi usage too in case BB is not on the classpath.

scordio: If I understand it correctly, the current config would force OSGi users to install Byte Buddy even when BB related features are not used. Is the installation a manual step or does it happen automatically? My OSGi experience is close to zero so please forgive the dumb question 🙂 

Thinking about the non-OSGi world, I would keep BB as a transitive dependency to make this change as transparent as possible. Otherwise, users would need to know that from now on soft assertions and assumptions require a manual declaration of BB to work properly, which is not really nice.

kriegfrj: > If I understand it correctly, the current config would force OSGi users to install Byte Buddy even when BB related features are not used.

Correct. 

> Is the installation a manual step or does it happen automatically?

If it has been declared as a non- optional dependency, then it is added automatically during the OSGi resolve process (at least, it is automatic in any sanely configured OSGi build). The AssertJ bundle will fail to run if BB is not present. 

If it is an optional dependency, then you need to add it explicitly. If you don't, then the AssertJ bundle will still run - but you'll get a ClassNotFoundException or a NoClassDefFoundError at runtime if you try and use any part of the code that requires BB.

> My OSGi experience is close to zero so please forgive the dumb question 🙂
> 

No such thing as a dumb question. 😀

> Thinking about the non-OSGi world, I would keep BB as a transitive dependency to make this change as transparent as possible. Otherwise, users would need to know that from now on soft assertions and assumptions require a manual declaration of BB to work properly, which is not really nice.

If that is the plan then best to keep the OSGi dependency policy the same and leave the dependency as non-optional.

The ideal solution is to factor out separate module(s) for soft assertions and assumptions. Then the core module can be declared without the BB dependencies, and the new module(s) with. Users can pick the modules they want.

scordio: I'd say we keep the dependency as non-optional for the current change and we might evaluate this topic again with AssertJ 4.

bjhargrave:
This looks good to me. 

We will soon release Bnd 6.2 which has a [new packaging mode for bnd-maven-plugin](https://github.com/bndtools/bnd/tree/master/maven/bnd-maven-plugin#jar-goal) which covers the jar packaging as well. So we will propose a PR to move to that once Bnd 6.2 is released.

PR #2493, opened 2022-02-11
Fix usingRecursiveComparison with usingOverriddenEquals ignores equal…

### LINKED ISSUE #2479: usingRecursiveComparison with usingOverriddenEquals ignores equals method of the compared object
#### Summary
Version 3.22.0
I expect if I use usingRecursiveComparison with usingOverriddenEquals, field "age" should be ignored.
But as result I get

> java.lang.AssertionError: 
> Expecting actual:
>   com.ComparatorDevices$Person@63be0d6
> to be equal to:
>   com.ComparatorDevices$Person@63be0b6
> when recursively comparing field by field, but found the following difference:
> 
> field/property 'age' differ:
> - actual value  : "22"
> - expected value: "11"
> 
> The recursive comparison was performed with this configuration:
> - overridden equals methods were used in the comparison
> - these types were compared with the following comparators:
>   - java.lang.Double -> DoubleComparator[precision=1.0E-15]
>   - java.lang.Float -> FloatComparator[precision=1.0E-6]
>   - java.nio.file.Path -> lexicographic comparator (Path natural order)
> - actual and expected objects and their fields were compared field by field recursively even if they were not of the same type, this allows for example to compare a Person to a PersonDto (call strictTypeChecking(true) to change that behavior).

#### Example

```java
class Person {
        String name;
        String age;

        public Person(String name, String age) {
            this.name = name;
            this.age = age;
        }

        @Override
            if (this == o) return true;
            Person person = (Person) o;
            return Objects.equals(name, person.name);
        }

        @Override
        public int hashCode() {
            return Objects.hash(name, age);
        }
    }

@Test()
    public void personTest() {
        Person expectedPerson = new Person("name", "11");
        Person actualPerson = new Person("name", "22");
        Assertions.assertThat(actualPerson)
                .usingRecursiveComparison()
                .usingOverriddenEquals()
                .isEqualTo(expectedPerson);
    }

    
```

uc4w6c:
Please do not merge as it is in progress

Fix #2479

#### Check List:
* Fixes #2479
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
+++ b/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
@@ -113,9 +113,9 @@ public class RecursiveComparisonDifferenceCalculator {
       if (!visitedDualValues.contains(dualValue)) dualValuesToCompare.addFirst(dualValue);
     }
 
-    private void initDualValuesToCompare(Object actual, Object expected, FieldLocation fieldLocation, boolean isRootObject) {
+    private void initDualValuesToCompare(Object actual, Object expected, FieldLocation fieldLocation) {
       DualValue dualValue = new DualValue(fieldLocation, actual, expected);
-      boolean mustCompareFieldsRecursively = mustCompareFieldsRecursively(isRootObject, dualValue);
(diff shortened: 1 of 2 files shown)


joel-costigliola:
Integrated thanks @uc4w6c!

PR #2500, opened 2022-02-21
Fix `containsExactly` not working properly with maps not using equals to compare keys

### LINKED ISSUE #2165: `containsExactly` does not work properly with maps not using equals to compare keys
#### Summary

This test should pass but fails.

```java
  @Test
  void should_pass_if_case_insensitive_actual_contains_given_entries_in_rder() {
    // GIVEN
    actual.put( "Color", "green" );
    actual.put( "NAME", "Yoda" );

    // THEN
    maps.assertContainsExactly(someInfo(), actual, entry("COLOR", "green"), entry("Name", "Yoda"));
  }
```

The underlying issue is when `containsExactly` verifies that the entries are in the same order, it ends up comparing `"Color"` against `"COLOR"` with `equals` instead of the actual map key comparison strategy (case insensitive string in the example).

It's not yet clear if there is a way to address this issue, if not we would have to mention the limitation in the javadoc.

wouterpolet:
#### Check List:
* Fixes #2165 
* Unit tests : YES
* PR meets the [contributing guidelines](https://github.com/assertj/assertj-core/blob/main/CONTRIBUTING.md : Check

Let me know if there is anything I missed.

I created tests that run on all case insensitive maps, but didn't add ones for normal ones. I wasn't sure if the existing test suite is already sufficient. I can still add that of course.


scordio:
> I created tests that run on all case insensitive maps, but didn't add ones for normal ones. I wasn't sure if the existing test suite is already sufficient. I can still add that of course.

I think the existing tests are not providing the needed coverage, especially for unmodifiable implementations.

The best would be having two parameterized tests like in `Maps_assertContainsOnlyKeys_Test`, so `should_pass`:

https://github.com/assertj/assertj-core/blob/8fa801095be09d75155bb6f159fb33f621512dcf/src/test/java/org/assertj/core/internal/maps/Maps_assertContainsOnlyKeys_Test.java#L86-L96

and `should_fail`:

https://github.com/assertj/assertj-core/blob/8fa801095be09d75155bb6f159fb33f621512dcf/src/test/java/org/assertj/core/internal/maps/Maps_assertContainsOnlyKeys_Test.java#L130-L144

wouterpolet:
I added the tests and fixed some errors that came from them :slightly_smiling_face:

joel-costigliola:
closing as it is addressed by https://github.com/assertj/assertj-core/pull/2623 which keeps @wouterpolet as the author.

PR #2502, opened 2022-02-22
Removed unnecessary unboxing in `ComparatorFactory`.

DiegoKrupitza:
Removing those unnecessary unboxing makes the clode cleaner, which only beneftis in the long term the project.


scordio:
Thanks, @DiegoKrupitza!

PR #2503, opened 2022-02-25
Fix MapEntry.hashCode() to conform to Map.Entry.hashCode() javadoc

bjhargrave:
The javadoc for Map.Entry.hashCode() specifies how to compute the
hashcode value for a Map Entry. This fix conforms to that
specification.


scordio:
Thanks, @bjhargrave. Good catch!

Actually, I think we should just let `MapEntry` extend `AbstractMap.SimpleImmutableEntry`, only keeping `entry()` and `toString()`. I'll take care of it during the merge.

PR #2512, opened 2022-03-02
Updated docs describing passing Fail.fail() as a lambda

armandino:
#### Check List:
* Fixes NA
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/Assertions.java
+++ b/src/main/java/org/assertj/core/api/Assertions.java
@@ -1752,7 +1752,7 @@ public class Assertions implements InstanceOfAssertFactories {
    * @param <T> dummy return value type
    * @param failureMessage error message.
    * @param args Arguments referenced by the format specifiers in the format string.
-   * @return nothing, it's just to be used in {@code doSomething(optional.orElseGet(() -> fail("b%s", ""oom)));}.
+   * @return nothing, it's just to be used in {@code doSomething(optional.orElseGet(() -> fail("b%s", "oom")));}.
    * @throws AssertionError with the given built message.
    */
   @CanIgnoreReturnValue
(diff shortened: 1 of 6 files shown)


src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -1752,7 +1752,7 @@ public static void setRemoveAssertJRelatedElementsFromStackTrace(boolean removeA
    * @param <T> dummy return value type
    * @param failureMessage error message.
    * @param args Arguments referenced by the format specifiers in the format string.
-   * @return nothing, it's just to be used in doSomething(optional.orElse(() -&gt; fail("b%s", ""oom)));.
+   * @return nothing, it's just to be used in {@code doSomething(optional.orElseGet(() -> fail("b%s", ""oom)));}.
```
scordio, line 1755:
```suggestion
   * @return nothing, it's just to be used in {@code doSomething(optional.orElseGet(() -> fail("b%s", "oom")));}.
```

src/main/java/org/assertj/core/api/WithAssertions.java
```diff
@@ -176,7 +176,7 @@
    * @param <T> dummy return value type
    * @param failureMessage error message.
    * @param args Arguments referenced by the format specifiers in the format string.
-   * @return nothing, it's just to be used in doSomething(optional.orElse(() -&gt; fail("b%s", ""oom)));.
+   * @return nothing, it's just to be used in {@code doSomething(optional.orElseGet(() -> fail("b%s", ""oom)));}.
```
scordio, line 179:
```suggestion
   * @return nothing, it's just to be used in {@code doSomething(optional.orElseGet(() -> fail("b%s", "oom")));}.
```

scordio:
Good catch. Thanks, @armandino!

PR #2513, opened 2022-03-03
Bnd 6.2.0

bjhargrave:
Update bnd maven plugins to 6.2.0 and move to use the new jar and test-jar goals of
bnd-maven-plugin.

Closes https://github.com/assertj/assertj-core/pull/2506

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@
     <byte-buddy.version>1.12.8</byte-buddy.version>
     <hamcrest.version>2.2</hamcrest.version>
     <opentest4j.version>1.2.0</opentest4j.version>
-    <bnd.version>6.1.0</bnd.version>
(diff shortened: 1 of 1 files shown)


bjhargrave:
cc: @rotty3000

joel-costigliola:
thanks @bjhargrave!

scordio:
@bjhargrave we have a [different configuration](https://github.com/assertj/assertj-guava/blob/541b47807205f3934db6687bfd966c3c54e542e5/pom.xml#L115-L134) in `assertj-guava` based on the `maven-bundle-plugin`. 

I guess we could use the `bnd-maven-plugin` instead, right?

bjhargrave:
> I guess we could use the `bnd-maven-plugin` instead, right?

Yes, I'll make a PR.

PR #2519, opened 2022-03-08
Add `assertThatCode(Callable)`

### LINKED ISSUE #1652: Assert return value of callable throwing exception

scordio:
Usage examples in [`CallableAssert_Demo_Test`](https://github.com/assertj/assertj-core/blob/callable-assert/src%2Ftest%2Fjava%2Forg%2Fassertj%2Fcore%2Fapi%2FCallableAssert_Demo_Test.java).

This introduces a breaking change for all the `assertThatCode()` assertions with a `Callable` that don't call `doesNotThrowAnyException()` but any other method from `AbstractThrowableAssert`.

#### Check List:
* Fixes #1652
* Unit tests : not yet
* Javadoc with a code example (on API only) : not yet

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractCallableAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractCallableAssert.java
@@ -1,49 +1,52 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2022 the original author or authors.
+ */
 package org.assertj.core.api;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.error.ShouldNotHaveThrown.shouldNotHaveThrown;
 
 import java.util.concurrent.Callable;
 
-import org.assertj.core.internal.Failures;
+import org.assertj.core.util.CheckReturnValue;
 
 public abstract class AbstractCallableAssert<SELF extends AbstractCallableAssert<SELF, V>, V>
-    extends AbstractAssert<SELF, Callable<V>> {
+    extends AbstractThrowingExecutableAssert<SELF, Callable<V>> {
 
-  private V result;
-  private Throwable thrown;
+  protected V result;
 
   protected AbstractCallableAssert(Callable<V> actual, Class<?> selfType) {
     super(actual, selfType);
-
-    try {
-      result = actual.call();
-    } catch (Throwable e) {
-      thrown = e;
-    }
   }
 
-  public SELF doesNotThrowAnyException() {
-    if (thrown != null) throw Failures.instance().failure(info, shouldNotHaveThrown(thrown));
-    return myself;
+  @Override
+  protected void execute(Callable<V> actual) throws Exception {
+    result = actual.call();
   }
 
+  @CheckReturnValue
   public AbstractObjectAssert<?, V> result() {
     return internalResult();
   }
 
+  @CheckReturnValue
   public <ASSERT extends AbstractAssert<?, ?>> ASSERT result(InstanceOfAssertFactory<?, ASSERT> assertFactory) {
     return internalResult().asInstanceOf(assertFactory);
   }
 
   private AbstractObjectAssert<?, V> internalResult() {
-    doesNotThrowAnyException(); // FIXME enough? better error message?
-    return assertThat(result).withAssertionState(myself);
-  }
-
-  public ThrowableAssertAlternative<Throwable> hasThrownException() {
-    if (thrown == null) throw Failures.instance().expectedThrowableNotThrown(Throwable.class);
-    return new ThrowableAssertAlternative<>(thrown).withAssertionState(myself);
+    doesNotThrowAnyException();
+    ObjectAssert<V> objectAssert = assertThat(result);
+    objectAssert.propagateAssertionInfoFrom(info);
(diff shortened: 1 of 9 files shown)


joel-costigliola:
LGTM :)

joel-costigliola:
I think we can drop my attempt https://github.com/assertj/assertj-core/pull/2516/files

src/main/java/org/assertj/core/api/AbstractCallableAssert.java
```diff
@@ -0,0 +1,49 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.error.ShouldNotHaveThrown.shouldNotHaveThrown;
+
+import java.util.concurrent.Callable;
+
+import org.assertj.core.internal.Failures;
+
+public abstract class AbstractCallableAssert<SELF extends AbstractCallableAssert<SELF, V>, V>
+    extends AbstractAssert<SELF, Callable<V>> {
```
scordio, line 11: @joel-costigliola I'm having second thoughts on extending `AbstractAssert` or `Assert` in general. I don't see any benefit on having the inherited assertions available, it will just pollute the code completion.

What if we would have a `ThrowingAssert` interface with two methods:
* `doesNotThrowAnyException`
* `hasThrownException`

My wish would be to use it here and also for `Runnable` related assertions (i.e., the existing `assertThatCode(ThrowingCallable)`), but I think the latter would be feasible only in version 4.

BTW in version 4 we could think about renaming `ThrowingCallable` to `ThrowingRunnable`.

joel-costigliola: I think the only drawback would be not being able to describe the assertions but I'm happy we try that, we can make `ThrowingAssert` to implement `Descriptable`,  the value of a clean API overweight the need to reimplement describing assertions

joel-costigliola (commented):
commented

joel-costigliola:
Slightly off topic, I think the existing `assertThatCode` is not great as it exposes throwable assertions, even if the code throws one, it does not read very nicely:
```java 
// why would we except to have an exception here?
assertThatCode(() -> doStuff()).hasMessage("Boom!");
```
your version with callable is much better in that sense.
The existing alternative is:
```
```java 
// reads nicely
assertThatThrownBy(() -> doStuff()).hasMessage("Boom!");
```

Ideally I would replace the existing `assertThatCode(ThrowingCallable)` by your version but that's probably something we should do for AssertJ 4 as it's a pretty big breaking change. What we could do though is deprecate `assertThatCode(ThrowingCallable)` in favor of `assertThatThrownBy` or  `assertThatCode(Callable)`.

scordio:
> Slightly off topic, I think the existing `assertThatCode` is not great as it exposes throwable assertions, even if the code throws one, it does not read very nicely

Yes, I tried to say that but with wrong wording 😆 

I'm exploring the direction of `ThrowingAssert` for `assertThatCode(Callable)` and maybe it's not so pricey to introduce `ThrowingRunnable` and the corresponding `assertThatCode` as an alternative for `assertThatCode(ThrowingCallable)`.

scordio:
@joel-costigliola I went with `ThrowingExecutableAssert`, maybe clearer than just `ThrowingAssert`. Could you have a look and let me know what you think?

I'd experiment the replacement of `assertThatCode(ThrowingCallable)` in a separate PR/branch once this one is stable. I have the feeling it would be a binary incompatible change but source compatible at the same time, as long as no one used weird patterns like `assertThatCode(...).hasMessage(...)`.

src/main/java/org/assertj/core/api/AbstractCallableAssert.java
```diff
@@ -0,0 +1,60 @@
...
+    extends AbstractThrowingExecutableAssert<SELF, Callable<V>> {
+
+  protected V result;
+
+  protected AbstractCallableAssert(Callable<V> actual, Class<?> selfType) {
+    super(actual, selfType);
+  }
+
+  @Override
+  protected void execute(Callable<V> actual) throws Exception {
```
ascopes, line 31: is it worth specifying that this throws `Throwable` directly? Shouldn't make much difference but semantically it'd make sense to be able to catch Errors and the likes if testing code that could raise errors directly, which I assume this would do already.

scordio: Right now, it wouldn't catch `Throwable` or `Error` instances as the input parameter is a `Callable`, and [`Callable::call`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/Callable.html#call()) allows throwing `Exception`s only.

Maybe we should consider if a new Callable-like interface that throws `Throwable` should be defined instead of relying on the JDK `Callable`.

@ascopes do you already have a concrete use case that would require it?

scordio: As suggested by @sbrannen in https://github.com/assertj/assertj/issues/3314#issuecomment-1874217792, a `ThrowingSupplier` could be defined if we decide that `Callable` throwing `Exception` is not flexible enough.

PR #2532, opened 2022-03-20
Fix support for generic comparable

### LINKED ISSUE #2557: `assertThat(comparable)` returns `ObjectAssert` instead of `AbstractComparableAssert`
### Summary
Given a `Comparable<T>` variable, it's currently not possible to use assertions from [`ComparableAssert`](https://www.javadoc.io/doc/org.assertj/assertj-core/latest/org/assertj/core/api/ComparableAssert.html) unless `T` is `Object`.

Also, if the type implementing `Comparable` targets a parent type that is not `Object`, `assertThat` always returns `ObjectAssert` for both strongly-typed and `Comparable<T>` variables.

### Example: JDK type not comparing with `Object`

```java
Comparable<String> comparable = "value";

assertThat(comparable)            // ObjectAssert<Comparable<String>>
  .isEqualByComparingTo("value"); // does not compile
```

### Example: Custom type comparing with self

```java
class ComparingWithSelf implements Comparable<ComparingWithSelf> {
  @Override
  public int compareTo(ComparingWithSelf other) {
    return 0;
  }
}

Comparable<ComparingWithSelf> comparable = new ComparingWithSelf();

assertThat(comparable)                            // ObjectAssert<Comparable<ComparingWithSelf>>
  .isEqualByComparingTo(new ComparingWithSelf()); // does not compile

ComparingWithSelf comparingWithSelf = new ComparingWithSelf();

assertThat(comparingWithSelf)                     // AbstractComparableAssert<?, ComparingWithSelf>
  .isEqualByComparingTo(new ComparingWithSelf()); // compiles
```

### Example: Custom type comparing with parent

```java
class Parent {}

class ComparingWithParent extends Parent implements Comparable<Parent> {
  @Override
  public int compareTo(Parent other) {
    return 0;
  }
}

Comparable<Parent> comparable = new ComparingWithParent();

assertThat(comparable)                              // ObjectAssert<Comparable<Parent>>
  .isEqualByComparingTo(new ComparingWithParent()); // does not compile

ComparingWithParent comparingWithParent = new ComparingWithParent();

assertThat(comparingWithParent)                     // ObjectAssert<ComparingWithParent>
  .isEqualByComparingTo(new ComparingWithParent()); // does not compile
```

### Example: Custom type comparing with `Object`

```java
class ComparingWithObject implements Comparable<Object> {
  @Override
  public int compareTo(Object other) {
    return 0;
  }
}

Comparable<Object> comparable = new ComparingWithObject();

assertThat(comparable)                              // AbstractComparableAssert<?, Comparable<Object>
  .isEqualByComparingTo(new ComparingWithObject()); // compiles

ComparingWithObject comparingWithObject = new ComparingWithObject();

assertThat(comparingWithObject)                     // AbstractComparableAssert<?, ComparingWithObject>
  .isEqualByComparingTo(new ComparingWithObject()); // compiles
```

### Workaround: Raw type

```java
Comparable comparable = "value";  // "rawtypes" warning

assertThat(comparable)            // AbstractComparableAssert<?, Comparable> with "unchecked" warning
  .isEqualByComparingTo("value"); // compiles with "unchecked" warning
```

### Workaround: `asInstanceOf`

This works only with non-null instances due to the null-check performed in `asInstanceOf`.

```java
Comparable<String> comparable = "value";

assertThat(comparable)                    // ObjectAssert<Comparable<String>>
  .asInstanceOf(comparable(String.class)) // AbstractComparableAssert<?, String>
  .isEqualByComparingTo("value");         // compiles
```
### Current status

| `assertThat` use case                 | `Comparable<T>` variable   | Strongly-typed variable    |
|--------------------------------------|----------------------------|----------------------------|
| JDK type not comparing with `Object` | `ObjectAssert`             | N/A (dedicated assertions) |
| Custom type comparing with self      | `ObjectAssert`             | `AbstractComparableAssert` |
| Custom type comparing with parent    | `ObjectAssert`             | `ObjectAssert`             |
| Custom type comparing with `Object`  | `AbstractComparableAssert` | `AbstractComparableAssert` |

joel-costigliola:
Adjust error classes generic signature to be compatible with AbstractGenericComparableAssert assertions

Fixes #2557


src/main/java/org/assertj/core/internal/Comparables.java
```diff
@@ -153,12 +154,12 @@ public String toString() {
...
-  public <T extends Comparable<? super T>> void assertEqualByComparison(AssertionInfo info, T actual, T expected) {
+  @SuppressWarnings("unchecked")
+  public <T> void assertEqualByComparison(AssertionInfo info, Comparable<? super T> actual, Comparable<? super T> expected) {
     assertNotNull(info, actual);
     // we don't delegate to comparisonStrategy, as this assertion makes it clear it relies on Comparable
-    if (actual.compareTo(expected) == 0)
-      return;
-    throw failures.failure(info, shouldBeEqual(actual, expected, info.representation()));
+    // TODO are we sure that any instance of Comparable<? super T> is always a T?
+    if (actual.compareTo((T) expected) != 0) throw failures.failure(info, shouldBeEqual(actual, expected, info.representation()));
```
joel-costigliola, line 162: @scordio as I changed the generic signature `expected` is now a `Comparable<T>` but not a `T` and thus `actual.compareTo(expected)` does not compile. I "fixed" that with a cast but I'm not 100% sure we won't get a class cast exception (maybe not as generics are erased at runtime).
I could not find a type that would be a `Comparable<T>` but not a `T`, do you think that cast can blow up?

src/main/java/org/assertj/core/api/WithAssertions.java
```diff
@@ -731,8 +731,8 @@ default <T extends Comparable<? super T>> AbstractComparableAssert<?, T> assertT
    * @return the created assertion object.
    * @since 3.23.0
    */
-  default <T extends Comparable<? super T>> AbstractComparableAssert<?, T> assertThatComparable(T actual) {
-    return assertThat(actual);
+  default RawComparableAssert assertThatComparable(Comparable<?> actual) {
```
scordio, line 734:
```suggestion
  default AbstractRawComparableAssert<?> assertThatComparable(Comparable<?> actual) {
```

scordio (commented):
Just some "cosmetic" feedback 🙂

I'll have a second round for the real stuff!

src/main/java/org/assertj/core/internal/Comparables.java
scordio: I was able to put together something weird enough that can throw a `ClassCastException`:

```java
static class BaseClass implements Comparable<Subclass> {
  @Override
  public int compareTo(Subclass o) {
    return 0;
  }
}

static class Subclass extends BaseClass {
}

@Test
void test() {
  // GIVEN
  Subclass s1 = new Subclass();
  BaseClass b1 = new BaseClass();
  // WHEN/THEN
  assertThatComparable(s1).isEqualByComparingTo(b1);
}
```

yields:

```
java.lang.ClassCastException: class org.assertj.core.api.Assertions_assertThatComparable_Test$BaseClass cannot be cast to class org.assertj.core.api.Assertions_assertThatComparable_Test$Subclass
```

Same result with the changes in #2587.

joel-costigliola: Good example, I think it is very unlikely to encounter a type like that because one could argue that defining a base class in term of its subclasses is not a great design.

scordio: I agree, it's a pretty rare use. I'll give it another try to see if we can prevent it.

scordio: I couldn't find a way to prevent:

```
assertThatComparable(s1).isEqualByComparingTo(b1);
```

at compile time without breaking other use cases, knowing that:

```
s1.compareTo(b1);
```

doesn't compile.

This is because we want to provide a `Comparable` as input parameter for the assertion method, while `compareTo` directly accepts the value type.

We could make it work but we would take out the option of providing `Comparable<T>` as assertion parameter, like in the following examples:

```java
Comparable<Name> name1 = new Name("ABC"); // object under assertion can be Comparable

Name name2 = new Name("ABC");
Comparable<Name> comparableName2 = name2;

assertThatComparable(name1).isEqualByComparingTo(name2); // compiles
assertThatComparable(name1).isEqualByComparingTo(comparableName2); // does not compile
```

I tend to favor type-safety at the price of not supporting the case with `comparableName2`.

What do you think @joel-costigliola ?

scordio:
I think we're almost there with this topic. I'd like to spend a bit more time on two topics:
* What I've mentioned in https://github.com/assertj/assertj-core/pull/2532#discussion_r871856650, to be evaluated as a separate PR on top of this branch
* Some refactoring of test classes

src/main/java/org/assertj/core/internal/Comparables.java
joel-costigliola: I agree, I think `assertThatComparable(name1).isEqualByComparingTo(name2);` is more commonly seen than `assertThatComparable(name1).isEqualByComparingTo(comparableName2);`

PR #2534, opened 2022-03-21
Fix file names to prevent MacOS casing issues

sarajuhosova:
Fixes the issue of MacOS not accepting files with the same name but different casing.

#### Check List:
* Fixes: NA
* Unit tests : NA
* PR meets the [contributing guidelines](https://github.com/assertj/assertj-core/blob/main/CONTRIBUTING.md): NA


scordio:
Very good catch, thanks @sarajuhosova!

PR #2562, opened 2022-04-10
Add assumptions methods to be in sync with assertions methods

### LINKED ISSUE #2531: Add assumptions methods to be in sync with assertions methods
#### Summary

Sync `assertThatXxx` methods with assumptions `assumeThatXxx`.

#### Missing ones

* `assumeThatCollection`
* `assumeThatIterable`
* `assumeThatIterator`
* `assumeThatList`
* `assumeThatPath`
* `assumeThatPredicate`
* `assumeThatStream`
* `assumeThatExceptionOfType`
* `assumeThatException`
* `assumeThatRuntimeException`
* `assumeThatNullPointerException`
* `assumeThatIllegalArgumentException`
* `assumeThatIOException`
* `assumeThatIndexOutOfBoundsException`
* `assumeThatReflectiveOperationException"`

Refactor `Assertions_sync_with_BDDAssumptions_Test` to be parameterized as `Assertions_sync_with_BDDAssertions_WithAssertions_and_soft_assertions_variants_Test`.

Don't add `assertThatComparable` it is going to be done as part of another ongoing issue.

HDreamer2:
#### Check List:
* Fixes #2531
* Unit tests :  NA


src/test/java/org/assertj/core/api/Assertions_sync_with_BDDAssumptions_Test.java
```diff
@@ -14,9 +14,16 @@
...
+
 import java.lang.reflect.Method;
+import java.util.stream.Stream;
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
 class Assertions_sync_with_BDDAssumptions_Test extends BaseAssertionsTest {
```
joel-costigliola, line 28: there is quite a few things to change, here's my suggestions:

- rename the test to `Assertions_sync_with_Assumptions_and_BDDAssumptions_Test` since we are going to cover everything here, 
- remove Assertions_sync_with_Assumptions_Test because of the previous point
- start with all `assertThatXxx` methods we should have an `assumeThatXxx` counterpart, that would be (if I haven't missed anything):
```
  private static Stream<String> assertion_methods() {
    return Stream.of("assertThat",
                     "assertThatCode",
                     "assertThatCollection",
                     "assertThatComparable",
                     "assertThatIterable",
                     "assertThatIterator",
                     "assertThatList",
                     "assertThatObject",
                     "assertThatPath",
                     "assertThatPredicate",
                     "assertThatStream",
                     "assertThatException",
                     "assertThatRuntimeException",
                     "assertThatNullPointerException",
                     "assertThatIllegalArgumentException",
                     "assertThatIOException",
                     "assertThatIndexOutOfBoundsException",
                     "assertThatReflectiveOperationException");
```
- then have a parameterized test that looks for each of them in Assertions and Assumptions with an helper method like `toAssumptionMethod` that replace `assertThat` by `assumeThat` (as you did for `toBDDAssumptionMethod`)
- remove the uneeded tests like code_assertions_and_bdd_assumptions_should_have_the_same_assertions_methods which is covered by having  `"assertThatCode",` in `assertion_methods`
- remove object_assertions_and_bdd_assumptions_should_have_the_same_assertions_methods 
- remove standard_assertions_and_bdd_assumptions_should_have_the_same_assertions_methods

I think we only need to parameterized tests:
- `standard_assertions_and_assumptions_should_have_similar_methods`
- `standard_assumptions_and_bdd_assumptions_should_have_the_same_assumptions_methods`

If `Assertions` is in sync with `Assumptions` and `Assumptions` with `BDDAssumptions` then `Assertions` is in sync with `BDDAssumptions` and we don't need a test for the latter.

Let me know if you have questions or clarifications, hopefully I haven't missed anything!

joel-costigliola (changes requested):
@HDreamer2 thanks for the PR, the main code is good (a javadoc comment to add), the main remaining work is on the test.

src/test/java/org/assertj/core/api/Assertions_sync_with_BDDAssumptions_Test.java
joel-costigliola: yes, we usually add assertion methods first and then the corresponding assumption ones so it makes sense that the assertions methods are the reference.

HDreamer2: Hi, I met some problems: The issue mentioned that don't add `assumeThatComparable` (hopefully I don't misunderstand) and in `assertion_methods` there is 'assertThatComparable' which is not sync in `Assumptions`. When I test `standard_assumptions_with_assumptions_should_have_simlilar_methods`, I found some `assertThat` methods and 'assertThatComparable' don't have corresponding 'assumeThat' methods, and 'assertThatList/Stream' and 'assumeThatList/Stream' don't have the same return type (the former is `<ELEMENT> ListAssert<ELEMENT>` and the latter is '<ELEMENT> AbstractListAssert<?, List<? extends ELEMENT>, ELEMENT, ObjectAssert<ELEMENT>>'). What should I do to solve these problems?

joel-costigliola: Ah ah, that means we need add `assumeThatComparable`.

When returned type are different, this can be ignored in methods comparison with `IGNORING_DECLARING_CLASS_AND_RETURN_TYPE`, see:
see 
- https://github.com/assertj/assertj-core/blob/09203199cf81e6482f8bab8d51f383ba1b207e92/src/test/java/org/assertj/core/api/BaseAssertionsTest.java#L40
- https://github.com/assertj/assertj-core/blob/main/src/test/java/org/assertj/core/api/Assertions_sync_with_BDDAssertions_WithAssertions_and_soft_assertions_variants_Test.java#L96

HDreamer2: Hi, I have one more question:

- In `standard_assumptions_with_assumptions_should_have_simlilar_methods` could I use `IGNORING_DECLARING_CLASS_RETURN_TYPE_AND_METHOD_NAME` since the method name `assertThatXXX` is different from 'assumeThatXXX' ? If it is OK, I found that there are some 'assertThat' methods with more than one parameters and they didn't match any 'assumeThat' methods. What should I do then?

joel-costigliola: you can ;-)
also if you could rebase your branch on top of main instead merging main into your branch, that would make it easier for us to integrate the PR, if it's too complicated that's alright we'll take care of it.

HDreamer2: Thanks! What about the `assertThat` methods with more than one parameter (In test they don't have any match) ? What can I do to handle them?

joel-costigliola: I think for these ones we can write one test per method as there are not not so many of them.

HDreamer2: Oops, I think I described the problem in a wrong way. I checked the test `standard_assertions_and_with_assumptions_should_have_similar_methods` and found that only test `assertThat` failed. That's because there are some ’assertThat‘ methods don't have corresponding version in 'assumeThat' methods. Here are  them:

- `public static <T> T assertThat(AssertProvider<T>)`
- `public static <T> ObjectAssert<T> assertThat(T)`
- `public static <ACTUAL extends List<? extends ELEMENT>,ELEMENT,ELEMENT_ASSERT extends AbstractAssert<ELEMENT_ASSERT, ELEMENT>> FactoryBasedNavigableListAssert<?, ACTUAL, ELEMENT, ELEMENT_ASSERT> assertThat(List<? extends ELEMENT>, AssertFactory<ELEMENT, ELEMENT_ASSERT>)`
- `public static <ELEMENT,ACTUAL extends List<? extends ELEMENT>,ELEMENT_ASSERT extends AbstractAssert<ELEMENT_ASSERT, ELEMENT>> ClassBasedNavigableListAssert<?, ACTUAL, ELEMENT, ELEMENT_ASSERT> assertThat(List<? extends ELEMENT>,java.lang.Class<ELEMENT_ASSERT>)`
- `public static <ACTUAL extends Iterable<? extends ELEMENT>,ELEMENT,ELEMENT_ASSERT extends AbstractAssert<ELEMENT_ASSERT, ELEMENT>> FactoryBasedNavigableIterableAssert<?, ACTUAL, ELEMENT, ELEMENT_ASSERT> assertThat(Iterable<? extends ELEMENT>,AssertFactory<ELEMENT, ELEMENT_ASSERT>)`
- `public static <ACTUAL extends Iterable<? extends ELEMENT>,ELEMENT,ELEMENT_ASSERT extends AbstractAssert<ELEMENT_ASSERT, ELEMENT>> ClassBasedNavigableIterableAssert<?, ACTUAL, ELEMENT, ELEMENT_ASSERT>Assertions.assertThat(ACTUAL,Class<ELEMENT_ASSERT>)`

(For the second method, it has two versions in 'Assertion' but there is only one in 'Assumption')
So the problem is not about methods with more than one parameter but about methods that don't have counterpart.

joel-costigliola: my bad, I did not understand your initial question.
I think it would be best for the time being to ignore these methods - except for `public static <T> ObjectAssert<T> assertThat(T)` -  as they are not widely used for assertions (I would wait for someone to ask the corresponding assumptions methods).

HDreamer2: Does ingnoring these methods mean I do nothing about these methods and they are ignored by the test (because I saw the 'assertThat' test was ignored in IDEA)?

joel-costigliola: I meant that when you get all `assertThat` methods you would need to remove the one we want to ignore or change the filter to exclude them directly. It will be easier to remove `"assertThat"` from `assertion_methods()` and have one test just for `assertThat`.

HDreamer2: I wrote a test `assertThat_and_assumeThat_should_have_the_similar_methods` and in this test I used a method `assertThat_filter_methods()` like `assertion_methods()` to filter out methods to be ignored as you said. Everything worked well except for `public static <T> ObjectAssert<T> assertThat(T)`. I checked the method and confirmed there is only one such method (not two as I said before) but in the test result there are two of them, one matched and the other mismatched. So I'm confused about this situation.

joel-costigliola: can you push your commits on your PR so that I can have look and run the code locally, it's going to be easier to find out what is going on, thanks!

HDreamer2: Thanks! I've pushed my commits.

joel-costigliola:
Integrated thanks @HDreamer2 for the good work!

PR #2566, opened 2022-04-12
Issue 2541: hasMonth() and hasMonthValue()

nith2001:
#### Check List:
* Fixes partially #2541
* Unit tests : YES
* Javadoc with a code example (on API only) : YES
* PR meets the [contributing guidelines](https://github.com/assertj/assertj-core/blob/main/CONTRIBUTING.md)

PR #2579, opened 2022-04-19
Fix javadoc error of `hasMonthValue` in `LocalDate` assertions

Zhou-Yicheng:
#### Check List:
* See #2566 the last comment
* Unit tests :  NA


joel-costigliola:
Good catch thanks @Zhou-Yicheng.

PR #2586, opened 2022-04-24
Overload returns() and add returnsValue() & returnsEmpty()

### LINKED ISSUE #2341: Extending AbstractAssert.returns
#### New variants of the returns assertions 

The returns assertion could be even more useful if there would be variations for Optionals. I suggest ``returnsValue(expected, function)`` as counterpart to the hasValue assertion to check Optional return values.

It would also be nice to have a description without writing ``.as("...")``. This also holds true for the normal ``returns`` assertion and is already done for the ``matches`` assertion. 

So in total I would suggest to add the following assertion functions:
* ``returns(expected, function, description)``
* ``returnsValue(expected, function, description)``
* ``returnsEmpty(function, description)``

#### Example

```java
class ObjectUnderTest {
  private Optional<String> optString;
  
  ObjectUnderTest(Optional<String> o) {
    optString = o;
  }

  Optional<String> getOptString() {
    return optString;
  }
}

@Test
void should_pass_optional_returns() {
  ObjectUnderTest out = new ObjectUnderTest(Optional.of("test string");
  assertThat(out).returnsValue("test string", ObjectUnderTest::getOptString);
}
```

SickDawn:
#### Check List:
* Fixes #2341
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractObjectAssert.java
@@ -74,6 +74,47 @@ public abstract class AbstractObjectAssert<SELF extends AbstractObjectAssert<SEL
   }
 
   /**
+   * This is to Overload returns() function.
+   * To combine returns() with as() function to customize the errorMessage.
+   */
+  public <T> SELF returns(T expected, Function<ACTUAL, T> from, String description){
+    requireNonNull(from,"The given getter method/Function must not be null");
+    as(description);
+    objects.assertEqual(info, from.apply(actual), expected);
+    return myself;
+  }
+
+  /**
+   * This is returnsValue() function.
+   * To directly show result of function from.
+   * Base on return().
+   */
+  public <T> SELF returnsValue(T expected, Function<ACTUAL, T> from, String description){
+    requireNonNull(from,"The given getter method/Function must not be null");
+    String value_words = from.toString();
+    description = description + value_words;
+    as(description);
+    objects.assertEqual(info, from.apply(actual), expected);
+    return myself;
+  }
+
+  /**
+   * This is returnsEmpty() function.
+   * To remove requireNonNull() based on return().
+   * To directly show that function from is empty instead of throws common error message.
+   * return actual = -1, expected = 0 as replacement of function check.
+   */
+  public <T> SELF returnsEmpty(T expected, Function<ACTUAL, T> from, String description){
+    if(from == null){
+      as("Function is null!");
+      objects.assertEqual(info, -1,0);
+    }
+    as(description);
+    objects.assertEqual(info, from.apply(actual), expected);
+    return myself;
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


sroeper:
@SickDawn thank you for your implementation!
I think there are some issues as far as I could see. Unfortunately I have not found a way to comment inside the code, so I will try to comment it here.

returnsEmpty is not implemented as suggested. The parameters does not make sense to me, the expected value is Optional.empty(), so I would suggest:
```java
public <T> SELF returnsEmpty(Function<ACTUAL, T> from, String description){
    requireNonNull(from,"The given getter method/Function must not be null");
    as(description);
    objects.assertEqual(info, from.apply(actual), Optional.empty());
    return myself;
  }
```

The tests for this method could look like:
```java
  /**
   * Test for empty() as value.
   */
  @Test
  public void test3_returnsEmpty(){
    // GIVEN
    ObjectUnderTest out = new ObjectUnderTest(Optional.empty());
    // THEN
    assertThat(out).returnsEmpty(out:getOptString, "Should be true for empty()");
  }
```

May be I am wrong, but for me it looks like the returnsValue method does not honour that the from function returns an optional but the expected value is not an optional. As a result I would expect that line 110 in AbstractObjectAssert should look like:
```java
objects.assertEqual(info, from.apply(actual), Optional.ofNullable(expected));
```

There might be also a need for this functions without description, similar to matches in AbstractAssert. What do you think?

sroeper:
After reading the contribution guideline it becomes clear that your code does not comply to them. You have to provide a test class for every assertion method and you should have also failing tests. Also the tests should be named with the prefix "should_pass_xxx" and "should_fail_xxx". Please read the guideline.

As an example you should add a test method to org.assertj.core.api.object.ObjectAssert_returns_Test.java like this for the overridden returns:
```java
  @Test
  void should_fail_with_given_description() {
    // WHEN
    assertions.returns("Yoga", Jedi::getName, "Jedi name");
    // THEN
    then(assertions.descriptionText()).isEqualTo("Jedi name");
  }
```

scordio:
@sroeper @SickDawn before moving forward, I suggest providing feedback to https://github.com/assertj/assertj-core/issues/2341#issuecomment-1108291520.

sroeper:
Done, see https://github.com/assertj/assertj-core/issues/2341#issuecomment-1108291520.

sroeper:
@SickDawn will you change your pull request according to the decision of the assertj team?
So, only implementing ```returns``` with a description?

joel-costigliola:
@SickDawn late late reply but are you considering addressing the latest comments https://github.com/assertj/assertj/issues/2341#issuecomment-1112051194 ? No pb if that's not the case, I can take over the PR

joel-costigliola:
@SickDawn thanks for the PR, integrated in https://github.com/assertj/assertj/commit/303704ef96cf4078b9ee315739d0cf2a6b5a8b8a (alright we were a bit late :wink: )

PR #2587, opened 2022-04-25
Avoid raw types

scordio:
Improvements for #2532.


scordio:
Rebased on the latest changes of #2532.

PR #2601, opened 2022-05-06
fix: Avoid `InputStream` manipulation when `mark` / `reset` are supported

### LINKED ISSUE #2596: `InputStream` supporting `mark` / `reset` should not be manipulated by assertions
Related to https://github.com/assertj/assertj-core/issues/2490#issuecomment-1035488261, the assertions of `AbstractInputStreamAssert` could avoid manipulating the stream position when the `InputStream` implementation supports [`mark`](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#mark-int-) / [`reset`](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#reset--) operations, tested by [`markSupported`](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#markSupported--).

jessicant:
#### Check List:
* Fixes #2596 
* Unit tests :  NO

This branch is failing tests. The idea for measuring `readLimit` is the `InputStream.available()` but it is causing `AbstractInputStreamAssert.java`'s `hasDigest` method to be unhappy - `Unhandled exception: java.io.IOException`. I am confused since `tryStreamReset` is catching that exception, so I am not sure why it is not bubbling up. I think this issue is probably simple, but any help would be much appreciated :)

### DIFF
--- a/src/main/java/org/assertj/core/api/AbstractInputStreamAssert.java
+++ b/src/main/java/org/assertj/core/api/AbstractInputStreamAssert.java
@@ -120,8 +120,8 @@ public abstract class AbstractInputStreamAssert<SELF extends AbstractInputStream
   /**
    * Verifies that the content of the actual {@code InputStream} is empty.
    * <p>
-   * <b>Warning: this will consume the first byte of the {@code InputStream}.</b>
-   * <p>
+   * <b>Warning: this will consume the first byte of the {@code InputStream} if the underlying implementation does
+   * not support {@link InputStream#markSupported() marking}.</b>
    * Example:
    * <pre><code class='java'> // assertion will pass
    * assertThat(new ByteArrayInputStream(new byte[] {})).isEmpty());
@@ -136,7 +136,7 @@ public abstract class AbstractInputStreamAssert<SELF extends AbstractInputStream
    * @since 3.17.0
    */
   public SELF isEmpty() {
-    inputStreams.assertIsEmpty(info, actual);
+    tryStreamReset(() -> inputStreams.assertIsEmpty(info, actual);
     return myself;
   }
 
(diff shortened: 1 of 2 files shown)


nith2001:
I think I fixed the IOException thing, but we probably have to deal with it no matter what since reset() has a potential IOException that needs to be caught as well, not just the available() function.

nith2001:
Screen Shot 2022-05-07 at 9 05 32 AM
Here are the tests we're failing. Keeping them here so that we have a place to reference them

nith2001:
Ok, turns out we needed to have an assertNotNull(info, actual) statement within the assertHasDigest(). I wonder if we need it for the other assertHasDigest() function variations.

nith2001:
Hi @scordio, I remember under the issue comments you said the following

> the mark/reset pair should be pushed down to InputStreams, being an implementation detail of the assertions

However, @jessicant and I have found including the private helper method for resetting the stream under `AbstractInputStreamAssert.java` rather than under `InputStreams.java` makes it much easier work with, since it allows to insert the `InputStreams.java` functions within the Runnable object and call them accordingly.

Is there a better reason for why this shouldn't happen?

nith2001:
Any updates on what the next steps can be?

scordio:
@nith2001, as mentioned in the other PR, we are now focusing on releasing 3.23.0 and we will come back to new features once done with that.

scordio:
@jessicant @nith2001 sorry for making you wait so long!

The base branch has now evolved so I had in mind to cherry-pick your changes, but I also noticed there is still some work to do to properly update all the `InputStream` assertions.

Would you still be interested in continuing this change? If not, we can take it over and keep you as authors.

nith2001:
Ahh, sorry, I don't think I'll have the time for this, but thank you!

scordio:
Superseded by #3713.

PR #2612, opened 2022-05-13
Only compare comparingOnlyFields if it is set (#2610)

### LINKED ISSUE #2610: Comparing only fields of different classes can pass with different values of the compared fields
#### Summary
When comparing different classes and specifying `comparingOnlyFields` the assertion could pass even if the compared values are different if one of the classes has more fields than the other.

I tested using the code in the main branch (ee52be2c).

#### Example

```java
public class ComparingOnlyTest {
  static class A1 {
    public final int a, b;
    public A1(int a, int b) {
      this.a = a;
      this.b = b;
    }
  }
  static class A2 {
    public final int a;
    public A2(int a) {
      this.a = a;
    }
  }

  @Test
  public void testComparingOnly() {
    A1 a1 = new A1(1, 2);
    A2 a2 = new A2(2);

    // This passes. Switching the places of a1 and a2 causes the assertion to fail as expected
    Assertions.assertThat(a1).usingRecursiveComparison().comparingOnlyFields("a").isEqualTo(a2);
  }
}
```

akutschera:
It looks like the comparedFields list was ignored when putting together the list of "non-ignored-actual-field-names".
I added a test to ensure that the order of objects is irrelevant when comparing unequal objects by field name only.

#### Check List:
* Fixes #2610
* Unit tests : YES
* Javadoc with a code example (on API only) :NA

### DIFF
--- a/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration.java
+++ b/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration.java
@@ -674,7 +674,7 @@ public class RecursiveComparisonConfiguration {
                             .filter(fieldName -> !shouldIgnoreFieldBasedOnFieldLocation(dualValue.fieldLocation.field(fieldName)))
                             .map(fieldName -> dualValueForField(dualValue, fieldName))
                             // evaluate field value ignoring criteria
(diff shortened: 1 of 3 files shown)


joel-costigliola:
Thanks @akutschera, I will review your PR shortly !

joel-costigliola:
Integrated thanks @akutschera!

PR #2623, opened 2022-05-22
Fix containsExactly that does not work properly with maps not using e…

### LINKED ISSUE #2165: `containsExactly` does not work properly with maps not using equals to compare keys

joel-costigliola:
…quals to compare keys.

Fix #2165

# Conflicts:
#	src/main/java/org/assertj/core/internal/Maps.java

### DIFF
--- a/src/main/java/org/assertj/core/internal/Maps.java
+++ b/src/main/java/org/assertj/core/internal/Maps.java
@@ -440,6 +440,10 @@ public class Maps {
     }
   }
 
+  private static boolean isSingletonMap(Map<?, ?> map) {
+    return map.size() == 1;
+  }
+
   private static boolean isMultiValueMapAdapterInstance(Map<?, ?> map) {
     return isInstanceOf(map, "org.springframework.util.MultiValueMapAdapter");
   }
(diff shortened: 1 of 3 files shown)


src/main/java/org/assertj/core/internal/Maps.java
```diff
@@ -559,11 +583,15 @@ public <K, V> void assertContainsExactly(AssertionInfo info, Map<K, V> actual, E
     if (notExpected.isEmpty() && notFound.isEmpty()) {
       // check entries order
       int index = 0;
+      // Create a map with the same type as actual to use the Map built-in comparison, ex: maps string case insensitive keys.
+      Map<K, V> emptyMap = createEmptyMap(actual);
       for (K keyFromActual : actual.keySet()) {
-        if (!deepEquals(keyFromActual, entries[index].getKey())) {
+        emptyMap.put(keyFromActual, null);
```
vlsi, line 589: User provided Map might fail to support null values

scordio:
Memo for merging: @sarajuhosova co-authored #2500.

PR #2624, opened 2022-05-23
Fix typos in Javadoc of ObjectEnumerableAssert

rweisleder:


src/main/java/org/assertj/core/api/ObjectEnumerableAssert.java
```diff
@@ -1397,7 +1397,7 @@
    *                                         character -&gt; assertThat(character.isMortal()).isTrue(),
    *                                         character -&gt; assertThat(character.getName()).isEqualTo("Legolas"));
    *
-   * // you can specify more that one assertion per requirements
+   * // you can specify more tha one assertion per requirement
```
scordio, line 1400:
```suggestion
   * // you can specify more than one assertion per requirement
```

rweisleder: Doh! 🙈 Reminder to myself: Before submitting, check that not only have the correct words been modified, but that the words have been modified correctly.

Thanks for proofreading. 😁

joel-costigliola:
Integrated thanks @rweisleder!

PR #2628, opened 2022-05-26
urlassert

### LINKED ISSUE #2196: Add `isReachable` / `isUnreachable` for URL assertions

weiyilei:
#### Check List:
* Fixes #2196 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/error/uri/ShouldBeReachable.java
+++ b/src/main/java/org/assertj/core/error/uri/ShouldBeReachable.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2022 the original author or authors.
+ */
+package org.assertj.core.error.uri;
+
+import static org.assertj.core.util.Strings.isNullOrEmpty;
+
+import java.net.URI;
+import java.net.URL;
+
+import org.assertj.core.error.BasicErrorMessageFactory;
+import org.assertj.core.error.ErrorMessageFactory;
+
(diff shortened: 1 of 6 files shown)


src/main/java/org/assertj/core/error/uri/ShouldBeReachable.java
```diff
@@ -0,0 +1,62 @@
...
+import java.net.URI;
+import java.net.URL;
+
+import org.assertj.core.error.BasicErrorMessageFactory;
+import org.assertj.core.error.ErrorMessageFactory;
+
+//CS304 Issue link: https://github.com/assertj/assertj-core/issues/2196
+public class ShouldBeReachable extends BasicErrorMessageFactory {
+
+  private static final String SHOULD_BE_REACHABLE = "%nExpecting actual:%n  <%s>%nbe reachable but it's not";
```
ascopes, line 26:
```suggestion
  private static final String SHOULD_BE_REACHABLE = "%nExpecting actual:%n  <%s>%nto be reachable but it's not";
```
Think there might be a missing word here

scordio:
Hi @weiyilei, thanks for raising the PR, and apologies for the extremely slow feedback!

Due to https://github.com/assertj/assertj/issues/2196#issuecomment-3343877472, we decided not to proceed with this feature, so I'm closing this PR.

Thanks again for your interest!

Issue #2424, opened 2021-11-28
Assertj multimodule build


joel-costigliola:
#### Summary

We tests assertj core with integration tests to validate junit 4, testng and osgi integration, this works but is more a workaround to simulate different environments.

A better solution consists in defining a multi module project where junit 4 and testng checks will be done in their own modules which are closer to real project using assertj.

#### Modules:

- assertj-core
- assertj-guava
- assertj-db
- assertj-joda
- assertj-neo4j
- assertj-examples
- assertj-core-junit4-with-opentest4j and remove the corresponding integration tests in assertj-core
- assertj-core-junit4-with-testng and remove the corresponding integration tests in assertj-core
- assertj-core-osgi-tests and remove the corresponding integration tests in assertj-core

#### Github actions

We need to update our CI actions to build the multimodule project.

A commit in `main` should trigger all the module build but the assertj core CI build should only reports errors when any of the following modules build fail: 

- assertj-core module build fails
- assertj-core-junit4-with-opentest4j and remove the corresponding integration tests in assertj-core
- assertj-core-junit4-with-testng and remove the corresponding integration tests in assertj-core
- assertj-core-osgi-tests and remove the corresponding integration tests in assertj-core

The assertj core should not fail if assertions modules fail (ex: assertj-guava), this should be reported by the assertion module build itself.

scordio:
Moving OSGi integration tests to a separate artifact would also address #2347.

scordio:
Proposal for a first milestone:

1. Update the project to be a multi-module:
```
assertj-build
 |--- assertj-core
```
2. Convert `assertj/assertj-parent-pom` to be a submodule ([keeping git history](https://gist.github.com/trongthanh/2779392)) and archive the source repository
```
assertj-build
 |--- assertj-core
 |--- assertj-parent-pom
```
4. Rename `assertj/assertj-core` repository to `assertj/assertj`

joel-costigliola:
I'd like to know that point 3. is :smile: but otherwise I like the MVP like approach.

One thing I'd like the multimodule build to support is a selection of module to build,ex 

- assertj-core
- assertj-examples
- assertj-core-junit4-with-opentest4j and remove the corresponding integration tests in assertj-core
- assertj-core-junit4-with-testng and remove the corresponding integration tests in assertj-core
- assertj-core-osgi-tests and remove the corresponding integration tests in assertj-core

We can achieve that with a shell script if we can't directly with maven.

scordio:
[...]

PR #2634, opened 2022-05-29
Add Error Prone static analysis

vlsi:
Error Prone has good patterns, and it prints good messages.

See https://errorprone.info/bugpatterns

Sample output:

```
[ERROR] /../assertj-core/src/main/java/org/assertj/core/api/AtomicReferenceArrayAssert.java:[2083,29] [CheckReturnValue] Ignored return value of 'usingElementComparator', which is annotated with @CheckReturnValue
[ERROR]     (see https://errorprone.info/bugpattern/CheckReturnValue)
[ERROR] /../assertj-core/src/main/java/org/assertj/core/internal/Paths.java:[438,3] [MethodCanBeStatic] A private method that does not reference the enclosing instance can be static
[ERROR]     (see https://errorprone.info/bugpattern/MethodCanBeStatic)
[ERROR]   Did you mean 'private static List<Path> sortedRecursiveContent(Path path) {'?
```

#### Check List:
* Unit tests : CI

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,6 +22,23 @@ jobs:
       - name: Test
         run: ./mvnw -V --no-transfer-progress -e verify javadoc:javadoc
 
+  errorprone:
+    name: Errorprone
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: add-opens for javac
+        # Error Prone with Java 16+ requires add-exports: https://errorprone.info/docs/installation
+        run: cp .mvn/jvm.errorprone.config .mvn/jvm.config
+      - name: Setup Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: 17
+#          cache: 'maven'
+      - name: Test
+        run: ./mvnw -V --no-transfer-progress -e -Perrorprone -DskipTests package
+
   sonar:
     name: Sonar code analysis
     runs-on: ubuntu-latest
--- a/.mvn/jvm.errorprone.config
+++ b/.mvn/jvm.errorprone.config
@@ -1,0 +1,10 @@
+--add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.model=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.processing=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
+--add-opens jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED
+--add-opens jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED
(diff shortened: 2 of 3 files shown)


vlsi:
Here's CI log in my fork: https://github.com/vlsi/assertj-core/runs/6645822760?check_suite_focus=true#step:5:46

joel-costigliola:
Interesting @vlsi! the team needs to have a look at the rules but looks promising

vlsi:
Of course, I don't suggest enabling and fixing all the bug patterns, however, I believe the errors produced by the default configuration should be fixed.

Even some of the "warnings by default" are worth fixing as well.

For instance:
1) https://github.com/assertj/assertj-core/blob/b44460623b9c0e83c2b311c1fc6b7bffa1a077b9/src/main/java/org/assertj/core/description/TextDescription.java#L54-L56

```
Error:  /../assertj-core/assertj-core/src/main/java/org/assertj/core/description/TextDescription.java:[56,24] [ArrayHashCode] hashcode method on array does not hash array contents
    (see https://errorprone.info/bugpattern/ArrayHashCode)
  Did you mean 'return Objects.hash(value, Arrays.hashCode(args));'?
```

2) https://github.com/assertj/assertj-core/blob/8a70b403490ba5799fc06b90dc8096c729b4da0e/src/main/java/org/assertj/core/internal/Iterables.java#L1315-L1321

```
Error:  /../assertj-core/assertj-core/src/main/java/org/assertj/core/internal/Iterables.java:[1321,31] [ReturnValueIgnored] Return value of 'orElseThrow' must be used
    (see https://errorprone.info/bugpattern/ReturnValueIgnored)
```

it might be questionable, however, I believe `Stream.anyMatch` would indeed be better:

```java
    if (stream(actual).anyMatch(predicate)) {
      throw failures.failure(info, anyElementShouldMatch(actual, predicateDescription));
    }
```

---

I remembered `errorprone` because it can detect `MethodCanBeStatic` (it verifies the method is private, so it does not break backward compatibility).

---

Then, there's https://github.com/palantir/assertj-automation which you might want to apply and promote :)

vlsi:
@joel-costigliola , one more question: for now, I include `errorprone` Maven plugin right into `assertj-core`, however, you might want to add it to the base configuration (https://github.com/assertj/assertj-parent-pom ?)

In my experience, moving assertj/assertj-parent-pom creates too much friction since Maven does not support cross-repository changes. I can't really have a pull request that would change `assertj-parent-pom` and apply the changed `assertj-parent-pom` into `assertj-core`.

We had the exact same issue with https://github.com/pgjdbc/pgjdbc-parent-poms when the PostgreSQL JDBC driver was Maven-based. Moving to Gradle was a huge relief so we could keep all the things in a single repository.

---

Even if you keep multiple repositories (e.g. keep `assertj-core`, `assertj-parent-whatever`, etc), Gradle makes it possible to use tools like https://melix.github.io/includegit-gradle-plugin/ when you can have local testing and even pull requests that cross repository bounds: https://twitter.com/CedricChampeau/status/1499019789033000962

---

Of course, I would yield to the team's decision, however, I strongly believe tools like ErrorProne, [Cherckerframework](https://checkerframework.org/), [AutoStyle](https://github.com/autostyle/autostyle) would improve contributor and developer experience.
At the same time, Maven makes many of those improvements impossible or extremely hard.

scordio:
> I can't really have a pull request that would change `assertj-parent-pom` and apply the changed `assertj-parent-pom` into `assertj-core`.
> 
> We had the exact same issue with https://github.com/pgjdbc/pgjdbc-parent-poms when the PostgreSQL JDBC driver was Maven-based. Moving to Gradle was a huge relief so we could keep all the things in a single repository.

Personally, I don't see how this is related to Maven. What you're mentioning is a multi-module setup, something we already plan to tackle in #2424. Even if we would switch to Gradle, most likely we would target the same setup.

I agree Gradle would allow multi-repository development but IMHO the fact that it's possible doesn't necessarily mean it's the direction we should take.

PR #2638, opened 2022-06-01
[ISSUE 2629] Improve readability of containsExactly

### LINKED ISSUE #2629: Display Improve readability of containsExactly when the order of elements does not match
#### Feature summary

```java
assertThat(Arrays.asList(1, 2, 3, 4)).containsExactly(1, 3, 2, 4);
```

yields the following error message:

```
java.lang.AssertionError: 
Actual and expected have the same elements but not in the same order, at index 1 actual element was:
  2
whereas expected element was:
  3
```

Note that `1` and `4` are not displayed in the error message.

**I suggest** that AssertJ should print the full expected and actual set of elements in case the order does not match.

For instance, when the value does not match, then AssertJ does print all the elements:

```java
assertThat(Arrays.asList(1, 2, 3, 4)).containsExactly(1, 2, 4, 4);
```

```
java.lang.AssertionError: 
Expecting actual:
  [1, 2, 3, 4]
  [1, 2, 4, 4]
  [4]
and others were not expected:
  [3]
```

#### Example

The following methods seems to be affected:

```java
org.assertj.core.api.AbstractIterableAssert#containsExactly(ELEMENT... values)
org.assertj.core.api.AbstractIterableAssert#containsExactlyElementsOf(Iterable<? extends ELEMENT> iterable)
```

Giovds:
Removes `elementsDifferAtIndex()` from `assertContainsExactly()` and re-use
`shouldContainExactlyWithDiffAssertionError()` in order to show the actual
and expected sets entirely. Making the `differencesFound()` check unnecessary.

It also introduces a new s`houldContainExactly()` and checks for both
the missing and unexpected diffs to be empty. Otherwise cases where the
actual and expected only differ in order, result into logging of missing
elements - which there are none.

#### Check List:
* Fixes #2629 
* Unit tests : YES

### DIFF
--- a/src/main/java/org/assertj/core/internal/Iterables.java
+++ b/src/main/java/org/assertj/core/internal/Iterables.java
@@ -1138,10 +1138,7 @@ public class Iterables {
       // if the objects are not equal, begin the error handling process
       if (!areEqual(actualAsList.get(i), values[i])) {
         IterableDiff<Object> diff = diff(actualAsList, asList(values), comparisonStrategy);
-        if (diff.differencesFound()) {
-          throw shouldContainExactlyWithDiffAssertionError(diff, actual, values, info);
-        }
-        throw failures.failure(info, elementsDifferAtIndex(actualAsList.get(i), values[i], i, comparisonStrategy));
+        throw shouldContainExactlyWithDiffAssertionError(diff, actual, values, info);
       }
     }
   }
--- a/src/main/java/org/assertj/core/internal/Iterables.java
+++ b/src/main/java/org/assertj/core/internal/Iterables.java
@@ -39,7 +39,7 @@ import static org.assertj.core.error.ShouldBeNullOrEmpty.shouldBeNullOrEmpty;
 import static org.assertj.core.error.ShouldBeSubsetOf.shouldBeSubsetOf;
 import static org.assertj.core.error.ShouldContain.shouldContain;
 import static org.assertj.core.error.ShouldContainAnyOf.shouldContainAnyOf;
-import static org.assertj.core.error.ShouldContainExactly.elementsDifferAtIndex;
+import static org.assertj.core.error.ShouldContainExactly.shouldContainExactlyWithIndexes;
 import static org.assertj.core.error.ShouldContainExactly.shouldContainExactly;
 import static org.assertj.core.error.ShouldContainExactlyInAnyOrder.shouldContainExactlyInAnyOrder;
 import static org.assertj.core.error.ShouldContainNull.shouldContainNull;
@@ -104,6 +104,7 @@ import java.util.function.Predicate;
 
 import org.assertj.core.api.AssertionInfo;
 import org.assertj.core.api.Condition;
+import org.assertj.core.configuration.Configuration;
 import org.assertj.core.error.UnsatisfiedRequirement;
 import org.assertj.core.error.ZippedElementsShouldSatisfy.ZipSatisfyError;
 import org.assertj.core.presentation.PredicateDescription;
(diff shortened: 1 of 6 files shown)


src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1137,10 +1137,7 @@ public void assertContainsExactly(AssertionInfo info, Iterable<?> actual, Object
       // if the objects are not equal, begin the error handling process
       if (!areEqual(actualAsList.get(i), values[i])) {
         IterableDiff<Object> diff = diff(actualAsList, asList(values), comparisonStrategy);
-        if (diff.differencesFound()) {
-          throw shouldContainExactlyWithDiffAssertionError(diff, actual, values, info);
-        }
-        throw failures.failure(info, elementsDifferAtIndex(actualAsList.get(i), values[i], i, comparisonStrategy));
+        throw shouldContainExactlyWithDiffAssertionError(diff, actual, values, info);
```
vlsi, line 1140: Does it make sense to include the first index of the diff elements into the message as well?

src/test/java/org/assertj/core/error/ShouldContainExactly_create_Test.java
```diff
@@ -30,6 +30,44 @@ class ShouldContainExactly_create_Test {
...
+
+    // WHEN
+    final String message = factory.create(new TextDescription("Test"));
+
+    // THEN
+    then(message).isEqualTo(format("[Test] %n"
+                                   + "Expecting actual:%n"
+                                   + "  [\"Yoda\", \"Han\", \"Luke\", \"Anakin\"]%n"
+                                   + "to contain exactly (and in same order):%n"
+                                   + "  [\"Yoda\", \"Luke\", \"Han\", \"Anakin\"]%n"));
```
joel-costigliola, line 47: I think what would be more helpful here in addition to this description is to have the list of mismatches, each mismatch showing the index, the actual element value and the expected element value.

On the example above we could add
```
but there were differences at these indexes:
- index 1: actual element "Han" but expected was "Luke"
- index 2: actual element "Luke" but expected was "Han"
```
or
On the example above we could add
```
but there were differences at these indexes:
- element at index 1: expected "Han" but was "Luke"
- element at index 2: expected "Luke" but was "Han"
```

joel-costigliola (changes requested):

src/test/java/org/assertj/core/error/ShouldContainExactly_create_Test.java
joel-costigliola: I prefer option 2, WDYT?

vlsi: I'm afraid the list of mismatches could be huge (e.g. both lists like 1000 elements and they are in the reverse order), so there should be a hard limit on the number of printed mismatches

vlsi: Option 3:

```diff
@@ -1,4 +1,4 @@
 Yoda
-Han
 Luke
+Han
 Anakin
```

vlsi: > Wouldn't almost all options be big and perhaps unreadable if the lists contain lots of elements?

Well, if the order of elements is completely different, then, I believe, any output that mentions all differences would be too verbose.

That is why I think it might be OK to show only the first (or the first N<=10) differences.
If the number of actual diffs is small, then all of them will be visible. If the number of diffs is big, then the first will be printed, and the user would understand "everything is different".

> without color differences to support the - and + signs it might be difficult to spot the differences

The idea behind + and - is to use unified diff. I moved it to a separate issue https://github.com/assertj/assertj-core/issues/2641

> I'd think the creator of the unit test should consider smaller data sets to prove the functionality is working

Databases support up to ~1000 columns, and I have functional tests that verify if the system is workable with that number of columns. I would still use regular assertions even for collections up to 100'000 elements.
Here's a recent case for a functional test with a collection having 65535 elements: https://github.com/pgjdbc/pgjdbc/pull/2525

joel-costigliola: Let's leave the length error aside (might be tackled in a different issue) and focus the error to produce when the compared iterables have the same number of elements.

I see two kind of errors from here:

1. elements don't match ignoring the order, there might be unexpected / missing elements in actual
2. elements match but the order is not right

For 1. we could show the diff as done in `should_display_full_expected_and_actual_sets_with_missing_and_unexpected_elements` and for 2. show the index mismatches, limiting them for brevity them number (could be configured with Configuration.maxElementsForPrinting which is set to 1000 be default).

joel-costigliola: 1000 is too much, I'm happy if we go with the first 10 and if we have more indicate in the error message that we only display the first 10

Giovds: I was playing around with the amount of indexes to display yesterday and the `MAX_ELEMENTS_FOR_PRINTING` is a lot indeed. The first 10 to 50 seems reasonable to me. It could be useful feedback to display something along the lines:
```text
but there were differences at these indexes:
  element at index 0: expected "Luke" but was "Han"
  ...
  element at index 9: expected "Anakin" but was "Leia"
Only showing the first 10 mismatches 
```
WDYT?

vlsi: What output would you suggest for the following call `assertThat(Arrays.asList(10, 1, 2, 3, 4, 5, 6, 7, 8, 9)).containsExactly(1, 2, 3, 4, 5, 6, 7, 8, 9, 10);` ?
I guess displaying all the indices as mismatching won't be that helpful.

joel-costigliola: @Giovds LGTM, maybe we should move `Only showing the first 10 mismatches` before showing the indexes to be sure users will see it (I'm fine with it at the end too, it's just a suggestion)
```
but there were differences at these indexes (only showing the first 10 mismatches):
  element at index 0: expected "Luke" but was "Han"
  ...
  element at index 9: expected "Anakin" but was "Leia"
```
If we have less than 10 mismatches, no need to mention `only showing the first 10 mismatches` 

this is a bit of a contrived example @vlsi, I feel the proposed error message is still helpful enough in this case.

joel-costigliola: I'm also happy to increase to 50 mismatches by default

vlsi: Why do you mean the example is contrieved? If the use-case is to assert elements in a strict order, then it would likely that failure cases would be "almost matching" like I show here. There was almost exact match, except the last element is misplaced

Giovds:
I'm not sure about this approach, but I'd like to hear your thoughts on it!

src/test/java/org/assertj/core/error/ShouldContainExactly_create_Test.java
```diff
@@ -14,22 +14,97 @@
 
 import static java.lang.String.format;
 import static org.assertj.core.api.BDDAssertions.then;
-import static org.assertj.core.error.ShouldContainExactly.elementsDifferAtIndex;
-import static org.assertj.core.error.ShouldContainExactly.shouldContainExactly;
+import static org.assertj.core.error.ShouldContainExactly.*;
```
Giovds, line 17: I'll probably undo this change and import just the things required.

joel-costigliola:
@Giovds is your PR ready for a review?

Giovds:
> @Giovds is your PR ready for a review?

@joel-costigliola Yes, that would be great!

joel-costigliola:
Integrated in 41179e555463021737468f71785ecca5aac152bb thanks @Giovds!

PR #2642, opened 2022-06-03
Runtime require opentest4j in JPMS

ascopes:
Change Opentest4J dependency to be non-static in module-info.java.

Since we rely on OpenTest4J for any assertion error, we should
include it as a runtime dependency to prevent the user needing
to also explicitly include that dependency.

This also ensures `requires transitive` propagates this dependency
automatically.

#### Check List
* Unit tests : N/A.
* Javadoc with a code example (on API only) : N/A.
* PR meets the [contributing guidelines](https://github.com/assertj/assertj-core/blob/main/CONTRIBUTING.md) - simple change - should be covered.

### DIFF
--- a/src/main/java9/module-info.java
+++ b/src/main/java9/module-info.java
@@ -43,7 +43,7 @@ module org.assertj.core {
   requires static net.bytebuddy;
   requires static org.hamcrest;
   requires static org.junit.jupiter.api;
-  requires static org.opentest4j; // to throw AssertionFailedError which is IDE friendly
+  requires org.opentest4j; // to throw AssertionFailedError which is IDE friendly
 
   // Services loaded by org.assertj.core.configuration.ConfigurationProvider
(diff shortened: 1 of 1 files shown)


src/main/java9/module-info.java
```diff
@@ -43,7 +43,7 @@
   requires static net.bytebuddy;
```
ascopes, line 43: Do we want to also apply this to net.bytebuddy? Since I believe this is needed for the assumption mechanism to work correctly.

```suggestion
  requires net.bytebuddy;
```

joel-costigliola:
opentest4j is not required (or at least it should not), I have looked for use of it in the main code and found in two places: `SoftAssertionsProvider` and `SoftAssertions` but the import is only used in the javadoc which we can remove.

ascopes:
Ahh, I was confusing the exception that provides the representation with this! Thanks for clarifying!

PR #2644, opened 2022-06-03
Swap out deprecated-for-removal methods in BDDAssertions_then_Test.java

ascopes:
The Character::new constructor has been deprecated for removal since Java 9.
All usages are directed to use Character#valueOf instead, which holds the same API
signature. This will prevent future compilation issues on newer JDKs when the
method is totally removed from OpenJDK.


joel-costigliola:
thanks @ascopes, good catch.
I would not be surprised if we have similar issues elsewhere

ascopes:
@joel-costigliola no problem. I couldn't spot other issues but if I can find any, I'll put a bulk PR up for them when I get time!

Issue #2639, opened 2022-06-02
Reflective assertions for Class Loading


ascopes:
#### Feature summary

I would like to be able to perform assertions on classes that I load dynamically from a given classloader. This would enable me to test code that deals with custom classloaders using assertj without needing to implement my own custom boilerplate to do this.

#### Example

```java
// Given
var mySpecialClassLoader = new MySpecialClassLoader(...);

// Then
assertThat(mySpecialClassLoader)
    .hasClass("org.example.MyClass1")
    .hasResource("...")
    .class("org.example.MyClass2")
    ... // returns a ClassAssert
```

scordio:
If I got it correctly, your proposal is to create assertions for `java.lang.ClassLoader`.

I think it would be a nice addition. Would you like to work on it?

As it would be a new assertion hierarchy (`AbstractClassLoaderAssert` and `ClassLoaderAssert`), I suggest doing that in two steps, with separate PRs:

1. Addition of `AbstractClassLoaderAssert` and `ClassLoaderAssert`, using `AbstractAssert` as parent of the former, together with the addition of the `assertThat` / `then` / `assumeThat` / `given` factory methods in the related classes.
2. Addition of assertions methods to `AbstractClassLoaderAssert`.

A few points:
* In general, we like having assertion names close to the JDK methods used behind (as you did already in the example), e.g., `hasClass` for `java.lang.ClassLoader::loadClass`, `hasResource` for `java.lang.ClassLoader::getResource`, etc. feel free to get in touch with us in case of any doubts about naming
* Navigational methods should have no prefixes as done already in the example, but should also return abstract types for extendability, e.g., class("org.example.MyClass2") having `AbstractClassAssert` as return type
* Constructors for the new `Assert` classes should be `protected`
* Avoid defining new classes in the internal layer (`org.assertj.core.internal`)
  * Favor private methods in `AbstractClassLoaderAssert`, or
  * Favor new package-private classes and methods in the `org.assertj.core.api` package in case something really has to be shared
* `OptionalAssert` / `AbstractOptionalAssert` might be a good reference for the implementation

ascopes:
Cool! Will try and take a look when I get some free time then. 

Thanks!

ascopes:
@scordio quick query.

ClassLoader has the method `loadClass` which returns a `Class<?>` object. What is the preferred navigable method name for this (since `... class(String binaryName)` wouldn't compile as it uses a keyword).

Issue #2573, opened 2022-04-16
Cannot link to assertj-core JavaDoc due to JPMS.


ascopes:
#### Summary

The JavaDocs for AssertJ appear to be being built under Java 8, which means the Java-9 `module-info.class` descriptor is not being taken into account, resulting in a JavaDoc being generated that behaves as if it were defined in the unnamed module (i.e. with no module name prefix on the path).

When linking another project to these JavaDocs for a JPMS-enabled project, `maven-javadoc-plugin` appears to be unable to link correctly.

#### Example

Under Maven 3.8.5, maven-javadoc-plugin 3.3.2, assertj-core 3.22.0, java-release.version 11:

```xml
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-javadoc-plugin</artifactId>
  <version>${maven-javadoc-plugin.version}</version>
  
  <configuration>
    <additionalOptions>
      <option>--override-methods=detail</option>
      <option>-version</option>
    </additionalOptions>
    <source>${java-release.version}</source>
    <links>
      <link>https://javadoc.io/doc/org.assertj/assertj-core/${assertj.version}</link>
    </links>
  </configuration>
  
  <executions>
    <execution>
      <id>default-javadoc</id>
      <phase>package</phase>
      <goals>
        <goal>jar</goal>
      </goals>
    </execution>
  </executions>
</plugin>
```

Running under JDK 11, JDK 12, JDK 13, and JDK 14; Maven will fail with the following error:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.3.2:jar (default-javadoc) on project java-compiler-testing: MavenReportException: Error while generating Javadoc: 
[ERROR] Exit code: 1 - javadoc: error - The code being documented uses modules but the packages defined in https://javadoc.io/doc/org.assertj/assertj-core/3.22.0/ are in the unnamed module.
[ERROR] 
[ERROR] Command line was: /opt/hostedtoolcache/Java_Zulu_jdk/12.0.2-3/x64/bin/javadoc @options @packages @argfile
[ERROR] 
[ERROR] Refer to the generated Javadoc files in '/home/runner/work/java-compiler-testing/java-compiler-testing/java-compiler-testing/target/apidocs' dir.
[ERROR] 
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[...]

PR #2645, opened 2022-06-03
GH-2639: Implement skeleton for ClassLoaderAssert

ascopes:
First PR for #2639.

- Implement skeleton type for AbstractClassLoaderAssert.
- Implement ClassLoaderAssert implementation.
- Hook up with Assertions, Assumptions, BDDAssertions, BDDAssumptions,
  BDDSoftAssertionsProvider, InstanceOfAssertFactories,
  StandardSoftAssertionsProvider, WithAssertions, and WithAssumptions.
- Write ClassLoaderAssert-bound tests for Assertions, Assumptions,
  BDDAssertions, BDDAssumptions, InstanceOfAssertFactories,
  WithAssertions.
- AssertionsUtil#expectAssumptionNotMetException now returns
  an AssumptionViolatedException rather than void, as to match with
  the API for AssertionsUtil#expectAssertionError. Since nothing
  relied on the return value before as this was a void method, this
  should not break anything existing.

I will add code examples to the second PR once a solid API has
been agreed on, as it prevents missing anything and creating
erroneous documentation.


ascopes:
I am out for the rest of today and part of tomorrow, so I will likely continue to look at this on Sunday at some point

scordio:
Sure, take your time @ascopes, and thank you!

ascopes:
Added a commit at the start that addresses a transitive issue with the Future tests on the Windows runner. I will cherry-pick this onto a separate branch so it can be reviewed separately first.

scordio:
Thanks @ascopes, I'll tackle this one as soon as I'm done with #2573

scordio:
This is now merged. Thanks, @ascopes!

I applied some changes like adding the version tags and some missing `assertThat` entry points.

Also, I reverted the changes for `expectAssumptionNotMetException` - the idea is not wrong, but after some GIVEN / WHEN / THEN refactoring the change was no longer needed and I preferred to avoid another style of testing for assumptions, an area that is already quite heterogeneous.

PR #2648, opened 2022-06-04
Work around/suppress around 95 warnings produced on JDK-17 with -Xlint:all

ascopes:
- Work around/suppress around 95 warnings produced on JDK-17 with -Xlint:all flag enabled.
- Fix a couple of warnings dotted around in tests to do with unchecked operations
- Fix incorrect javadoc reference in Floats_AssertLessThanOrEqualTo_test

Some suppressions are being marked as being redundant by IntelliJ when they are not. For these, I have added an IntelliJ-specific suppression about redundant suppressions to prevent a risk of the original suppression being accidentally removed in the future due to the false positive.

Xlint:all was run by adding 

```xml
<compilerArgs>
  <arg>-Xlint:all</arg>
</compilerArgs>
```

...to the `<configuration />` block for maven-compiler-plugin in the POM.

### DIFF
--- a/src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
+++ b/src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
@@ -89,7 +89,9 @@ public interface InstanceOfAssertFactories {
    *
    * @see #PREDICATE
    */
-  @SuppressWarnings({ "rawtypes", "unused" }) // rawtypes: using Class instance, unused: parameter needed for type inference
+  @SuppressWarnings({"rawtypes", "unused", "unchecked", "RedundantSuppression"})
+  // rawtypes+unchecked: using Class instance, unused: parameter needed for type inference.
+  // IntelliJ can warn that this is redundant when it is not.
   static <T> InstanceOfAssertFactory<Predicate, PredicateAssert<T>> predicate(Class<T> type) {
     return new InstanceOfAssertFactory<>(Predicate.class, Assertions::<T> assertThat);
   }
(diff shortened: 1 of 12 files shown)


joel-costigliola:
Integrated thanks @ascopes!

src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
```diff
@@ -89,7 +89,9 @@ public interface InstanceOfAssertFactories {
    *
    * @see #PREDICATE
    */
-  @SuppressWarnings({ "rawtypes", "unused" }) // rawtypes: using Class instance, unused: parameter needed for type inference
+  @SuppressWarnings({"rawtypes", "unused", "unchecked", "RedundantSuppression"})
+  // rawtypes+unchecked: using Class instance, unused: parameter needed for type inference.
+  // IntelliJ can warn that this is redundant when it is not.
```
scordio, line 94: @ascopes do you happen to remember if this was happening when opening AssertJ code in IntelliJ, or when using these factories in client code?

At least, it seems that the former doesn't happen anymore.

ascopes: This was quite a long time ago. I can't remember seeing it more recently though so they probably have fixed it since!

scordio: Great and thanks a lot for the quick feedback!

ascopes: No problem. Just so happened to pop up on my phone while I was doing something!

PR #2649, opened 2022-06-05
Fix spurious Future tests failure on Windows runners

ascopes:
Addresses an issue where FutureAssert tests in CI would sometimes fail on the Windows runner due to kernel thread sleep calls.

I have increased the timeout for that call, and also implemented a custom ThreadFactory for the abstract base class which will ensure that:

1. the threads in the executor are daemon threads
2. the threads in the executor are named appropriately to assist in debugging in the future (e.g. through Java Mission Control, JMX, JConsole, or similar)
3. all uncaught exceptions get reported to the logs
4. the threads emulate PrivilegedThreadFactory in that they hold a reference to the parent thread context class loader (no need to run this in an access control block for tests).
5. the threads run with the highest priority where possible.

### DIFF
--- a/src/test/java/org/assertj/core/api/future/AbstractFutureTest.java
+++ b/src/test/java/org/assertj/core/api/future/AbstractFutureTest.java
@@ -12,23 +12,63 @@
  */
 package org.assertj.core.api.future;
 
+import static java.lang.String.format;
+
+import java.lang.Thread.UncaughtExceptionHandler;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.platform.commons.logging.Logger;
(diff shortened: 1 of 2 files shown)


joel-costigliola:
That's super awesome @ascopes, we were looking into it without really finding the issue (I work with linux so hard to investigate windows issues)

ascopes:
@joel-costigliola no worries! I use Linux too. My assumption when I saw it was that the GitHub runners for Windows are a bit sluggish due to how they are timesliced. I'd noticed that one of my repos that has to test under every JDK for Windows and Linux was about 2x as slow on the Windows runners as on Linux. Guessing the OS isn't waking up the JVM threads quickly enough from sleep which was causing the assertion to fail. Using a longer duration seems to fix it.

If it happens again, it might be worth looking into whether there is a way to mock the clock that is being used for the timing operations instead. Alternatively it may be useful to write a `@Flake` annotation that acts a bit like `@RepeatedTest` does for JUnit5 which retries the test several times.

Failing that, it may be easier to just have a test case where the future is immediately complete, and one where it is never complete, and check that the former succeeds and the latter fails. This would probably be fine since it is mostly relying on JRE internals to compare these times anyway. Not ideal, admittedly.

Anyway, hopefully that fixes it!

PR #2662, opened 2022-06-16
feat: use randomized matrix for CI verifications for better test coverage

### LINKED ISSUE #2660: Consider randomized matrix for GitHub Actions testing
#### Feature summary

@scordio , what do you think of testing **various** Java vendors for better test coverage?
I believe it would help both assertj (e.g. avoid accidental dependency on OpenJDK+HotSpot implementation), Java vendors (you'll test their implementations and hopefully report bugs), and the Java community (everybody wins from better AssertJ and better Java).

Here's a script that automatically generates random subset of the parameters for CI testing: https://github.com/vlsi/github-actions-random-matrix

Here's how it looks like in pgjdbc project: https://github.com/pgjdbc/pgjdbc/runs/6876908328?check_suite_focus=true
It enables to capture bugs like:
* Bugs in timezone-related code (matrix can shuffle timezones)
* Bugs in OS-dependent code (matrix can shuffle operating systems)
* Accidental dependency on `Obejct#hashCode` uniqueness
* Accidental dependency on HotSpot JIT implementation (matrix can add "randomize JIT decisions" options to surface issues like "wrong usage of volatile")
* Combinations of the above!

At the same time, the CI duration is easy to configure: you can specify the number of jobs you need, so you don't end up with "PR test takes hours".
At the same time, you can hard-code "known interesting" cases. For instance, you can ask matrix to always include "lowest supported Java", "always include a test with `Object#hashCode` returning the same value for all objects", etc.
At the same time, if you happen to run into an obscure failure (e.g. CI failure in Pacific/Chatham timezone on Windows), then you can test it with a couple of lines. 

For instance, in pgjdbc we have a flaky test that is related to replication. I've submitted a small PR that basically requests the matrix to produce "only replication-related test": https://github.com/pgjdbc/pgjdbc/pull/2545
Note that we test against 13 PostgreSQL versions, and some of the do not support replication. The matrix handles all that just fine. The user says "I want replication only", and the matrix obeys and produces a variety of matching cases.

In case you wonder, the total number of valid permutations for pgjdbc is ~1'000'000.

#### Example

See https://github.com/pgjdbc/pgjdbc/runs/6876908328?check_suite_focus=true
See https://github.com/cbeust/testng/runs/6720757608?check_suite_focus=true

vlsi:
fixes #2660

Known issues:
- [x] javac 17, 18, 19, 20 fail to compile assertj-core with `-XX:hashCode=2`, so I excluded that parameter for now. It should be re-introduced once assertj splits "build jar" from "run tests" into different JVMs. Here's the OpenJDK issue: https://bugs.openjdk.org/browse/JDK-8288590
- [x] https://github.com/assertj/assertj-core/issues/2663
- [ ] https://github.com/assertj/assertj-core/issues/2666
Sample failures:

```
Expecting actual:
  ["Manu", "Magic"]
  ["MANU", "MAGIC"]
  ["MAGIC"]
and others were not expected:
  ["Magic"]
when comparing values using CaseInsensitiveStringComparator
	at org.assertj.core.error.AssertionErrorCreator.assertionError(AssertionErrorCreator.java:93)
	at org.assertj.core.internal.Failures.failure(Failures.java:126)
	at org.assertj.core.internal.Iterables.shouldContainExactlyWithDiffAssertionError(Iterables.java:1150)
	at org.assertj.core.internal.Iterables.assertContainsExactly(Iterables.java:1141)
	at org.assertj.core.api.AbstractIterableAssert.containsExactlyForProxy(AbstractIterableAssert.java:426)
	at org.assertj.core.api.ListAssert$ByteBuddy$RpsLqv5u.containsExactlyForProxy$accessor$VeEh08EE(Unknown Source)
	at org.assertj.core.api.ListAssert$ByteBuddy$RpsLqv5u$AssertJ$SoftProxies$ExMVMBlJ.call(Unknown Source)
	at org.assertj.core.api.ErrorCollector.intercept(ErrorCollector.java:56)
	at org.assertj.core.api.ListAssert$ByteBuddy$RpsLqv5u.containsExactlyForProxy(Unknown Source)
	at org.assertj.core.api.ListAssert$ByteBuddy$RpsLqv5u.containsExactlyForProxy(Unknown Source)
	...(71 remaining lines not displayed - this can be changed with Assertions.setMaxStackTraceElementsDisplayed)]
```

```
Error:    Assertions_assertThat_with_Stream_startsWith_Test.should_pass_if_actual_and_sequence_are_equal_according_to_custom_comparison_strategy:150 
Expecting actual:
  "Stream under test"
to start with:
  ["LUKE", "lEIA"]
when comparing values using CaseInsensitiveStringComparator
```

```
Error:    Strings_assertIsXmlEqualCase_Test.should_pass_if_both_Strings_are_XML_equals_case_insensitively:74 
expected: "<rss version="2.0">
    <channel>
        <TITLE>JAVA Tutorials</TITLE>
    </channel>
</rss>
"
 but was: "<rss version="2.0">
    <Channel>
        <title>Java Tutorials</title>
    </Channel>
</rss>
"
when comparing values using CaseInsensitiveStringComparator
```


vlsi:
The solution is to replace `s1.toLowerCase().compareTo(s2.toLowerCase())` with `s1.compareToIgnoreCase(s2)` in https://github.com/assertj/assertj-core/blob/main/src/test/java/org/assertj/core/util/CaseInsensitiveStringComparator.java#L26

ascopes:
It might also be beneficial to consider enabling random test execution in surefire. It won't necessarily improve coverage, but it will help identify potentially flaky tests!

vlsi:
@scordio , @joel-costigliola , would you please review the PR and conclude if you accept or reject it?

I remember you suggested raising a PR: https://github.com/assertj/assertj/issues/2660#issuecomment-1156982075

joel-costigliola:
it's low priority for us at the moment, so we'll leave that open for now

PR #2667, opened 2022-06-18
[CI fix] Allow add-label to fail in binary-compatibility.yml

ascopes:
Binary-compatibility is currently failing when attempting to add a label to 
the resource behind this integration when the person running the CI does not have
permission to manipulate that resource.

This allows that error to be ignored.

### DIFF
--- a/.github/workflows/binary-compatibility.yml
+++ b/.github/workflows/binary-compatibility.yml
@@ -35,6 +35,9 @@ jobs:
       - name: Add label
         if: failure()
         uses: actions/github-script@v6
+        # This will occur when the user running the CI job lacks permission to manipulate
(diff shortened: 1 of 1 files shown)


joel-costigliola:
Integrated thanks @ascopes!

scordio:
@ascopes under which PR did you get the failure? I'm wondering if it's related to the default token permissions and if it's something we could fix.

If this doesn't work with PR raised from fork branches, it kills a bit the purpose of this label.

ascopes:
@scordio it is under the WIP PR for GH-2369 on the most recent build

scordio:
Thanks, will look into it.

PR #2668, opened 2022-06-19
Run cross-version CI on JRE 8 and 11 to verify compatibility

ascopes:
Just a suggestion that I thought might be beneficial for CI builds:
considering running CI builds on JRE 8 and JRE 11 explicitly.

This reduces the risk of missing differences between older JREs and
JRE 17 which CI currently runs on. The main risk by avoiding this
is coming across bugs that were fixed in newer JRE versions.

One of the things this can impact is the differences in default
classloaders in newer versions.

### DIFF
--- a/.github/workflows/cross-version.yml
+++ b/.github/workflows/cross-version.yml
@@ -4,7 +4,24 @@ on: [push, pull_request]
 
 jobs:
 
-  test_java:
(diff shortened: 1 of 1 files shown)


ascopes:
Closed since this causes additional issues with `--release` under JRE 8

PR #2670, opened 2022-06-19
Reset global state more aggressively in tests

ascopes:
Noticed when I run the tests in my IDE, sometimes there are cases where loads of tests are failing.

image

Seems to be related to the global state being changed, so I have added an annotation that can be placed on any test cases that manipulate the global state. This annotation will ensure the test runs in global isolation to prevent race conditions, and it will enforce resetting the global state after each test and at the end of the test execution for the class (to handle cases such as `@AfterAll` which run outside the scope of regular test cases).

I have also made Surefire now execute tests in complete random order, which should assist in identifying any other places this kind of thing is happening in

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -372,6 +372,7 @@
             <exclude>org/assertj/core/osgi/**</exclude>
             <exclude>org/assertj/core/internal/objects/Objects_assertHasOnlyFields_Test*</exclude>
           </excludes>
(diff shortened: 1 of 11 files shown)


joel-costigliola:
Good stuff @ascopes, thanks heaps!

ascopes:
@joel-costigliola no worries!

PR #2686, opened 2022-06-25
Fix incomplete test for given(Duration) in BDDAssumptionsTest

ascopes:
Just a quick one while I remember. Spotted a BDD assumption test that was not doing anything.


joel-costigliola:
Good catch, thanks

PR #2690, opened 2022-06-28
Fix locale handling for ignoring case assertions

### LINKED ISSUE #2666: assertContainsIgnoringCase fails to compare i and I in tr_TR locale
See `org.assertj.core.internal.Strings#assertContainsIgnoringCase`

https://github.com/assertj/assertj-core/blob/9051a958e6ab0a750bb243060aef57001ab97e6e/src/main/java/org/assertj/core/internal/Strings.java#L528-L531

I would suggest adding https://github.com/policeman-tools/forbidden-apis verification to just ban `toLowerCase()`, `toUpperCase()` and other unsafe methods: https://github.com/assertj/assertj-core/issues/2664

### LINKED ISSUE #2688: Assertj must include commons-lang3 copyright notice to comply with its license
**Describe the bug**

https://github.com/assertj/assertj-core/pull/2685 introduced code borrowed from `org.apache.commons.lang3`, so you must include NOTICE in both source and binary re-distributions of assertj (see https://github.com/apache/commons-lang/blob/master/NOTICE.txt, https://github.com/apache/commons-lang/blob/master/LICENSE.txt#L37-L38).

scordio:
Fixes #2666.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/internal/Strings.java
+++ b/assertj-core/src/main/java/org/assertj/core/internal/Strings.java
@@ -16,6 +16,7 @@ import static java.lang.Character.isDigit;
 import static java.lang.Character.isWhitespace;
 import static java.lang.String.format;
 import static java.util.Arrays.stream;
+import static java.util.Locale.ROOT;
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.toCollection;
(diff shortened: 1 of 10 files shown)


scordio:
Keeping it as a draft until it's confirmed how the licensing aspects for `commons-lang` should be covered.

scordio:
`commons-lang` code is no longer used.

PR #2692, opened 2022-06-29
Add building block to create custom precise assertions

alex859:
#### Check List:
* Unit tests : YES

The goal is to be able to make precise soft assertions on complex objects (i.e. where we don't need to assert the equality of the full object) and to have easy to read test code and assertion error messages.

Example:

```java
assertThat(customer).is(
    customer(
        name(
            first("Boris"),
            last("Johnson")),
            address(
                firstLine("10, Downing Street"),
                postcode("SW2A 2AA")
         )
    )
)
``` 

and on failure get an error message like:

```
[✗] customer:[
   [✗] name:[
      [✓] first: Boris,
      [✗] last: Johnson but was Johnstone
   ],
   [✗] address:[
      [✓] first line: 10, Downing Street,
      [✗] postcode: SW2A 2AA but was SW1A 2AA
   ]
```

I tried using custom soft assertions, but I found the error message to be less readable (and the implementation more complex).
Maybe there is already another way to achieve this?

Also `NestableCondition` is not a great name :relaxed:.

Looking forward to getting your feedback.

Thanks!

### DIFF
--- a/src/main/java/org/assertj/core/condition/NestableCondition.java
+++ b/src/main/java/org/assertj/core/condition/NestableCondition.java
@@ -1,0 +1,185 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2022 the original author or authors.
+ */
+
+package org.assertj.core.condition;
+
+import static java.util.stream.Collectors.toList;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+import org.assertj.core.api.Condition;
+import org.assertj.core.description.Description;
+
+/**
+ * Building block to define a precise soft assertion about a complex object.
+ * It allows to create readable assertions and produces beautiful assertion error messages
+ * <p>
+ *   <pre><code class='java'>
+ *   Example:
+ *   class Customer {
+ *      final String name;
+ *      final Address address;
+ *
+ *      Customer(String name, Address address) {
+ *        this.name = name;
+ *        this.address = address;
+ *      }
+ *   }
+ *
+ *   class Address {
+ *      final String firstLine;
+ *      final String postcode;
+ *
+ *      Address(String firstLine, String postcode) {
+ *        this.firstLine = firstLine;
+ *        this.postcode = postcode;
+ *      }
+ *    }
+ *
+ *   static Condition{@literal <Customer>} name(String expected) {
+ *     return new Condition<>(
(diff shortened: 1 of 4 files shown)


joel-costigliola:
Thanks @alex859 , I will review your PR in the next couple weeks!

vlsi:
This assertion API looks lika Hamcrest, and it is known for its poor discoverability.
In other words, developer who writes `customer(` has no possibility to figure out the available continuations.

Have you considered using lambdas instead of nesting function results? Even though Java's nested lambdas are a bit verbose, they are discoverable.

E.g.
```
assertThat(customer).is(c -> {
    c.name.is(n -> {
        n.first.isEqualTo("Boris") // this allows to use full AssertJ power
        n.last("Jonson") // it might be helpful for literals
    })
})
```

With Kotlin, it would become

```
assertThat(customer).`is` { // "is" is reserved, so alternative is required
    name.`is` {
        first.isEqualTo("Boris") // this allows to use full AssertJ power
        last("Jonson") // it might be helpful for literals
    }
}
```

The key difference from the proposal in PR is that users get autocomplete after they type `c.`

joel-costigliola:
Integrated thanks @alex859!

alex859:
Thanks @joel-costigliola!

alex859:
@vlsi Thanks for your comment! I'm not sure how the first Java example could be implemented, what type is `c.name`?

Kotlin devs can define a DSL to build nestable conditions that are discoverable.
Something like:

```kotlin
@DslMarker
annotation class CustomerDsl

@CustomerDsl
class CustomerConditions {
    private val conditions = mutableListOf<Condition<Customer>>()

    fun name(nameConditions: NameConditions.() -> Unit) {
        conditions += NameConditions().apply(nameConditions).build()
    }

    fun build(): Condition<Customer> =
        nestable("customer", *conditions.toTypedArray())
}

@CustomerDsl
class NameConditions {
    private val conditions = mutableListOf<Condition<Name>>()

    fun first(expected: String) {
        conditions += verboseCondition(
            { expected == it.first },
            "first: $expected",
            { " but was ${it.first}" }
        )
    }

    fun last(expected: String) {
        conditions += verboseCondition(
            { expected == it.last },
            "last: $expected",
            { " but was ${it.last}" }
        )
    }

    fun build(): Condition<Customer> =
        nestable("name", { it.name }, *conditions.toTypedArray())
}
```
And then use `invoke` as you suggested


```kotlin
operator fun ObjectAssert<Customer>.invoke(
    conditions: CustomerConditions.() -> Unit
): ObjectAssert<Customer> =
    `is`(CustomerConditions().apply(conditions).build())
```
to get:

```kotlin
(assertThat(customer)) {
    name {
        first("Boris")
        last("Unknown")
    }
}
```

vlsi:
>Kotlin devs can define a DSL to build nestable conditions that are discoverable.

@alex859 , I guess the same (or even better DSL) can be created without resorting to `Condition` and `Nestable`.
Already existing `assertSoftly` is more-or-less fine for creating DSL like you suggested, and the implementation would not be limited by the small amount of existing `Condition` implementations.

alex859:
@vlsi the main advantage of the proposed approach is that on failure the error message is more readable.

vlsi:
Technically speaking, "address" string should come together with function that extracts the address. Otherwise it is just wrong to link "address" and allOf.
There might be anyOf, noneOf, and so on, so labelling allOf with "address" is wrong.

If the functions are labelled properly, then the library can produce nice messages.

PR #2735, opened 2022-08-12
Custom formatters now supports any CharSequence.

### LINKED ISSUE #2734: Support for GString in `Assertions.registerFormatterForType`
When registering custom formatter in Groovy and using the interpolated string we are getting `ClassCastException` on [StandarardRepresentation:311](https://github.com/assertj/assertj/blob/main/assertj-core/src/main/java/org/assertj/core/presentation/StandardRepresentation.java#L311) because we are trying to cast `GStringImpl` to `String`.

Here is the simple Spock spec to demonstrate this:

```groovy
import org.assertj.core.api.Assertions
import spock.lang.Specification

class SomeSpec extends Specification {
    def setupSpec() {
        Assertions.registerFormatterForType(Long) { "\$$it\$" }
    }

    def "should fail"() {
        expect:
        Assertions.assertThat(1L).isEqualTo(2L)
    }
}
```

There are two straightforward workarounds for this issue:

- Explicitly invoking `toString` on interpolated string like: `{ "\$$it\$".toString() }`
- Using plain concatenation: `{'$' + it + '$'}`

If you interested in this change, I would like to contribute the patch.

nstdio:
#### Check List:
* Fixes #2734
* Unit tests : YES

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,7 +10,7 @@ insert_final_newline = true
 charset = utf-8
 
 # 2 space indentation for java, xml and yml files
-[*.{java,xml,yml,sh}]
+[*.{java,groovy,xml,yml,sh}]
 indent_style = space
 indent_size = 2
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,7 +10,7 @@ insert_final_newline = true
 charset = utf-8
 
 # 2 space indentation for java, xml and yml files
-[*.{java,groovy,xml,yml,sh}]
+[*.{java,xml,yml,sh}]
 indent_style = space
 indent_size = 2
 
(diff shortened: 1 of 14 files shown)


scordio:
Due to the fact that Groovy doesn't yet support Java 20, we should move Groovy tests to a separate Maven module under [assertj-tests/assertj-integration-tests](https://github.com/assertj/assertj/tree/main/assertj-tests/assertj-integration-tests) and enable that module via a profile enabled according to a JDK condition.

The module could be named `assertj-core-groovy-tests`.

In case you don't have time to deal with such a refactoring, we'll take care of it during the merge.

scordio:
Merged. Thank you, @nstdio!

PR #2737, opened 2022-08-12
Move integration tests into separate modules

scordio:
Step 1 of https://github.com/assertj/assertj/issues/2424#issuecomment-1202145883


scordio:
OSGi will be moved in a separate PR.

PR #1778, opened 2020-02-06
Add "it" infrastructure for integration tests

sormuras:
This PR adds an optional [maven-invoker-plugin](https://maven.apache.org/plugins/maven-invoker-plugin/) based profile to the build configuration that builds a set of Maven projects located in the newly added `"src/it"` directory.

No dedicated verification script is evaluated after such an `"it"` project build for the time being. A successful build is considered as successful test case.

PR #2759, opened 2022-08-31
Add integration test infrastructure for the module path

scordio:
Related to #2424, closes #1778.


scordio:
IDEA doesn't seem to like the setup, but Maven works fine.

Raised [IDEA-301032](https://youtrack.jetbrains.com/issue/IDEA-301032).

scordio:
Superseded by #3771.

PR #2772, opened 2022-09-08
style: fix Bad Smells in org.assertj.core.internal.RealNumbers

MartinWitt:
# Repairing Code Style Issues
## Non-Protected-Constructor-in-Abstract-Class

A non-protected constructor in an abstract class is not needed because only subclasses can be instantiated
## Changes: 
* Constructor `org.assertj.core.internal.RealNumbers()` is now protected instead of public

* Constructor `org.assertj.core.internal.RealNumbers(org.assertj.core.internal.ComparisonStrategy)` is now protected instead of public

#### Check List:
* Unit tests :  NA


MartinWitt:
Reading the failure message, I have to ask if this is a bug in your test code.
```
Expecting actual:
  23:59:30.415119537Z
to be strictly before:
  00:01:10.427289267Z
  ```
 It seems like you compare two dates generated while running the CI and if the CI start short 0 UTC, the second date is generated after 0 UTC and the test fails?

scordio:
Integrated. Thanks, @MartinWitt!

Issue #2760, opened 2022-08-31
`extractingResultOf` fails on the module path


scordio:
## Describe the bug

* AssertJ Core version: 3.23.1
* Java version: 11.0.16+8 Temurin
* Test framework version:

From https://github.com/assertj/assertj/issues/1504#issuecomment-1230743158 by @beatngu13 and @aepfli:

> We are facing similar issues since we test on the module path, see:
> 
> https://github.com/junit-pioneer/junit-pioneer/runs/8060473637?check_suite_focus=true#step:4:68
> 
> This is one of the failing tests:
> 
> https://github.com/junit-pioneer/junit-pioneer/blob/main/src/test/java/org/junitpioneer/internal/PioneerAnnotationUtilsTests.java#L516
> 
> Looks like the problematic path comes from `extractingResultOf`, @aepfli made it work switching to `extracting`:
> 
> https://github.com/junit-pioneer/junit-pioneer/pull/614/files#diff-169ff4622006e6406c530aec2c75d78a8b07f70493494a260e9fa40784a71cdb
> 
> We are using AssertJ 3.22.0. Note that the changes are not on `main`, this is the corresponding PR:
> 
> https://github.com/junit-pioneer/junit-pioneer/pull/651

## Test case reproducing the bug

Test case at: https://github.com/junit-pioneer/junit-pioneer/blob/main/src/test/java/org/junitpioneer/internal/PioneerAnnotationUtilsTests.java#L516

which yields:

```
> Task :test
WARNING: module-info.class ignored in patch: /home/runner/work/junit-pioneer/junit-pioneer/build/classes/java/main
WARNING: module-info.class ignored in patch: /home/runner/work/junit-pioneer/junit-pioneer/build/classes/java/main

Pioneer Annotation utilities > getting annotated annotations > finds annotated annotations on a method FAILED
    java.lang.IllegalStateException: java.lang.IllegalAccessException: class org.assertj.core.util.introspection.MethodSupport (in module org.assertj.core) cannot access class com.sun.proxy.jdk.proxy2.$Proxy30 (in module jdk.proxy2) because module jdk.proxy2 does not export com.sun.proxy.jdk.proxy2 to module org.assertj.core
        at org.assertj.core@3.23.1/org.assertj.core.util.introspection.MethodSupport.invokeMethod(MethodSupport.java:60)
        at org.assertj.core@3.23.1/org.assertj.core.util.introspection.MethodSupport.methodResultFor(MethodSupport.java:53)
[...]

Issue #2347, opened 2021-09-22
Test JAR should not be deployed to Central during releases


scordio:
c9807b9c9a300758ee262ea5026acd3be68781cf added the test JAR to the `maven-jar-plugin` configuration for the OSGi integration tests. As a side effect, the test JAR was deployed starting from version 3.17.2.

We should prevent the test JAR from being deployed during releases.

scordio:
Fixed by #2742.

PR #2773, opened 2022-09-09
Draft: Testing Module Testing with Maven for #2760

aepfli:
Sorry i simply started experimenting out of curiosity regarding #2760 , and I want to share what I tried and what worked and did not work for such modular tests. Please just close this pr, if you do not think it is fitting. Although if you think, this needs just polishing, let me know, i can adapt and help to make it better.

This pr works, but only has one test - so theoretically it is usable :) 

BUT... i wanted to experiment a little bit more, and maybe make the original unit tests work within this modular module. 

## test-jar approach

I tried adding following to the assertj-core to generate a test jar 
```
<plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-jar-plugin</artifactId>
        <executions>
          <execution>
            <goals>
              <goal>test-jar</goal>
            </goals>
          </execution>
        </executions>
      </plugin>
``` 
and tried to apply it as a dependency within my modular test. Sadly this ended up with a lot of package errors
<details>
  <summary>errors</summary>

```
[ERROR] the unnamed module reads package org.assertj.core.api from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.api.filter from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.api.iterable from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.api.junit.jupiter from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.api.recursive.comparison from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.condition from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.configuration from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.data from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.description from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.error from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.error.future from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.error.uri from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.extractor from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.groups from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.matcher from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.presentation from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.util from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.util.diff from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.util.diff.myers from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.util.introspection from both assertj.core and org.assertj.core
[ERROR] the unnamed module reads package org.assertj.core.util.xml from both assertj.core and org.assertj.core
[ERROR] module assertj.core reads package org.assertj.core.util from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.util.introspection from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.util.diff from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.util.diff.myers from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.util.xml from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.groups from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.configuration from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.condition from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.description from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.extractor from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.api from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.api.filter from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.api.junit.jupiter from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.api.iterable from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.api.recursive.comparison from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.matcher from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.data from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.error from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.error.future from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.error.uri from both org.assertj.core and assertj.core
[ERROR] module assertj.core reads package org.assertj.core.presentation from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.util from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.util.introspection from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.util.diff from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.util.diff.myers from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.util.xml from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.groups from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.configuration from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.condition from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.description from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.extractor from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.api from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.api.filter from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.api.junit.jupiter from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.api.iterable from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.api.recursive.comparison from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.matcher from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.data from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.error from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.error.future from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.error.uri from both org.assertj.core and assertj.core
[ERROR] module org.hamcrest reads package org.assertj.core.presentation from both org.assertj.core and assertj.core
[INFO] 63 errors 
[INFO] -------------------------------------------------------------
```
</details>

## additional source directory

Then i thought maybe extracting them into `target/test-classes` could also do the trick - sadly not
Then i thought maybe i could simply add the test source directory to the modularTest module, but this somehow needed more than 5 minutes and did not finish, so i also stepped away from this approach.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/BDDAssumptions.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/BDDAssumptions.java
@@ -3019,7 +3019,7 @@ public final class BDDAssumptions extends Assumptions {
    * Executed test:
    * <pre><code class='java'> {@literal @Test}
    * public void given_the_assumption_is_met_the_test_is_executed() {
-   *   given(new File("file.ext").toPath()).isRelative();
+   *   given(ResourceUtil.getResource("file.ext")).isRelative();
    *   // the remaining code is executed
    *   // ...
    * }</code></pre>
@@ -3027,7 +3027,7 @@ public final class BDDAssumptions extends Assumptions {
    * Skipped test:
    * <pre><code class='java'> {@literal @Test}
    * public void given_the_assumption_is_not_met_the_test_is_skipped() {
-   *   given(new File("file.ext").toPath()).isAbsolute();
+   *   given(ResourceUtil.getResource("file.ext")).isAbsolute();
    *   // the remaining code is NOT executed.
    *   // ...
    *}</code></pre>
--- a/assertj-core/src/test/java/org/assertj/core/api/Assertions_linesOf_Test.java
+++ b/assertj-core/src/test/java/org/assertj/core/api/Assertions_linesOf_Test.java
@@ -22,6 +22,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
 
+import org.assertj.core.util.ResourceUtil;
 import org.junit.jupiter.api.Test;
 
 class Assertions_linesOf_Test {
@@ -30,14 +31,14 @@ class Assertions_linesOf_Test {
 
   @Test
   void should_read_lines_of_file_with_UTF8_charset() {
-    File file = new File("src/test/resources/utf8.txt");
+    File file = ResourceUtil.getResource("utf8.txt").toFile();
     assertThat(linesOf(file, "UTF-8")).isEqualTo(EXPECTED_CONTENT);
     assertThat(linesOf(file, StandardCharsets.UTF_8)).isEqualTo(EXPECTED_CONTENT);
   }
 
   @Test
   void should_read_lines_of_path_with_UTF8_charset() {
-    Path path = Paths.get("src", "test", "resources", "utf8.txt");
+    Path path = ResourceUtil.getResource("utf8.txt");
     assertThat(linesOf(path, "UTF-8")).isEqualTo(EXPECTED_CONTENT);
     assertThat(linesOf(path, StandardCharsets.UTF_8)).isEqualTo(EXPECTED_CONTENT);
   }
--- a/assertj-core/src/test/java/org/assertj/core/api/BDDSoftAssertionsTest.java
+++ b/assertj-core/src/test/java/org/assertj/core/api/BDDSoftAssertionsTest.java
@@ -99,6 +99,7 @@ import org.assertj.core.test.Animal;
 import org.assertj.core.test.CartoonCharacter;
 import org.assertj.core.test.Name;
 import org.assertj.core.test.Person;
+import org.assertj.core.util.ResourceUtil;
 import org.assertj.core.util.VisibleForTesting;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
@@ -1877,7 +1878,7 @@ class BDDSoftAssertionsTest extends BaseAssertionsTest {
   @Test
   void path_soft_assertions_should_report_errors_on_methods_that_switch_the_object_under_test() {
     // GIVEN
-    Path path = new File("src/test/resources/actual_file.txt").toPath();
+    Path path = ResourceUtil.getResource("actual_file.txt");
     // WHEN
     softly.then(path)
           .overridingErrorMessage("error message")
@@ -1900,7 +1901,7 @@ class BDDSoftAssertionsTest extends BaseAssertionsTest {
   @Test
   void file_soft_assertions_should_report_errors_on_methods_that_switch_the_object_under_test() {
     // GIVEN
-    File file = new File("src/test/resources/actual_file.txt");
+    File file = ResourceUtil.getResource("actual_file.txt").toFile();
     // WHEN
     softly.then(file)
           .overridingErrorMessage("error message")
@@ -1923,7 +1924,7 @@ class BDDSoftAssertionsTest extends BaseAssertionsTest {
   @Test
   void file_soft_assertions_should_work_with_navigation_methods() {
     // GIVEN
-    File file = new File("src/test/resources/actual_file.txt");
+    File file = ResourceUtil.getResource("actual_file.txt").toFile();
     // WHEN
     softly.then(file)
           .overridingErrorMessage("error message")
--- a/assertj-core/src/test/java/org/assertj/core/api/EntryPointAssertions_contentOf_Test.java
+++ b/assertj-core/src/test/java/org/assertj/core/api/EntryPointAssertions_contentOf_Test.java
@@ -22,6 +22,7 @@ import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.stream.Stream;
 
+import org.assertj.core.util.ResourceUtil;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
@@ -33,7 +34,7 @@ class EntryPointAssertions_contentOf_Test extends EntryPointAssertionsBaseTest {
   @MethodSource("fileContentOfWithCharsetFunctions")
   void should_read_file_content_with_charset(BiFunction<File, Charset, String> contentOfWithCharsetFunction) {
     // GIVEN
-    File sampleFile = new File("src/test/resources/utf8.txt");
+    File sampleFile = ResourceUtil.getResource("utf8.txt").toFile();
     // WHEN
     String content = contentOfWithCharsetFunction.apply(sampleFile, UTF_8);
     // THEN
@@ -48,7 +49,7 @@ class EntryPointAssertions_contentOf_Test extends EntryPointAssertionsBaseTest {
   @MethodSource("fileContentOfWithCharsetAsStringFunctions")
   void should_read_file_content_with_charset_as_string(BiFunction<File, String, String> contentOfWithCharsetFunction) {
     // GIVEN
-    File sampleFile = new File("src/test/resources/utf8.txt");
+    File sampleFile = ResourceUtil.getResource("utf8.txt").toFile();
     // WHEN
     String content = contentOfWithCharsetFunction.apply(sampleFile, "UTF8");
     // THEN
@@ -63,7 +64,7 @@ class EntryPointAssertions_contentOf_Test extends EntryPointAssertionsBaseTest {
   @MethodSource("fileContentOfWithDefaultCharsetFunctions")
   void should_read_file_content_with_default_charset(Function<File, String> contentOfWithDefaultCharsetFunction) {
     // GIVEN
-    File sampleFile = new File("src/test/resources/ascii.txt");
+    File sampleFile = ResourceUtil.getResource("ascii.txt").toFile();
     // WHEN
     String content = contentOfWithDefaultCharsetFunction.apply(sampleFile);
     // THEN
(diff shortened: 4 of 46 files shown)


scordio:
Hi Simon, thanks for the PR! This probably overlaps with #2759, which is already stable from the Maven perspective but:
* Has an issue with IntelliJ, see [IDEA-301032](https://youtrack.jetbrains.com/issue/IDEA-301032)
* Lacks a test case for #2760 and the Pioneer issue - I'm still looking at your branch and thinking how to get it, maybe I'll "steal" the whole `PioneerAnnotationUtils` as a test target

scordio:
> make the original unit tests work within this modular module.

This is a cool point I didn't consider yet: it would be much better to re-run our entire test suite in the modular world instead of adding dedicated tests for it.

aepfli:
> Hi Simon, thanks for the PR! This probably overlaps with #2759

yes it does, sorry I did not see that one :)

> Lacks a test case for https://github.com/assertj/assertj/issues/2760 and the Pioneer issue - I'm still looking at your branch and thinking how to reproduce the issue, maybe I'll steal the whole PioneerAnnotationUtils as a test target

I added one specially for this case within this pr.

> This is a cool point I didn't consider yet: it would be much better to re-run our entire test suite in the modular world instead of adding dedicated tests for it.

maybe somebody has suggestions for this error `the unnamed module reads package org.assertj.core.api from both assertj.core and org.assertj.core` and how we can easily fix it, but i have no clue who i can ask here. This error only applies as soon as i add the test-jar as a dependency. Most likely because it does not have a module-info.java but on the other hand, i also did not find a way to produce it with a module-info.java . As i am hooked i will experiment a little bit more. and see what i might discover.

scordio:
> maybe somebody has suggestions for this error `the unnamed module reads package org.assertj.core.api from both assertj.core and org.assertj.core` and how we can easily fix it, but i have no clue who i can ask here.

Are you getting this error in the IDE or with Maven?

I'll try to play with it as well this evening.

aepfli:
I pushed the changes to use a `test-jar`. i get this error with maven during `testCompile` of the modular-testing module

scordio (commented):
I'm still checking what could be done in this area. Below are just a few comments but feel free to ignore them until it's clear where we want to go.

About these errors:

```
the unnamed module reads package org.assertj.core.api from both assertj.core and org.assertj.core
```

I believe this is caused by these three actors:

1. The `assertj-core` multi-release JAR with this [`module-info`](https://github.com/assertj/assertj/blob/main/assertj-core/src/main/java9/module-info.java)
2. The `assertj-core` test JAR that has no module descriptor so it becomes an automatic module
3. The `assertj-core-moduletest` with the `module-info` under `src/test/java`

I assume the unnamed module in the error message is 3, while `assertj.core` corresponds to 2 and `org.assertj.core` corresponds to 1.

I believe the `assertj-core` test JAR should be created with a descriptor, but again, right now I believe that having a test JAR would be a backup option (#2347).

I'm trying another strategy with the branch of #2759:
* Setting the `project.build.testSourceDirectory` to `../../../assertj-core/src/test/java`
* Defining an additional execution of the `maven-compiler-plugin` where the `module-info` under `src/test/java` gets compiled

This should work when running the `maven-surefire-plugin` under the `assertj-core-module-path` module, I just need to fix the necessary declarations in the `module-info`.

Also, it requires a bit of duplication for all the test scoped dependencies of `assertj-core` and also the `assertj-core/src/test/java` folder is detected by IntelliJ as a test source root for two separate modules.

It might be OK but I still need to digest the idea 🙂

aepfli:
I tried the path you want to use, I did not manage to get it to compile. Not sure why but test compile took ages ;)

I even abused the approach from the mutli release jar and patched the jar with a module-info out of the Modulartest module. But as said I had some kind of compile issues, as it tool ages and I was too impatient to wait more than 5 minutes. 

I will also try to experiment some more in the evening. Looks really interesting.

> Setting the project.build.testSourceDirectory to ../../../assertj-core/src/test/java

Be aware to also set the test resources ;)

scordio:
I've pushed my WIP changes in e13846080b8974c18969167c9c1ec9ef1de3745e, I'll try to finalize them later in the day.

aepfli:
i used part of your code, and I figured out it is just hamcrest - by adding the optional junit dependency and excluding hamcrest, I made it work locally :) with your testdirectory approach. i will check later again, but it looks promising

aepfli:
please do not be overwhelmed by my recent changes

I discovered an issue with all files which have been accessed during the test, as the path was always module specific and not loaded via resources, i adapted the tests to do so (most likely there is a good reason for that, i just thought a fast replace in the IDE could do the trick to see which errors are left)

now i am left with 6 errors, and one of those is my test for those methods. strangely enough there has not been a lot of other detections regarding package violations, except the deepdifference tests.

```
[ERROR] Errors: 
[ERROR]   RecursiveAssertionDriver_JavaClassLibraryRecursionTest.should_assert_over_and_recurse_into_jcl_classes_when_configured_to_recurse_into_JCL:45 » Introspection
[ERROR]   DeepDifference_Test.testBigDecimals:80->assertHaveNoDifferences:327->assertHaveNoDifferences:331 » Introspection
[ERROR]   DeepDifference_Test.testEquivalentCollections:249->assertHaveNoDifferences:327->assertHaveNoDifferences:331 » Introspection
[ERROR]   DeepDifference_Test.testSameObject:66->assertHaveNoDifferences:327->assertHaveNoDifferences:331 » Introspection
[ERROR]   DeepDifference_Test.testUnorderedCollection:140->assertHaveNoDifferences:327->assertHaveNoDifferences:331 » Introspection
[ERROR]   IterableAssert_extracting_Test.should_allow_assertions_on_property_values_extracted_from_given_iterable:36 » IllegalState
[INFO] 
[ERROR] Tests run: 18673, Failures: 0, Errors: 6, Skipped: 16
```

I know, those changes are now a lot and blew up the whole PR, and they might not be desired, this was again just exploration and fact finding. So please if you do not see those fitting the need, ignore them :), but we would need to find another solution for the tests and the resources.

scordio:
Thank you so much for taking the time to do this already!

I'm using your attempts as inspiration and I'm also trying to take a slightly different direction to avoid the weird IDE behavior I was mentioning: instead of setting `testSourceDirectory` and `testResources`, I keep the configuration in the corresponding plugins so the IDE doesn't get influenced (066cc49bb000d902745b726e1bf566cbdf703c30).

I'm not sure how much time I'll have in the next days but I'll continue on this topic as soon as I can.

aepfli:
if I can support you somehow further, or if there is anything I should test/do, etc. please let me know :) I am glad if I can help you.

aepfli:
just something which popped to my head, instead of recompiling the tests why not just reuse the compiled files?

```
            <plugin>
                <artifactId>maven-resources-plugin</artifactId>
                <groupId>org.apache.maven.plugins</groupId>
                <executions>
                    <execution>
                        <id>copy-resource-one</id>
                        <phase>generate-test-resources</phase>
                        <goals>
                            <goal>copy-resources</goal>
                        </goals>

                        <configuration>
                            <outputDirectory>${project.build.testOutputDirectory}</outputDirectory>
                            <resources>
                                <resource>
                                    <directory>../../assertj-core/target/test-classes</directory>
                                    <includes>
                                        <include>**/*</include>
                                    </includes>
                                </resource>
                            </resources>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
```

this way the test is executed faster, and it should actually not cause any problems?

aepfli:
also an interesting twitter thread regarding this topic: https://twitter.com/lukaseder/status/1389236184472301583

scordio:
> also an interesting twitter thread regarding this topic: https://twitter.com/lukaseder/status/1389236184472301583

Thanks! I think I've seen it last year and I agree especially with:

> Instead, there can be proper test library modules. Modules that are only used by other test modules, and will never be shipped to users.
>
> Perhaps, it's time to let go of the "old" Maven directory layouts?

which is somehow the strategy I'm trying to use with #2424.

It might be we'll go with the approach of moving all the tests to a separate module. I need to digest the idea a bit more 🙂

scordio:
One more thought: I'm not really thrilled to move tests to a separate module-based artifact until [IDEA-301032](https://youtrack.jetbrains.com/issue/IDEA-301032) is fixed, otherwise it will be an unpleasant experience for contributors using IDEA (myself included 😄).

scordio:
Superseded by #3771.

PR #2776, opened 2022-09-12
Improving null-safety of isEqualToNormalizingNewlines

### LINKED ISSUE #2775: Improve null-safety of isEqualToNormalizingNewlines
AssertJ version: 3.23.1
Java version: 17

When comparing strings with isEqualToNormalizingNewlines, the handling of `null` can be improved.

### Example 1

```java
@Test
void compare_null_and_null() {
    String actual = null;
    String expected = null;
    assertThat(actual).isEqualToNormalizingNewlines(expected);
}
```

I would expect this test to succeed since `null == null` and `assertThat(actual).isEqualTo(expected)` succeeds. Instead, it throws a NullPointerException:
```
java.lang.NullPointerException: Cannot invoke "java.lang.CharSequence.toString()" because "actual" is null

	at org.assertj.core.internal.Strings.normalizeNewlines(Strings.java:681)
	at org.assertj.core.internal.Strings.assertIsEqualToNormalizingNewlines(Strings.java:673)
	at org.assertj.core.api.AbstractCharSequenceAssert.isEqualToNormalizingNewlines(AbstractCharSequenceAssert.java:1692)
	at de.rweisleder.DemoTest.compare_null_and_null(DemoTest.java:13)
```

### Example 2

```java
@Test
void compare_null_and_empty() {
    String actual = null;
    String expected = "";
    assertThat(actual).isEqualToNormalizingNewlines(expected);
}
```

I would expect this test to fail with `AssertionError: Expecting actual not to be null`. Instead, it throws a NullPointerException same as above.

### Example 3

```java
@Test
void compare_empty_and_null() {
    String actual = "";
    String expected = null;
    assertThat(actual).isEqualToNormalizingNewlines(expected);
}
```

I would expect this test to fail with something like `AssertionError: expected: null, but was: ""`. Instead it throws the same NullPointerException:
```
java.lang.NullPointerException: Cannot invoke "java.lang.CharSequence.toString()" because "actual" is null

	at org.assertj.core.internal.Strings.normalizeNewlines(Strings.java:681)
	at org.assertj.core.internal.Strings.assertIsEqualToNormalizingNewlines(Strings.java:674)
	at org.assertj.core.api.AbstractCharSequenceAssert.isEqualToNormalizingNewlines(AbstractCharSequenceAssert.java:1692)
	at de.rweisleder.DemoTest.compare_empty_and_null(DemoTest.java:27)
```

Notice the misleading "because **"actual"** is null" while in fact **expected** is null. (due to Javas improved NullPointerException in Java 14+)

Nacho321:
Improving the null-safety in isEqualToNormalizingNewlines to avoid NPE errors when actual and/or expected are null. 

#### Check List:
* Fixes #2775
* Unit tests : YES

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/internal/Strings.java
+++ b/assertj-core/src/main/java/org/assertj/core/internal/Strings.java
@@ -59,6 +59,7 @@ import static org.assertj.core.error.ShouldNotBeEmpty.shouldNotBeEmpty;
 import static org.assertj.core.error.ShouldNotBeEqualIgnoringCase.shouldNotBeEqualIgnoringCase;
 import static org.assertj.core.error.ShouldNotBeEqualIgnoringWhitespace.shouldNotBeEqualIgnoringWhitespace;
 import static org.assertj.core.error.ShouldNotBeEqualNormalizingWhitespace.shouldNotBeEqualNormalizingWhitespace;
+import static org.assertj.core.error.ShouldNotBeNull.shouldNotBeNull;
 import static org.assertj.core.error.ShouldNotContainAnyWhitespaces.shouldNotContainAnyWhitespaces;
 import static org.assertj.core.error.ShouldNotContainCharSequence.shouldNotContain;
 import static org.assertj.core.error.ShouldNotContainCharSequence.shouldNotContainIgnoringCase;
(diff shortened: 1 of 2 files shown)


marcela-cardona-s:
Hello, I was reviewing the proposed solution and found that the cyclomatic complexity was increased from 3 to 14, so I propose a more optimal solution, keeping the complexity at 3. You can use the "CodeMetrics" plugin in IntelliJ Idea to validate this value. I leave you the [commit](https://github.com/marcela-cardona-s/assertj/commit/eb3bb586c4af9d20c3dea9458ea0f00d554ad193) of my repository where I propose the solution.

joel-costigliola:
Thanks @Nacho321 and @marcela-cardona-s, I have integrated this PR wiht both of you as authors.

PR #2790, opened 2022-09-27
test: add usingRecursiveComparison tests

dsankouski:
Note, I created those tests, when investigated, why my RecursiveComparisonConfiguration with ignoreFieldsMatchingRegexes doesn't work. My regex was `"\\w*[iI]d"`, which doesn't work on lists, because fields named like `list[0].field`


joel-costigliola:
Integrated thanks @dsankouski ! (sorry for the long delay)

PR #2802, opened 2022-10-09
Return cache in `setup-java` action

drakulavich:
Returns cache from `actions/setup-java`.

According to https://github.com/actions/setup-java/issues/305 should be fixed.


joel-costigliola:
Integrated thanks @drakulavich!

PR #2803, opened 2022-10-10
Fix test results summary

drakulavich:
#### Check List:
* Updates `EnricoMi/publish-unit-test-result-action` to v2;
* Fixes files pattern to publish JUnit test results
* Updates workflows to prevent double execution from PR (as `push` and `pull-request`)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,9 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [ main ]
+  pull_request:
 
 jobs:
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,12 +39,12 @@ jobs:
 
       # Add test results to the PR.
       - name: Publish unit test results
-        uses: EnricoMi/publish-unit-test-result-action@v1
+        uses: EnricoMi/publish-unit-test-result-action@v2
         if: ${{ matrix.comment-test-results }}
         continue-on-error: true
         with:
           check_name: "JUnit test results"
-          files: "target/surefire-reports/TEST-*.xml"
+          junit_files: "assertj*/**/target/surefire-reports/TEST-*.xml"
 
   sonar:
     name: Sonar code analysis
--- a/.github/workflows/cross-version.yml
+++ b/.github/workflows/cross-version.yml
@@ -1,6 +1,9 @@
 name: Cross-Version
 
-on: [push, pull_request]
+on:
+  push:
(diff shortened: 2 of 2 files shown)


drakulavich:
image

Fix for that summary ↑

joel-costigliola:
That one was bothering but I did have the time to look into, thanks @drakulavich!

.github/workflows/main.yml
```diff
@@ -1,6 +1,9 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [ main ]
```
scordio, line 5: This way the workflow doesn't run for branches unless there is a pull request for it. Is there a reason why you disabled that?

joel-costigliola: merged too quickly I guess :sweat_smile:

scordio: 🚨👮‍♂️🚨😄 

It's anyway a small adjustment, I'll take care of it.

drakulavich: Sorry to break the flow for local branches. I intended to skip multiple executions when you start working inside the fork's PR. Before the change each commit triggers the workflows two times.

If you need the experiments with local branches I suggest adding some additional pattern like:
```yml
on:
  push:
    branches:
      - main
      - local-pattern-here-*
```

drakulavich:
```

  ┌────────────────────────────┐       ┌─────────────────────────────┐
  │   Push feature branch      ├──────►│   Create Pull Request       │
  └─────────────┬──────────────┘       └──────────────┬──────────────┘
                │                                     │
                │                                     │
                │                                     │
 ┌──────────────▼──────────────────┐  ┌───────────────▼────────────┐
 │   Push event triggers workflow  │  │ PR event triggers workflow │
 └─────────────────────────────────┘  └────────────────────────────┘

```
I mean just pipeline execution, the tests summary step will be executed in each pipeline separately.

scordio: This is true only for branches of this repo but not for forks.

Check [this PR](https://github.com/assertj/assertj/pull/2794/checks) for example, `CI` and `Cross-Version` are running only for `pull-request` events, leading to a total of 12 checks.

However, if you check a [dependabot PR](https://github.com/assertj/assertj/pull/2787/checks), `CI` and `Cross-Version` are running for both `push` and `pull-request` events, leading to a total of 22 checks.

scordio: Let's properly address this topic in #2810.

PR #2813, opened 2022-10-15
Use getParameterCount rather than getParameterTypes().length

snuyanzin:
The issue with `java.lang.reflect.Constructor#getParameterTypes()` is that each time it creates a new array by `clone` method.
It does not make sense to do it in order to only get number of parameter types

#### Check List:
* Unit tests : NO


scordio:
Good catch and thanks for your first contribution, @snuyanzin!

PR #2823, opened 2022-10-27
Deprecate `catchThrowableOfType(ThrowingCallable, Class)` in favor of `catchThrowableOfType(Class, ThrowingCallable)`

### LINKED ISSUE #2821: Make the assertion lambdas Kotlin friendly
#### Feature summary
Some assertions accept a functor:
```java
public static <THROWABLE extends Throwable> THROWABLE catchThrowableOfType(ThrowingCallable shouldRaiseThrowable,
                                                                             Class<THROWABLE> type) {
    return AssertionsForClassTypes.catchThrowableOfType(shouldRaiseThrowable, type);
  }
```

Kotlin has a syntax for lambdas, so it could call:
```kotlin
catchThrowableOfType(NotFoundException::class.java) { 
    connectionService.findBy(hubId = "hub2", connId = "conn1") 
}
```
But as it is, it needs to stick with the less readable:
```kotlin
catchThrowableOfType( { 
    connectionService.findBy(hubId = "hub2", connId = "conn1") 
}, NotFoundException::class.java)
```
Could you please add some overrides, moving the functors to the last place? Like,

```java
public static <THROWABLE extends Throwable> THROWABLE catchThrowableOfType(
         Class<THROWABLE> type, ThrowingCallable shouldRaiseThrowable) {
    return AssertionsForClassTypes.catchThrowableOfType(shouldRaiseThrowable, type);
  }
```

Thanks for considering.

AssertJ 3.23.1

java-coding-prodigy:
#### Check List:
*  Fixes #2821 
* Unit tests : TODO


scordio:
Yes, I'd copy the existing documentation.

java-coding-prodigy:
> Yes, I'd copy the existing documentation.

Should I also mention the new method in the old methods description?

scordio:
> Should I also mention the new method in the old methods description?

Yes, you could add a `@deprecated` entry in the Javadoc similarly to what we have with [`AbstractThrowableAssert::getCause`](https://github.com/assertj/assertj/blob/main/assertj-core/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java#L184).

java-coding-prodigy:
Now for unit tests should I also use similar unit tests as the old methods do? Or should I change the method calls in the tests to the new method?

java-coding-prodigy:
> Now for unit tests should I also use similar unit tests as the old methods do? Or should I change the method calls in the tests to the new method?

So I'll do the second thing?

assertj-core/src/main/java/org/assertj/core/api/ThrowableAssert.java
```diff
@@ -68,6 +68,7 @@ public static Throwable catchThrowable(ThrowingCallable shouldRaiseThrowable) {
   }
 
   @SuppressWarnings("unchecked")
+  @Deprecated(since = "3.23.11")
   public static <THROWABLE extends Throwable> THROWABLE catchThrowableOfType(ThrowingCallable shouldRaiseThrowable,
                                                                              Class<THROWABLE> type) {
     Throwable throwable = catchThrowable(shouldRaiseThrowable);
```
scordio, line 74: I suggest replacing the body of this method with:

```java
return ThrowableAssert.catchThrowableOfType(type, shouldRaiseThrowable);
```

scordio:
> > Now for unit tests should I also use similar unit tests as the old methods do? Or should I change the method calls in the tests to the new method?
> 
> So I'll do the second thing?

Yes, changing the method calls in the tests to the new method is fine. Sorry for late feedback!

assertj-core/src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -1415,11 +1416,54 @@ public static Throwable catchThrowable(ThrowingCallable shouldRaiseThrowable) {
    * @see #catchThrowable(ThrowingCallable)
    * @since 3.9.0
    */
+  @Deprecated(since = "3.23.11")
```
scordio, line 1419:
```suggestion
  @Deprecated
```

The `since` attribute is supported on Java 9+ but we are still targeting Java 8.

scordio (changes requested):
As per my comment, please remove the unsupported attributes so that the build can go forward.

assertj-core/src/main/java/org/assertj/core/api/Assertions.java
java-coding-prodigy: Should I put that bit in the documentation then?

java-coding-prodigy:
I don't understand why some of the checks are failing

scordio:
It's caused by the `standard_assertions_and_bdd_assertions_should_have_the_same_non_assertions_methods` test method in `Assertions_sync_with_BDDAssertions_WithAssertions_and_soft_assertions_variants_Test`.

To fix it, you need to add new entry point methods in `BDDAssertions`, in line with the changes you did in `Assertions`.

java-coding-prodigy:
> It's caused by the `standard_assertions_and_bdd_assertions_should_have_the_same_non_assertions_methods` test method in `Assertions_sync_with_BDDAssertions_WithAssertions_and_soft_assertions_variants_Test`.
> 
> To fix it, you need to add new entry point methods in `BDDAssertions`, in line with the changes you did in `Assertions`.

Thank you, I had overlooked the `BDDAssertions.java` file previously.
I also added a `<p>` tag in the documentations of the old methods after the deprecated marker, as the method description was coming on the same line otherwise.

java-coding-prodigy:
@scordio Could you please approve the workflows? Thank you.

java-coding-prodigy:
@scordio Approve the workflows, and also review the code in case any changes still need to made. 
Thank you :)

assertj-core/src/main/java/org/assertj/core/api/ThrowableAssert.java
java-coding-prodigy: This has been done, please resolve the conservation so the PR can move forward.

assertj-core/src/main/java/org/assertj/core/api/Assertions.java
java-coding-prodigy: This has been done, please resolve the conservation so the PR can move forward.

assertj-core/src/main/java/org/assertj/core/api/AssertionsForClassTypes.java
```diff
@@ -897,13 +897,15 @@ public static NotThrownAssert assertThatNoException() {
...
    * @return The captured exception or <code>null</code> if none was raised by the callable.
-   * @see AssertionsForClassTypes#catchThrowableOfType(ThrowableAssert.ThrowingCallable, Class)
+   * @see AssertionsForClassTypes#catchThrowableOfType(Class, ThrowableAssert.ThrowingCallable)
    */
   public static Throwable catchThrowable(ThrowingCallable shouldRaiseThrowable) {
     return ThrowableAssert.catchThrowable(shouldRaiseThrowable);
   }
 
   /**
+   * @deprecated use {@link #catchThrowableOfType(Class, ThrowingCallable)} instead.
```
vlsi, line 907: Please move deprecated javadoc after the documentation. The doc should not start with `@deprecated`

assertj-core/src/main/java/org/assertj/core/api/AssertionsForClassTypes.java
```diff
@@ -940,11 +942,53 @@ public static Throwable catchThrowable(ThrowingCallable shouldRaiseThrowable) {
...
+   *
+   *   public CustomParseException(String msg, int l, int c) {
+   *     super(msg);
+   *     line = l;
+   *     column = c;
+   *   }
+   * }
+   *
+   * CustomParseException e = catchThrowableOfType(() -&gt; { throw new CustomParseException("boom!", 1, 5); },
+   *                                               CustomParseException.class);
```
vlsi, line 971: Please fix arguments order in the sample

assertj-core/src/main/java/org/assertj/core/api/BDDAssertions.java
```diff
@@ -1905,9 +1907,55 @@ public static Throwable catchThrowable(ThrowingCallable shouldRaiseThrowable) {
...
+   * // assertions succeed
+   * assertThat(textException).hasMessage("boom!");
+   * assertThat(textException.line).isEqualTo(1);
+   * assertThat(textException.column).isEqualTo(5);
+   *
+   * // succeeds as catchThrowableOfType returns null when the code does not thrown any exceptions
+   * assertThat(catchThrowableOfType(Exception.class, () -&gt; {})).isNull();
+   *
+   * // fails as TextException is not a RuntimeException
+   * catchThrowableOfType(() -&gt; { throw new TextException("boom!", 1, 5); }, RuntimeException.class);</code></pre>
```
vlsi, line 1946: Please fix argument order

vlsi:
Technically speaking, the PR is not something "Kotlin requires".
The PR makes the code easier to use in Java as well: lambda body can be big, so the code is easier to follow if smaller arguments (e.g. `Class`) come before lambdas.

So I would suggest renaming the PR (and commits) to something like `Move lambdas to last arguments of ....`.

scordio:
I agree with @vlsi, also mentioned in https://github.com/assertj/assertj/issues/2821#issuecomment-1291911809:

> We believe that the readability of Java code can also benefit from this change.

PR renamed.

java-coding-prodigy:
Bump

assertj-core/src/main/java/org/assertj/core/api/AssertionsForClassTypes.java
scordio: Indeed, `@deprecated` should be in the last position according to [Order of Tags](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#orderoftags) from Oracle.

I'll fix it here although the order of the tags is generally not consistent in the codebase.

Is there anything between the `maven-javadoc-plugin` and Spotless that could help enforce an order in line with Oracle guidelines?

assertj-core/src/main/java/org/assertj/core/api/ThrowableAssert.java
scordio: I don't see it done but I'll do it now 😉

scordio:
We should also add a Kotlin-based test as the whole topic started from there. I'll take care of it.

scordio (approved):

assertj-core/src/main/java/org/assertj/core/api/AssertionsForClassTypes.java
scordio: Raised #3469.

PR #2824, opened 2022-10-27
Throw org.opentest4j.AssertionFailedError for containsExactly

### LINKED ISSUE #2697: Throw org.opentest4j.AssertionFailedError for containsExactly
#### Feature summary

Throw `org.opentest4j.AssertionFailedError` for `org.assertj.core.api.AbstractIterableAssert#containsExactly` failures so IDEs would show `<Click to see difference>` feature.

I think it is a sub-task of https://github.com/assertj/assertj-core/issues/1335, however, #1335 looks too broad to ever finished, so I suggest creating sub-tasks.

I suggest adding a task list to #1335 as described in https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists#creating-task-lists

#### Example

```java
assertThat(java.util.Arrays.asList("world", "hello")).containsExactly("hello", "world");
```

ChrisZoet:
#### Check List:
* Fixes #2697 
* Unit tests :  Yes (edited existing ones).

Following the [contributing guidelines](https://github.com/assertj/assertj/blob/main/CONTRIBUTING.md) will make it easier for us to review and accept your  #PR.

**Changes:**

Calling a new failure method inside Iterables that eventually invokes assertionFailedError from AssertionErrorCreator with the expected and actual values needed for the "Click to see difference" feature when AssertionFailedError is thrown. 

Also made modifications on Iterables_assertContainsExactly_tests to call the same failure method.

**New changes:**

Implementation for arrays.

* Rewritten array tests to make sure they call the right failure method.
* assertContainsExactly calls failure method with actual and expected values.
* Actual and expected values are converted to a list when the failure method is called so they show the right values when the 
  toString method is called under water.
* The click to see differences feature will show the actual and expected values side by side.
* Fixing format at Iterables.

--------------------------------------------------------------------------------------------------------------------------------------------

**Results**

**Old:**

image

**New:**

image

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/internal/Iterables.java
+++ b/assertj-core/src/main/java/org/assertj/core/internal/Iterables.java
@@ -1161,7 +1161,7 @@ public class Iterables {
   }
 
   private AssertionError shouldContainExactlyWithIndexAssertionError(Iterable<?> actual, Object[] values,
-                                                                                 List<IndexedDiff> indexedDiffs, AssertionInfo info) {
+                                                                     List<IndexedDiff> indexedDiffs, AssertionInfo info) {
     return failures.failure(info, shouldContainExactlyWithIndexes(actual, list(values), indexedDiffs, comparisonStrategy), actual, list(values));
   }
 
--- a/assertj-core/src/main/java/org/assertj/core/internal/Iterables.java
+++ b/assertj-core/src/main/java/org/assertj/core/internal/Iterables.java
@@ -1161,13 +1161,13 @@ public class Iterables {
   }
 
   private AssertionError shouldContainExactlyWithIndexAssertionError(Iterable<?> actual, Object[] values,
-                                                                     List<IndexedDiff> indexedDiffs, AssertionInfo info) {
-    return failures.failure(info, shouldContainExactlyWithIndexes(actual, list(values), indexedDiffs, comparisonStrategy));
+                                                                                 List<IndexedDiff> indexedDiffs, AssertionInfo info) {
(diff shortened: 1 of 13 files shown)


assertj-core/src/main/java/org/assertj/core/internal/Iterables.java
```diff
@@ -1161,13 +1161,13 @@ private List<IndexedDiff> compareOrder(Object[] values, List<Object> actualAsLis
   }
 
   private AssertionError shouldContainExactlyWithIndexAssertionError(Iterable<?> actual, Object[] values,
-                                                                     List<IndexedDiff> indexedDiffs, AssertionInfo info) {
-    return failures.failure(info, shouldContainExactlyWithIndexes(actual, list(values), indexedDiffs, comparisonStrategy));
+                                                                                 List<IndexedDiff> indexedDiffs, AssertionInfo info) {
```
joel-costigliola, line 1164: minor: keep original formatting

joel-costigliola (changes requested):
@ChrisZoet thanks for the PR, it looks good.
You have done the changes for iterables, we should also do them for arrays in https://github.com/assertj/assertj/blob/main/assertj-core/src/main/java/org/assertj/core/internal/Arrays.java#L276

ChrisZoet:
> @ChrisZoet thanks for the PR, it looks good. You have done the changes for iterables, can you also do them for arrays in https://github.com/assertj/assertj/blob/main/assertj-core/src/main/java/org/assertj/core/internal/Arrays.java#L276, thanks!

Will do!

joel-costigliola:
Integrated thanks @ChrisZoet for the good work!

PR #2848, opened 2022-11-18
Fix AbstractAsset#satisfies(Consumer...) double evaluation

### LINKED ISSUE #2846: AbstractAssert.satisfies(Consumer...) evaluate assertions twice if any condition fail
**Describe the bug**

AbstractAssert.satisfies(Consumer...)  executes each consumer twice if any of the consumer fail. It comes from statistiesForProxy implementation

```java
  protected SELF satisfiesForProxy(Consumer<? super ACTUAL>[] assertionsGroups) throws AssertionError {
    checkArgument(stream(assertionsGroups).allMatch(java.util.Objects::nonNull), "No assertions group should be null");
    if (stream(assertionsGroups).allMatch(this::satisfiesAssertions)) return myself;
    // some assertions groups were not met! let's report all the errors
    List<AssertionError> assertionErrors = stream(assertionsGroups).map(this::catchOptionalAssertionError)
                                                                   .filter(Optional::isPresent)
                                                                   .map(Optional::get)
                                                                   .collect(toList());
    throw multipleAssertionsError(assertionErrors);
  }
```

Executing the consumer twice is usually OK as most consumers are idempotent. However, some are not. I stumbled upon this is issue writing the following test involving a JAX-RS response

```java
              .run( webTarget -> {
                        var response = webTarget.get( );

                        assertThat( response )
                                .satisfies( isClientErrorWithMessage( "filter", "foo" ) );
                    } );


 Consumer<Response> isClientErrorWithMessage( String field, String... messageParts ) {
        return response -> {
            assertThat( response.getStatus( ) ).isEqualTo( 400 );

            var jsonResponse = response.readEntity( JsonNode.class );
            [...]
    }
```

A response can only be read once. `java.lang.IllegalStateException: Entity input stream has already been closed .` is displayed rather than the actual error. 
 
It seems easy to change satisfiesForProxy implementation to avoid the superfluous execution. I can try to submit a merge request if you acknowledge the issue and that my reasoning is correct. 

* assertj core version: 3.22.0
* java version: 17

```java
    @Test
    void reproducer( ) {
        Consumer<String> consumer = s -> {
            System.out.println( s );
            assertThat( s ).isEqualTo( "bar" );
        };
        assertThat( "Foo" ).satisfies( consumer );
    }
```

```
Foo
Foo


expected: "bar"
 but was: "Foo"
Comparison Failure: 
Expected :"bar"
Actual   :"Foo"
```

cykl:
Fixes #2846

#### Check List:
* Fixes #2846
* Unit tests : No. I assume that satisfies behavior is already checked and adding a specific check for double evaluation is not required. I can add this test if you wish.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -866,13 +866,16 @@ public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, 
   // in order to avoid compiler warning in user code
   protected SELF satisfiesForProxy(Consumer<? super ACTUAL>[] assertionsGroups) throws AssertionError {
     checkArgument(stream(assertionsGroups).allMatch(java.util.Objects::nonNull), "No assertions group should be null");
-    if (stream(assertionsGroups).allMatch(this::satisfiesAssertions)) return myself;
     // some assertions groups were not met! let's report all the errors
     List<AssertionError> assertionErrors = stream(assertionsGroups).map(this::catchOptionalAssertionError)
                                                                    .filter(Optional::isPresent)
                                                                    .map(Optional::get)
                                                                    .collect(toList());
-    throw multipleAssertionsError(assertionErrors);
+    if (assertionErrors.isEmpty()) {
+      return myself;
+    } else {
+      throw multipleAssertionsError(assertionErrors);
+    }
   }
 
   private Optional<AssertionError> catchOptionalAssertionError(Consumer<? super ACTUAL> assertions) {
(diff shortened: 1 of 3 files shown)


cykl:
_CodeQL / Analyze (java) (pull_request)_ failure seems unrelated to this patch: `   Error: 1-18 14:24:53] [autobuild] [ERROR]     class file has wrong version 61.0, should be 55.0` (Java 17  vs 11)

scordio:
> _CodeQL / Analyze (java) (pull_request)_ failure seems unrelated to this patch: ` Error: 1-18 14:24:53] [autobuild] [ERROR] class file has wrong version 61.0, should be 55.0` (Java 17 vs 11)

Yes, absolutely unrelated and now it should be fixed on `main`.

joel-costigliola:
@cykl I think it would be good to have a test with a consumer that fails if executed a second time so that if we refactor this code in the future the test will fail if the refactoring leads to execute the consumer multiple times.

cykl:
@joel-costigliola Tests added. 

Not too sure about the style or class name but that seems to comply with the CONTRIBUTING recommendations. Feel free to amend or comment, whatever is the fastest to you.

joel-costigliola:
Integrated after minor code improvements, thanks @cykl!

PR #2853, opened 2022-11-20
Add method hasMessageMatching(Pattern)

asolntsev:
Pattern might be preferred over String regex at least for two reasons:
1. It's more performant (pre-compiled Pattern instance can be reused in many tests)
2. It may be more readable. For example, `Pattern.compile("foo", DOTALL)` may be more readable than `"(?s)foo"`.


joel-costigliola:
Integrated thanks @asolntsev!

Issue #2061, opened 2020-12-04
Add ability to use String Supplier when setting the assertion description.


tkvangorder:
#### Summary

I am working on a [project that will migrate JUnit 5 style assertions to AssertJ's assertion library.](https://github.com/openrewrite/rewrite-testing-frameworks/issues/8) 

One issue I have run into is that JUnit 5 has a variant of their assertion methods that accept a Supplier<String> to lazily instantiate the message: 

```
assertEquals("joe", "fred", () -> "These names should be the same");
```
Ideally, I would like to convert this to: 

```
assertThat("fred").as(() -> "These names should be the same").isEqualTo("joe");
```

There is no obvious, generic way to convert the supplier into a Description. We envision many projects using our "recipe" to perform this migration and it would be helpful if there was an overloaded `as` method within the core assertJ library. 

Right now, I am using .`withFailMessage` which already has an overload that accepts a supplier but, of course, this will 
completely override the message.

I think it would be as easy as adding something like the following:

```java
	public SELF as(final Supplier<String> messageSupplier) {
		super.as(
			new Description() {
			
			@Override
			public String value() {
				return messageSupplier.get();
			}
		});
	}
```

Thanks

joel-costigliola:
Fair enough!

PR #2855, opened 2022-11-29
add Descriptable#describedAs(Supplier<String>)

vlsi:
`describedAs(Supplier<String>)` makes it eaiser to use lazy descriptions from Kotlin

This is a follow-up to #2061

An alternative option might be to make `Description implements Supplier<String>` and migrate to `Supplier<String>` instead of `Description`, however, that is a more involved change.


joel-costigliola:
looking at it in the next few days

vlsi:
Could this be merged? 😉

joel-costigliola:
Integrated thanks @vlsi!

PR #2884, opened 2022-12-18
Comparables::assertIsBetween String alloc. removed

DirkToewe:
`Comparables::checkBoundsValidity`, which is used by `Comparables::assertIsBetween`, does build an error message String even if the assertion does not fail. This results in noticeable overhead. The JVM can hardly eliminate the String building since `start.toString()`, `operator.toString()` and `end.toString()` may have side effects.

This PR moves the String building inside a `Supplier<String>` which is only evaluated on failure.


joel-costigliola:
Integrated thanks @DirkToewe!

PR #2902, opened 2023-01-06
build: Use Bnd feature to write resolved bndrun files to output folder

bjhargrave:
Bnd 6.4.0 allows writing and reading bndrun files from different folders. So we change to use this feature so that the resolver plugin writes the resolved bndrun file to the target folder and the testing plugin uses the resolved bndrun file from the target folder.

This change means that running the maven build does not modify checked in source (bndrun files).

#### Check List:
* Unit tests : NA

### DIFF
--- a/assertj-guava/pom.xml
+++ b/assertj-guava/pom.xml
@@ -132,8 +132,9 @@
               <goal>resolve</goal>
             </goals>
             <configuration>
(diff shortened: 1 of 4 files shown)


scordio:
Thanks, @bjhargrave!

PR #2755, opened 2022-08-29
Add missing description to nested condition

alex859:
#### Check List:
* Unit tests : YES
* Javadoc with a code example (on API only) : NA
* PR meets the [contributing guidelines](https://github.com/assertj/assertj/blob/main/CONTRIBUTING.md)

Calling the `description()` method of a nested condition is currently returning empty strings, e.g.:
```java
customer(
   name(
      first("John")),
      address(
         firstLine("11, Downing Street"),
         postcode("SW1A 2AA"))).description()
```
is
```
customer:[
     name:[
        
     ],
     address:[
        ,
        
     ]
  ]
```
but we should have:
```
customer:[
     name:[
        first: John
     ],
     address:[
        first line: 11, Downing Street,
        postcode: SW1A 2AA
     ]
  ]
```
This PR fixes the issue.

Issue #2495, opened 2022-02-12
Soft assertions with failed navigation methods should prevent further assertions


scordio:
We started introducing `isNotNull` checks in some navigation methods (#2401, #2411, #2412) but this created a side effect for soft assertions.

For example, the following:

```java
// GIVEN
SoftAssertions softly = new SoftAssertions();
// WHEN
softly.assertThat(null)
      .extracting(Object::getClass) // inner isNotNull() throws AssertionError 
      .isEqualTo(...);
// THEN
then(softly.errorsCollected()).hasSize(1);
```

fails at `extracting` with an NPE, as the `AssertionError` thrown by `isNotNull` is swallowed by the soft assertions' error collector, and the rest of the `extracting` method is executed.

We need to rethink what to expect in these cases. Ideally:
* `extracting` should be completed and one `AssertionError` due to `isNotNull` should be collected without further failures
* Any other chained assertions should be skipped without collecting additional errors

This might require some refactoring in the ByteBuddy configuration and in the `ErrorCollector` implementation.

joel-costigliola:
With the recent work done by @filiphr, I think we can close this issue since we don't execute assertions after a failed navigation.

This test succeeds now, we capture the null actual error but not the `isEqualTo("foo")` one as it is not skipped.
```java
// GIVEN
SoftAssertions softly = new SoftAssertions();
// WHEN
softly.assertThat((Object) null)
      .extracting(Object::getClass) // inner isNotNull() throws AssertionError
      .isEqualTo("foo");
// THEN
then(softly.errorsCollected()).hasSize(1);
then(softly.errorsCollected().get(0)).hasMessageContaining("Expecting actual not to be null");
```

#2784 is also addressed, the test results in the two expected errors

```java
@Test
public void optionalSoftAssertion() {
    Optional<String> empty = Optional.empty();
    
    SoftAssertions.assertSoftly(softly -> {
        softly.assertThat(0).isEqualTo(1);
        softly.assertThat(empty).get();
    });
}
```


```
org.assertj.core.error.MultipleAssertionsError: 2 assertion errors:

-- error 1 --
expected: 1
 but was: 0

-- error 2 --
Expecting Optional to contain a value but it was empty.
```

@scordio can we close this or was there something else to tackle ?

PR #2905, opened 2023-01-08
Make nestable condition factory accept conditions on super types

alex859:
#### Check List:
* Unit tests : YES

Makes it possible to create a `NestableCondition<T>` by passing in a condition on a super type of `T`.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/condition/NestableCondition.java
+++ b/assertj-core/src/main/java/org/assertj/core/condition/NestableCondition.java
@@ -142,12 +142,12 @@ public class NestableCondition<ACTUAL, NESTED> extends Join<ACTUAL> {
     return new NestableCondition<ACTUAL, ACTUAL>(descriptionPrefix, stream(conditions));
   }
 
-  private NestableCondition(String descriptionPrefix, Stream<Condition<? super NESTED>> conditions, Function<? super ACTUAL, ? extends NESTED> extractor) {
+  private NestableCondition(String descriptionPrefix, Stream<? extends Condition<? super NESTED>> conditions, Function<? super ACTUAL, ? extends NESTED> extractor) {
     super(compose(conditions, extractor));
     this.descriptionPrefix = descriptionPrefix;
   }
 
-  private NestableCondition(String descriptionPrefix, Stream<Condition<? super ACTUAL>> conditions) {
+  private NestableCondition(String descriptionPrefix, Stream<? extends Condition<? super ACTUAL>> conditions) {
     super(conditions.collect(toList()));
     this.descriptionPrefix = descriptionPrefix;
   }
@@ -162,7 +162,7 @@ public class NestableCondition<ACTUAL, NESTED> extends Join<ACTUAL> {
     return descriptionPrefix;
   }
 
-  private static <ACTUAL, NESTED> List<Condition<? super ACTUAL>> compose(Stream<Condition<? super NESTED>> conditions,
+  private static <ACTUAL, NESTED> List<Condition<? super ACTUAL>> compose(Stream<? extends Condition<? super NESTED>> conditions,
                                                                   Function<? super ACTUAL, ? extends NESTED> extractor) {
     return conditions.map(condition -> compose(condition, extractor)).collect(toList());
   }
--- a/assertj-core/src/main/java/org/assertj/core/condition/NestableCondition.java
+++ b/assertj-core/src/main/java/org/assertj/core/condition/NestableCondition.java
@@ -125,8 +125,8 @@ public class NestableCondition<ACTUAL, NESTED> extends Join<ACTUAL> {
    * @param <NESTED> the type of object nested into {@literal K}
    */
   @SafeVarargs
-  public static <ACTUAL, NESTED> Condition<ACTUAL> nestable(String descriptionPrefix, Function<ACTUAL, NESTED> extractor,
-                                                            Condition<NESTED>... conditions) {
+  public static <ACTUAL, NESTED> Condition<ACTUAL> nestable(String descriptionPrefix, Function<? super ACTUAL, ? extends NESTED> extractor,
+                                                            Condition<? super NESTED>... conditions) {
     return new NestableCondition<>(descriptionPrefix, stream(conditions), extractor);
   }
 
@@ -138,16 +138,16 @@ public class NestableCondition<ACTUAL, NESTED> extends Join<ACTUAL> {
    * @param <ACTUAL> the type of object the resulting condition accepts
    */
   @SafeVarargs
-  public static <ACTUAL> Condition<ACTUAL> nestable(String descriptionPrefix, Condition<ACTUAL>... conditions) {
-    return new NestableCondition<>(descriptionPrefix, stream(conditions));
+  public static <ACTUAL> Condition<ACTUAL> nestable(String descriptionPrefix, Condition<? super ACTUAL>... conditions) {
+    return new NestableCondition<ACTUAL, ACTUAL>(descriptionPrefix, stream(conditions));
   }
 
-  private NestableCondition(String descriptionPrefix, Stream<Condition<NESTED>> conditions, Function<ACTUAL, NESTED> extractor) {
+  private NestableCondition(String descriptionPrefix, Stream<Condition<? super NESTED>> conditions, Function<? super ACTUAL, ? extends NESTED> extractor) {
     super(compose(conditions, extractor));
     this.descriptionPrefix = descriptionPrefix;
   }
 
-  private NestableCondition(String descriptionPrefix, Stream<Condition<ACTUAL>> conditions) {
+  private NestableCondition(String descriptionPrefix, Stream<Condition<? super ACTUAL>> conditions) {
     super(conditions.collect(toList()));
     this.descriptionPrefix = descriptionPrefix;
   }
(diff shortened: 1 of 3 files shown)


assertj-core/src/main/java/org/assertj/core/condition/NestableCondition.java
```diff
@@ -138,16 +138,16 @@ public static <ACTUAL, NESTED> Condition<ACTUAL> nestable(String descriptionPref
...
    */
   @SafeVarargs
-  public static <ACTUAL> Condition<ACTUAL> nestable(String descriptionPrefix, Condition<ACTUAL>... conditions) {
-    return new NestableCondition<>(descriptionPrefix, stream(conditions));
+  public static <ACTUAL> Condition<ACTUAL> nestable(String descriptionPrefix, Condition<? super ACTUAL>... conditions) {
+    return new NestableCondition<ACTUAL, ACTUAL>(descriptionPrefix, stream(conditions));
   }
 
-  private NestableCondition(String descriptionPrefix, Stream<Condition<NESTED>> conditions, Function<ACTUAL, NESTED> extractor) {
+  private NestableCondition(String descriptionPrefix, Stream<Condition<? super NESTED>> conditions, Function<ACTUAL, NESTED> extractor) {
```
vlsi, line 145: Function<a, n> should be Function<? … a, ? … n> as well

alex859: Hi @vlsi good point, thanks :+1: .

I've changed it for the function's input, but I can't find an example where changing the boundary of the function output is required :thinking: .

vlsi: `Function<ACTUAL, NESTED>` means function that `consumes ACTUAL` and `produces NESTED`.
With Java, you have to put `extends` and `super` all over the place (== in parameter declaration and return types).

For instance, `Function<A, B>` must always (except small edge cases) should be used as `Function<?  super A, ? extends B>`.
Then the method would accept `Function<.., ...>` that does not exactly match `Function<ACTUAL, NESTED>`.

See https://gist.github.com/vlsi/3e7c6a0aa3f13ec0cfc94c0f7447ceb7

assertj-core/src/test/java/org/assertj/core/condition/NestableCondition_matches_Test.java
```diff
@@ -75,4 +79,35 @@ void should_not_match_if_any_condition_in_nested_level_does_not_match() {
...
+  void should_accept_conditions_on_supertypes() {
+    // GIVEN
+    final ValueCustomer boris = new ValueCustomer(new Name("Boris", "Johnson"),
+                                                  new Address("10, Downing Street",
+                                                              "SW1A 2AA",
+                                                              new Country("United Kingdom")),
+                                                  12);
+    Condition<ValueCustomer> valueCustomer = nestable("value customer", name(
+                                                                         first("Boris")),
+                                                  value(12));
```
vlsi, line 93: It would be nice if the lines could be wrapped. For instance, put `new ValueCustomer` on a new line, and put `new Name` on a new line as well. Currently, the code contains lots of whitespace, and it is hard to follow.

assertj-core/src/test/java/org/assertj/core/condition/NestableConditionFixtures.java
```diff
@@ -63,6 +63,12 @@ static Condition<Customer> customer(Condition<Customer>... conditions) {
     return nestable("customer", conditions);
   }
 
+  static Condition<ValueCustomer> value(Integer expected) {
+    return verboseCondition(valueCustomer -> expected.equals(valueCustomer.value),
+      "value: " + expected,
+      valueCustomer -> " but was " + valueCustomer.value);
```
vlsi, line 69: This seems to duplicate the logic of `assertEquals`. Can it be de-duplicated?

assertj-core/src/main/java/org/assertj/core/condition/NestableCondition.java
alex859: Thanks :+1:

assertj-core/src/test/java/org/assertj/core/condition/NestableCondition_matches_Test.java
alex859: I agree, but that is the format I get applying the Eclipse format.

alex859: Tried to improve it.

assertj-core/src/test/java/org/assertj/core/condition/NestableConditionFixtures.java
alex859: Not sure about this, what `assertEquals` method are you referring to?

assertj-core/src/main/java/org/assertj/core/condition/NestableCondition.java
```diff
@@ -138,16 +138,16 @@ public static <ACTUAL, NESTED> Condition<ACTUAL> nestable(String descriptionPref
...
    */
   @SafeVarargs
-  public static <ACTUAL> Condition<ACTUAL> nestable(String descriptionPrefix, Condition<ACTUAL>... conditions) {
-    return new NestableCondition<>(descriptionPrefix, stream(conditions));
+  public static <ACTUAL> Condition<ACTUAL> nestable(String descriptionPrefix, Condition<? super ACTUAL>... conditions) {
+    return new NestableCondition<ACTUAL, ACTUAL>(descriptionPrefix, stream(conditions));
   }
 
-  private NestableCondition(String descriptionPrefix, Stream<Condition<NESTED>> conditions, Function<ACTUAL, NESTED> extractor) {
+  private NestableCondition(String descriptionPrefix, Stream<Condition<? super NESTED>> conditions, Function<? super ACTUAL, ? extends NESTED> extractor) {
```
vlsi, line 145: `Stream<? extends Condition<…` ?

assertj-core/src/test/java/org/assertj/core/condition/NestableConditionFixtures.java
vlsi: `assertThat(valueCustomer.value).as("value").isEqualTo(expected)`

alex859: My understanding is that when creating a condition we need a predicate returning true/false, while calling `isEqual` would throw an exception when the condition is not satisfied. 

It would be nice to reuse assertions as conditions. Is there a way of doing that? (a way better than catching the exception an returning false?)

vlsi:
`NestableCondition` seems to duplicate logic of the already existing `org.assertj.core.condition.MappedCondition`.
Have you considered reusing `MappedCondition` instead?

/cc @joel-costigliola

vlsi:
The generified code looks good to me, however, it is sad that `nestable` duplicates logic of `MappableCondition` and it lacks support for `isEqualTo` builders.

So the users of `nestable` can't really leverage `assertj` power and they resort to `Objects.equals` and all its quircks like String and StringBuilder looking the same and returning false on equals.

alex859:
I considered using `MappingCondition` instead of adding the `compose` method. The issue I had was that the description generated had some extra text I didn't like, e.g. in the assertion message from `NestableCondition_assertionMessage_Test#should_show_correct_error_message_with_two_nested_objects` instead of 
```
[✗] postcode: SW2A 2AA but was SW1A 2AA
```
we would get 
```
[✗] mapped
   from: <Customer> org.assertj.core.condition.Customer@68c72235
   to:   <Address> org.assertj.core.condition.Address@34f7cfd9
   then checked:
         [✗] postcode: SW2A 2AA but was SW1A 2AA
```

I agree on the lack of using the power of `assertj`, isn't that a general issue when using conditions?

vlsi:
>I considered using MappingCondition instead of adding the compose method. The issue I had was that the description generated had some extra text I didn't like, e.g. in the assertion message from

Have you considered making MappedCondition less verbose or making the verbosity configurable?
The issue is that both `MappedCondition` and `NestedCondition` appear as implementations of `Condition`, so it makes the API hard to reason about. Personally, I often check "find all implementations" to select the one I need.

> I agree on the lack of using the power of assertj, isn't that a general issue when using conditions?

One of the ways out could be suggesting that users should resort to something like `org.assertj.core.api.HamcrestCondition.matching(org.hamcrest.Matchers.is(expected))` instead of `verboseCondition`.


An alternative way could be decouple `assertThat(...).isEqualTo()` into `AssertBuilder` and `assert verification` steps.
Of course, it might require a more elaborate integration with assertj, however, it might be in line with https://github.com/assertj/assertj/issues/2495.
In other words, it might be nice if "navigation and assertion methods" could be decoupled from "triggering the verification". Then "navigation" could be reused across regular asserts, soft asserts, and nested asserts.

alex859:
I'm sorry I think I misunderstood your previous comment. Are you suggesting that, instead of adding `NestableCondition`, we could modify `MappedCondition` to achieve the same result and have only one of them?

Regarding `isEqual` maybe another option could be to add a condition that internally uses `StandardComparisonStrategy#areEqual`?

vlsi:
>Are you suggesting that, instead of adding NestableCondition, we could modify MappedCondition to achieve the same result and have only one of them?

I think so.

>Regarding isEqual maybe another option could be to add a condition that internally uses StandardComparisonStrategy#areEqual?

There are multiple implementations of `StandardComparisonStrategy`.
A short-term solution might be adding something like `AreEqualCondition` instead of the current `verboseCondition`.
However, it would be sad to start adding conditions for all the methods in `ComparisonStrategy`

alex859:
Hi @vlsi , I think you are right, the same behaviour can be achieved by having:
* a less verbose/configurable `MappedCondition`
* a configurable `AllOf` description prefix

With those changes there would be no need for `NestableCondition`, and for example `NestableConditionFixtures#address`  could be defined as something like:

```java
  static Condition<Customer> address(Condition<? super Address>... conditions) {
    return AllOf.allOf(
      "address", 
      Arrays.stream(conditions)
        .map(condition -> lessVerboseMappedCondition((Customer customer) -> customer.address, condition))
        .collect(Collectors.toList())
    );
  }
```

Thoughts?

vlsi:
I believe "address" string should be a part of lessVerboseCondition, not an argument to allof

alex859:
Good point. I think that some improvements can be discussed and addressed with a separate PR, the goal of this one was just to improve the current state of `NestableCondition`.

vlsi (commented):
The improved generics look good to me

joel-costigliola:
Will look at it this weekend, thanks for your patience

alex859:
Thank you! It would be great if you could also review #2755.

alex859:
Thank you! It would be great if you could also review #2755.

joel-costigliola:
Integrated thanks @alex859!

alex859:
Thank you! ☺️

PR #2907, opened 2023-01-09
Add KeyStoreAssert

chriskilding:
Add `KeyStoreAssert` with the following assertions:

- `containsAlias(String)`
- `doesNotContainAlias(String)`
- TODO add others

#### Check List:
* [x] Unit tests : YES
* [ ] Javadoc with a code example (on API only) : YES / NO / NA
* [ ] PR meets the [contributing guidelines](https://github.com/assertj/assertj/blob/main/CONTRIBUTING.md)

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/KeyStoreAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/KeyStoreAssert.java
@@ -1,0 +1,83 @@
+package org.assertj.core.api;
+
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.cert.Certificate;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.assertj.core.error.KeyStoreShouldContainAlias.shouldContainAlias;
+import static org.assertj.core.error.KeyStoreShouldNotContainAlias.shouldNotContainAlias;
+
+public class KeyStoreAssert extends AbstractAssert<KeyStoreAssert, KeyStore> {
+  public KeyStoreAssert(KeyStore keyStore) {
+    super(keyStore, KeyStoreAssert.class);
+  }
+
+  public KeyStoreAssert containsEntry(String alias, Certificate[] certificateChain) {
+    isNotNull();
+
+    // FIXME implement and write tests
+
+    return this;
+  }
+
+  /**
+   * Verifies that the actual {@code KeyStore} contains the specified alias.
+   * @param alias the expected alias to find
+   * @return {@code this} assertion object.
+   */
+  public KeyStoreAssert containsAlias(String alias) {
+    isNotNull();
+
+    if (!doesContainAlias(actual, alias)) {
+      throwAssertionError(shouldContainAlias(alias));
+    }
+
+    return this;
+  }
+
+  /**
+   * Verifies that the actual {@code KeyStore} does not contain the specified alias.
+   * @param alias the alias that should be absent
+   * @return {@code this} assertion object.
+   */
+  public KeyStoreAssert doesNotContainAlias(String alias) {
+    isNotNull();
+
+    if (doesContainAlias(actual, alias)) {
+      throwAssertionError(shouldNotContainAlias(alias));
+    }
+
+    return this;
+  }
+
+  private static boolean doesContainAlias(KeyStore keyStore, String alias) {
+    final boolean doesContainAlias;
+    try {
+      doesContainAlias = keyStore.containsAlias(alias);
+    } catch (KeyStoreException e) {
+      // FIXME is there a better unchecked exception type for this?
+      throw new RuntimeException(e);
+    }
+    return doesContainAlias;
+  }
+
+  private static Map<String, Certificate[]> keystoreToMap(KeyStore keyStore) {
+    final Map<String, Certificate[]> ks = new HashMap<>();
+
+    try {
+      final Enumeration<String> aliases = keyStore.aliases();
+      while (aliases.hasMoreElements()) {
+        final String a = aliases.nextElement();
+        final Certificate[] certificateChain = keyStore.getCertificateChain(a);
+        ks.put(a, certificateChain);
+      }
+    } catch (KeyStoreException e) {
+      throw new RuntimeException(e);
+    }
+
+    return ks;
+  }
+}
--- a/assertj-core/src/main/java/org/assertj/core/error/KeyStoreShouldContainAlias.java
+++ b/assertj-core/src/main/java/org/assertj/core/error/KeyStoreShouldContainAlias.java
@@ -1,0 +1,12 @@
+package org.assertj.core.error;
+
+public class KeyStoreShouldContainAlias extends BasicErrorMessageFactory {
+
+  public KeyStoreShouldContainAlias(String alias) {
+    super("Expected KeyStore to contain alias <%s>, but it did not", alias);
+  }
+
+  public static KeyStoreShouldContainAlias shouldContainAlias(String alias) {
+    return new KeyStoreShouldContainAlias(alias);
+  }
+}
--- a/assertj-core/src/main/java/org/assertj/core/error/KeyStoreShouldNotContainAlias.java
+++ b/assertj-core/src/main/java/org/assertj/core/error/KeyStoreShouldNotContainAlias.java
@@ -1,0 +1,12 @@
+package org.assertj.core.error;
+
+public class KeyStoreShouldNotContainAlias extends BasicErrorMessageFactory {
+
+  public KeyStoreShouldNotContainAlias(String alias) {
+    super("Expected KeyStore not to contain alias <%s>, but it did", alias);
+  }
+
+  public static KeyStoreShouldNotContainAlias shouldNotContainAlias(String alias) {
+    return new KeyStoreShouldNotContainAlias(alias);
+  }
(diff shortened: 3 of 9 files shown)


chriskilding:
@joel-costigliola would you be able to give me some early stage pointers on this PR? In particular

- Where do I need to export the KeyStoreAssert for library users to see it? (So far I've put it in `org.assertj.core.api.Assertions` and `org.assertj.core.api.BDDAssertions` - is there anywhere else I should export it?)
- Do I need to make a separate `AbstractKeyStoreAssert` superclass and have the `KeyStoreAssert` extend it? (So far I've just got the one concrete class.)
- Any thoughts on the assertion error message structure? (I created new error message classes under `org.assertj.core.error` - emulating what some of the other classes did, but wasn't sure if this is the right thing to do.)

joel-costigliola:
@chriskilding will be looking at it this week, thanks for your patience 😉

assertj-core/src/main/java/org/assertj/core/api/KeyStoreAssert.java
```diff
@@ -0,0 +1,95 @@
...
+  public KeyStoreAssert containsEntry(String alias, Certificate[] certificateChain) {
+    isNotNull();
+
+    // FIXME implement and write tests
+
+    return this;
+  }
+
+  /**
+   * Verifies that the actual {@code KeyStore} contains the specified alias.
```
joel-costigliola, line 39: add a code example showing a successful assertions and if that helps a failed one (probably not needed in this case but that's a general advice)

joel-costigliola: this comment applies to all assertion methods in this file

assertj-core/src/main/java/org/assertj/core/api/KeyStoreAssert.java
```diff
@@ -0,0 +1,95 @@
...
+      throwAssertionError(shouldNotContainAlias(alias));
+    }
+
+    return this;
+  }
+
+  private static boolean doesContainAlias(KeyStore keyStore, String alias) {
+    final boolean doesContainAlias;
+    try {
+      doesContainAlias = keyStore.containsAlias(alias);
```
joel-costigliola, line 71: you might be able to return directly here and avoid having to define `doesContainAlias`

assertj-core/src/main/java/org/assertj/core/api/KeyStoreAssert.java
```diff
@@ -0,0 +1,95 @@
...
+   * @param alias the expected alias to find
+   * @return {@code this} assertion object.
+   */
+  public KeyStoreAssert containsAlias(String alias) {
+    isNotNull();
+
+    if (!doesContainAlias(actual, alias)) {
+      throwAssertionError(shouldContainAlias(alias));
+    }
+
```
joel-costigliola, line 49: minor: remove empty lines (comment applies elsewhere too)

assertj-core/src/main/java/org/assertj/core/api/KeyStoreAssert.java
```diff
@@ -0,0 +1,95 @@
...
+    return this;
+  }
+
+  private static boolean doesContainAlias(KeyStore keyStore, String alias) {
+    final boolean doesContainAlias;
+    try {
+      doesContainAlias = keyStore.containsAlias(alias);
+    } catch (KeyStoreException e) {
+      // FIXME is there a better unchecked exception type for this?
```
joel-costigliola, line 73: not sure but I don't think it matters a lot since the real cause is passed

assertj-core/src/main/java/org/assertj/core/api/KeyStoreAssert.java
```diff
@@ -0,0 +1,95 @@
...
+
+import static org.assertj.core.error.KeyStoreShouldContainAlias.shouldContainAlias;
+import static org.assertj.core.error.KeyStoreShouldNotContainAlias.shouldNotContainAlias;
+
+public class KeyStoreAssert extends AbstractAssert<KeyStoreAssert, KeyStore> {
+  public KeyStoreAssert(KeyStore keyStore) {
+    super(keyStore, KeyStoreAssert.class);
+  }
+
+  public KeyStoreAssert containsEntry(String alias, Certificate[] certificateChain) {
```
joel-costigliola, line 30: Maybe I'm wrong but I feel that `containsEntry` is not a term related to keystore , I would prefer something like `containsCertificateChainForAlias`, longer but clearer IMO

assertj-core/src/main/java/org/assertj/core/api/KeyStoreAssert.java
```diff
@@ -0,0 +1,95 @@
...
+    try {
+      doesContainAlias = keyStore.containsAlias(alias);
+    } catch (KeyStoreException e) {
+      // FIXME is there a better unchecked exception type for this?
+      throw new RuntimeException(e);
+    }
+    return doesContainAlias;
+  }
+
+  private static Map<String, Certificate[]> keystoreToMap(KeyStore keyStore) {
```
joel-costigliola, line 79: will be used for the first assertion I assume

assertj-core/src/main/java/org/assertj/core/error/KeyStoreShouldContainAlias.java
```diff
@@ -0,0 +1,24 @@
...
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2023 the original author or authors.
+ */
+package org.assertj.core.error;
+
+public class KeyStoreShouldContainAlias extends BasicErrorMessageFactory {
+
+  public KeyStoreShouldContainAlias(String alias) {
+    super("Expected KeyStore to contain alias <%s>, but it did not", alias);
```
joel-costigliola, line 18: typo `Expected` -> `Expecting`

assertj-core/src/main/java/org/assertj/core/error/KeyStoreShouldContainAlias.java
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  public static KeyStoreShouldContainAlias shouldContainAlias(String alias) {
```
joel-costigliola, line 21: rename to `keyStoreShouldContainAlias` to avoid any ambiguity as from where the alias comes from

assertj-core/src/main/java/org/assertj/core/error/KeyStoreShouldNotContainAlias.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2023 the original author or authors.
+ */
+package org.assertj.core.error;
+
+public class KeyStoreShouldNotContainAlias extends BasicErrorMessageFactory {
```
joel-costigliola, line 15: similar comments as those in `KeyStoreShouldContainAlias` ;-)

assertj-core/src/test/java/org/assertj/core/api/keystore/KeyStoreAssert_containsAlias_Test.java
```diff
@@ -0,0 +1,55 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.BDDAssertions.then;
+import org.assertj.core.test.KeyStores;
+import static org.assertj.core.error.KeyStoreShouldContainAlias.shouldContainAlias;
+import static org.assertj.core.util.AssertionsUtil.expectAssertionError;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class KeyStoreAssert_containsAlias_Test {
```
joel-costigliola, line 29: minor: test classes don't need to be `public`, let's make it package private (no keyword)

joel-costigliola (changes requested):
Thanks @chriskilding, first round of comment done, no major issue but some improvements to do

assertj-core/src/main/java/org/assertj/core/error/KeyStoreShouldContainAlias.java
chriskilding: Updated

assertj-core/src/main/java/org/assertj/core/error/KeyStoreShouldContainAlias.java
chriskilding: Done

assertj-core/src/main/java/org/assertj/core/error/KeyStoreShouldNotContainAlias.java
chriskilding: Done

assertj-core/src/test/java/org/assertj/core/api/keystore/KeyStoreAssert_containsAlias_Test.java
chriskilding: Updated

PR #2921, opened 2023-01-15
Add a friendly warning to usages of .isEqualTo and .isNotEqualTo on a…

ascopes:
…ssertions.

Currently, if I were to perform the following code...

    var assertion = assertThat(something);
    assertThat(assertion).isEqualTo(assertion);

... I would get an error about the use of .equals being unsupported, and that I should consider using .isEqualTo instead. This is somewhat confusing to the reader, since they *ARE* using .isEqualTo. This is due to the fact that .isEqualTo internally will invoke calls to .equals on the 'actual' member which will trigger the aforementioned warning.

I have added two new tests to .isEqualTo and .isNotEqualTo that check if the 'actual' member is an assertion itself, and then raise an error if the condition is met. The advisory note is to use .isSameAs and .isNotSameAs instead.

This is useful when using assertj to test custom assertion types for extension libraries, where we may have wanted to use ".satisfies" instead of ".isEqualTo".

An example of where I hit this error:

```java
      // Given
      var assertions = new JavaFileObjectKindAssert(kind);

      // Then
      assertThat(assertions.extension())
          .isInstanceOf(AbstractStringAssert.class)
          .isEqualTo(kind.extension);
```
```
java.lang.UnsupportedOperationException: 'equals' is not supported...maybe you intended to call 'isEqualTo'
```

...of course, what I really meant to do was this:

```java
      // Given
      var assertions = new JavaFileObjectKindAssert(kind);

      // Then
      assertThat(assertions.extension())
          .isInstanceOf(AbstractStringAssert.class)
          .satisfies(extensionAssertion -> extensionAssertion.isEqualTo(kind.extension));
```


joel-costigliola:
Integrated thanks @ascopes and @scordio !

PR #2924, opened 2023-01-20
Embed module version into module descriptor

ascopes:
This makes the module descriptor aware of the module version.

By doing this, stacktraces in Java 9+ will contain the specific version of AssertJ next to the module name, allowing bug reports to contain the AssertJ version in use automatically when a stacktrace is present.

### DIFF
--- a/assertj-core/pom.xml
+++ b/assertj-core/pom.xml
@@ -319,6 +319,9 @@
             </goals>
             <configuration>
               <release>9</release>
+              <compilerArgs>
+                <compilerArg>--module-version=${project.version}</compilerArg>
+              </compilerArgs>
               <compileSourceRoots>
                 <compileSourceRoot>${project.basedir}/src/main/java9</compileSourceRoot>
               </compileSourceRoots>


scordio:
Isn't this done by default since version 3.8.1 of the `maven-compiler-plugin`? See apache/maven-compiler-plugin@2ba3f260c524c226dfd0e94fb22ab2b007519bb1.

What could be a good reproducer to see the result before and after the change?

ascopes:
Aha, I didn't realise this had been added into the plugin directly! I must have been checking the non JPMS version of it when I checked for this in the past!

My mistake! Thanks for the info

scordio:
No problem at all! Actually if you have the chance please check if the latest AssertJ version looks good from your perspective, not sure if our MRJAR setup could influence this somehow.

ascopes:
From what I can see on my side, everything is perfect with my builds using JPMS with AssertJ (on the newest version).

Haven't had any issues at all, and nothing is jumping out at me.

If I do spot anything though, I will put a PR up/open an issue!

Issue #2932, opened 2023-01-28
Consider adding `isPast()` and `isFuture()` methods for asserting dates


armandino:
#### Feature summary
The new methods would be syntactic sugar for assertions similar to:

```java
assertThat(LocalDate.now().minusDays(1)).isBefore(LocalDate.now());
```

#### Example

```java
assertThat(LocalDate.now().minusDays(1)).isPast();
assertThat(Instant.now().plusSeconds(3)).isFuture();
```

Edit: tasks to track adding `isInThePast()` and `isInTheFuture()`:

- [ ] Instant
- [x] LocalDate
- [ ] LocalDateTime
- [ ] OffsetDateTime
- [ ] ZonedDateTime

joel-costigliola:
We have `isInThePast()` but that's a [date assertion](https://www.javadoc.io/doc/org.assertj/assertj-core/latest/org/assertj/core/api/AbstractDateAssert.html#isInThePast()) so it would make sense to add them for the date/time introduced in java 8.

StefanBratanov:
Hi @joel-costigliola is it possible to work on this issue?

joel-costigliola:
@StefanBratanov thanks! please pick one of the task and raise a PR for it.

armandino:
Awesome. Thank you @joel-costigliola, and @StefanBratanov for taking this on :)

joel-costigliola:
thanks @StefanBratanov for your contribution.

StefanBratanov:
> thanks @StefanBratanov for your contribution.

No worries. Thank you for maintaining the library. :)

PR #2933, opened 2023-01-30
Add `isInThePast` and `isInTheFuture` to LocalDate assertions

StefanBratanov:
#### Check List:
* Fixes one of the tasks in #2932 
* Unit tests : YES

Two things I am unsure about:
- The correctness of the `@since` version
- `ShouldBeInTheFuture/ThePast` error message contains the class name (java.time.LocalDate). This seems to be the behaviour of the `MessageFormatter` handling dates classes. Is the error message ok like this?


joel-costigliola:
Integrated thanks @StefanBratanov for the good work! 
I did some minor changes before integrating it
- no need to use `@DisplayName` this is now done automatically with a junit custom test naming config
- updated the javadoc as the example for `isInTheFutture`  was not that nice (came from date assertions that I also updated)

StefanBratanov:
> Integrated thanks @StefanBratanov for the good work! I did some minor changes before integrating it
> 
> * no need to use `@DisplayName` this is now done automatically with a junit custom test naming config
> * updated the javadoc as the example for `isInTheFutture`  was not that nice (came from date assertions that I also updated)

Thank you @joel-costigliola . Will keep in mind for any future contributions. :)

PR #2968, opened 2023-02-28
Add isRecord() and isNotRecord() Class assertions

ljrmorgan:
#### Check List:
* Fixes NA
* Unit tests : YES

Add `isRecord()` and `isNotRecord()` Class assertions.

As `record` only landed in Java 16 I'm having to use reflection to check whether the class is a record or not. I'm also only able to add unit tests for the negative cases, because I can't compile code that uses `record`.

If this is accepted I thought I'd also open a PR that adds a `hasRecordComponents` assertion.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractClassAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractClassAssert.java
@@ -13,10 +13,14 @@
 package org.assertj.core.api;
 
 import static org.assertj.core.error.ShouldBeAssignableTo.shouldBeAssignableTo;
+import static org.assertj.core.error.ShouldBeRecord.shouldBeRecord;
+import static org.assertj.core.error.ShouldBeRecord.shouldNotBeRecord;
 import static org.assertj.core.error.ShouldNotBeNull.shouldNotBeNull;
 import static org.assertj.core.util.Arrays.array;
 
 import java.lang.annotation.Annotation;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.Objects;
 
 import org.assertj.core.internal.Classes;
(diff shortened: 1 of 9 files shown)


joel-costigliola:
Thanks @ljrmorgan, this one is better targeted to assertj 4.0 which will use java 17

assertj-core/src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -211,6 +211,48 @@ public SELF isNotAnnotation() {
...
+   *
+   * // this assertion fails:
+   * assertThat(String.class).isRecord();</code></pre>
+   *
+   * @return {@code this} assertions object
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} is not a record.
+   */
+  public SELF isRecord() {
+    classes.assertIsRecord(info, actual);
```
scordio, line 231: No need to have the inner Classes layer as it's unlikely that this assertion will be shared somewhere else, please move the actual implementation here as a private method.

assertj-core/src/main/java/org/assertj/core/api/AbstractClassAssert.java
```diff
@@ -211,6 +211,48 @@ public SELF isNotAnnotation() {
...
+   *
+   * // this assertion fails:
+   * assertThat(Jedi.class).isNotRecord();</code></pre>
+   *
+   * @return {@code this} assertions object
+   * @throws AssertionError if {@code actual} is {@code null}.
+   * @throws AssertionError if the actual {@code Class} is a record.
+   */
+  public SELF isNotRecord() {
+    classes.assertIsNotRecord(info, actual);
```
scordio, line 252: Similar to the other comment, let's have the implementation directly here.

assertj-core/src/main/java/org/assertj/core/internal/Classes.java
```diff
@@ -170,6 +172,44 @@ public void assertIsAnnotation(AssertionInfo info, Class<?> actual) {
...
+    if (!isRecord(actual)) throw failures.failure(info, shouldBeRecord(actual));
+  }
+
+  private boolean isRecord(Class<?> actual) {
+    try {
+      Method isRecordMethod = Class.class.getMethod("isRecord");
+      return (boolean) isRecordMethod.invoke(actual);
+    } catch (NoSuchMethodException e) {
+      // Definitely not a record if we're running on a JVM that doesn't support records
+      return false;
```
scordio, line 207: If we go with reflection, we could deliver this as part of 3.x but we should also add an integration test using Java 17 as language level.

Would you mind creating a new `assertj-core-java-17` Maven module under https://github.com/assertj/assertj/tree/main/assertj-tests/assertj-integration-tests? You can tune the Java version in the corresponding POM via the `java.version` property, similar to what we have here:

https://github.com/assertj/assertj/blob/066cc49bb000d902745b726e1bf566cbdf703c30/assertj-tests/assertj-integration-tests/assertj-core-module-path/pom.xml#L15-L17

assertj-core/src/main/java/org/assertj/core/internal/Classes.java
```diff
@@ -170,6 +172,44 @@ public void assertIsAnnotation(AssertionInfo info, Class<?> actual) {
...
+    } catch (Exception e) {
```
scordio, line 208: I'd prefer to keep specific exception types so it's clear what we want to catch.

assertj-core/src/test/java/org/assertj/core/api/classes/ClassAssert_isRecord_Test.java
```diff
@@ -0,0 +1,36 @@
...
+import org.assertj.core.api.ClassAssertBaseTest;
+
+import static org.mockito.Mockito.verify;
+
+/**
+ * Tests for <code>{@link org.assertj.core.api.ClassAssert#isRecord()}</code>.
+ *
+ * @author Louis Morgan
+ */
+public class ClassAssert_isRecord_Test extends ClassAssertBaseTest {
```
scordio, line 25: Once the implementation is moved to AbstractClassAssert, this test class should contain concrete test cases that can run on Java 8 like you've done in Classes_assertIs*Record_Test.

scordio:
> A Java 17 baseline would also remove all the reflection code, which would be great. Is there a `4.x` branch or anything that I could rebase onto?

Not yet but as I mentioned I still think we could ship this in 3.x and change the implementation in 4.x.

@joel-costigliola how do you see it?

joel-costigliola:
I can live with that

assertj-core/src/main/java/org/assertj/core/internal/Classes.java
ljrmorgan: Thanks @scordio, I've added a module as requested. I wasn't quite sure what naming conventions to follow for the test class that I was adding there, I called it `RecordTest` in the end to keep it simple like the other IT modules. I thought that was a reasonable approach since I don't imagine many more tests will be added to that module, but happy to revisit.

scordio:
Thanks for your first contribution, @ljrmorgan!

PR #2969, opened 2023-03-01
Add `hasNullValue` and `doesNotHaveNullValue` to `AtomicReferenceAssert`

ghkim3221:
#### Check List:
* Fixes NA
* Unit tests : YES

`hasNullValue` and `doesNotHaveNullValue` will be useful for checking whether an `AtomicReference` has the null value or not. These are shortcuts of `hasVallue(null)` and `doesNotHaveValue(null)` and would be useful when using package which non-null is the default since IDE may warn that `hasVallue(null)` receives `null`.

### DIFF
--- a/assertj-core/src/test/java/org/assertj/core/api/atomic/reference/AtomicReferenceAssert_doesNotHaveNullValue_Test.java
+++ b/assertj-core/src/test/java/org/assertj/core/api/atomic/reference/AtomicReferenceAssert_doesNotHaveNullValue_Test.java
@@ -1,0 +1,26 @@
+package org.assertj.core.api.atomic.reference;
+
+import org.junit.jupiter.api.Test;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.assertj.core.util.FailureMessages.actualIsNull;
+
+class AtomicReferenceAssert_doesNotHaveNullValue_Test {
+
+  @Test
+  void should_pass_when_actual_does_not_have_the_null_value() {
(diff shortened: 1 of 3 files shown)


assertj-core/src/test/java/org/assertj/core/api/atomic/reference/AtomicReferenceAssert_doesNotHaveNullValue_Test.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.assertj.core.error.ShouldNotContainValue.shouldNotContainValue;
+
+class AtomicReferenceAssert_doesNotHaveNullValue_Test {
```
joel-costigliola, line 23: Can you follow the testing guidelines documented in https://github.com/assertj/assertj/blob/main/CONTRIBUTING.md notably using GIVEN WHEN THEN and `AssertionUtil.expectAssertionError`.

this comment applies to the other test too, thanks!

joel-costigliola (changes requested):
Thanks  @ghkim3221, first round of code review done.

assertj-core/src/test/java/org/assertj/core/api/atomic/reference/AtomicReferenceAssert_doesNotHaveNullValue_Test.java
ghkim3221: Sure!! thanks.

joel-costigliola:
Integrated thanks @ghkim3221!

PR #2995, opened 2023-03-22
Add hasRecordComponents() Class assertion

ljrmorgan:
#### Check List:
* Fixes NA
* Unit tests : YES

Add a `hasRecordComponents()` Class assertion, following on from #2968.

I've tried to take a similar approach to assertions like `hasPublicFields()`, but as with `isRecord()` and `isNotRecord()` I've left the implementation of the assertion in `AbstractClassAssert` rather than extracting it to `Classes`.


assertj-core/src/main/java/org/assertj/core/error/ShouldHaveRecordComponents.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  /**
+   * Creates a new <code>{@link org.assertj.core.error.ShouldHaveRecordComponents}</code>
+   *
+   * @param actual the actual value in the failed assertion.
+   * @param actualRecordComponents the actual record components for this class
+   * @return the created {@code ErrorMessageFactory}.
+   */
+  public static ErrorMessageFactory shouldHaveNoRecordComponents(Class<?> actual, Set<String> actualRecordComponents) {
```
scordio, line 43: I think this deserves its own `ShouldHaveNoRecordComponents` class

scordio (commented):
Thanks @ljrmorgan, just two comments

ljrmorgan:
Thanks for reviewing @scordio!

I was trying to copy the behaviour of `hasPublicFields()`/`hasDeclaredFields()` etc. which when they are called with no arguments assert that the class has no public/declared field. TBH though I found that behaviour a bit surprising, and I prefer your suggestion of just requiring that at least one parameter is passed to `hasRecordComponents()` (presumably by changing the signature to something like `hasRecordComponents(String first, String... rest)`?)

I'm less sure about adding a `hasNoRecordComponents()` assertion though. I can add it if you feel strongly,  but I'm wondering if that's actually a useful thing to assert? I can understand why people might want to assert that a class has no public fields, I can't really think of a scenario where someone would to check that a record has no record components (or even why someone would want to create a record with no record components). I'm happy to go with whatever you think though.

Thanks!

scordio:
> I'm less sure about adding a `hasNoRecordComponents()` assertion though.

Fair enough, nothing prevents us to add it later if users come back with a concrete use case. For now, let's skip it.

scordio:
This is now merged. Thanks, @ljrmorgan!

PR #3003, opened 2023-03-25
Fix ignoringFields with different map size

### LINKED ISSUE #2988: usingRecursiveComparison().ignoringFields("fieldName") is not working properly when maps sizes are different
#### Summary
usingRecursiveComparison().ignoringFields("fieldName") is not working properly when comparing two maps with different sizes.

#### Example
using the following code the ingested node will be the same size as the expected one after ignoring "ingestionDate". So I suppose that start comparing the size of the maps makes the method more limited such as in this case for example.
```java
assertThat(ingestedNode).usingRecursiveComparison().ignoringFields("ingestionDate").isEqualTo(expectedNode);
```

OlivierCavadenti:
Fix ignoringFields behavior that doesn't work properly when using maps of different sizes with usingRecursiveComparison and isEqualTo for example.

Fix #2988

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
@@ -15,6 +15,7 @@ package org.assertj.core.api.recursive.comparison;
 import static java.lang.String.format;
 import static java.util.Objects.deepEquals;
 import static java.util.stream.Collectors.joining;
+import static java.util.stream.Collectors.toMap;
 import static org.assertj.core.api.recursive.comparison.ComparisonDifference.rootComparisonDifference;
 import static org.assertj.core.api.recursive.comparison.DualValue.DEFAULT_ORDERED_COLLECTION_TYPES;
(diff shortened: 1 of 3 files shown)


joel-costigliola:
@OlivierCavadenti have you got time to finish this PR, if not, that's ok I can finish (you'll get full credit for it anyway :wink:)

joel-costigliola:
I'm taking over this one @OlivierCavadenti as the fix is more complex than I thought, but you'll get credit :wink: .

The fix should check if the key is to be ignored according to the configuration, that includes:

- ignored fields 
- ignored fields of type
- compared types (if the value is not a compared type, it must be ignored)
- compared fields (if the value is not a compared type, it must be ignored)

OlivierCavadenti:
> I'm taking over this one @OlivierCavadenti as the fix is more complex than I thought, but you'll get credit 😉 .
> 
> The fix should check if the key is to be ignored according to the configuration, that includes:
> 
> * ignored fields
> * ignored fields of type
> * compared types (if the value is not a compared type, it must be ignored)
> * compared fields (if the value is not a compared type, it must be ignored)

Hi ! 
Thanks and sorry for not finish this one... !

joel-costigliola:
Finally integrated, thanks @OlivierCavadenti!

PR #3007, opened 2023-03-30
Add getters for field path in ComparisonDifference

aakatov:
Check List:
Fixes absense of getters for field path in ComparisonDifference
Unit tests : NA
Javadoc with a code example (on API only) : NA
PR meets the [contributing guidelines](https://github.com/assertj/assertj/blob/main/CONTRIBUTING.md)

org.assertj.core.api.recursive.comparison.ComparisonDifference unnecessery closes access to both decomposedPath and concatenatedPath.
org.assertj.core.api.recursive.comparison.RecursiveComparisonDifferenceCalculator#determineDifferences is a public method but you cannot get a field path from its result. It would be nice to have getters on it to use in an application code.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/ComparisonDifference.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/ComparisonDifference.java
@@ -92,6 +92,10 @@ public class ComparisonDifference implements Comparable<ComparisonDifference> {
     return additionalInformation;
   }
 
+  public List<String> getDecomposedPath() {
+    return decomposedPath;
+  }
+
   @Override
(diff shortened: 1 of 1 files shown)


scordio:
Would you have a concrete use case to share?

aakatov:
@scordio I have an application where as a part of regression testing we create a lot of entity objects, and we need to compare them field by field with entities we already have in the system. By the end, we generate an Excel report with differences:
image


I use RecursiveComparisonDifferenceCalculator for it, but I can't get field names and their path from the result of comparison.

aakatov:
@joel-costigliola I don't need both. I'll keep a getter for `decomposedPath` then.

PR #3021, opened 2023-04-18
Use hash code for compareUnorderedIterables to improve performance in…

### LINKED ISSUE #3020: .usingRecursiveComparison().ignoringCollectionOrder() is very slow for medium to large unordered collections
Comparing two unordered collections recursively is very slow if the two collections are not small and their elements are not in a similar order.

I've had a look through the implementation in `RecursiveComparisonDifferenceCalculator#compareUnorderedIterables` and it looks like the complexity in the worst case is quadratic.

I think this could be improved by using a hash map to avoid unnecessary comparisons, but this would require a deep hash implementation that is consistent with the deep equals configuration. (There was a `deepHashCode` method in an earlier version of `RecursiveComparisonDifferenceCalculator`, but it was removed in commit 7406133f.)

I've tested this with assertj version 3.24.2.

Some example timed comparisons:

```java
@Test
public void test_usingRecursiveComparison() {
    List<Integer> range100 = IntStream.range(0, 100).boxed().toList();
    List<Integer> range1000 = IntStream.range(0, 1000).boxed().toList();
    List<Integer> range10000 = IntStream.range(0, 10000).boxed().toList();

    // everything works fine for comparisons of ordered collections
    Stopwatch stopwatch = Stopwatch.createStarted();
    assertThat(range100)
            .usingRecursiveComparison()
            .isEqualTo(range100);
    System.out.println(stopwatch);  // 176.6 ms
    stopwatch = Stopwatch.createStarted();
    assertThat(range1000)
            .usingRecursiveComparison()
            .isEqualTo(range1000);
    System.out.println(stopwatch);  // 53.10 μs
    stopwatch = Stopwatch.createStarted();
    assertThat(range10000)
            .usingRecursiveComparison()
            .isEqualTo(range10000);
    System.out.println(stopwatch);  // 32.40 μs

    List<Integer> reversed100 = IntStream.range(0, 100).boxed()
            .sorted(comparing(x -> -x)).toList();
    List<Integer> reversed1000 = IntStream.range(0, 1000).boxed()
            .sorted(comparing(x -> -x)).toList();
    List<Integer> reversed10000 = IntStream.range(0, 10000).boxed()
            .sorted(comparing(x -> -x)).toList();

    // but comparisons of unordered collections become slow very quickly
    stopwatch = Stopwatch.createStarted();
    assertThat(range100)
            .usingRecursiveComparison()
            .ignoringCollectionOrder()
            .isEqualTo(reversed100);
    System.out.println(stopwatch);  // 153.8 ms
    stopwatch = Stopwatch.createStarted();
    assertThat(range1000)
            .usingRecursiveComparison()
            .ignoringCollectionOrder()
            .isEqualTo(reversed1000);
    System.out.println(stopwatch);  // 1.471 s
    stopwatch = Stopwatch.createStarted();
    assertThat(range10000)
            .usingRecursiveComparison()
            .ignoringCollectionOrder()
            .isEqualTo(reversed10000);
    System.out.println(stopwatch);  // 1.639 min
}
```

JozsefKutas:
… some common cases. See #3020

#### Check List:
* Fixes #3020
* Unit tests : NO

This changes `RecursiveComparisonDifferenceCalculator#compareUnorderedIterables` to use a hash map to keep track of elements of the actual iterable. When comparing elements of the expected iterable, actual values with the same hash code are compared first. This still has quadratic time complexity in the worst cases (if the hash implementation of the elements is not compatible with the recursive comparison, or if the two iterables have lots of different elements), but will be faster in some common cases.

I've checked that this fixes the particular test case provided in #3020, but I'm not sure this needs a unit test as it's just a performance issue.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
@@ -14,12 +14,12 @@ package org.assertj.core.api.recursive.comparison;
 
 import static java.lang.String.format;
 import static java.util.Objects.deepEquals;
+import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.joining;
 import static org.assertj.core.api.recursive.comparison.ComparisonDifference.rootComparisonDifference;
 import static org.assertj.core.api.recursive.comparison.DualValue.DEFAULT_ORDERED_COLLECTION_TYPES;
 import static org.assertj.core.api.recursive.comparison.FieldLocation.rootFieldLocation;
 import static org.assertj.core.util.IterableUtil.sizeOf;
-import static org.assertj.core.util.IterableUtil.toCollection;
 import static org.assertj.core.util.Lists.list;
 import static org.assertj.core.util.Sets.newHashSet;
 
@@ -30,9 +30,9 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


joel-costigliola:
Integrated thanks @JozsefKutas !

PR #3036, opened 2023-05-07
Fixes a bug when failing assertions on DirectoryStream types

ascopes:
The DirectoryStream class is somewhat special because the API allows it to only support .iterator() being called at most once on any implementations. Due to the nature of how the StandardRepresentation currently works for iterable types, any assertion error that is thrown as a result of assertions failing on a DirectoryStream object can result in IllegalStateExceptions being propagated to the caller if .iterator() has already been called on the class. This can break further if using soft assertions, as the state of the class could change unexpectedly resulting in different exceptions being raised depending on the execution order.

I've added a mechanism to deal with allowing the definition of so-called 'blacklisted' iterable types, just in case other cases like this appear in the JRE standard library anywhere in the future. This will prevent unexpected errors occuring when dealing with the NIO FileSystem APIs.

This was discovered as a result of https://github.com/ascopes/java-compiler-testing/pull/450, where @marschall pointed out that the `#iterator()` on a `DirectoryStream` should only be called once in some implementations.

### DIFF
--- a/assertj-core/src/test/java/org/assertj/core/api/iterable/IterableAssert_doNotBreakOnFailingTestsForDirectoryStreams.java
+++ b/assertj-core/src/test/java/org/assertj/core/api/iterable/IterableAssert_doNotBreakOnFailingTestsForDirectoryStreams.java
@@ -1,0 +1,100 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2023 the original author or authors.
+ */
+package org.assertj.core.api.iterable;
+
+import org.assertj.core.api.IterableAssert;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.SecureDirectoryStream;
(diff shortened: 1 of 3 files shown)


joel-costigliola:
Thanks for the analysis and the PR @ascopes

assertj-core/src/test/java/org/assertj/core/api/iterable/IterableAssert_doNotBreakOnFailingTestsForDirectoryStreams.java
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Checks that we can fail an assertion on DirectoryStream types without accidentally
+ * dereferencing the items in the iterable. This is important for this class as the
+ * iterators provided by directory streams can only be used once. API contracts
+ * allow this class to throw exceptions if {@code #iterator()} is called multiple times.
+ *
+ * @author Ashley Scopes
+ */
+public class IterableAssert_doNotBreakOnFailingTestsForDirectoryStreams {
```
joel-costigliola, line 40: Replace `assertThat` by `then`from `BDDAssertions`

assertj-core/src/test/java/org/assertj/core/api/iterable/IterableAssert_doNotBreakOnFailingTestsForDirectoryStreams.java
```diff
@@ -0,0 +1,100 @@
...
+    OneShotDirectoryStream stream = new OneShotDirectoryStream();
+    IterableAssert<Path> assertions = new IterableAssert<>(stream);
+
+    // When
+    // Some code that makes use of the iterator() method prior to running the assertions
+    stream.iterator();
+
+    // Then
+    assertThatThrownBy(() -> assertions.isInstanceOf(SecureDirectoryStream.class))
+      .isInstanceOf(AssertionError.class);
```
joel-costigliola, line 67: use `expectAssertionError(() -> assertions.isInstanceOf(SecureDirectoryStream.class))` from `AssertionUtil`

joel-costigliola (commented):
looks good, just minor comments

joel-costigliola:
Integrated thanks @ascopes !

Issue #3043, opened 2023-05-14
Improve 'Expecting code to raise a throwable' when lambda returns a value


findepi:
`assertThatThrownBy` throws "Expecting code to raise a throwable" when called code doesn't throw.
When test assertion fails, it's tempting to know what did the code do instead.

Enhance "Expecting code to raise a throwable" to mention produced value in case when the lambda is a producer.

scordio:
The related change has been reverted in #3318 and the topic should be evaluated again together with #1652 and #2519.

findepi:
@scordio apologies for https://github.com/assertj/assertj/issues/3314 and thank you for the quick revert.

scordio:
No problem at all, @findepi! Unfortunately, we missed that potential breaking change but now we learned 🙂

sbrannen:
- Related to https://github.com/junit-team/junit5/issues/1576

PR #3044, opened 2023-05-14
Provide value when assertThatThrownBy/thenThrownBy fail

findepi:
#### Check List:
* Fixes https://github.com/assertj/assertj/issues/3043
* Unit tests : YES
* Javadoc with a code example (on API only) : NO, not yet
* PR meets the [contributing guidelines](https://github.com/assertj/assertj/blob/main/CONTRIBUTING.md) -- i hope so!

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/Assertions.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/Assertions.java
@@ -75,6 +75,7 @@ import java.util.stream.LongStream;
 import java.util.stream.Stream;
 
 import org.assertj.core.api.ThrowableAssert.ThrowingCallable;
+import org.assertj.core.api.ThrowableAssert.ThrowingCallableWithValue;
 import org.assertj.core.api.filter.FilterOperator;
 import org.assertj.core.api.filter.Filters;
 import org.assertj.core.api.filter.InFilter;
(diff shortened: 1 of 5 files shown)


assertj-core/src/main/java/org/assertj/core/api/Assertions.java
```diff
@@ -1210,6 +1212,17 @@ public static <T extends Throwable> AbstractThrowableAssert<?, T> assertThat(T a
...
+  @CanIgnoreReturnValue
+  public static AbstractThrowableAssert<?, ? extends Throwable> assertThatThrownBy(ThrowingCallableWithValue shouldRaiseThrowable) {
+    Object value;
+    try {
+      value = shouldRaiseThrowable.call();
+    } catch (Throwable throwable) {
+      return assertThat(throwable);
+    }
+    throw Failures.instance().failure(String.format("Expecting code to raise a throwable, but it returned [%s] instead", value));
+  }
```
findepi, line 1224: I am pretty sure that this is not how it should eventually look like.
Very minimum would be a javadoc.
This is for issue illustration / POC at this stage.

Happy to improve if this is at least directionally correct.

joel-costigliola:
Thanks @findepi, I will review your PR shortly!

assertj-core/src/main/java/org/assertj/core/api/Assertions.java
joel-costigliola: This looks good to me.

We need to do the same in BDDAssertions

joel-costigliola (changes requested):

findepi:
Thanks @joel-costigliola for your review!
added similar change in BDDAssertions. Also added this for overloads that take `String description, Object... args`.

joel-costigliola:
Integrated thanks @findepi!

findepi:
thank you Joel!

PR #3080, opened 2023-06-12
Add `isSealed` and `isNotSealed` to `Class` assertions

scordio:


scordio:
Good idea, @geneva78. I raised #3263 to track it.

PR #3088, opened 2023-06-17
[mvn] Update maven wrapper

hazendaz:
- use secure license url
- make sure releases include sha512 and sha256 checksums in addition to defaults
- make sure build considers downloading stronger checksums
- bump to maven 3.9.2

note: 
* running wrapper to update will not produce secure license url at this time.
* by default, maven now uses header data to determine what checksum to download.  Until all package managers catch up, that means it will fall back to sha1 or md5.  The smart checksums false allows it to check for secure checksums.
* The checksum algorithms will be deployed in additionl to the defaults sha1 and md5 with this change.

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,0 +1,2 @@
+-Daether.checksums.algorithms=SHA-512,SHA-256,SHA-1,MD5
+-Daether.connector.smartChecksums=false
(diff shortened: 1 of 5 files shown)


joel-costigliola:
Integrated thanks @hazendaz!

Issue #2978, opened 2023-03-09
Add convenient values() extractor method in MapAssert


Skym0sh0:
#### Feature summary
It would be nice if there was a method in MapAssert that returns a (Abstract)CollectionAssert over the values contained in the map.

Example:
```java
Assertions.assertThat(Map.of("str", 1))
	.values()
	.contains(1);
```

Currently I have to use the following:

```java
Assertions.assertThat(Map.of("str", 1))
	.extracting(Map::values, as(collection(Integer.class)))
	.contains(1);
```

joel-costigliola:
We are ok with this improvement, likely will provide `values()` to switch the object under test to a `Collection<Object>` and `values(Class valueType)` to a `Collection<Class>`.

kang-mansher:
Can I work on this improvement

joel-costigliola:
Sure go ahead @kang-mansher, thanks!

igmtz:
Hello @joel-costigliola can I contribute? 😄

joel-costigliola:
@kang-mansher are you still interested in contributing? if not @igmtz would be keen to.

scordio:
Closing as duplicate of #3291.

PR #3106, opened 2023-07-12
add value() method to MapAssert to allow assertions on values set

mmsikorski:
**Feature summary:**
The aim of this PR is to add the methods values() and values(Class<T> clazz) to enable assertions on a set of values within a given map.

The issue that prompted this enhancement can be found at the following link:
https://github.com/assertj/assertj/issues/2978 .


scordio:
Apologies, @mmsikorski, this slipped under our radar, and the same topic has been addressed in #3297 🙁

Issue #3065, opened 2023-06-02
OOME while just trying to assert long array.length


onacit:
**Describe the bug**

Given, an instance of `byte[]` whose length is, incidentally, `98007000`,
I tried to verify its length.

```java
final var bytes = getBytes();
final var expectedSize = whatever-not-98007000;
assertThat(bytes).isNotNull().hasSize(expectedSize);
```

And an `OutOfMermoryError` thrown, seems it tried to create a nice message.

```
Exception in thread "main" java.lang.OutOfMemoryError: Java heap space
	at java.base/java.lang.Integer.toString(Integer.java:457)
	at java.base/java.lang.Byte.toString(Byte.java:441)
	at org.assertj.core.presentation.StandardRepresentation.toStringOf(StandardRepresentation.java:348)
	at org.assertj.core.presentation.StandardRepresentation.toStringOf(StandardRepresentation.java:250)
	at org.assertj.core.presentation.StandardRepresentation.safeStringOf(StandardRepresentation.java:621)
	at org.assertj.core.presentation.StandardRepresentation.lambda$representElements$0(StandardRepresentation.java:628)
	at org.assertj.core.presentation.StandardRepresentation$$Lambda$508/0x00000008003b4a40.apply(Unknown Source)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:1006)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
```

etellman:
Here's what I came up with: https://github.com/assertj/assertj/pull/3109. Let me know what you think of the approach.

This is my first open source contribution, so I may have messed something up with the logistics. Also me know if I created the pull request correctly, etc.

PR #3109, opened 2023-07-16
Format large arrays as strings

etellman:
The process for converting an array of primitives to a string was:
1. convert the array to an array of objects
2. convert the objects to strings
3. select up to 1000 elements to print

This is fine if there aren't too many elements, but if there are 2^30 elements, all the copying may result in an OOM exception.

With this change, instead of copying the arrays, create a collection that proxies the original collection as a list of Strings. The strings aren't created unless they're actually needed.

Testing:

Originally, on my computer, trying to turn big arrays into strings resulted in OOM error after about 10 seconds of computation. Now, this completes in a fraction of a second.

#### Check List:
* Fixes https://github.com/assertj/assertj/issues/3065
* Unit tests : YES
* PR meets the [contributing guidelines](https://github.com/assertj/assertj/blob/main/CONTRIBUTING.md) I hope so :^)

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/presentation/PrimitiveArrayList.java
+++ b/assertj-core/src/main/java/org/assertj/core/presentation/PrimitiveArrayList.java
@@ -1,0 +1,44 @@
+package org.assertj.core.presentation;
+
+import org.assertj.core.util.Arrays;
+import org.assertj.core.util.Preconditions;
+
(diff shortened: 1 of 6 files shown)


joel-costigliola:
Thanks @etellman, I think your approach makes a lot of sense.

joel-costigliola:
Integrated thanks @etellman! 
I mainly improved the tests and did some cosmetic changes.

Issue #3087, opened 2023-06-17
NPE in StandardRepresentation for maps with null entries


tkruse:
**Describe the bug**

When using a non-standard Java map where entries can be null, NPEs happen when assertJ tries to generate failure messages representing the map content.

Note that Map code allowing null map entries in `.entrySet()` are probably buggy, but I cannot tell if it really break a contract.

```
java.lang.NullPointerException
	at org.assertj.core.presentation.StandardRepresentation.toStringOf(StandardRepresentation.java:491)
	at org.assertj.core.presentation.StandardRepresentation.toStringOf(StandardRepresentation.java:236)
	at org.assertj.core.presentation.CompositeRepresentation.toStringOf(CompositeRepresentation.java:40)
        ...
```

* assertj core version: 3.24.2
* java version: 8
* test framework version: Junit4
* os (if relevant): Ubuntu

**Test case reproducing the bug**

To reproduce:
```
public class NullEntryMapTest {

    static class NullEntryMap<K, V> extends ConcurrentHashMap<K, V> {
        @Override
        public Set<Entry<K, V>> entrySet() {
            Set<Map.Entry<K, V>> set = new HashSet<>();
            set.add(null);
            return set;
        }
    }

    @Test
    public void testStandardRepresentationWithNullEntryMap() {
        StandardRepresentation.STANDARD_REPRESENTATION.toStringOf(new NullEntryMap<>());
    }

    @Test
    public void testContains() {
        assertThat(new NullEntryMap<>()).containsKey("foo"); // actual assertion does not matter
    }
}
```

To fix, StandardRepresentation should check the entry is not null.
It might be good to also check other occurences of entry.getKey() /entry.getValue() in the code.

I notice IntelliJ as an example handles this case in it's representation of maps.

joel-costigliola:
Thanks for reporting this @tkruse!

etellman:
I was looking for something easy to do, and this looks pretty easy. Can I take this one?

joel-costigliola:
Thanks @etellman, it should not be to complicated since it is only related to maps.

Cheers

joel-costigliola:
We only want to fix the NPE in the `StandardRepresentation`, I would not change the behavior of `containsKey` as part of this issue, this could be discussed in a different ticket if it is based on an actual use case.

etellman:
I probably didn't explain this clearly, but I don't think there should make any change for this. The code as written now seems fine to me.

[...]

PR #3110, opened 2023-07-16
Handle null values in map entry sets when formatting

etellman:
If the map entry set contains a null value, print "null" instead of throwing a NPE.

see: https://github.com/assertj/assertj/issues/3087

#### Check List:
* Fixes[ #??? (ignore if not applicable)](https://github.com/assertj/assertj/issues/3087)
* Unit tests : YES 


joel-costigliola:
Integrated thanks @etellman!

Issue #3034, opened 2023-05-03
Null pointer using withEqualsForFieldsMatchingRegexes when values are null


UncleInf:
When using `withEqualsForFieldsMatchingRegexes` and both original and compare values are null NPE is thrown in: 
```
/assertj-core-3.24.2-sources.jar!/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java:172
```
`getCustomErrorMessage` does not have custom message registered and tries getting from value's class, and since value is null we get a NPE.

Since `withErrorMessageForFields` does not work for regexp values there is no way to walkaround this. 
Having ability to have custom message for regexp values would be also great, but I guess this is out of scope for this ticket.

* assertj core version: 3.24.2
* java version: 17

etellman:
I can take a look at this one.

etellman:
I think this fixes it: https://github.com/assertj/assertj/pull/3112

PR #3112, opened 2023-07-17
Fix NPE when expected and actual field values are null

etellman:
When both the expected and actual field values are null, fall back to the default message instead of throwing a NPE when trying to find a custom message for the value type.

see: https://github.com/assertj/assertj/issues/3034

#### Check List:
* Fixes https://github.com/assertj/assertj/issues/3034
* Unit tests : YES /


joel-costigliola:
Integrated thanks @etellman! :smiley_cat:

PR #3120, opened 2023-07-24
Normalize breakingSpaces like regular whiteSpaces

### LINKED ISSUE #3113: isEqualToNormalizingWhitespace doesn't normalise non breaking space characters
**Describe the bug**
isEqualToNormalizingWhitespace doesn't normalise non breaking space characters

* assertj core 3.21:
* java version: 17
* Kotlin 1.9.0
* test framework version: JUnit 5.9.1
* os (if relevant): 

**Compare 2 strings with one having a [NBSP](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/-typography/nbsp.html) character instead of a normal space**

```Kotlin
val result = "10,00 € to 100NBSP000,00 € per year"
val expected = "10,00 € to 100 000,00 € per year"
assertThat(result).isEqualToNormalizingWhitespace(expected)
```

maximedezette:
#### Check List:
* Fixes #3113 
* Unit tests : YES

# Note

I think that the breaking-space management could be improved by using a public constant so that we don't have to update all the tests if some of the breaking space codes change over time. It would look like this:

```java
    Stream<Arguments> breakingSpaces = Arrays.stream(Strings.BREAKING_SPACES)
      .map(breakingSpace -> Arguments.of("my" + breakingSpace + "foo bar", "my foo bar"));
    Stream<Arguments> regularSpaces = Stream.of(Arguments.of("my   foo bar", "my foo bar"),
      Arguments.of("  my foo bar  ", "my foo bar"),
      Arguments.of(" my\tfoo bar ", " my foo bar"),
      Arguments.of(" my foo    bar ", "my foo bar"),
      Arguments.of(" my foo    bar ", "  my foo bar   "),
      Arguments.of("       ", " "),
      Arguments.of(" my\tfoo bar ", new String(arrayOf(' ', 'm', 'y', ' ', 'f', 'o', 'o', ' ', 'b', 'a', 'r'))),
      Arguments.of(" my\tfoo bar ", " my\tfoo bar "),   // same
      Arguments.of(null, null),   // null
      Arguments.of(" \t \t", " "),
      Arguments.of(" abc", "abc ")
    );
    return Stream.concat(regularSpaces, breakingSpaces);

```
Instead of having to update all the tests:

image

What do you think ?


assertj-core/src/main/java/org/assertj/core/internal/Strings.java
```diff
@@ -469,6 +475,11 @@ private static String normalizeWhitespace(CharSequence toNormalize) {
     return result.toString().trim();
   }
 
+  private static boolean isBreakingSpace(Character character) {
+    return stream(BREAKING_SPACES)
+      .anyMatch(character::equals);
```
maximedezette, line 480: I could have written:

`java.util.Arrays.asList(BREAKING_SPACES).contains(character);` which is equivalent but I don't think that its more readable. If you think that's better, I can change.

assertj-core/src/main/java/org/assertj/core/internal/Strings.java
```diff
@@ -458,7 +464,7 @@ private static String normalizeWhitespace(CharSequence toNormalize) {
     boolean lastWasSpace = true;
     for (int i = 0; i < toNormalize.length(); i++) {
       char c = toNormalize.charAt(i);
-      if (isWhitespace(c)) {
+      if (isWhitespace(c) || isBreakingSpace(c)) {
```
maximedezette, line 467: I didn't change the `isWhiteSpace` method because the `isNotBlank` assertion concider breakingSpaces as regular character and uses `isWhiteSpace` method. 

If there are more methods that need to concider that breakingSpace are the same as regular ones, we should concider doing a dedicated methods like "isRegularWhiteSpace" and "isWhiteSpace" which could includes breakingSpaces 🤔

assertj-core/src/test/java/org/assertj/core/internal/strings/Strings_assertEqualsNormalizingWhitespace_Test.java
```diff
@@ -80,6 +80,10 @@ public static Stream<Arguments> equalNormalizingWhitespaceGenerator() {
                      Arguments.of(" my\tfoo bar ", " my\tfoo bar "),   // same
                      Arguments.of(null, null),   // null
                      Arguments.of(" \t \t", " "),
-                     Arguments.of(" abc", "abc "));
+                     Arguments.of(" abc", "abc "),
+                     Arguments.of("my\u00A0foo bar", "my foo bar"), // non-breaking space
+                     Arguments.of("my\u2007foo bar", "my foo bar"), // non-breaking space
+                     Arguments.of("my\u202Ffoo bar", "my foo bar")); // non-breaking space
+
```
scordio, line 87: > I think that the breaking-space management could be improved by using a public constant so that we don't have to update all the tests if some of the breaking space codes change over time.

Feel free to pull constants into `StringsBaseTest`, similar to what `MapsBaseTest` does with map instances.

scordio (changes requested):
Thank you, @maximedezette! I've put some comments.

Could you please also enhance the Javadoc as mentioned in https://github.com/assertj/assertj/issues/3113#issuecomment-1650584857? 

Lastly, the build is failing due to formatting issues. You can run `./mvnw spotless:apply` to make sure that all changes adhere to the expected code style.

assertj-core/src/main/java/org/assertj/core/internal/Strings.java
maximedezette: My comment was very confused sorry "^^

I was concidering defining a custom method named `isWhitespace`  that could be used instead of `Character::isWhitespace` but I was concerned that doing so would break the `isNotBlank` assertion (if all `Character::isWhitespace` call are replaced with a custom  `isWhitespace`  method defined as:

```java  
private static boolean isWhitespace(char c) {
    return Character.isWhitespace(c) || NON_BREAKING_SPACES.contains(c);
  }
```
)

So I concidered adding an other method called `isRegularWhiteSpace` to not break assertions like `isNotBlank` to have a method when we need to concider non-breaking spaces and an other when we need to ignore them.

```java  
private static boolean isRegularWhitespace(char c) {
    return Character.isWhitespace(c);
  }
```

assertj-core/src/main/java/org/assertj/core/internal/Strings.java
```diff
@@ -108,6 +108,12 @@ public class Strings {
   private static final String EMPTY_STRING = "";
   private static final Strings INSTANCE = new Strings();
   private static final String PUNCTUATION_REGEX = "\\p{Punct}";
+
+  private static final Character[] BREAKING_SPACES = new Character[]{
+    '\u00A0',
+    '\u2007',
+    '\u202F'
+  };
```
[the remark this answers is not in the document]
maximedezette: You're right  :slightly_smiling_face: Done :+1:

assertj-core/src/test/java/org/assertj/core/internal/strings/Strings_assertEqualsNormalizingWhitespace_Test.java
maximedezette: I've done that in an other commit :+1:

assertj-core/src/main/java/org/assertj/core/internal/Strings.java
scordio: Superseded by the usage of a static `Set`.

scordio:
@maximedezette This PR has now several merge conflicts but you don't need to worry about them: I've taken the opportunity to do some refactoring in the surrounding area to better evaluate if we should also change other methods relying on whitespace.

Once I'm done with the refactoring, I'll cherry-pick your changes maintaining you as the author.

maximedezette:
@scordio I have missed your comment mentioning the cherry-pick and I've done the rebase :sweat_smile: 

I have thrown away a bit of code about the refactoring of non-breaking spaces tests but I'll do an other PR if I feel that it could improve the tests after your refactoring :+1:

scordio:
@maximedezette sorry for the slow feedback!

I just refreshed your branch and made some cosmetic changes.

Also, I noticed that this change will also affect `isNotEqualToNormalizingWhitespace` and `isEqualToNormalizingPunctuationAndWhitespace`. I already aligned the Javadoc but we should also have tests for those assertions.

Would you like to add the missing tests? No problem if you don't have time, I can take care of them next week.

assertj-core/src/main/java/org/assertj/core/internal/Strings.java
scordio: For now, let's keep the implementation as it is. If we get feedback that more should be changed, we can think about how to refactor this area.

maximedezette:
@scordio I don't have time at the moment but if it can wait a few days, I should be able to add those tests this week :+1:

scordio (approved):

scordio:
Thanks for your first contribution, @maximedezette!

PR #3124, opened 2023-07-25
Fix PrimitiveArralList/TransformlingList test file names

etellman:
The test classes for PrimitiveArrayList and TransformingList follow the '_Test' naming convention but the file names were missing the underscore. It doesn't really matter, but it looked a little weird in the IDE.

#### Check List:
* Fixes # : just cosmetic
* Unit tests : just file rename


joel-costigliola:
Merged, good catch @etellman!

Issue #3123, opened 2023-07-25
Converting large collections and streams to strings can be more efficient


etellman:
**Describe the bug**
Converting a large collection with a time-consuming string conversion for each element takes longer than it needs to because AssertJ converts all the elements in the collection to strings but only uses the first and last 500 strings.

This is related to https://github.com/assertj/assertj/issues/3065, but for Streams and Iterables instead of arrays.

* assertj core version: 3.25.0-SNAPSHOT
* java version: openjdk 17.0.7
* test framework version: JUnit5 
* os (if relevant): OSX, but probably not relevant

**Test case reproducing the bug**

Converting this collection a string takes 27 minutes on my MacBook:
```
    int elementsPerArray = 1000;
    List<int[]> numbers = new ArrayList<>();
    for (int i = 0; i < 1 << 20; i++) {
      numbers.add(new int[elementsPerArray]);
    }
```

It's also possible to get OOM error with more elements in the list and smaller arrays in each element.

A possible fix would be to narrow the stream down to the elements are actually needed and then just convert those to strings. I think this can be done by changing `StandardRepresentation.representElements()` to:

```
  private List<String> representElements(Stream<?> elements, String start, String end, String elementSeparator,
                                         String indentation, Object root) {
    // new
   final PrintingAccumulator accumulator = new PrintingAccumulator(maxElementsForPrinting);
   elements.forEach(accumulator::add);

   // same, but uses narrowed down list of elements from the accumulator instead of the original stream
   return accumulator.toList().stream().map(element -> safeStringOf(element, start, end, elementSeparator, indentation, root))
                  .collect(toList());
  }
```
`PrintingAccumulator` only retains the elements that will be used in the final string conversion and discards everything in the middle of the collection.

I gave this a quick try and it produced the correct string in around 2 seconds. I can clean this up a bit and attach a PR with the proposed solution.

etellman:
I just noticed this very trivial thing to fix...I failed to follow the test class name convention in my last PR, leaving out the underscore before 'Test'. You fixed the class names before committing, but the file names were still off, so this PR makes the file names match the class names: https://github.com/assertj/assertj/pull/3124

I would have attached to the previous issue, but that one is already closed and it didn't seem worth re-opening just for this.

etellman:
I created a new PR with a much smaller test, since the previous one causes OOM for the automated build system: https://github.com/assertj/assertj/pull/3133

[...]

PR #3126, opened 2023-07-27
Improve efficiency of iterable string conversion

etellman:
Instead of converting all the elements to strings and then narrowing the strings down to the head and tail of the list, narrow the elements down first.

see: https://github.com/assertj/assertj/issues/3123

#### Check List:
* Fixes https://github.com/assertj/assertj/issues/3123
* Unit tests : YES 


assertj-core/src/test/java/org/assertj/core/presentation/StandardRepresentation_iterable_format_Test.java
```diff
@@ -260,6 +264,22 @@ public void should_use_overridden_toString_over_iterable_representation_in_colle
     then(formatted).isEqualTo("[{\"a\":1}]");
   }
 
+  @Test
+  void should_format_big_list() {
+    // GIVEN
+    int elementsPerArray = 1000;
+    List<int[]> numbers = new ArrayList<>();
+    for (int i = 0; i < 1 << 20; i++) {
```
joel-costigliola, line 272: I guess `1 << 20` is too much for the current github actions tests memory settings since the test ends with an OOM.

etellman: Oh, OK. It's pretty extreme, so you could dial it back to 2^15 elements, with each element an array of 500, or something like that.

etellman:
It looks like this one finally passes all the tests. It's gotten pretty messy, however, since main changed in the meantime, and I merged main into it, and used multiple checkins to fix license files and formatting. Let me know if this is too confusing, and I can start over with a new PR with all the changes in a single commit from the latest main.

etellman:
@joel-costigliola - It looks like this PR may have gotten lost in the shuffle. Let me know if you need anything else from me.

joel-costigliola:
Integrated thanks @etellman for the good work!

I noticed you merged main into your branch, it's not big deal but we prefer contributors to rebase their branch on main since at the end we squash all commits into one (I cherry picked your commits and did a bit of refactoring so no problem at all)

etellman:
That makes sense. I'll rebase next time.

PR #3132, opened 2023-07-31
Introduce assertThatCharSequence to disambiguate Groovy's GString. (Fixes #2791)

### LINKED ISSUE #2791: Introduce `assertThatCharSequence` to disambiguate Groovy's `GString`
#### Feature summary

Currently, Groovy's interpolated expressions cannot be a direct input of `assertThat` as `groovy.lang.GString` implements both `Comparable` and `CharSequence`.

We might add `assertThatCharSequence(CharSequence)` / `assumeThatCharSequence(CharSequence)` and the corresponding BDD variants to disambiguate these calls.

#### Example

This should compile:

```groovy
  @Test
  void should_accept_interpolated_expressions() {
    // GIVEN
    def foo = "foo"
    def actual = /.*${foo}.*/
    // WHEN/THEN
    assertThatCharSequence(actual).isEqualTo(".*foo.*")
  }
```

Implementation notes:
* The implementation can delegate to the existing `assertThat(CharSequence)`
* The new method should also be added to `Java6` classes whenever possible. `Assertions_sync` and `Java6Assertions_sync` can help figure out where the gaps are.

#### Workaround

Use `toString`:

```groovy
  @Test
  void should_accept_interpolated_expressions() {
    // GIVEN
    def foo = "foo"
    def actual = /.*${foo}.*/
    // WHEN/THEN
    assertThat(actual.toString()).isEqualTo(".*foo.*")
  }
```

amodolo:
Added `assertThatCharSequence(CharSequence)`/`assumeThatCharSequence(CharSequence)` method and corresponding BDD and Java6 variants.

#### Check List:
* Fixes #2791 
* Unit tests : YES


scordio:
This is now merged. Thanks for your first contribution, @amodolo!

PR #3133, opened 2023-08-01
Make test for long iterable smaller

etellman:
Creating the test data for the long iterable causes an out of memory error in the GitHub automated build system. This change uses a much smaller example, with fewer things to iterate through and smaller items. It adds a timeout, however, so the test still should fail without the related change that makes this more efficient.

#### Check List:
* Fixes https://github.com/assertj/assertj/issues/3123
* Unit tests : YES

### DIFF
--- a/assertj-core/src/test/java/org/assertj/core/presentation/StandardRepresentation_iterable_format_Test.java
+++ b/assertj-core/src/test/java/org/assertj/core/presentation/StandardRepresentation_iterable_format_Test.java
@@ -25,11 +25,16 @@ import static org.mockito.Mockito.withSettings;
 
 import java.nio.file.DirectoryStream;
 import java.nio.file.SecureDirectoryStream;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Stream;
 
+import org.apache.commons.lang3.StringUtils;
+import org.assertj.core.configuration.Configuration;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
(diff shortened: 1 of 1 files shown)


assertj-core/src/test/java/org/assertj/core/presentation/StandardRepresentation_iterable_format_Test.java
```diff
@@ -260,6 +265,23 @@ public void should_use_overridden_toString_over_iterable_representation_in_colle
     then(formatted).isEqualTo("[{\"a\":1}]");
   }
 
+  @Test
+  @Timeout(value = 1, unit = TimeUnit.SECONDS)
+  void should_format_big_list() {
+    // GIVEN
+    int elementsPerArray = 200;
+    List<int[]> numbers = new ArrayList<>();
+    for (int i = 0; i < 1 << 18; i++) {
```
joel-costigliola, line 274: the build fails with a timeout https://github.com/assertj/assertj/actions/runs/5721061463/job/15502060632?pr=3133 maybe use `15` instead of `18` ?

etellman: 15 instead of 18 sounds OK. 2^15 is only around 32000, so you could just use 32000 instead of (1 << 15). I did it as (1 << 18) because the numbers were pretty big.

I was trying to make the test so it timed out with the original implementation and passed with the related PR that makes it more efficient, but I'm a little fuzzy on whether that test run was just the test change or the test change and the related change that would make the test pass. If it's just the test change, then it's supposed to timeout, and when the other change is added, it should start passing.

This test only takes around 200 ms on my laptop with the related code change. If it's timing out on the GitHub hosts, that makes me happy about how fast my laptop is. :^)

joel-costigliola: Can you push the changes with `32000` so that we can see if it succeeds on github ?

etellman: No problem.

Sorry, I'm new to contributing to open source, so I messed up the mechanics of this. I didn't realize you could update the original PR by pushing more changes to the PR branch, so I created a new PR from a different branch (this PR), which really confused things.

I'll update the original PR with this change, and get everything to pass there. I just started doing this, but realized there are some formatting and missing license issues, as well, so I'll also fix those.

So I think we can abandon this one and I'll just fix the original one.

joel-costigliola: Hey @etellman no problem at all!

 I'm closing this one then but let me know which is the correct one to review then, is that https://github.com/assertj/assertj/issues/3123 ?

etellman: The original PR passes the tests now: https://github.com/assertj/assertj/pull/3126

Feel free to close this one.

etellman: > Hey @etellman no problem at all!
> 
> I'm closing this one then but let me know which is the correct one to review then, is that #3123 ?

No, it's https://github.com/assertj/assertj/pull/3126. #3123 was for arrays.

PR #3135, opened 2023-08-01
Change the way the properties are collected using the Class.getMethod…

### LINKED ISSUE #3125: Recursion comparison using properties ignores default interface methods.
If an Object has methods defined as default methods in an interface, these are not detected by the ComparingProperties() introspection strategy:

for example:

```
public interface Bar {
  default String getName() {
     return "bar";
  }
}

public class Foo implements Bar {
}
```

expected:

ComparingProperties should also detect properties defined in as default methods in interfaces implemented by the target object.

amodolo:
I change the way the properties are collected using the `Class.getMethods()` api instead iterating the class hierarchy using `Class.declaredMethods().` I think that collect the properties using the `getMethods` api should be  a better approach because:
-  it returns all public methods considering both super classes and interfaces
-  there's no need to filter public methods
-  there's no need to iterate through the class hierarchy 

Moreover, in this way the properties declared in the interfaces will not be ignored.  

#### Check List:
* Fixes #3125 (ignore if not applicable)
* Unit tests : YES

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/ComparingProperties.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/ComparingProperties.java
@@ -12,9 +12,7 @@
  */
 package org.assertj.core.api.recursive.comparison;
 
-import static java.util.Arrays.stream;
-import static java.util.stream.Collectors.toCollection;
-import static java.util.stream.Collectors.toSet;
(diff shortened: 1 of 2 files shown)


joel-costigliola:
Thanks @amodolo, I will do my best to review your PR this weekend !

joel-costigliola:
Integrated thanks @amodolo!

PR #3137, opened 2023-08-03
Fix/3129 recursive comparison fails if comparing only fileds has non existing fields

aindriu-aiven:
#### Check List:
* Fixes # 3129 Partially
* Unit tests : YES

This PR fixes
This PR checks when using recursive comparison if all fields specified in "comparingOnlyFields" exist on the actual object being compared. This prevents misnaming of the fields and silent failures.

This PR does not fix
Items specified in containers (lists/arrays/etc) these items are currently compared using "Deeply compare two Collections that must be same length and in same order." checking all field entries and ignoring the specified fields.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
@@ -152,6 +152,30 @@ public class RecursiveComparisonDifferenceCalculator {
       }
     }
 
+    private static Set<String> getAllChildFieldsSpecifiedForCompare(RecursiveComparisonConfiguration recursiveComparisonConfiguration,
+                                                                    DualValue dualValue) {
+      return recursiveComparisonConfiguration.getComparedFields().stream()
+                                             // Remove all specified fields that are not children of this DualValue
+                                             .filter(field -> isChildOfSpecifiedComparatorField(dualValue, field))
+                                             // Map the next FieldLocation to the fieldName
+                                             .map(field -> getChildFieldForValidation(field, dualValue.fieldLocation))
+                                             .collect(Collectors.toSet());
+    }
+
+    private static boolean isChildOfSpecifiedComparatorField(DualValue dualValue, FieldLocation field) {
+
+      return field.getPathToUseInRules()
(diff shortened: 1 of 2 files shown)


joel-costigliola:
@aindriu-aiven thanks for the PR.

I'm looking at it and think of another approach that could be easier, I believe.

If compared fields are specified then we can check before even starting any field comparison that actual has the mentioned fields, if not then we should return an illegal argument exception since what the user has specified is a mistake, in that sense it's not an assertion error, just wrong arguments.

If we goes that way, I would add a method like `checkCompareFieldsExist(actual)`  if comparedFields is not empty.

WDYT ?

aindriu-aiven:
> @aindriu-aiven thanks for the PR.
> 
> I'm looking at it and think of another approach that could be easier, I believe.
> 
> If compared fields are specified then we can check before even starting any field comparison that actual has the mentioned fields, if not then we should return an illegal argument exception since what the user has specified is a mistake, in that sense it's not an assertion error, just wrong arguments.
> 
> If we goes that way, I would add a method like `checkCompareFieldsExist(actual)` if comparedFields is not empty.
> 
> WDYT ?

Hey Yeah let me check that out and I'll try get an updated option up shortly with those changes 👍

aindriu-aiven:
Hey @joel-costigliola I have an update to this PR but before I go ahead and push it I wanted to ask for a clarification.

I have a PR now that will check for any incorrect/misspelt fields.

If you send in a collection it will also check the objects in the collection. 
Including if say you have a List<Map<?,?>> it will get to the root values in the Map and check them.

However having completed my unit tests I noticed this is overkill, currently the code would not do a comparison on the fields in a Map/List. Do you want me to leave that piece of code out or leave it in as some future proofing?


Also P.S. sorry got caught up in a release, but think i should be able to get this completed in a couple hours once I have feedback.

joel-costigliola:
@aindriu-aiven sorry I have forgotten to telI you but have taken over this piece of work, I'm actually quite close to finish it.
my bad really

aindriu-aiven:
> @aindriu-aiven sorry I have forgotten to telI you but have taken over this piece of work, I'm actually quite close to finish it. my bad really

No problem, I should have let you know I was tied up with another release. I pushed the changes I made to this PR so you can check them out if you would like.

Otherwise feel free to comment.

Thanks

Issue #3105, opened 2023-07-12
make `asList()` type safe


Illutax:
#### Feature summary

The method `asList()` should return a correctly typed list.

#### Example

Assume a type `Bar` with the filed `strings: List<String>`. And a type `Foo` with the field `bar: List<Bar>` (both with default getter and all-args-constructor).

```java
class Foo {
  private final List<Bar> bars;
}

class Bar {
  private final List<String> strings;
}
```

The following block is type-safe
```java
import static java.util.Arrays.asList;

Bar bar = new Bar(asList("a","b"));
assertThat(bar)
  .extracting(Bar::getStrings)
  .asList()
  .containsExactly("a", "b");
```

When working with *nested* structures and using `Assert.asList()` the returned list will be of type Object.

```java
Foo myComplexFoo = new Foo(asList(new Bar(asList("a", "b")), new Bar(asList("a", "b"))));
assertThat(myComplexFoo)
  .extracting(Foo::getBars)
  .asList()
  .extracting(o -> ((Bar)o).getStrings())
  .containsExactly(asList("a", "b"), asList("a", "b"));
```

We had to cast the Object back to a Bar for working further with it.

The proposal creates an overload to the asList() which gets a type such that we can further assert the structure like:

```java
Foo myComplexFoo = new Foo(asList(new Bar(asList("a", "b")), new Bar(asList("a", "b"))));
assertThat(myComplexFoo)
  .extracting(Foo::getBars)
  .asTypedList(Bar.class)
  .extracting(Bar::getStrings)
  .containsExactly(asList("a", "b"), asList("a", "b"));
```

Illutax:
I tried to wrestle with Javas' generic type system but it was fruitless.

I came up with a generic method where the type is bound by an additionally passed type argument.

```java
<T> AbstractListAssert<?, List<? extends T>, T, ObjectAssert<T>> asTypedList(Class<T> type);
```

Maybe it would be the better way to overload the `extracting` method instead?

Illutax:
Ok, this works. But it is quite a mouthful and I'm sure that I will have to return to this issue for future reference.

I see that you implemented it in 2019, so I can imagine that you can remember it rather quickly.

Foo::getBars already returns Bar.class implicitly, why can't we utilize it instead?

Illutax:
Would the assertj community benefit from the aforementioned `asTypedList(Class<T> type)`?

scordio:
[...]

PR #3138, opened 2023-08-03
Deprecate `asList` in favor of `asInstanceOf`

scordio:
Triggered by https://github.com/assertj/assertj/issues/3105#issuecomment-1633150460.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -591,12 +591,11 @@ public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, 
   }
 
   /** {@inheritDoc} */
-  @SuppressWarnings("unchecked")
+  @Deprecated
   @Override
   @CheckReturnValue
   public AbstractListAssert<?, List<?>, Object, ObjectAssert<Object>> asList() {
-    objects.assertIsInstanceOf(info, actual, List.class);
-    return newListAssertInstance((List<Object>) actual).as(info.description());
+    return asInstanceOf(InstanceOfAssertFactories.LIST);
   }
 
   /** {@inheritDoc} */
--- a/assertj-core/src/main/java/org/assertj/core/api/Assert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/Assert.java
@@ -650,7 +650,9 @@ public interface Assert<SELF extends Assert<SELF, ACTUAL>, ACTUAL> extends Descr
    * assertThat(unsortedListAsObject).asList().isSorted();</code></pre>
    *
    * @return a list assertion object
+   * @deprecated use {@link #asInstanceOf(InstanceOfAssertFactory) asInstanceOf(InstanceOfAssertFactories.LIST)} instead
    */
+  @Deprecated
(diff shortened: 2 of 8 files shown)


manusa:
Hi, there are no JavaDocs pointing to the proposed alternative. I had to scavenge the source code to reach to this PR.

Might be interesting to provide the suggested approach.
Which I'm still not sure if is to use:

```java
asInstanceOf(InstanceOfAssertFactories.LIST)
```

instead

scordio:
Hi @manusa, which Javadoc were you looking at exactly?

[`Assert::asList`](https://www.javadoc.io/doc/org.assertj/assertj-core/latest/org/assertj/core/api/Assert.html#asList()) Javadoc has the information you're looking for:

Screenshot 2024-01-09 at 13 39 48

but I just realized [`AbstractAssert::asList`](https://www.javadoc.io/doc/org.assertj/assertj-core/latest/org/assertj/core/api/AbstractAssert.html#asList()) doesn't have it:

Screenshot 2024-01-09 at 13 41 59

manusa:
:facepalm: sorry.

Yes, I was meaning the method in the `AbstractAssert` class.

scordio:
Thanks for the confirmation. There seems to be a gap in the Javadoc generation, I'm not sure what could be done in AssertJ but I'll have a look.

manusa:
No worries. I'm not sure if `{@inheritDoc}` would work in this case and help avoid doc duplication, maybe worth giving a try.

If not, I'm happy to send a patch with the missing section in `AbstractAssert` class ;)

manusa:
OK, I think the problem is (TIL) that inheritDoc needs to be used in the tags too.

https://docs.oracle.com/javase/6/docs/technotes/tools/solaris/javadoc.html#@inheritDoc


>Inherits (copies) documentation from the "nearest" inheritable class or implementable interface into the current doc comment at this tag's location. This allows you to write more general comments higher up the inheritance tree, and to write around the copied text.
>
>This tag is valid only in these places in a doc comment:
>
> - In the main description block of a method. In this case, the main description is copied from a class or interface up the hierarchy.
> - In the text arguments of the @return, @param and @throws tags of a method. In this case, the tag text is copied from the corresponding tag up the hierarchy.

So I believe something like this is required: 
```java
  /**
   * {@inheritDoc}
   * @return {@inheritDoc}
   * @deprecated {@inheritDoc}
   */
```

scordio:
Thanks, I remembered something like that but couldn't find the official reference quickly 🙂 

It doesn't mention anything about the `@deprecated` tag, though. Would you like to give it a try and submit a patch, in case it works?

manusa:
> It doesn't mention anything about the `@deprecated` tag, though. Would you like to give it a try and submit a patch, in case it works?

I'll submit a patch to add the deprecation suggestion manually.

I've done some checks in my projects and the behavior differs depending on the engine used to generate Javadocs.

For example, in IntelliJ, if the following if the `@return {@inheritDoc}` is not provided, then the tag is not shown in the editor-rendered Javadoc. However, it is actually shown in the Quick doc viewer (Ctrl+q).
On the other hand, the `@return` tag is propagated in the javadoc jar too.

In any case, and a common denominator for all is that the `@deprecated` tag is never propagated, and from what I've seen, it doesn't support the `inheritDoc` either.

PR #3151, opened 2023-08-18
Fix NPE thrown by `anyMatch(Objects::isNull)` given an iterable containing a null element

armandino:
This PR fixes a null pointer exception produced by the following usage:

```java
assertThat(Arrays.asList("foo", null)).anyMatch(Objects::isNull);
```

I came across the exception while trying to assert that a collection contains both, null and non-null values:

```java
assertThat(Arrays.asList("foo", null))
   .anyMatch(Objects::isNull)
   .anyMatch(Objects::nonNull);
```

The alternative is to use `containsNull()`:

```java
assertThat(Arrays.asList("foo", null)
    .containsNull()
    .anyMatch(Objects::nonNull);
```

#### Check List:
* Unit tests : YES

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/internal/Iterables.java
+++ b/assertj-core/src/main/java/org/assertj/core/internal/Iterables.java
@@ -1352,9 +1352,9 @@ public class Iterables {
                                  PredicateDescription predicateDescription) {
     assertNotNull(info, actual);
     predicates.assertIsNotNull(predicate);
-    stream(actual).filter(predicate)
-                  .findFirst()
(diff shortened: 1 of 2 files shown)


scordio:
Note that `List.of()` doesn't allow `null` elements – I replaced it with `Arrays.asList` in your example to avoid confusion.

scordio:
Thank you, @armandino!

armandino:
> Note that `List.of()` doesn't allow `null` elements – I replaced it with `Arrays.asList` to avoid confusion.

Good catch! Reminder for me to test snippets before posting :) Thank you, @scordio!

PR #3166, opened 2023-09-01
#3159 Fix misleading subsequence failure messages in string assertions

### LINKED ISSUE #3159: Error message for `containsSubsequence` is misleading for duplicate subsequence
Version: AssertJ 3.24.2

Given the following assertion:

```java
assertThat("v1 : v2 : v3").containsSubsequence("v2", "v2", "v3");
```

The failure message is:

```
Expecting actual:
  "v1 : v2 : v3"
to contain the following CharSequences in this order:
  ["v2", "v2", "v3"]
but "v2" was found before "v2"
```

The `but "v2" was found before "v2"` message is misleading since the actual string does not contain two "v2" strings.

JohnBryte:
- Resolved ambiguity in error messages when a subsequence element is repeated in the expectation but not in the actual string.
      - Refactored `assertContainsSubsequence` to handle different subsequence scenarios.
      - Improved error handling by introducing `handleNotFound` to distinctly report missing elements and discrepancies in expected duplicates.
      - Enhanced `getNotFoundSubsequence` to correctly compute and return missing or inadequately represented subsequences.

      This change ensures clearer and more accurate failure messages when asserting string subsequences.

#### Check List:
* Fixes #3159 
* Unit tests : YES

**Description**:

The core assertContainsSubsequence method has undergone a significant overhaul to better cater to a broader range of subsequence scenarios. The driving force behind this refactoring is twofold:

- Enhanced Subsequence Handling:

    The refactoring ensures that the method can adeptly handle cases where there might be duplicate subsequences present. The essence of this handling lies in distinguishing between two different situations:
        Where the actual string contains the subsequence but not as many times as the subsequence array dictates.
        Where the actual string doesn't contain a particular subsequence at all.

- Introduction of handleNotFound Method:

    A fresh method, handleNotFound, has been extracted to specifically address the nuances of missing subsequences. A core task of this new method is:
        To scrutinize whether missing subsequences arise because the actual string didn't cater to the repeated instances as specified by the subsequence array.

- Behavioural Precedence:

In instances where at least one subsequence is completely absent from the actual string, this missing subsequence takes precedence over any discrepancies in the count of other subsequences. As illustrated in the example:
```
String actual = "Yoda";
String[] subsequence = { "Yo", "da", "Han", "Foo" };
```

This will yield the same output as:
```
String actual = "Yoda";
String[] subsequence = { "Yo", "da", "Han", "Foo" , "da"};
```
Even though "da" appears twice in the latter example, the predominant error message would be about the absence of the strings "Han" and "Foo" in the actual string. This decision was made to ensure that the most pertinent errors (complete absence) are flagged prior to counting discrepancies.

### DIFF
--- a/assertj-core/src/test/java/org/assertj/core/error/ShouldContainSubsequenceOfCharSequence_create_Test.java
+++ b/assertj-core/src/test/java/org/assertj/core/error/ShouldContainSubsequenceOfCharSequence_create_Test.java
@@ -24,7 +24,6 @@ import org.assertj.core.presentation.StandardRepresentation;
 import org.assertj.core.test.CaseInsensitiveStringComparator;
 import org.junit.jupiter.api.Test;
 
-
 /**
  * Tests for <code>{@link ShouldContainSubsequenceOfCharSequence#create(org.assertj.core.description.Description, org.assertj.core.presentation.Representation)}</code>.
  *
@@ -43,7 +42,8 @@ class ShouldContainSubsequenceOfCharSequence_create_Test {
     // THEN
     then(message).isEqualTo(format("[Test] %nExpecting actual:%n" +
                                    "  \"" + actual + "\"%n" +
-                                   "to contain the following CharSequences in this order (possibly with other values between them):%n" +
+                                   "to contain the following CharSequences in this order (possibly with other values between them):%n"
+                                   +
                                    "  [\"{\", \"author\", \"title\", \"}\"]%n" +
                                    "but \"title\" was found before \"author\"%n"));
   }
@@ -60,7 +60,8 @@ class ShouldContainSubsequenceOfCharSequence_create_Test {
     // THEN
     then(message).isEqualTo(format("[Test] %nExpecting actual:%n" +
                                    "  \"" + actual + "\"%n" +
-                                   "to contain the following CharSequences in this order (possibly with other values between them):%n" +
+                                   "to contain the following CharSequences in this order (possibly with other values between them):%n"
+                                   +
                                    "  [\"{\", \"author\", \"title\", \"}\"]%n" +
                                    "but \"title\" was found before \"author\"%n" +
                                    "when comparing values using CaseInsensitiveStringComparator"));
@@ -78,7 +79,8 @@ class ShouldContainSubsequenceOfCharSequence_create_Test {
     // THEN
     then(message).isEqualTo(format("[Test] %nExpecting actual:%n" +
                                    "  \"" + actual + "\"%n" +
-                                   "to contain the following CharSequences in this order (possibly with other values between them):%n" +
+                                   "to contain the following CharSequences in this order (possibly with other values between them):%n"
+                                   +
                                    "  [\"{\", \"title\", \"author\", \"title\", \"}\"]%n" +
                                    "But%n" +
                                    "2nd occurrence of \"title\" was not found%n" +
--- a/assertj-core/src/test/java/org/assertj/core/error/ShouldContainSubsequenceOfCharSequence_create_Test.java
+++ b/assertj-core/src/test/java/org/assertj/core/error/ShouldContainSubsequenceOfCharSequence_create_Test.java
@@ -23,6 +23,10 @@ import org.assertj.core.internal.ComparatorBasedComparisonStrategy;
 import org.assertj.core.presentation.StandardRepresentation;
 import org.assertj.core.test.CaseInsensitiveStringComparator;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
 
 /**
  * Tests for <code>{@link ShouldContainSubsequenceOfCharSequence#create(org.assertj.core.description.Description, org.assertj.core.presentation.Representation)}</code>.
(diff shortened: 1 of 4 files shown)


JohnBryte:
Just a quick question, currently it seems my PR fails the checks because of:
```
[ERROR] /home/runner/work/assertj/assertj/assertj-core/src/test/java/org/assertj/core/error/ShouldContainSubsequenceOfCharSequence_create_Test.java:[73,87] cannot find symbol
  symbol:   method of(java.lang.String,int)
  location: interface java.util.Map
```
I was wondering if I need to use ImmutableMap, as I've seen that ImmutableMap is used almost exclusively
Or should I do something like
```
Map<CharSequence, Integer> notFound = new HashMap<>();
notFound.put(key, value);
```
Please let me know, so I can try to solve the conflicts

assertj-core/src/test/java/org/assertj/core/error/ShouldContainSubsequenceOfCharSequence_create_Test.java
```diff
@@ -57,10 +59,28 @@ void should_create_error_message_with_custom_comparison_strategy() {
...
                                    "but \"title\" was found before \"author\"%n" +
                                    "when comparing values using CaseInsensitiveStringComparator"));
   }
 
+  @Test
+  void should_create_error_message_indicating_duplicate_subsequence() {
+    // GIVEN
+    String[] sequenceValues = { "{", "title", "author", "title", "}" };
+    String actual = "{ 'title':'A Game of Thrones', 'author':'George Martin'}";
+    ErrorMessageFactory factory = shouldContainSubsequence(actual, sequenceValues, Map.of("title", 1),
```
scordio, line 73: As the language level is 8, you could use instead `org.assertj.core.test.Maps.mapOf` and `org.assertj.core.api.Assertions.entry`:

```suggestion
    ErrorMessageFactory factory = shouldContainSubsequence(actual, sequenceValues, mapOf(entry("title", 1)),
```

JohnBryte:
I now (hopefully) fixed the formatting issues. If the checks could be run again on occasion I would greatly appreciate it

joel-costigliola:
checks look ok, windows build failure is not related.

assertj-core/src/test/java/org/assertj/core/error/ShouldContainSubsequenceOfCharSequence_create_Test.java
```diff
@@ -57,10 +60,30 @@ void should_create_error_message_with_custom_comparison_strategy() {
...
+    // WHEN
+    String message = factory.create(new TextDescription("Test"), new StandardRepresentation());
+    // THEN
+    then(message).isEqualTo(format("[Test] %nExpecting actual:%n" +
+                                   "  \"" + actual + "\"%n" +
+                                   "to contain the following CharSequences in this order (possibly with other values between them):%n"
+                                   +
+                                   "  [\"{\", \"title\", \"author\", \"title\", \"}\"]%n" +
+                                   "But%n" +
+                                   "2nd occurrence of \"title\" was not found%n" +
```
joel-costigliola, line 86: why do we go the next line ? I feel it would be better like this (unless I'm missing something) 
```
 "But the 2nd occurrence of \"title\" was not found%n" +
```
I'm adding `the` before 2nd as it sounds to me to be better English wise

assertj-core/src/test/java/org/assertj/core/error/ShouldContainSubsequenceOfCharSequence_create_Test.java
```diff
@@ -57,10 +60,30 @@ void should_create_error_message_with_custom_comparison_strategy() {
...
+                                   "when comparing values using CaseInsensitiveStringComparator"));
+  }
```
joel-costigliola, line 88: It would be great to have more cases showing other ordinals (like 1st, 3rd, 4th) otherwise the `ordinal` method test coverage would be quite incomplete.  A parameterized test would be a good fit for this kind of tests ;-)

joel-costigliola (changes requested):
@JohnBryte thanks for the PR, first round of review done.

assertj-core/src/test/java/org/assertj/core/error/ShouldContainSubsequenceOfCharSequence_create_Test.java
JohnBryte: Hey Joel, thanks for all the feedback!
The idea behind the next line was (and I should have made that clean with an explicit test) if there are multiple duplicates in they would all be listed underneath each other and form a "block".
E.g.:
```
    String[] sequenceValues = { "{", "title", "George", "title", "title", "George", "}" };
    String actual = "{ 'title':'A Game of Thrones', 'author':'George Martin', 'title':'The Kingkiller Chronicle', 'author':'Patrick Rothfuss'}";
```
The expected outcome (with newline) would be something like:
```
"But"
"the 3rd occurrence of \"title\" was not found%n"
"the 2nd occurrence of \"George\" was not found%n"
```

Without the new line:
```
"But the 3rd occurrence of \"title\" was not found%n"
"the 2nd occurrence of \"George\" was not found%n"
```

But this can be easily adapted to you preference!
Or should this scenario handled in another way?
For example, fail with the first duplicate encountered?

joel-costigliola:
Thanks for the quick update, I'll do a second review and likely merge the PR unless I find something significant.

assertj-core/src/test/java/org/assertj/core/error/ShouldContainSubsequenceOfCharSequence_create_Test.java
joel-costigliola: Thanks for the clarification on the wording, I would say it is better to report as many errors as possible and for the wording my preference would go to this:
```
But: 
- the 3rd occurrence of \"title\" was not found
- the 2nd occurrence of \"George\" was not found
```
and if there is only one error: 
```
But the 3rd occurrence of \"title\" was not found
```

joel-costigliola:
@JohnBryte only one comment on the PR about the wording on the message.

joel-costigliola:
Integrated thanks @JohnBryte for the good work!

sbrannen:
Thanks for addressing this, @JohnBryte! 👏

PR #3174, opened 2023-09-14
Add `hasPrecisionOf(int)` assertion for `BigDecimal`

armandino:
#### Check List:
* Fixes N/A
* Unit tests : YES

AssertJ provides a `hasScaleOf()` assertion for the `BigDecimal` type.  This PR adds a companion assertion `hasPrecisionOf` for verifying the `BigDecimal.precision()` method. I followed the `hasScaleOf()` implementation and tests in this PR.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractBigDecimalAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractBigDecimalAssert.java
@@ -12,17 +12,18 @@
  */
 package org.assertj.core.api;
 
-import static java.util.Objects.requireNonNull;
-
-import java.math.BigDecimal;
-import java.util.Comparator;
-
 import org.assertj.core.data.Offset;
 import org.assertj.core.data.Percentage;
 import org.assertj.core.internal.BigDecimals;
 import org.assertj.core.internal.ComparatorBasedComparisonStrategy;
 import org.assertj.core.util.CheckReturnValue;
 import org.assertj.core.util.VisibleForTesting;
+
+import java.math.BigDecimal;
+import java.util.Comparator;
+
+import static java.util.Objects.requireNonNull;
+import static org.assertj.core.error.ShouldHavePrecision.shouldHavePrecision;
 
 /**
  * Base class for all implementations of assertions for {@link BigDecimal}s.
@@ -276,12 +277,35 @@ public abstract class AbstractBigDecimalAssert<SELF extends AbstractBigDecimalAs
    * // assertion will fail
    * assertThat(new BigDecimal(&quot;8.00&quot;)).hasScaleOf(3);
    * assertThat(new BigDecimal(&quot;8.00&quot;).setScale(4)).hasScaleOf(2);</code></pre>
-   * 
+   *
    * @param expectedScale the expected scale value.
    * @return {@code this} assertion object.
    */
   public SELF hasScaleOf(int expectedScale) {
     bigDecimals.assertHasScale(info, actual, expectedScale);
+    return myself;
+  }
+
+  /**
+   * Verifies the {@code BigDecimal} under test has the given precision.
+   * <p>
+   * Example:
+   * <pre><code class='java'> // assertions will pass
+   * assertThat(new BigDecimal(&quot;8.00&quot;)).hasPrecisionOf(3);
+   * assertThat(new BigDecimal(&quot;0.05&quot;)).hasPrecisionOf(1);
+   *
+   * // assertion will fail
+   * assertThat(new BigDecimal(&quot;8.00&quot;)).hasPrecisionOf(1);
+   * assertThat(new BigDecimal(&quot;0.05&quot;)).hasPrecisionOf(2);</code></pre>
+   *
+   * @param expectedPrecision the expected precision value.
+   * @return {@code this} assertion object.
+   */
+  public SELF hasPrecisionOf(int expectedPrecision) {
+    isNotNull();
+    if (actual.precision() != expectedPrecision) {
+      throwAssertionError(shouldHavePrecision(actual, expectedPrecision));
+    }
     return myself;
   }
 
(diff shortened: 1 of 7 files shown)


assertj-core/src/main/java/org/assertj/core/internal/BigDecimals.java
```diff
@@ -84,4 +85,10 @@ public void assertHasScale(AssertionInfo info, BigDecimal actual, int expectedSc
     throw failures.failure(info, shouldHaveScale(actual, expectedScale));
   }
 
+  public void assertHasPrecision(AssertionInfo info, BigDecimal actual, int expectedPrecision) {
+    assertNotNull(info, actual);
+    if (areEqual(actual.precision(), expectedPrecision)) return;
+    throw failures.failure(info, shouldHavePrecision(actual, expectedPrecision));
+  }
```
scordio, line 92: I imagine you've taken inspiration from `assertHasScale`. However, these assertions are hardly reusable somewhere else so I suggest pulling this to `AbstractBigDecimalAssert`, similar to what we have for `AbstractCharSequenceAssert::isAlphabetic`:

https://github.com/assertj/assertj/blob/fd20abf93fd58a93f02ff9f1353dd6ae6e62f755/assertj-core/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java#L2018-L2022

assertj-core/src/main/java/org/assertj/core/error/ShouldHavePrecision.java
```diff
@@ -0,0 +1,30 @@
...
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2023 the original author or authors.
+ */
+package org.assertj.core.error;
+
+import java.math.BigDecimal;
+
+public class ShouldHavePrecision extends BasicErrorMessageFactory {
```
scordio, line 17: Could you please add a unit test for this class, similar to `ShouldBeVisible_create_Test`?

assertj-core/src/test/java/org/assertj/core/api/bigdecimal/BigDecimalsAssert_assertHasPrecisionOf_Test.java
```diff
@@ -0,0 +1,32 @@
...
+  protected BigDecimalAssert invoke_api_method() {
+    return assertions.hasPrecisionOf(1);
+  }
+
+  @Override
+  protected void verify_internal_effects() {
+    verify(bigDecimals).assertHasPrecision(getInfo(assertions), getActual(assertions), 1);
+  }
+
+}
```
scordio, line 32: Once the implementation is pulled up, this test class should contain actual test cases like `CharSequenceAssert_isVisible_Test` (and there is no need to extend `BigDecimalAssertBaseTest`).

assertj-core/src/main/java/org/assertj/core/api/BigDecimalPrecisionAssert.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Copyright 2012-2023 the original author or authors.
+ */
+package org.assertj.core.api;
+
+public class BigDecimalPrecisionAssert<T> extends AbstractBigDecimalPrecisionAssert<BigDecimalAssert> {
```
scordio, line 15: `T` seems to be unused, isn't it?

assertj-core/src/test/java/org/assertj/core/internal/bigdecimals/BigDecimalAssert_precision_Test.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import java.math.BigDecimal;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatNullPointerException;
+import static org.assertj.core.api.BDDAssertions.then;
+
+class BigDecimalAssert_precision_Test {
```
scordio, line 23: `precision` would be a new navigation method and we defined a base test interface for such cases.

Could you try to add the following?

```suggestion
class BigDecimalAssert_precision_Test implements NavigationMethodBaseTest<BigDecimalAssert<BigDecimal>> {
```

assertj-core/src/main/java/org/assertj/core/api/AbstractBigDecimalAssert.java
```diff
@@ -523,4 +543,30 @@ public AbstractBigDecimalScaleAssert<SELF> scale() {
...
+   * // assertions fails
+   * assertThat(bgDecimal).precision().isBetween(6, 8);</code></pre>
+   *
+   * @return AbstractBigDecimalPrecisionAssert built with the {@code BigDecimal}'s precision.
+   * @throws NullPointerException if the given {@code BigDecimal} is {@code null}.
+   */
+  @SuppressWarnings({ "unchecked", "rawtypes" })
+  public AbstractBigDecimalPrecisionAssert<SELF> precision() {
+    requireNonNull(actual, "Can not perform assertions on the precision of a null BigDecimal");
+    return new BigDecimalPrecisionAssert(myself);
```
scordio, line 570: We briefly discussed this approach with the team.

We should add a bit of context to the error messages by showing the root object we navigated from.

For example, assuming:

```
assertThat(myBigDecimal).scale().isGreaterThan(5L)
```

If the assertion fails, it will show something like

```
expecting 2 to be greater than 5 but was not
```

where the fact that a `BigDecimal` precision was checked is forgotten.

Instead, the error message should have a better description and be more like:

```
[checking 2.123 precision] expecting 2 to be greater than 5 but was not
```

Unfortunately, we don't do this consistently in AssertJ so I didn't manage to find a good example quickly. I'll try to have another look in the next few days.

scordio (changes requested):
Thanks @armandino! My comments are mostly due to the `hasScale()` / `scale()` patterns you probably followed. We would prefer to follow a different style for new assertions.

We might also improve `hasScale()` / `scale()` but this can be taken up in a separate issue or PR.

assertj-core/src/main/java/org/assertj/core/api/BigDecimalPrecisionAssert.java
armandino: I tried removing it, but looks like it is required here:

```java
  public AbstractBigDecimalPrecisionAssert<SELF> precision() {
    isNotNull();
    return new BigDecimalPrecisionAssert(myself);
  }
```

armandino:
@scordio thanks for the feedback! It was very helpful. I addressed most of the comments, but there are still a few outstanding items.

assertj-core/src/test/java/org/assertj/core/internal/bigdecimals/BigDecimalAssert_precision_Test.java
```diff
@@ -0,0 +1,82 @@
...
+import static org.assertj.core.api.BDDAssertions.then;
+import static org.assertj.core.util.AssertionsUtil.expectAssertionError;
+import static org.assertj.core.util.FailureMessages.actualIsNull;
+import static org.mockito.Mockito.verify;
+
+class BigDecimalAssert_precision_Test implements NavigationMethodBaseTest<BigDecimalAssert> {
+
+  @Override
+  public void should_honor_registered_comparator() {
+    // TODO
```
armandino, line 34: I've overridden this test as it was failing. Any suggestions on how to handle it?

assertj-core/src/test/java/org/assertj/core/internal/bigdecimals/BigDecimalAssert_precision_Test.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  @Override
+  public void should_keep_existing_assertion_state() {
+    // TODO
```
armandino, line 39: I've overridden this test as it was failing. Any suggestions on how to handle it?

armandino:
Hi @scordio, I updated the PR. Please let me know if I missed anything.

PR #3180, opened 2023-09-15
Remove duplicate maven version

Bananeweizen:
It's defined in the parent POM already via https://github.com/assertj/assertj/blob/55356a5254e9a50881301bbfaaa32f1a23d1dc2d/assertj-parent/pom.xml#L236-L238


scordio:
Thanks, @Bananeweizen. Good catch!

PR #3182, opened 2023-09-15
Remove default goal

### LINKED ISSUE #3181: Maven default goal is not set at the right place
**Describe the bug**
The assertj parent module contains a maven default goal. That's the wrong place to define a default goal, because the default goal is not something that gets inherited by the children of that parent. Instead it must be set in the reactor (the root directory pom.xml).

Bananeweizen:
This makes it possible to invoke maven without arguments in the root module.

#### Check List:
* Fixes #3181
* Unit tests : NA

### DIFF
--- a/assertj-parent/pom.xml
+++ b/assertj-parent/pom.xml
@@ -80,7 +80,6 @@
   </dependencyManagement>
 
   <build>
-    <defaultGoal>clean install</defaultGoal>
     <pluginManagement>
       <plugins>
         <plugin>


scordio:
As mentioned https://github.com/assertj/assertj/issues/3181#issuecomment-1721620118, I believe we should have no default goal. Would you mind updating the PR accordingly?

scordio:
Thanks, @Bananeweizen!

PR #3185, opened 2023-09-16
Disable tests failing on Windows

### LINKED ISSUE #3183: Tests failing on Windows
**Describe the bug**
There are 20+ tests failing on Windows because of using unsupported features.
* the assertion conversion is implemented in shell scripts, and the test tries to run them
* many path related tests do positive and negative testing for all kinds of assertions with a kind of variation of test cases, one or two being on symbolic links each time. As the javadoc of the symbolic link creation method states, that may fail on some systems because of missing privileges. That's the case for Windows.

I have a local change that disables these test methods (for the symbolic links) or test classes (for the conversion) on Windows and with a useful message. Would that be a reasonable fix as PR? The much harder alternative would be to recognize the IOException for the missing privileges in the tests, to verify their message and that it's Windows. I'm not a fan of such tests (and the message is localized for me, so not sure whether that even works).

Bananeweizen:
* The conversion shell scripts cannot be executed on Windows.
* Path related tests trying to create symbolic links fail with an exception on Windows, as documented in
java.nio.file.Files.createSymbolicLink(Path, Path, FileAttribute<?>...)

#### Check List:
* Fixes #3183 
* Unit tests : NA

With this change I can build the complete Maven reactor. Without it, there are around 50 test failures, exactly on the disabled test classes and methods.

### DIFF
--- a/assertj-core/src/test/java/org/assertj/core/internal/paths/Paths_assertEndsWithRaw_Test.java
+++ b/assertj-core/src/test/java/org/assertj/core/internal/paths/Paths_assertEndsWithRaw_Test.java
@@ -26,6 +26,8 @@ import java.nio.file.Paths;
 
 import org.assertj.core.internal.PathsBaseTest;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledOnOs;
+import org.junit.jupiter.api.condition.OS;
 
 class Paths_assertEndsWithRaw_Test extends PathsBaseTest {
 
@@ -71,6 +73,7 @@ class Paths_assertEndsWithRaw_Test extends PathsBaseTest {
   }
 
   @Test
+  @DisabledOnOs(value = OS.WINDOWS, disabledReason = "Fails on Windows due to missing privileges")
   void should_fail_if_actual_is_not_canonical() throws IOException {
     // GIVEN
     Path file = createFile(tempDir.resolve("file"));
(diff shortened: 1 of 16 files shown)


scordio:
Hi @Bananeweizen, apologies for getting back so late to this PR!

> The conversion shell scripts cannot be executed on Windows.

You're totally right and I wonder why our build doesn't fail on the Windows runner – I'll dig a bit and get back to you.

> Path related tests trying to create symbolic links fail with an exception on Windows, as documented in
java.nio.file.Files.createSymbolicLink(Path, Path, FileAttribute<?>...)

I couldn't find a specific reference to Windows at [`Files::createSymbolicLink`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/file/Files.html#createSymbolicLink(java.nio.file.Path,java.nio.file.Path,java.nio.file.attribute.FileAttribute...)). Nevertheless, I experienced the same in a Windows environment with a unprivileged user.

As these tests can be successfully executed on the GitHub runner (and I suppose in any other Windows environment with the appropriate permissions), it might be a good idea to introduce a custom JUnit condition to address this constellation. I will take care of that on top of your changes.

scordio:
> The conversion shell scripts cannot be executed on Windows.

I refactored how we do script testing in #3785: this is now a separate Maven module running on non-Windows platforms only, and using JUnit `@TempDir` instead of the manual handling we had before.

Although my PR has taken a different direction from this one, I truly appreciated that you looked into the problem, somehow hidden before, therefore I'm adding you as a coauthor 😉

scordio:
Superseded by #3785 and #3788.

PR #2621, opened 2022-05-21
Align javadoc stylesheet to be compatible with Java 17

scordio:
Fixes #2576

PR #3191, opened 2023-09-20
ci: update workflows to Java 21

### LINKED ISSUE #3187: Upgrade workflows from Java 17 to 21
Today Java 21 has become GA. We should upgrade our workflows from Java 17 to 21.

We may need to redefine the Javadoc stylesheet as happened during the transition to Java 11 (#1570 / #1600) and 17 (#2576 / #2621).

grigala:
#### Check List:
* Fixes #3187
* Unit tests : NA


.github/workflows/cross-version.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [21, 22]
+        java: [22, 23]
```
scordio, line 15: I don't think there is any early-access build for 23, I think we keep 22 only for now.

grigala: Yea, I checked after the pipeline failed. Removed 23 :)

scordio:
@grigala thanks for raising the PR!

I've added more references to #3187 about the Javadoc stylesheet issues we encountered with the past upgrades. Do you plan to take care of it as well?

It's basically about generating the Javadoc locally and checking if anything breaks in the resulting pages. If that's the case, we would need to figure out how to apply the same customizations to the Java 21 base stylesheet (like what I've done in #2621).

martinfrancois:
> @grigala thanks for raising the PR!
> 
> I've added more references to #3187 about the Javadoc stylesheet issues we encountered with the past upgrades. Do you plan to take care of it as well?
> 
> It's basically about generating the Javadoc locally and checking if anything breaks in the resulting pages. If that's the case, we would need to figure out how to apply the same customizations to the Java 21 base stylesheet (like what I've done in #2621).

Unfortunately, you were right, the CSS was broken again in Java 21... But we finally managed to fix it, see the last commit I pushed.

scordio:
It seems the search results lack highlighting.

[Current Javadoc](https://www.javadoc.io/doc/org.assertj/assertj-core/latest/index.html):

Screenshot 2023-09-28 at 18 15 52

With the changes from this PR:

Screenshot 2023-09-28 at 18 15 43

[JDK 21 Javadoc](https://docs.oracle.com/en/java/javase/21/docs/api/index.html):

Screenshot 2023-09-28 at 18 17 48

As a minor point, we might want to remove the bold style from the SEARCH label/hyperlink.

scordio:
We would like to generate the Javadoc with Java 21 already in 3.25.0 so I'll have a look at the style for the search results.

scordio:
I noticed only now that the stylesheet went a bit far from the baseline that Java 21 provides so I started again from the base stylesheet in 2715b982a4ab461e40875abb1bb50fc139854549 and applied the customizations in fc234d6c69ae5d416bbe05f5b8c422636beaaf68.

scordio:
Thanks for your contribution, @grigala, @martinfrancois, and @georgebax!

PR #3195, opened 2023-09-23
Speed up for BinaryDiff used by hasSameBinaryContentAs for example. Fixes #3193.

### LINKED ISSUE #3193: BinaryDiff could be much faster
**org.assertj.core.internal.BinaryDiff could be much faster**

* assertj core version: 3.24.2
* java version: Java 11
* test framework version: JUnit 5

```
Path expected = ...
Path actual = ...
assertThat(actual).hasSameBinaryContentAs(expected);
```

**Solution**

Change implementation as follows:

  ```
public BinaryDiffResult diff(Path actual, byte[] expected) throws IOException {
    try(InputStream actualStream = new BufferedInputStream(Files.newInputStream(actual))) {
      return diff(actualStream, expected);
    }
  }
```

(add BufferedInputStream)

Resulting speedup for 2,5MB: from 8 sec down to 0.2 sec.

matthiaskraaz:
#### Check List:
* Fixes #3193
* Unit tests : NA (not added methods, hoping there are unit tests in place already)
* Javadoc with a code example (on API only) : NA (not added a method)

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/internal/BinaryDiff.java
+++ b/assertj-core/src/main/java/org/assertj/core/internal/BinaryDiff.java
@@ -12,6 +12,7 @@
  */
 package org.assertj.core.internal;
 
+import java.io.BufferedInputStream;
 import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


joel-costigliola:
Integrated thanks @matthiaskraaz !

PR #3206, opened 2023-10-03
Add default constructor for `RecursiveComparator`

scordio:
This allows a simpler creation of a `RecursiveComparator` instance for cases where a custom `RecursiveComparisonConfiguration` is not required.

Originally inspired by https://stackoverflow.com/questions/77151593.


scordio:
Jobs failing due to maxcellent/javadoc.io#180

Issue #1403, opened 2019-01-19
Fix javadoc generation on open jdk 11 and 12


joel-costigliola:
#### Summary

Typical error in travis see https://travis-ci.org/joel-costigliola/assertj-core/jobs/481842510:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.0.1:javadoc (default-cli) on project assertj-core: An error has occurred in Javadoc report generation: 
[ERROR] Exit code: 1 - javadoc: error - The code being documented uses modules but the packages defined in https://docs.oracle.com/javase/8/docs/api/ are in the unnamed module.
```

#### Fix

https://issues.apache.org/jira/browse/MJAVADOC-562

PascalSchumacher:
Open JDK issue (https://bugs.openjdk.java.net/browse/JDK-8212233) was reopened and fixed in 12.0.1 and 13. I hope this will be backported to Java 11.

PascalSchumacher:
Open JDK issue is now marked as fixed in Java 11.0.3.

joel-costigliola:
there is hope, we need to wait for travis to use it I guess.

hazendaz:
@joel-costigliola I'm pretty sure this one was fixed by maven with javadoc plugin 3.1.0 and not an issue with java itself.

joel-costigliola:
@hazendaz thanks, looks good. I think I'll bump the version in the parent pom and release it for use in assertj-core.

PascalSchumacher:
Was fixed in https://github.com/joel-costigliola/assertj-maven-parent-pom/commit/9b521bc33ac81b9a809de39c00599a9f928934d8 by the way.

joel-costigliola:
@PascalSchumacher thanks, I forgot about this

joel-costigliola:
Unfortunately the maven-javadoc-plugin 3.10.0 did not fix the issue as per this build https://travis-ci.org/joel-costigliola/assertj-core/builds/543057440.

I'm going to disable the javadoc again for java 11+ projects

joel-costigliola:
Fixed by adding specifying that source is 8 in the maven-javadoc-plugin configuration.

csaltos:
Also adding `<detectJavaApiLink>false</detectJavaApiLink>` to the Maven javadoc pluging configuration section helps

hemanthsridhar:
works with the below configuration.

`<plugin>
[...]

PR #3231, opened 2023-10-22
Fix javadoc options / Use legacy mode to overcome <source> level issue

hazendaz:
- options need to have additional arguments as an array rather than string
- removed old source 8 on javadocs as no longer needed here.
- Per javadoc 3.6.0 plugin, maven introduced ```<legacyMode>``` which overcomes bugs with javadocs that prevented source from being set higher as documented [here](https://github.com/assertj/assertj/issues/1403).  To replace the source attribute this is set to 'true'


scordio:
Hi @hazendaz, sorry for getting back to you so late!

Some of your changes have been applied as part of #3324. Still, I would apply the changes for `additionalOptions` you proposed.

However, I'm unsure if [`legacyMode`](https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#legacyMode) is required. The build already excludes the module path and we haven't noticed any side effects since #3324.

hazendaz:
closing this, will open new one separately against main.

hazendaz:
@scordio How exactly is modular path excluded?  Javadoc have always ran modular checks.  The legacyMode flag (poorly named) disables that.

hazendaz:
Rebased this and created new PR to main since this got switched to 3.x since last time it was worked on.

PR #3238, opened 2023-10-25
Gather performance tests

### LINKED ISSUE #3170: Some Flaky tests
**Describe the bug**
We found tests flaky failing more frequently when running in a constraint environment.

* Assert core version: using sha commit 9b653c5 
* java version: 17
* test framework version: ? 
* docker image: maven:3.8.3-openjdk-17

The test scenario is showed below.

## Commnet 
Hello, We tried running your project and discovered that it contains some flaky tests (i.e., tests that nondeterministically pass and fail). We found these tests to fail more frequently when running them on certain machines of ours.

To prevent others from running this project and its tests in machines that may result in flaky tests, we suggest adding information to the README.md file indicating the minimum resource configuration for running the tests of this project as to prevent observation of test flakiness.

If we run this project in a machine with 1cpu and 1gb ram, we observe flaky tests. We found that the tests in this project did not have any flaky tests when we ran it on machines with 2cpu and 2gb ram.
  
Here is a list of the tests we have identified and their likelihood of failure on a system with less than the recommended 2 CPUs and 2 GB RAM.

1. CompletableFutureAssert-succeedsWithin#should-allow-assertion-on-future-result-when-completed-normally-within-timeout
2. ByNameSingleExtractor#should-extract-property-with-bare-name-method
3. FutureAssert-succeedsWithin#should-allow-assertion-on-future-result-when-completed-normally-within-timeout
4. Iterables-assertDoesNotHaveDuplicates#should-pass-within-time-constraints
5. PropertyOrFieldSupport-getSimpleValue-With-Map-input#should-extract-field-value-even-if-map-key-matches-given-name

Thank you for your attention to this matter. We hope that our recommendations will be helpful in improving the quality and performance of your project, especially for others to use.

## Reproducing

~~~Dockerfile

FROM maven:3.8.3-openjdk-17

WORKDIR /home/

RUN git clone https://github.com/assertj/assertj && \
  cd assertj && \
  git checkout 9b653c5
  
WORKDIR /home/assertj

RUN mvn install -DskipTests

ENTRYPOINT ["mvn", "test", "-fn"]

~~~

Build the image:
  
~~~bash
$> mkdir tmp

$> cp Dockerfile tmp

$> cd tmp

$> docker build -t assertj . # estimated time of build 3m
~~~
  
Running:
this configuration likely prevents flakiness (no flakiness in 10 runs)
  
~~~bash
$> docker run --rm --memory=2g --cpus=2 --memory-swap=-1 assertj | tee output.txt
$> grep "Failures:"  output.txt # checking results
~~~

 this other configuration –similar to the previous– can’t prevent flaky tests (observation in 10 runs)
~~~bash
$> docker run --rm --memory=1g --cpus=1 --memory-swap=-1 assertj | tee output2.txt
$> grep "Failures:"  output2.txt # checking results
~~~

nicokosi:
Use `flaky` and `performance` JUnit `@Tag` annotations. Cf. https://junit.org/junit5/docs/current/user-guide/#writing-tests-tagging-and-filtering

Use `@Nested` annotation when mixing common tests with performance tests. Cf. https://junit.org/junit5/docs/current/user-guide/#writing-tests-nested

Fixes #3170.

#### Check List:
* Fixes #3170
* Unit tests : NA (no new tests)

### DIFF
--- a/assertj-core/src/test/java/org/assertj/core/api/abstract_/AbstractAssert_withFailMessage_performance_Test.java
+++ b/assertj-core/src/test/java/org/assertj/core/api/abstract_/AbstractAssert_withFailMessage_performance_Test.java
@@ -17,10 +17,12 @@ import static org.assertj.core.api.Assertions.assertThat;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
-// only run for perf metrics
-@Disabled
+@Disabled("only run on demand")
+@Tag("performance")
+@Tag("flaky")
 class AbstractAssert_withFailMessage_performance_Test {
 
   @Test
--- a/assertj-core/src/test/java/org/assertj/core/internal/iterables/Iterables_assertDoesNotHaveDuplicates_Test.java
+++ b/assertj-core/src/test/java/org/assertj/core/internal/iterables/Iterables_assertDoesNotHaveDuplicates_Test.java
@@ -33,17 +33,19 @@ import java.util.stream.Stream;
 import org.assertj.core.api.AssertionInfo;
 import org.assertj.core.internal.Iterables;
 import org.assertj.core.internal.IterablesBaseTest;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
 /**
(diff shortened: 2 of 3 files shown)


nicokosi:
I first gave a try segregating/deactivating tests like `Iterables_assertDoesNotHaveDuplicates_Test#should_pass_within_time_constraints` and `Iterables_assertDoesNotHaveDuplicates_Test#should_pass_within_time_constraints_with_custom_comparison_strategy` but it may not be the best idea (thanks @jcgay for the hint)... :sweat_smile: 

Then I tried to move those tests into the `assertj-performance-tests` module but I failed. :innocent: 

I close this PR. :wave:

PR #3251, opened 2023-11-03
Better error message instead of "actual is not empty while group of values to look for is" #3230

### LINKED ISSUE #3230: Better error message instead of "actual is not empty while group of values to look for is"
#### Feature summary
I want to receive detailed error messages so I can better understand what went wrong in my test.
That is why I chose AssertJ (thank you so much for creating this great lib) but in some edge-cases it does not satisfy this requirement.

#### Example

This assertion:
```
    assertThat(new String[] {"foo", "bar"}).as("detailed extra hint I want to see if this fails").containsExactly(new String[0]);
```
is giving me:
```
java.lang.AssertionError: actual is not empty while group of values to look for is.
	at org.assertj.core.internal.Arrays.failIfEmptySinceActualIsNotEmpty(Arrays.java:774)
	at org.assertj.core.internal.Arrays.commonChecks(Arrays.java:498)
	at org.assertj.core.internal.Arrays.assertContainsExactly(Arrays.java:277)
	at org.assertj.core.internal.ObjectArrays.assertContainsExactly(ObjectArrays.java:298)
	at org.assertj.core.api.AbstractObjectArrayAssert.containsExactlyForProxy(AbstractObjectArrayAssert.java:619)
	at org.assertj.core.api.AbstractObjectArrayAssert.containsExactly(AbstractObjectArrayAssert.java:612)
```

My expected result would be:
```
org.opentest4j.AssertionFailedError: [detailed extra hint I want to see if this fails] 
Expecting actual:
  ["foor", "bar"]
  []
but others were not expected:
  ["foo", "bar"]
...
```
(This is just an example aligned to what I get if the expected array/collection is not empty. Feel free to define the error message as prefer as long as you include the message given via `as` and the actual values.)

pbacz:
This changes the message for arrays and iterables as mentioned in #3230

#### Check List:
* Fixes #3230
* Unit tests : YES

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/error/ActualIsNotEmpty.java
+++ b/assertj-core/src/main/java/org/assertj/core/error/ActualIsNotEmpty.java
@@ -1,0 +1,24 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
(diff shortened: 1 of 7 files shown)


joel-costigliola:
Integrated thanks @pbacz !

PR #3256, opened 2023-11-07
Correct javadoc to reflect method bahavior

### LINKED ISSUE #3249: Wrong javadoc in AbstractObjectArrayAssert (and similar)
I noticed that currently `org.assertj.core.api.AbstractObjectArrayAssert#startsWith` and similar methods in this and other array related classes say

`@throws IllegalArgumentException if the given argument is an empty array.`

while in fact they do throw `AssertError` in that case (and only if the actual is not empty, the check pass if both arrays passed are empty).

This is also reflected in tests like here: `org.assertj.core.internal.objectarrays.ObjectArrays_assertContainsExactly_Test#should_fail_if_array_of_values_to_look_for_is_empty_and_actual_is_not`

```
  @Test
  void should_fail_if_array_of_values_to_look_for_is_empty_and_actual_is_not() {
    assertThatExceptionOfType(AssertionError.class).isThrownBy(() -> arrays.assertContainsExactly(someInfo(), actual, array()));
  }
```

It would be great to fix the javadoc. I can contribute to this.

pbacz:
#### Check List:
* Fixes #3249
* Unit tests : NA

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractBooleanArrayAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractBooleanArrayAssert.java
@@ -205,7 +205,7 @@ public abstract class AbstractBooleanArrayAssert<SELF extends AbstractBooleanArr
    * @param values the given values.
    * @return {@code this} assertion object.
    * @throws NullPointerException if the given argument is {@code null}.
-   * @throws IllegalArgumentException if the given argument is an empty array.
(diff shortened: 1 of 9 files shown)


pbacz:
@joel-costigliola Looks like `CI / OS macOS-latest` failed. Can this be restarted? I haven't changed any code related to the test that fialed.

joel-costigliola:
Integrated thanks @pbacz !

Issue #3288, opened 2023-12-07
Unexpected UnknownFormatConversionException when doing containsExactly on ListAssert


BernhardBln:
**Describe the bug**
When doing a `containsExactly` on a ListAssert, 

**only for the case that all elements are there but the order is wrong** (like, `containsExactlyInAnyOrder` would pass but `containsExactly` would not),

there will be a `UnknownFormatConversionException` if a `%` is part of the data.

* assertj core version: 3.24.2
* java version: 17.0.9
* test framework version: junit 5.10

**Test case reproducing the bug**

```java
    @Test
    void fails() {

        var actual = List.of("xxx", "%");

        // passes
        assertThat(actual).containsExactlyInAnyOrder("%", "xxx");

        // fails with UnknownFormatConversionException
        assertThat(actual).containsExactly("%", "xxx");

    }
```

The stacktrace:

```
java.util.UnknownFormatConversionException: Conversion = '"'

	at java.base/java.util.Formatter.parse(Formatter.java:2750)
	at java.base/java.util.Formatter.format(Formatter.java:2671)
	at java.base/java.util.Formatter.format(Formatter.java:2625)
	at java.base/java.lang.String.format(String.java:4145)
	at org.assertj.core.util.Strings.formatIfArgs(Strings.java:86)
	at org.assertj.core.error.MessageFormatter.format(MessageFormatter.java:62)
	at org.assertj.core.error.BasicErrorMessageFactory.create(BasicErrorMessageFactory.java:107)
	at org.assertj.core.internal.Failures.assertionErrorMessage(Failures.java:144)
	at org.assertj.core.internal.Failures.failure(Failures.java:133)
	at org.assertj.core.internal.Iterables.shouldContainExactlyWithIndexAssertionError(Iterables.java:1169)
	at org.assertj.core.internal.Iterables.assertElementOrder(Iterables.java:1150)
	at org.assertj.core.internal.Iterables.assertContainsExactly(Iterables.java:1137)
	at org.assertj.core.api.AbstractIterableAssert.containsExactlyForProxy(AbstractIterableAssert.java:427)
	at org.assertj.core.api.AbstractIterableAssert.containsExactly(AbstractIterableAssert.java:420)
```

joel-costigliola:
Thanks for reporting this @BernhardBln !

PR #3290, opened 2023-12-08
Escape percentage sign from elements to build format

quaff:
Fix GH-3288


joel-costigliola:
Integrated thanks for the good work @quaff !

PR #3313, opened 2023-12-29
Support multiple AfterAssertionErrorCollected callbacks

Achitheus:
#### Check List:
- [ ] Fixes -> ignore
- [x] Unit tests : YES
- [ ] Javadoc with a code example (on API only) : NA
- [x] PR meets the [contributing guidelines](https://github.com/assertj/assertj/blob/main/CONTRIBUTING.md)

### Let AssertJ integrations safely rely on this fuctionality!
Without this multiple callbacks support there is a problem with using `DefaultAssertionErrorCollector.setAfterAssertionErrorCollected()` method.
It can be used only by user at the moment, but not by another frameworks developers .  
Why do I think so?  Because `callback` field in the `DefaultAssertionErrorCollector` class is not instance of any `Collection\<AfterAssertionErrorCollected\>` type, but is just `AfterAssertionErrorCollected`. Thus `setAfterAssertionErrorCollected()` method can be executed only once for each `SoftAssertions` object (by one exact actor - user), because every next call leads to overriding and erasing previously set callback.


joel-costigliola:
@Achitheus that's interesting, could you give us a real life use case to show why this could be useful ?

Achitheus:
> @Achitheus that's interesting, could you give us a real life use case to show why this could be useful ?

Yes. Im gonna add AssertJ soft assertions support for java Allure reports. I am already done with it and even tested, but haven't suggested PR at the moment. My solution relys on this functionality to detect assertion error and mark current step as failed and then escalate this status to all outer steps of the step chain.
But there is a question. What if user also try to set his own callback for logs for example or what if he add dependency which also use this feature? Conflicts will definitely happen. And in this conflict, whoever is last is the winner because he rewrited all previous callbacks (only the last one to be precise). That's why I need this change

Achitheus:
I should have pushed several commits at once. Sorry. I hope you don't have actions proccesing time restrictions or something :)

Achitheus:
@joel-costigliola what do you think? Could you please run actions one more time?

joel-costigliola:
Sorry I missed your reply, I'll think about it this week end, but gut feeling is it should be ok

Achitheus:
@joel-costigliola I'm not rushing you but just reminding :)

joel-costigliola:
Thanks for the reminder, I'll put it for the next version 3.26.0

assertj-core/src/main/java/org/assertj/core/api/DefaultAssertionErrorCollector.java
```diff
@@ -28,12 +30,13 @@ public class DefaultAssertionErrorCollector implements AssertionErrorCollector {
   private volatile boolean wasSuccess = true;
   private List<AssertionError> collectedAssertionErrors = synchronizedList(new ArrayList<>());
 
-  private AfterAssertionErrorCollected callback = this;
+  private Set<AfterAssertionErrorCollected> callbacks = new HashSet<>();
```
joel-costigliola, line 33: We probably want an ordered collection here in case the ordering of callback is important, so a list or a linked hash set.
Why would we need a set here ? in case you want the registration process to be idempotent ? any use case in min ?

assertj-core/src/main/java/org/assertj/core/api/DefaultAssertionErrorCollector.java
```diff
@@ -114,7 +117,7 @@ public List<AssertionError> assertionErrorsCollected() {
    * @since 3.17.0
    */
   public void setAfterAssertionErrorCollected(AfterAssertionErrorCollected afterAssertionErrorCollected) {
-    callback = afterAssertionErrorCollected;
+    callbacks.add(afterAssertionErrorCollected);
```
joel-costigliola, line 120: This is misleading to me, I would not expect adding a callback but defining the unique callback, ex:
```java
   softly.setAfterAssertionErrorCollected(err -> stringList.add(err.getMessage()));
    softly.setAfterAssertionErrorCollected(throwableList::add);
```
I expect `throwableList::add` to be the callback and not both of them, like when you do:
```java
person.setName("Jack");
person.setName("Joe");
```
you don't expect `person` to have two names.

For backward compatibility, I would define a new method `addAfterAssertionErrorCollected`.

assertj-core/src/test/java/org/assertj/core/api/DefaultAssertionErrorCollector_setAfterAssertionErrorCollected_Test.java
```diff
@@ -0,0 +1,71 @@
...
+
+class DefaultAssertionErrorCollector_setAfterAssertionErrorCollected_Test {
+  private List<String> stringList;
+  private List<Throwable> throwableList;
+  private SoftAssertions softly;
+
+  @BeforeEach
+  void given() {
+    stringList = new ArrayList<>();
+    throwableList = new ArrayList<>();
```
joel-costigliola, line 31: Rename to `errorMessages` (more intention revealing) and `throwables` (no need to know that it is a list)

assertj-core/src/test/java/org/assertj/core/api/DefaultAssertionErrorCollector_setAfterAssertionErrorCollected_Test.java
```diff
@@ -0,0 +1,71 @@
...
+  void should_perform_both_specified_actions_on_each_assertion_error() {
+    // WHEN
+    softly.setAfterAssertionErrorCollected(err -> stringList.add(err.getMessage()));
+    softly.setAfterAssertionErrorCollected(throwableList::add);
+    softly.assertThat(1)
+          .isEqualTo(1_000)
+          .isBetween(7, 15)
+          .isEven();
+    // THEN
+    then(stringList).hasSameSizeAs(throwableList).hasSize(3);
```
joel-costigliola, line 45: cosmetic: new line after the first assertion to get all assertions aligned:
```java
then(stringList).hasSameSizeAs(throwableList)
                .hasSize(3);
```

assertj-core/src/test/java/org/assertj/core/api/DefaultAssertionErrorCollector_setAfterAssertionErrorCollected_Test.java
```diff
@@ -0,0 +1,71 @@
...
+    softly.setAfterAssertionErrorCollected(err -> stringList.add(err.getMessage()));
+    softly.setAfterAssertionErrorCollected(throwableList::add);
+    softly.collectAssertionError(new AssertionError("hello"));
+    softly.collectAssertionError(new AssertionError("world"));
+    // THEN
+    then(stringList).hasSameSizeAs(throwableList).hasSize(2);
+  }
+
+  @Test
+  void should_pass_if_the_callback_object_can_be_set_only_one_time() {
```
joel-costigliola, line 60: rename to `should_register_the_same_callback_only_once`

joel-costigliola (changes requested):

joel-costigliola:
@Achitheus thanks, PR commented.

assertj-core/src/main/java/org/assertj/core/api/DefaultAssertionErrorCollector.java
Achitheus: Maybe I was too focused on my own case while choosing Collection. I used AOP and it seems like my `pointcut` catched creating the same soft assertion object twice. I dont remember details to be honest.
I expected this question and was ready for it, but I found a problem. According to JLS-15.27.4:

> the identity of the result of evaluating a lambda expression is unpredictable, and therefore identity-sensitive operations may produce different results in different implementations of the Java programming language, or even upon different lambda expression evaluations in the same implementation.

So it looks risky and as you said I'd suggest good old `ArrayList`. But should I make it as usual or like it done for `collectedErrors` field?
`private List<AssertionError> collectedAssertionErrors = synchronizedList(new ArrayList<>());`
Unfortunatelly I'm not good at concurrency issues.

joel-costigliola: Safer to use a `synchronizedList`, won't be a performance issue since the callback registration won't happen often

assertj-core/src/main/java/org/assertj/core/api/DefaultAssertionErrorCollector.java
Achitheus: I got it. I'd suggest just rename this method to `addAfterAssertionErrorCollected()`, because according to generally accepted understanding the `set` version should do something like:
```
public void setAfterAssertionErrorCollected(AfterAssertionErrorCollected afterAssertionErrorCollected) {
    callbacks.clear();
    callbacks.add(this);
    callbacks.add(afterAssertionErrorCollected);
}
```
Or even without adding `this`.
But you are right about compatibility, so maby there should be some warning in `javadoc` at least. Something like:

> Same as {@link DefaultAssertionErrorCollector#addAfterAssertionErrorCollected(AfterAssertionErrorCollected)}, but
>  is also deletes all previously set callbacks except the {@code onAssertionErrorCollected()} of {@code this} object.
>Please consider using {@link DefaultAssertionErrorCollector#addAfterAssertionErrorCollected(AfterAssertionErrorCollected)}
>  instead, because another frameworks and integrations can also use this functionality.
> 
>  @param afterAssertionErrorCollected the callback.
> 
>  @since 3.17.0

(this is a draft. I can do better with dictionary and translator :) )
Thus, the previous `javadoc` should be moved to `add` method I suppose.
What do you think?

Achitheus: The previous version of the method overwrites even "this.onAssertionErrorCollected()" calllback. So I won't add `callbacks.add(this)` command to `set` method version.

assertj-core/src/test/java/org/assertj/core/api/DefaultAssertionErrorCollector_setAfterAssertionErrorCollected_Test.java
Achitheus: For some reason I thought it's more readable when you see `someList.size()` or `someList.isEmpty()`. But I've always noticed that it's too expensive in terms of name lenghts. Thank you!

Achitheus:
@joel-costigliola Thank you for your insightful comments! Let me know if there is anything else I can do/fix.

Achitheus:
@joel-costigliola I finally realized my mistake. I should use `mvn verify` instead of `mvn test` to see locally format violations and license issues.
I hope next time everything will be "green" :) Sorry

assertj-core/src/main/java/org/assertj/core/api/DefaultAssertionErrorCollector.java
```diff
@@ -74,6 +75,22 @@ public List<AssertionError> assertionErrorsCollected() {
...
+  /**
+   * Same as {@link DefaultAssertionErrorCollector#addAfterAssertionErrorCollected(AfterAssertionErrorCollected)}, but
+   * also removes all previously added callbacks. Please consider using
+   * {@link DefaultAssertionErrorCollector#addAfterAssertionErrorCollected(AfterAssertionErrorCollected)}
+   * instead, because another frameworks and integrations can also use this functionality.
+   *
+   * @param afterAssertionErrorCollected the callback.
+   *
+   * @since 3.17.0
+   * @deprecated
```
joel-costigliola, line 87: I don't see it as deprecated, users that already register only one callback don't want to migrate to `addAfterAssertionErrorCollected` so we should keep the initial javadoc just adding that it clears all callbacks previously added with `addAfterAssertionErrorCollected` (as you did)

assertj-core/src/test/java/org/assertj/core/api/DefaultAssertionErrorCollector_addAfterAssertionErrorCollected_Test.java
```diff
@@ -0,0 +1,73 @@
...
+    // GIVEN
+    AfterAssertionErrorCollected callback = throwables::add;
+    // WHEN
+    for (int i = 0; i < 10; i++) {
+      softly.addAfterAssertionErrorCollected(callback);
+    }
+    softly.collectAssertionError(new AssertionError("hello"));
+    // THEN
+    then(throwables).hasSize(10);
+  }
```
joel-costigliola, line 72: add a unit test to check that setAfterAssertionErrorCollected replace all registered callback with the one specified

joel-costigliola (commented):
Second round of comments

assertj-core/src/test/java/org/assertj/core/api/DefaultAssertionErrorCollector_addAfterAssertionErrorCollected_Test.java
Achitheus: Done. But I would like to clarify whether this test should be moved from this class 
`DefaultAssertionErrorCollector_addAfterAssertionErrorCollected_Test`
to this new one
`DefaultAssertionErrorCollector_setAfterAssertionErrorCollected_Test`
?
Or even moved to `org/example/test/DefaultAssertionErrorCollector_Test.java`
Or let it be in the current class, but rename the class to
`DefaultAssertionErrorCollector_afterAssertionErrorCollected_methods_Test`
or
`DefaultAssertionErrorCollector_callbacks_Test`
or something?

assertj-core/src/main/java/org/assertj/core/api/DefaultAssertionErrorCollector.java
Achitheus: added to commit

Achitheus:
@joel-costigliola build success locally. 
As I mentioned above, the queston of the right place for `setAfter...()` method test is still open for me. I left it in the current test class for now.
I'm ready for the third round :) Let me know if there is anything else I can do!

joel-costigliola:
Integrated thanks @Achitheus !

PR #3319, opened 2024-01-02
Fix deep comparison for singleton maps in `assertContainsExactly`

### LINKED ISSUE #3315: AssertJ cannot compare arrays in maps
**Describe the bug**

* assertj core version: 3.25.0
* java version: 17
* test framework version: JUnit 5.10.1
* os (if relevant): MacOS Sonoma 14.2.1

This code worked in AssertJ 3.24.2, but fails in AssertJ 3.25.0:

```java
import java.util.Map;

  @Test
  void fails() {
    assertThat(Map.of("firstName", new String[] {"Bob"}))
      .containsExactlyEntriesOf(Map.of("firstName", new String[] {"Bob"}));
  }
```

This test fails with the error message:
```java
    java.lang.AssertionError: 
      "firstName"=["Bob"]
    whereas expected element was:
      "firstName"=["Bob"]
        at play.utils.UtilsTest.fails(UtilsTest.java:50)
```

scordio:
Fixes #3315.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/internal/Maps.java
+++ b/assertj-core/src/main/java/org/assertj/core/internal/Maps.java
@@ -554,7 +554,7 @@ public class Maps {
     // Stream API avoided for performance reasons
     for (Entry<? extends K, ? extends V> entry : entries) {
       // must perform deep equals comparison on values as Map.remove(Object, Object) relies on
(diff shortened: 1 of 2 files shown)


scordio:
Superseded by #3321.

PR #3324, opened 2024-01-05
Fix unresolvable Javadoc stylesheet URLs, simplify configuration

### LINKED ISSUE #3323: Unresolvable stylesheet URLs in latest Javadoc
When visiting https://www.javadoc.io/doc/org.assertj/assertj-core/latest/index.html from Chrome, a few errors are logged out to the console:

```
index.html:1
Refused to apply style from 'https://www.javadoc.io/static/org.assertj/assertj-core/3.25.1/assertj-javadoc.css' because its MIME type ('application/xml') is not a supported stylesheet MIME type, and strict MIME checking is enabled.
index.html:1
Refused to apply style from 'https://www.javadoc.io/static/org.assertj/assertj-core/3.25.1/resources/fonts/dejavu.css' because its MIME type ('application/xml') is not a supported stylesheet MIME type, and strict MIME checking is enabled.
index.html:110
Refused to apply style from 'https://www.javadoc.io/static/org.assertj/assertj-core/3.25.1/assertj-javadoc.css' because its MIME type ('application/xml') is not a supported stylesheet MIME type, and strict MIME checking is enabled.
``` 

and the following issues are reported:

Screenshot 2024-01-05 at 09 22 21

scordio:
Fixes #3323.


scordio:
After the changes, there is still one issue to resolve:

```
index.html:1 
Refused to apply style from 'http://localhost:63342/assertj/assertj-core/target/site/apidocs/resources/fonts/dejavu.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled.
```

Screenshot 2024-01-05 at 18 21 59

scordio:
The `dejavu.css` issue appears also with other projects (e.g., https://javadoc.io/doc/org.springframework/spring-core/latest/index.html) and seems to be related to [JDK-8227487](https://bugs.openjdk.org/browse/JDK-8227487).

scordio:
> The `dejavu.css` issue also occurs with other projects (e.g., https://javadoc.io/doc/org.springframework/spring-core/latest/index.html) and seems to be related to [JDK-8227487](https://bugs.openjdk.org/browse/JDK-8227487) and [JDK-8324774](https://bugs.openjdk.org/browse/JDK-8324774).
> 
> Not something we can currently fix in AssertJ.

Since Java 23, DejaVu fonts are included by default, and the `--no-fonts` option can be used to prevent it.

Starting with version 3.11.3 of the `maven-javadoc-plugin`, `--no-fonts` [is applied by default](https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#disableNoFonts), which makes sense in our use case as we override the related stylesheet properties:

https://github.com/assertj/assertj/blob/584ecc0962c7b198836e24fc04865ef3fb64bc81/javadoc-theme/assertj-theme.css#L3-L4

However, this still results in a 404, as the default Javadoc stylesheet still tries to `GET /resource-files/fonts/dejavu.css`. I've reached out to Oracle about it.

PR #3325, opened 2024-01-07
Add `InstanceOfAssertFactory` for `Set` instances

### LINKED ISSUE #3322: Improve interoperability of `asInstanceOf` with `Set` instances
Right now, there is no nice way to call `asInstanceOf` when a `Set` instance is expected.

Current solutions:

```java
Object actual = Set.of("first", "second");

assertThat(actual)
  .asInstanceOf(InstanceOfAssertFactories.type(Set.class))
  .satisfies(set -> assertThat(set).containsExactly("first", "second")); // unchecked warning

assertThat(actual)
  .asInstanceOf(InstanceOfAssertFactories.COLLECTION) // no implicit `isInstanceOf(Set.class)`
  .containsExactly("first", "second"));

assertThat(map.get("value"))
  .asInstanceOf(new InstanceOfAssertFactory<>(Set.class, Assertions::<String> assertThat)) // verbose
  .containsExactly("first", "second");
```

By design, we didn't introduce any `InstanceOfAssertFactory` for types that don't have more specific assertions, i.e., there is no `AbstractSetAssert` today because it wouldn't have any added value compared to [`AbstractCollectionAssert`](https://www.javadoc.io/doc/org.assertj/assertj-core/latest/org/assertj/core/api/AbstractCollectionAssert.html).

We should investigate how to improve such use cases.

Triggered by: https://github.com/spring-projects/spring-framework/blob/2784f6008e97c7283b2e57e98d18a89fe8f530ae/spring-beans/src/test/java/org/springframework/beans/factory/config/YamlProcessorTests.java#L144-L153

scordio:
Closes #3322.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
@@ -37,6 +37,7 @@ import java.util.Optional;
 import java.util.OptionalDouble;
 import java.util.OptionalInt;
 import java.util.OptionalLong;
+import java.util.Set;
 import java.util.Spliterator;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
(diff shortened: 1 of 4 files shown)


assertj-core/src/main/java/org/assertj/core/api/InstanceOfAssertFactories.java
```diff
@@ -908,6 +909,32 @@ static <ELEMENT> InstanceOfAssertFactory<List, ListAssert<ELEMENT>> list(Class<E
     return new InstanceOfAssertFactory<>(List.class, Assertions::<ELEMENT> assertThat);
   }
 
+  /**
+   * {@link InstanceOfAssertFactory} for a {@link Set}, assuming {@code Object} as element type.
+   *
+   * @see #set(Class)
+   * @since 3.21.0
```
sbrannen, line 916:
```suggestion
   * @since 3.26.0
```

PR #3333, opened 2024-01-10
Preserves the original order of elements when returning the duplicates

### LINKED ISSUE #3331: assertThat(collection).doesNotHaveDuplicates() is not sorted, even if the collection is
The error message of `assertThat(collection).doesNotHaveDuplicates()` lists the duplicates in a different order than the elements of the collection actually have. That is not an error per-se, but can make it harder to find the root cause of the failing assertion. So this might also be labeled improvement instead of bug.

* assertj core version: 3.25.1

```java
    @Test
    void sortOrder() throws Exception {
        var list = List.of("org.objectweb.asm.source",
                "org.eclipse.emf.compare.diagram.gmf.source.feature.group",
                "org.eclipse.emf.compare.rcp.ui.source.feature.group",
                "org.objectweb.asm",
                "org.eclipse.jgit.feature.group",
                "org.eclipse.emf.compare.diagram.gmf.source.feature.group",
                "org.objectweb.asm",
                "org.apache.httpcomponents.httpclient",
                "org.eclipse.emf.compare.rcp.ui.source.feature.group",
                "org.eclipse.jgit.feature.group",
                "org.eclipse.emf.compare.ide.ui.source.feature.group");
        assertThat(list.stream().sorted()).doesNotHaveDuplicates();
    }
```
Note that the list gets sorted in the assertion. Still the output is not sorted, because the duplicates are collected in a set which is not insertion order stable.
```
Found duplicate(s):
  ["org.eclipse.emf.compare.diagram.gmf.source.feature.group",
    "org.eclipse.emf.compare.rcp.ui.source.feature.group",
    "org.objectweb.asm",
    "org.eclipse.jgit.feature.group"]
in:
  ["org.apache.httpcomponents.httpclient",
    "org.eclipse.emf.compare.diagram.gmf.source.feature.group",
    "org.eclipse.emf.compare.diagram.gmf.source.feature.group",
    "org.eclipse.emf.compare.ide.ui.source.feature.group",
    "org.eclipse.emf.compare.rcp.ui.source.feature.group",
    "org.eclipse.emf.compare.rcp.ui.source.feature.group",
    "org.eclipse.jgit.feature.group",
    "org.eclipse.jgit.feature.group",
    "org.objectweb.asm",
    "org.objectweb.asm",
    "org.objectweb.asm.source"]
```

ranjitshinde91:
#### Check List:
* Fixes #3331
* Unit tests : YES

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/internal/AbstractComparisonStrategy.java
+++ b/assertj-core/src/main/java/org/assertj/core/internal/AbstractComparisonStrategy.java
@@ -40,10 +40,10 @@ public abstract class AbstractComparisonStrategy implements ComparisonStrategy {
         noDuplicates.add(element);
       }
     }
-    return duplicatesInOriginalOrder(iterable, duplicates);
+    return duplicatesWithOrderPreserved(iterable, duplicates);
   }
 
-  private Iterable<?> duplicatesInOriginalOrder(Iterable<?> iterable, Set<Object> duplicates) {
+  private Iterable<?> duplicatesWithOrderPreserved(Iterable<?> iterable, Set<Object> duplicates) {
     Set<Object> duplicatesInOriginalOrder  = new LinkedHashSet<>();
     // for elements of type Set collections, ensure each element is added only once
     Set<Object> alreadyAddedDuplicates = newSetUsingComparisonStrategy();
--- a/assertj-core/src/main/java/org/assertj/core/internal/AbstractComparisonStrategy.java
+++ b/assertj-core/src/main/java/org/assertj/core/internal/AbstractComparisonStrategy.java
@@ -44,11 +44,11 @@ public abstract class AbstractComparisonStrategy implements ComparisonStrategy {
   }
 
   private Iterable<?> duplicatesWithOrderPreserved(Iterable<?> iterable, Set<Object> duplicates) {
-    Set<Object> duplicatesInOriginalOrder  = new LinkedHashSet<>();
+    Set<Object> duplicatesInOriginalOrder = new LinkedHashSet<>();
     // for elements of type Set collections, ensure each element is added only once
     Set<Object> alreadyAddedDuplicates = newSetUsingComparisonStrategy();
     for (Object element : iterable) {
-      if (duplicates.contains(element)&& !alreadyAddedDuplicates.contains(element)) {
+      if (duplicates.contains(element) && !alreadyAddedDuplicates.contains(element)) {
         duplicatesInOriginalOrder.add(element);
         alreadyAddedDuplicates.add(element);
       }
--- a/assertj-core/src/main/java/org/assertj/core/internal/AbstractComparisonStrategy.java
+++ b/assertj-core/src/main/java/org/assertj/core/internal/AbstractComparisonStrategy.java
@@ -22,7 +22,7 @@ import java.util.Set;
 
 /**
  * Base implementation of {@link ComparisonStrategy} contract.
- * 
+ *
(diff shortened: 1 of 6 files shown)


joel-costigliola:
Please add a unit test proving that the order is honored.

assertj-core/src/main/java/org/assertj/core/internal/AbstractComparisonStrategy.java
```diff
@@ -39,7 +40,20 @@ public Iterable<?> duplicatesFrom(Iterable<?> iterable) {
...
-    return duplicates;
+    return duplicatesWithOrderPreserved(iterable, duplicates);
+  }
+
+  private Iterable<?> duplicatesWithOrderPreserved(Iterable<?> iterable, Set<Object> duplicates) {
+    Set<Object> duplicatesInOriginalOrder  = new LinkedHashSet<>();
+    // for elements of type Set collections, ensure each element is added only once
+    Set<Object> alreadyAddedDuplicates = newSetUsingComparisonStrategy();
+    for (Object element : iterable) {
+      if (duplicates.contains(element)&& !alreadyAddedDuplicates.contains(element)) {
```
Bananeweizen, line 51: you can get rid of addedAddDuplicates entirely, if you make the second part of the if test `!duplicatesInOriginalOrder.contains(element)`

ranjitshinde91: For elements of array type e.g. newArrayList(array("Merry"), array("Frodo"),  array("Merry")) TreeSet uses custom comparator to identify the duplicates. This is been done by checking the deeply equals for two array types( areEquals method). So elements array("Merry") and array("Merry") will be treated as duplicates.

since we don't have this comparator in LinkedHashSet, it will not be able to identify two arrays with same elements as duplicates.

Bananeweizen (commented):
looks okay to me. I just don't understand why so many tests have been adapted. I would expect that those can remain as before, can't they?

ranjitshinde91:
Tests are changed because order of duplicate elements in response is changed now. 

e.g if the list has (6, -8, 6, 8) previous implementation with TreeSet + AbsoluteValueComparator would  return (6, 8)
now with order preserved it returns (6, -8)

joel-costigliola:
@ranjitshinde91 there are tests failing, can you look at them ?

assertj-core/src/main/java/org/assertj/core/internal/AbstractComparisonStrategy.java
kulgan: I might be missing a lot of context here, so ignore me if that is the case. Instead of introducing a private method, can this be done in the `duplicatesFrom` method instead, like replacing Lines 34 with a LinkedHashSet?

ranjitshinde91:
@joel-costigliola there were few formatting errors, fixed those.

assertj-core/src/main/java/org/assertj/core/internal/AbstractComparisonStrategy.java
ranjitshinde91: earlier it was not possible but with switch option-2 (first duplicate first in response) implementation its possible. 

I have made that change.

Bananeweizen (commented):
looks good to me.

kulgan:
Looks good to me

joel-costigliola:
Integrated thanks @ranjitshinde91, @Bananeweizen and @kulgan !

Issue #3320, opened 2024-01-02
Regression in Recursive Comparison


austinarbor-wk:
**Describe the bug**
It appears that 3.25.0 has changed behavior of recursive comparison

* assertj core version: 3.25.0
* java version: 17
* test framework version: junit platform 5.10.1

The below test succeeds in 3.24.2 but fails in 3.25.0

```java
  @Test
  void testRecursiveCompare() {
    var a1 = new A();
    a1.name = "test1";
    var b1 = new B();
    a1.b = b1;
    b1.name = "test2";

    var a2 = new A();
    a2.name = "test1";
    a2.b = b1;

    var list1 = List.of(a1);
    var list2 = List.of(a2);

    var cfg = RecursiveComparisonConfiguration.builder().withComparedFields("name", "b").build();
    cfg.useOverriddenEquals();

    assertThat(list1)
        .usingRecursiveFieldByFieldElementComparator(cfg)
        .containsExactlyElementsOf(list2);
  }

  private static class A {
    private String name;

    private String guid = UUID.randomUUID().toString();

    private B b;

    @Override
    public boolean equals(Object o) {
      if (this == o) {
        return true;
      }
      if (!(o instanceof A a)) {
        return false;
      }
      return Objects.equals(name, a.name) && Objects.equals(guid, a.guid) && Objects.equals(b, a.b);
    }

    @Override
    public int hashCode() {
      return Objects.hash(name, guid, b);
    }
  }

  private static class B {
    private String name;

    private String guid = UUID.randomUUID().toString();

    @Override
    public boolean equals(Object o) {
      if (this == o) {
        return true;
      }
      if (!(o instanceof B b)) {
        return false;
      }
      return Objects.equals(name, b.name) && Objects.equals(guid, b.guid);
    }
  }
```

scordio:
Thanks for reporting it, @austinarbor-wk!

scordio:
Probably related to 8379cac6fa83229b5ff0d516ccb1114096700816.

Still unsure if the behavior in version 3.24.2 was correct, though.

austinarbor-wk:
I think my expectation from the original code sample above based on the javadoc is that the comparison would be equivalent to the compound assertion of
```java
assertThat(a1.name).isEqualTo(a2.name)
assertThat(a1.b).isEqualTo(a2.b)
```

Is my understanding incorrect?

scordio:
[...]

PR #3350, opened 2024-01-26
Add test case for slowness in deeply-nested usingRecursiveComparison().isEqualTo() uses in 

ash211:
First, thank you for the assertj project!  It's been a joy to work with and we use it extensively across my company's Java ecosystem.

I'm opening this PR to demonstrate a performance regression we've observed between assertj 3.24.2 and 3.25.2.  The regression is significant enough that we're unable to take 3.25.2 and are holding back on 3.24.2 for the time being.  This is inconvenient though, so we'd like to work together to identify and implement a fix that restores the previous performance characteristics.

There are 3 new test cases:
1. `FieldLocation_Test.should_build_from_long_nested_path_in_reasonable_time()` <-- PASS
2. `RecursiveComparisonConfiguration_getActualNonIgnoreFields_Test.should_return_fields_in_a_reasonable_amount_of_time_for_deeply_nested_object()` <-- PASS
3. `RecursiveComparisonAssert_isEqualTo_Test.can_compare_deeply_nested_objects_in_reasonable_time()` <-- FAIL

The first two test cases pass locally for me, as they execute in a reasonable amount of time.  But the third one fails because it takes too long.  When copying this test case onto the old 3.24.2 release, it passes.  This is the shape of assertion that is breaking for us.

In the new 3.25.2 release, this test case seems to be triggering quadratic behavior in time.  Uncommenting each additional link in the list doubles the duration of the test.  On my machine the test as written takes 15sec, when I would expect it to be under a second.  For reference, on 3.24.2 it takes ~275ms.

I think it was introduced by https://github.com/assertj/assertj/commit/0eb44cc99f1d0e106e2335d0b1e851a85bf73466 but have not confirmed.

cc @joel-costigliola

cc @schlosna we thought 3.25.2 would fix this perf regression -- https://github.com/assertj/assertj/issues/3320#issuecomment-1901099549 -- but it sadly doesn't

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/FieldLocation.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/FieldLocation.java
@@ -14,14 +14,17 @@ package org.assertj.core.api.recursive.comparison;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.unmodifiableList;
+import static java.util.Collections.unmodifiableSet;
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.joining;
-import static java.util.stream.Collectors.toList;
 import static org.assertj.core.util.Lists.list;
+import static org.assertj.core.util.Sets.newLinkedHashSet;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 import java.util.regex.Pattern;
 
 /**
@@ -33,7 +36,7 @@ public final class FieldLocation implements Comparable<FieldLocation> {
 
   private final String pathToUseInRules;
   private final List<String> decomposedPath;
-  private final List<String> pathsHierarchyToUseInRules;
+  private final Set<String> pathsHierarchyToUseInRules;
 
   public FieldLocation(List<String> path) {
     decomposedPath = unmodifiableList(requireNonNull(path, "path cannot be null"));
@@ -163,7 +166,10 @@ public final class FieldLocation implements Comparable<FieldLocation> {
 
   @Override
   public int hashCode() {
-    return Objects.hash(pathToUseInRules, decomposedPath, pathsHierarchyToUseInRules);
+    int result = Objects.hashCode(pathToUseInRules);
+    result = 31 * result + Objects.hashCode(decomposedPath);
+    result = 31 * result + Objects.hashCode(pathsHierarchyToUseInRules);
+    return result;
   }
 
   @Override
@@ -195,6 +201,10 @@ public final class FieldLocation implements Comparable<FieldLocation> {
   public boolean isRoot() {
     // root is the top level object compared or in case of the top level is a iterable/array the elements are considered as roots.
     // we don't do it for optional it has a 'value' field so for the moment
+    return isRootPath(pathToUseInRules);
+  }
+
+  private boolean isRootPath(String pathToUseInRules) {
     return pathToUseInRules.isEmpty();
   }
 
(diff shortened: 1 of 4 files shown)


joel-costigliola:
Thanks for reporting this @ash211, I'll look into it within the next few days.

schlosna:
@ash211 I created https://github.com/ash211/assertj/pull/1 on top of your branch & this PR with some improvements that pass the tests you added and seem to bring things back in line with 3.24.2 performance of recursive comparisons.

joel-costigliola:
@schlosna the additions you did look good, do you mind having your version in this PR so that I start integrating it ?

schlosna:
> @schlosna the additions you did look good, do you mind having your version in this PR so that I start integrating it ?

Sounds good to me. If we want it in this PR I think @ash211 has to merge https://github.com/ash211/assertj/pull/1 into his branch. I could also open a separate PR directly to assertj that includes both @ash211 test cases and the fix. I'm happy to do either.

ash211:
Merged David’s PR into this one. Thank you!

assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/FieldLocation.java
```diff
@@ -163,7 +166,10 @@ public boolean equals(Object obj) {
 
   @Override
   public int hashCode() {
-    return Objects.hash(pathToUseInRules, decomposedPath, pathsHierarchyToUseInRules);
+    int result = Objects.hashCode(pathToUseInRules);
+    result = 31 * result + Objects.hashCode(decomposedPath);
+    result = 31 * result + Objects.hashCode(pathsHierarchyToUseInRules);
+    return result;
```
joel-costigliola, line 172: @schlosna is there a reason not to only use `pathToUseInRules` as I suggested ? Was there any issue doing that ?

joel-costigliola: No biggie as this implementation is fine anyway, and the performance looks good on my machine.

schlosna: The [equals verifier test was failing](https://github.com/schlosna/assertj-core/actions/runs/7671553403/job/20910160863#step:4:6666) as `hashCode()` did not use all of the fields that `equals` does, so I rolled `hashCode` back per https://github.com/ash211/assertj/pull/1#discussion_r1468652427 .

If we really want, we could do something like following so that the other fields contribute to hash code in a cheap way, but I'm inclined to leave it as is unless `Objects.hashCode` becomes a further bottleneck.


```
  @Override
  public int hashCode() {
    int result = Objects.hashCode(pathToUseInRules);
    // The other fields are derivations of pathToUseInRules, so not truly necessary for hashCode.
    // To make EqualsVerifier happy, we can use collection sizes as cheap mechanism to contribute to hash code.
    result = 31 * result + (decomposedPath == null ? 0 : decomposedPath.size());
    result = 31 * result + (pathsHierarchyToUseInRules == null ? 0 : pathsHierarchyToUseInRules.size());
    return result;
  }
```

joel-costigliola: Fine by me, I tried using only `pathToUseInRules` but it did not make a noticeable improvement performance wise.

joel-costigliola:
Integrated thanks for the good work @ash211 and @schlosna!

ash211:
Much appreciated, thank you @joel-costigliola :D

schlosna:
> Integrated thanks for the good work @ash211 and @schlosna!

Excellent! @joel-costigliola thanks again for the great library and quick turnaround!

PR #3371, opened 2024-02-19
Avoid duplicating maven-javadoc-plugin configuration and CSS files

armandino:
#### Check List:
* Fixes N/A
* Unit tests : NA

**Summary**
The `assertj-core` and `assertj-guava` modules contain identical Javadoc CSS files and almost identical  `maven-javadoc-plugin` configuration. This PR modifies the duplication in pom and moves CSS files to another directory to allow them to be referenced in sub-modules.


pom.xml
```diff
@@ -164,6 +164,26 @@
           <artifactId>maven-clean-plugin</artifactId>
           <version>${maven-clean-plugin.version}</version>
         </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-javadoc-plugin</artifactId>
+          <configuration>
+            <additionalOptions>-html5 --allow-script-in-comments</additionalOptions>
+            <javadocDirectory>${maven.multiModuleProjectDirectory}/javadoc-theme</javadocDirectory>
```
scordio, line 172: I'm not sure we should rely on this property, see [MNG-6589](https://issues.apache.org/jira/browse/MNG-6589).

Maven 4 will bring better options, though ([MNG-7038](https://issues.apache.org/jira/browse/MNG-7038)).

armandino: @scordio wasn't aware of that. Thanks for the links!

scordio: You could use the `rootDirectory` property that I introduced for a similar challenge with Spotless:

```suggestion
            <javadocDirectory>${rootDirectory}/javadoc-theme</javadocDirectory>
```

Would that help?

armandino: Done, thanks!

scordio:
Thanks, @armandino!

PR #3377, opened 2024-02-25
Add `createAssert(ValueProvider)` to `AssertFactory`

### LINKED ISSUE #3368: AssertFactory with the ability to convert the value upfront
#### Feature summary
While working on the AssertJ support for Spring MVC, I've realized that assertJ does not provide a lot of conversion utilities for the actual value, which is more than fair, but can be helpful higher in the stack. 

One of the support we have is for json path where you can craft an expression against the body result of an HTTP request and further assert the content. Our json path feature an "extractingPath" that is similar to AssertJ's extracting feature. But that operates on the raw value and we want to be able to assert that the raw JSON can be deserialized to a DTO.

#### Example

Let's take a silly little example with a family API that returns this:

```json
{
  "familyMembers": [
    { "name": "Homer" },
    { "name": "Marge" },
    { "name": "Bart" },
    { "name": "Lisa" },
    { "name": "Maggie" }
  ]
}
```

Ignoring the setup, this is the pseudo code I am talking about:

```java
@Test
void convertToTargetType() {
    assertThat(mvc.perform(get("/family")).body.jsonPath()
            .extractingPath("$.familyMembers[0]").convertTo(Member.class)
            .satisfies(member -> assertThat(member.name).isEqualTo("Homer"));
}
```

This works but I am returning an `AbstractObjectAssert` against my DTO that represent a family member. What's I'd like to do is to express a type and, at the same time, provide a dedicated `Assert` implementation for it.

`InstanceOfAssertFactory` does what I need but `getType` is package private so I can't access it. Also I like the idea of having an interface for this concept and expose a `Type` so that generic types can be taken into account perhaps?

scordio:
Closes #3368.

Open points:

- [x] Proper test coverage for the new `InstanceOfAssertFactory` constructor
- [x] Proper test coverage for the new `createAssert(ValueProvider)` of `InstanceOfAssertFactory`
- [x] Proper test coverage for all the existing `InstanceOfAssertFactories` instances and related `createAssert(ValueProvider)` behavior.
- [x] Integration tests based on `DefaultConversionService` from `spring-core`

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AssertFactory.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AssertFactory.java
@@ -12,6 +12,10 @@
  */
 package org.assertj.core.api;
 
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.function.Function;
+
 /**
  * A single method factory interface to create an {@link Assert} for a given value.
  * This factory method typically wraps a call to <code>assertThat(actual)</code>
@@ -29,12 +33,72 @@ package org.assertj.core.api;
 public interface AssertFactory<T, ASSERT extends Assert<?, ?>> {
 
   /**
-   * Creates the custom {@link Assert} instance for the given element value.
+   * Creates the custom {@link Assert} instance for the given value.
    * <p>
-   * Typically, this will just invoke <code>assertThat(actual)</code>
+   * Typically, this will just invoke <code>assertThat(actual)</code>.
+   *
    * @param actual the input value for the {@code Assert} instance
-   * @return returns the custom {@code Assert} instance for the given element value
+   * @return the custom {@code Assert} instance for the given value
    */
   ASSERT createAssert(T actual);
+
+  /**
+   * Creates the custom {@link Assert} instance for the value provided by the
+   * given {@code valueProvider}.
+   * <p>
+   * Overriding implementations might provide a {@link Type} instance to express the desired type
+   * of the value returned by the provider. When doing so, the factory is required to be consistent
+   * with type parameter {@link T}.
+   * <p>
+   * The default implementation always requests a value compatible with {@code Object}.
+   *
+   * @param valueProvider the value provider for the {@code Assert} instance
+   * @return the custom {@code Assert} instance for the value provided by the given value provider
+   * @since 3.26.0
+   */
+  default ASSERT createAssert(ValueProvider<? extends T> valueProvider) {
+    T actual = valueProvider.apply(Object.class);
+    return createAssert(actual);
+  }
+
+  /**
+   * A provider that provides a value compatible with a given {@link Type}.
+   * 
+   * @param <T> the type of provided value.
+   * @since 3.26.0
+   */
+  @FunctionalInterface
+  interface ValueProvider<T> extends Function<Type, T> {
+
+    /**
+     * Provides a value compatible with the given {@code type}.
+     * <p>
+     * The compatibility definition depends on the actual {@code type} instance:
+     * <ul>
+     *   <li>If {@code type} is a {@link Class}, the provided value must be an instance of it.</li>
+     *   <li>If {@code type} is a {@link ParameterizedType}, the provided value must be an instance of its {@link ParameterizedType#getRawType() raw type}.</li>
+     * </ul>
+     * All other {@link Type} subtypes are unsupported.
+     *
+     * @param type the given type, either a {@link Class} or a {@link ParameterizedType} instance
+     * @return the provided value
+     */
+    T provide(Type type);
+
+    /**
+     * Provides a value compatible with the given {@code type}.
+     *
+     * @param type the given type
+     * @return the provided value
+     * @throws IllegalArgumentException if {@code type} is neither a {@link Class} nor a {@link ParameterizedType} instance
+     */
+    default T apply(Type type) {
+      if (!(type instanceof Class || type instanceof ParameterizedType)) {
+        throw new IllegalArgumentException("Unsupported type: " + type.getClass());
+      }
+      return provide(type);
+    }
+
+  }
 
 }
--- a/assertj-core/src/main/java/org/assertj/core/api/AssertFactory.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AssertFactory.java
@@ -12,13 +12,9 @@
  */
 package org.assertj.core.api;
 
-import java.lang.reflect.Type;
-import java.util.Optional;
-
 /**
- * A functional interface to create an {@link Assert} for a given value.
- * <p>
- * The factory typically wraps a call to <code>assertThat(actual)</code>
+ * A single method factory interface to create an {@link Assert} for a given value.
+ * This factory method typically wraps a call to <code>assertThat(actual)</code>
  * to produce a concrete assert type {@code ASSERT} for the input element of type {@code T}.
  * <p>
  * This interface is typically used by navigation assertions on iterable types like {@link AbstractIterableAssert}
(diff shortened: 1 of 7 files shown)


scordio:
@snicoll aside from enhancing test coverage, I don't expect major changes to the productive code.

Could you check if this is what you expect?

About:

> - Evaluate the addition of a new interface for getType() / getRawClass() contract
> - Evaluate the impact of using the new interface on the existing API (e.g., asInstanceOf, extracting, etc.)

I have the feeling such a change would be binary incompatible — not the interface per se but applying it consistently in the rest of the library. I'll have a better look in the upcoming days.
If it's proven to be binary incompatible, I'd rather address this topic in AssertJ 4.

About: 
> Evaluate the addition of a type similar to [ParameterizedTypeReference](https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/ParameterizedTypeReference.java) and the corresponding method factories into InstanceOfAssertFactories

I haven't introduced it yet but I still have the feeling it would be valuable. I'll probably do it once the rest is covered.

snicoll:
> Could you check if this is what you expect?

I want to acknowledge I've seen this but I wasn't able to spend time on this yet. It's high on my list. Thanks a lot for your efforts!

scordio:
> I want to acknowledge I've seen this but I wasn't able to spend time on this yet. It's high on my list. Thanks a lot for your efforts!

No problem at all and good luck with your speech tomorrow! I'll be around if you'd like to chat about this topic 😉

snicoll:
Looking at the Javadoc of `InstanceOfAssertFactory` and the reason that made this PR to exist, I am starting to wonder if things couldn't be restructured a little bit differently.

What we need is an `AssertFactory` that expresses the `Type` it's expecting for its `createAssert` method. It already does so using the generic type, but the actual generic information may be lost or not available. From that perspective, I wonder if a `default Type getType()` on `AssertFactory` could be a way out. If the type is not exposed, then it's fine and we can reject such instances upfront if we need to the information to validate/convert the `actual` value.

Doing this means that `InstanceOfAssertFactory` largely becomes an implementation detail still and additional `AssertFactory` can join the club as long as they expose their type.

The problem currently is that `InstanceOfAssertFactory` has to say something along the lines of "it's going to cast the `actual` to the target type" while this new methods makes it so that we can use such a type without doing a casting at all.

I also wonder if the `rawClass` has to be exposed. It could be left package private from convenience as you can rebuild the raw type from `Type`.

scordio:
> What we need is an `AssertFactory` that expresses the `Type` it's expecting for its `createAssert` method.

Locally, I introduced an interface that extends `AssertFactory` and defines the `getType` method. However, I've been holding that back because of the overall impact on the existing methods that today expect `InstanceOfAssertFactory`, like `asInstanceOf` and `extracting` variants. Such a change is probably binary-incompatible and I would rather deliver it in version 4.

I'll polish my changes and raise them as an additional PR on top of this one so that we can look at it.

However, reading the rest of your message, I have the impression that you don't want to rely on the existing available methods but only take advantage of the instances in `InstanceOfAssertFactories`, right?

> If the type is not exposed, then it's fine and we can reject such instances upfront if we need to the information to validate/convert the `actual` value

What do you mean by "if the type is not exposed"? Or, more generally, what kind of implementation do you imagine for `default Type getType()`?

In https://github.com/assertj/assertj/issues/3368#issuecomment-1951891247, you mentioned an approach with default method but that was about having a `conversionFactory` available.

> I also wonder if the `rawClass` has to be exposed. It could be left package private from convenience as you can rebuild the raw type from `Type`.

Indeed, that's not mandatory and can be kept package-private.

scordio:
> Locally, I introduced an interface that extends `AssertFactory` and defines the `getType` method. However, I've been holding that back because of the overall impact on the existing methods that today expect `InstanceOfAssertFactory`, like `asInstanceOf` and `extracting` variants. Such a change is probably binary-incompatible and I would rather deliver it in version 4.

As expected, the binary compatibility of #3389 is a disaster 🙃 

We might introduce the interface but applying it consistently would be postponed until version 4.

snicoll:
> As expected, the binary compatibility of https://github.com/assertj/assertj/pull/3389 is a disaster 🙃

I don't think we need a separate interface. This is what I was suggesting:

```java
@FunctionalInterface
public interface AssertFactory<T, ASSERT extends Assert<?, ?>> {

  /**
   * Creates the custom {@link Assert} instance for the given element value.
   * <p>
   * Typically, this will just invoke <code>assertThat(actual)</code>
   * @param actual the input value for the {@code Assert} instance
   * @return returns the custom {@code Assert} instance for the given element value
   */
  ASSERT createAssert(T actual);

  /**
   * Return the target {@link Type} this factory expects, or {@code null} if
   * the information is not available.
   * @return the element type.
   */
  default Type getType() {
    return null;
  }

}
```

With that, you need to rename `getType` in `InstanceOfAssertFactory` to `getRawType` or something but it being package private does not matter.

As far as I can see this is fully backward compatible. I would not introduce a complete new interface to expose the type as it's required for a very specific case (the one discussed here).

If you don't like the getter there in such a high level interface, which I understand, here's an alternative that's also 100% backward compatible:

```java
@FunctionalInterface
public interface AssertFactory<T, ASSERT extends Assert<?, ?>> {

  /**
   * Creates the custom {@link Assert} instance for the given element value.
   * <p>
   * Typically, this will just invoke <code>assertThat(actual)</code>
   * @param actual the input value for the {@code Assert} instance
   * @return returns the custom {@code Assert} instance for the given element value
   */
  ASSERT createAssert(T actual);

  /**
   * Creates the custom {@link Assert} instance for the value provided by the
   * given {@code valueProvider}.
   * @param valueProvider the value provider for the {@code Assert} instance
   * @return returns the custom {@code Assert} instance for the value produced by
   * the given value provider
   */
  default ASSERT createAssert(Function<Type, T> valueProvider) {
    throw new UnsupportedOperationException();
  }

}
```

assertj-core/src/main/java/org/assertj/core/api/AssertFactory.java
```diff
@@ -29,12 +33,28 @@
...
    * @return returns the custom {@code Assert} instance for the given element value
    */
   ASSERT createAssert(T actual);
 
+  /**
+   * Return the {@link Type} of the input this factory expects, or an empty {@code Optional}
+   * if the information is not available.
+   * 
+   * @implSpec Overriding implementations providing a non-empty optional are required to
+   * return a {@code Type} instance consistent with the value of the factory type parameter {@link T}.
```
scordio, line 49: The need to explain this highlights the fact that the solution is suboptimal. I'll experiment with the `createAssert(Function)` alternative.

scordio:
> Evaluate the addition of a type similar to [`ParameterizedTypeReference`](https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/ParameterizedTypeReference.java) and the corresponding method factories into `InstanceOfAssertFactories`

We discussed that we'll wait for a concrete demand.

scordio:
@snicoll could you try the latest changes? It should be in line with what we discussed, although I still need to spend some time on it for proper test coverage.

snicoll:
Looks good. I like the interface that keeps a link between `Type` and the expected generic!

scordio:
BTW, we already have `spring-core` in the test classpath so I can try to put together an integration test close to the usage pattern you expect.

snicoll:
Cool, I hijacked this in `InstanceOfAssertFactoriesTest`:

```java
private static final DefaultConversionService conversionService = new DefaultConversionService();

@Test
void springConversionTest() {
	List<String> values = List.of("123", "456", "789");
	// WHEN
	ListAssert<Integer> result = list(Integer.class).createAssert(type -> convert(values, type));
	// THEN
	result.containsExactly(123, 456, 789);

}

private <T> T convert(Object instance, Type type) {
	return (T) conversionService.convert(instance,
			new TypeDescriptor(ResolvableType.forType(type), null, null));
}
```

This is one example of many. The json path that was the inspiration for this ask is relying on Jackson to do the conversion based on a `Type` but it's a very similar arrangement. Thanks again!

assertj-core/src/main/java/org/assertj/core/api/AssertFactory.java
```diff
@@ -12,6 +12,10 @@
  */
 package org.assertj.core.api;
 
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.function.Function;
+
 /**
  * A single method factory interface to create an {@link Assert} for a given value.
```
scordio, line 20: This should be improved as it's no longer a single-method factory interface. In addition, adding some usage examples for the two `createAssert` methods would be beneficial.

scordio:
I'm still unsure how much the Javadoc should elaborate on the concrete usage of the new `createAssert`.

@snicoll I will merge the changes and release 3.26.0 soon. I would be very interested in seeing how this will be used so that I can derive a meaningful example for our Javadoc or docs 🙂

PR #3378, opened 2024-02-26
Add `isNotThrownBy()` method as opposite to `isThrownBy()`

quaff:
It can be used like:
```java
    assertThatExceptionOfType(IllegalArgumentException.class).isNotThrownBy(() -> {
      // no exception
    });

    assertThatExceptionOfType(IllegalArgumentException.class).isNotThrownBy(() -> {
      throw new IllegalStateException();
    });
```

#### Check List:
* Unit tests : YES

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/ThrowableTypeAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/ThrowableTypeAssert.java
@@ -25,7 +25,7 @@ import org.assertj.core.util.VisibleForTesting;
  * <p>
  * The class itself does not do much, it delegates the work to {@link ThrowableAssertAlternative} after calling {@link #isThrownBy(ThrowingCallable)}.
  *
- * @param <T> type of throwable to be thrown.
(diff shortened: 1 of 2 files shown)


quaff:
Do I need create a separate issue before submit PR? @scordio

scordio:
No, that's ok @quaff, and thanks a lot for the proposal!

Your intention is clear. I would like to double-check the user experience in case of errors, i.e., when the code doesn't throw anything vs. when the code throws the exception that shouldn't be thrown.

Would you have time to try these two cases and post the results here? It would help us get on board faster 🙂

quaff:
> Your intention is clear. I would like to double-check the user experience in case of errors, i.e., when the code doesn't throw anything vs. when the code throws the exception that shouldn't be thrown.
> 
> Would you have time to try these two cases and post the results here? It would help us get on board faster 🙂

If I understand correctly, the test case should cover that, the assertion will success if the code doesn't throw anything or throws the exception that shouldn't be thrown.

joel-costigliola:
Integrated thanks @quaff !

PR #3389, opened 2024-03-07
Introduce `TypeBasedAssertFactory`

scordio:


scordio:
Closing as it's not the direction we want to take (see https://github.com/assertj/assertj/pull/3377#issuecomment-1983146581).

PR #3405, opened 2024-03-21
Throwable assertion support for `hasMessageNormalizingNewlines(String)`

armandino:
I propose adding `hasMessageNormalizingNewlines(String)` method for asserting `Throwable` messages.

Currently, this can be done using workarounds, e.g.

```java
Throwable ex = catchThrowable(() -> someMethod());
      
assertThat(ex.getMessage()).isEqualToNormalizingNewlines("multiline\nmessage");
```

#### Check List:
* Unit tests : YES

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
@@ -326,6 +326,32 @@ public abstract class AbstractThrowableAssert<SELF extends AbstractThrowableAsse
   }
 
   /**
+   * Verifies that the message of the actual (@code Throwable) is equal to the given one
+   * after normalizing new line characters (i.e. '\r\n' == '\n').
+   * <p>
+   * Example:
+   * <pre><code class='java'> Throwable invalidArgException = new IllegalArgumentException("Error:\r\ninvalid input");
+   * Throwable throwable = new Throwable(invalidArgException);
+   *
+   * // This assertion succeeds:
+   * assertThat(throwable).hasMessageNormalizingNewlines("Error:\r\ninvalid input");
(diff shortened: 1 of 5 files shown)


joel-costigliola:
@armandino sorry for the late reply, you already have a way to do that (nicely I think) with [message()](https://www.javadoc.io/static/org.assertj/assertj-core/3.25.3/org/assertj/core/api/AbstractThrowableAssert.html#message()): 
```java
assertThat(ex).message().isEqualToNormalizingNewlines("multiline\nmessage");
```

armandino:
@joel-costigliola thank you for the suggestion. That definitely works. The only downside is that it requires getting an instance of the thrown exception, as in my snippet above.

I usually use `assertThatThrownBy`:

```java
assertThatThrownBy(() -> someMethod())
    .isInstanceOf(SomeException.class)
    .hasMessage("...");
```

but this doesn't does not support  the following usage:

```java
assertThatThrownBy(() -> someMethod())
    .isInstanceOf(SomeException.class)
    .hasMessageNormalizingNewlines("...");  // would be nice to have
```

With the introduction  of text blocks (https://openjdk.org/jeps/378), `hasMessageNormalizingNewlines()` would be a useful addition, unless there's another AssertJ APi that supports doing something like this that I'm not aware of.

joel-costigliola:
@armandino the following snippet works because `assertThatThrownBy` returns `AbstractThrowableAssert` which expose `message()`, that should work for you too, right ?
```java
assertThatThrownBy(() -> {})
   .isInstanceOf(Exception.class)
   .message()
   .isEqualToNormalizingNewlines("...");
```

armandino:
@joel-costigliola yes, that does work. I wasn't aware the `message()` method is exposed there. Thanks again! I'll close the PR.

PR #3441, opened 2024-04-24
Add doesNot[Start/End]WithWhitespace methods to AbstractCharSequenceAssert

### LINKED ISSUE #3433: Add AbstractCharSequenceAssert#doesNotHaveAny(Leading|Trailing)Whitespaces
#### Feature summary

Sometimes, I need to verify that a `String` shouldn't have any leading/trailing whitespaces which means the [trim()](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#trim--) or [strip()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#strip()) should be applied to the actual value.

#### Example

```java
String s = parseValue();
assertThat(s)
        .isNotNull()
        .satisfies(v -> {
            assertThat(v.strip()).isEqualTo(v); // any alternatives?
        })
       .doesNotHaveAnyLeadingWhitespaces()
       .doesNotHaveAnyTrailingWhitespaces()
;
```

yyytir777:
#### Check List:
* Fixes #3433 
* Unit tests : YES

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractCharSequenceAssert.java
@@ -2182,11 +2182,7 @@ public abstract class AbstractCharSequenceAssert<SELF extends AbstractCharSequen
   }
 
   private void assertDoesNotStartWithWhitespace(CharSequence actual) {
-    isNotNull();
-    isNotEmpty();
-    char first = actual.charAt(0);
-
-    if (Character.isWhitespace(first)) throwAssertionError(shouldNotStartWithWhitespace(actual));
+    if (Character.isWhitespace(actual.codePoints().findFirst().getAsInt())) throwAssertionError(shouldNotStartWithWhitespace(actual));
   }
 
   /**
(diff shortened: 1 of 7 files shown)


onacit:
Can't we use [CharSequence#codepoints()](https://docs.oracle.com/javase/8/docs/api/java/lang/CharSequence.html#codePoints--)?

```
Character.isWhiteapce(actual.codePoints().findFirst().getAsInt());
Character.isWhiteapce(actual.codePoints().reduce((v1, v2) -> v2).getAsInt());
```

yyytir777:
Thanks for better method. I'll fix it right now.

joel-costigliola:
@onacit why would using code points be better than char to detect whitespaces ?
do you have an example where using code points would detect whitespaces that char don't ?

onacit:
@joel-costigliola As documented in [Character.isWhitespace(char)](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isWhitespace-char-), it'll, I think, be better when we use codePoints.

> Note: This method cannot handle [supplementary characters](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#supplementary). To support all Unicode characters, including supplementary characters, use the [isWhitespace(int)](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isWhitespace-int-) method.

Thanks.

scordio:
> Can't we use [CharSequence#codePoints()](https://docs.oracle.com/javase/8/docs/api/java/lang/CharSequence.html#codePoints--)?
> 
> ```
> Character.isWhiteapce(actual.codePoints().findFirst().getAsInt());
> Character.isWhiteapce(actual.codePoints().reduce((v1, v2) -> v2).getAsInt());
> ```

Best would be to compose a test that demonstrates the need for `codePoints`, for example using Unicode whitespace that wouldn't be detected by `Character.isWhitespace(char)` but would otherwise work with `Character.isWhitespace(int)`. The test would also prevent regressions if we changed the implementation to `Character.isWhitespace(char)`.

Taking inspiration from https://stackoverflow.com/a/18169122/9714611, maybe [a value from this list](https://www.fileformat.info/info/unicode/category/Zs/list.htm) could help?

joel-costigliola:
I'm on it @scordio ! @yyytir777  you'll get full credit for the PR don't worry :)

joel-costigliola:
Integrated thanks @yyytir777 !

Issue #3442, opened 2024-04-24
Move to Java 17


scordio:
We want AssertJ 4 to use and require Java 17.

This mostly means increasing the [`java.version`](https://github.com/assertj/assertj/blob/bd895e7cdc38f7168381c77a4a2015650664c58f/assertj-parent/pom.xml#L19) Maven property to 17.

In addition, we can:
* Move the tests from [`assertj-core-java-17`](https://github.com/assertj/assertj/tree/3.x/assertj-tests/assertj-integration-tests/assertj-core-java-17) to `assertj-core` and delete `assertj-core-java-17`
* Replace any usage of `List.of`, `Set.of`, and `Map.of` from [`Jdk11`](https://github.com/assertj/assertj/blob/bd895e7cdc38f7168381c77a4a2015650664c58f/assertj-core/src/test/java/org/assertj/core/test/jdk11/Jdk11.java) with the corresponding JDK API and delete the `org.assertj.core.test.jdk11` package

scordio:
> Move the tests from [`assertj-core-java-17`](https://github.com/assertj/assertj/tree/3.x/assertj-tests/assertj-integration-tests/assertj-core-java-17) to `assertj-core` and delete `assertj-core-java-17`

We should rather go in the opposite direction due to using the module path only.

We'll take care of this in #3446.

scordio:
We are discussing with the team whether to stay on Java 8 or upgrade to Java 11 instead of upgrading to Java 17.

The main reason is to keep our effort reasonable and continue providing support to users who cannot upgrade yet (e.g., all RHEL 7 users under [ELS](https://www.redhat.com/en/blog/announcing-4-years-extended-life-cycle-support-els-red-hat-enterprise-linux-7)).

If we target Java 11, AssertJ 3 will enter maintenance mode and not receive new features.

Is there any strong reason why we should keep upgrading to Java 17?

anbusampath:
> Is there any strong reason why we should keep upgrading to Java 17?

Because most of the Java ecosystem trying to stay Java 17 as new baseline and support later LTSs.

Java 17 is becoming new Java 8.

Issue #3382, opened 2024-03-01
isEqualTo on date can lead to unexpected failures when switching time zones


wimdeblauwe:
**Describe the bug**
The behaviour of `isEqualTo(String)` when asserting a java.util.Date can fail if you switch the default time zone during a test run.

* assertj core version: 3.24.2
* java version: 21
* test framework version: JUnit 5
* os (if relevant): 

**Test case reproducing the bug**

Add a test case showing the bug that we can run

```java
public class Test {
 @Test
    void testWithIsEqualTo() {
//      TimeZone.setDefault(TimeZone.getTimeZone("CET"));
//      assertThat(Date.from(Instant.parse("2024-03-01T00:00:00.000+01:00")))
//          .describedAs("Using CET time zone")
//          .isEqualTo("2024-03-01");
      TimeZone.setDefault(TimeZone.getTimeZone("WET"));
      assertThat(Date.from(Instant.parse("2024-03-01T00:00:00.000+00:00")))
          .describedAs("Using WET time zone")
          .isEqualTo("2024-03-01");
    }

    @Test
    void test() {
//      TimeZone.setDefault(TimeZone.getTimeZone("CET"));
//      assertThat(Date.from(Instant.parse("2024-03-01T00:00:00.000+01:00")))
//          .describedAs("Using CET time zone")
//          .hasYear(2024)
//          .hasMonth(3)
//          .hasDayOfMonth(1);
      TimeZone.setDefault(TimeZone.getTimeZone("WET"));
      assertThat(Date.from(Instant.parse("2024-03-01T00:00:00.000+00:00")))
          .describedAs("Using WET time zone")
          .hasYear(2024)
          .hasMonth(3)
          .hasDayOfMonth(1);
    }
}
```

Run this test with the code commented out. The test will be ok.

Now enable the code in comments and run the test again. The test `testWithIsEqualTo` now fails, but on the part that was already active before.

(I found the issue with using [@DefaultTimeZone](https://junit-pioneer.org/docs/default-locale-timezone/), but I can reproduce it without it, so I believe it might be an AssertJ issue).

joel-costigliola:
Thanks for reporting this @wimdeblauwe, we will have a look at it shotly
.

joel-costigliola:
AssertJ uses various date format to get parse the date from the given string, in the test case, it is the `yyyy-MM-dd` date pattern (as expected).

[...]

PR #3467, opened 2024-05-09
Fix `Instant` and `String` conversion with `Date` assertions

### LINKED ISSUE #3359: Unable to compare some timestamp types

scordio:
This reverts the changes introduced with #3410.

Fixes #3359.


assertj-core/src/main/java/org/assertj/core/api/AbstractDateAssert.java
```diff
@@ -27,6 +27,7 @@
 import static org.assertj.core.util.Lists.list;
 import static org.assertj.core.util.Lists.newArrayList;
 
+import java.sql.Timestamp;
```
scordio, line 30: This probably has an underlying impact on the module descriptor as `java.sql` would become required whenever an assertion with an `Instant` parameter is called.

I'll see how we can prove it (related to #2759).

assertj-core/src/test/java/org/assertj/core/api/date/DateAssert_isEqualTo_Test.java
```diff
@@ -12,42 +12,125 @@
...
+      // THEN
+      then(error).isInstanceOf(AssertionFailedError.class);
+    }
+
+    Arguments[] should_fail() {
+      return new Arguments[] {
+          arguments(Date.from(parse("1970-01-01T00:00:00.000000001Z")),
+                    "1970-01-01T00:00:00.000000001Z"),
+          arguments(Timestamp.from(parse("1970-01-01T00:00:00.000000001Z")),
+                    "1970-01-01T00:00:00.000000001Z")
```
scordio, line 130: The fact that these cases currently belong to failures might not be 100% correct.

However, I don't see an easy way forward right now so I'd rather document the current behavior with them and we can address this topic in a second step.

joel-costigliola: I think we are hitting java `Date` limitation as `SimpleDateFormat` does not handle nanos.
https://stackoverflow.com/questions/30135025/java-date-parsing-with-microsecond-or-nanosecond-accuracy

The best course of action is to document that and point to not to use date assertion accepting string, which was a handy shortcut for users at the time to easily create dates but now with the java.time API it is not necessary and I would even say something we could deprecate.

scordio: I wonder if cases like this one and #3382 should be addressed differently, for example dropping the internal usage of `SimpleDateFormat` and favoring `DateTimeFormatter` instead.

For now, I propose to merge these changes as they are, for sure they don't make anything worse, and we can follow up on the `SimpleDateFormat` topic separately.

WDYT?

joel-costigliola: Let's merge this and use `DateTimeFormatter` as it handles nanoseconds

assertj-core/src/main/java/org/assertj/core/api/AbstractDateAssert.java
scordio: I'm not ready yet with #2759 and #3442, which would help demonstrate whether we have a dependency. I'd go ahead with the merge now and fix it later if something pops up.

assertj-core/src/test/java/org/assertj/core/api/date/DateAssert_isEqualTo_Test.java
scordio: #3482 raised for tracking.

PR #3474, opened 2024-05-14
[javadocs] Fix javadocs

hazendaz:
option needed to have additional arguments, removed old source 8 on javadocs as no longer needed here.

- Replacement of #3231


scordio:
`main` is not yet open for contributions, please target `3.x` for now (where             `<additionalOption>` is already used).

What I meant in the other ticket is that I don't get the problem we're trying to solve with the introduction of:

```xml
<legacyMode>true</legacyMode>
```

As far as I can judge from the logs and the output at https://www.javadoc.io/doc/org.assertj/assertj-core/latest/index.html, we don't have any visible issue.

In addition, compared to the JUnit Javadoc, we don't have the `MODULE` entry in the top header. I honestly don't know why. Could it be because we currently compile `module-info.java` in a separate `maven-compiler-plugin` execution?

https://github.com/assertj/assertj/blob/4f072c99641691f67f4217eff44d4d1866ddc274/assertj-core/pom.xml#L321-L333

Could it be related to the `multiReleaseOutput` parameter which will put the `module-info.class` under `META-INF/versions/9`?

Honestly, I didn't dig into the `maven-javadoc-plugin` to cross-check the sources used.

hazendaz:
Ah, I see.  Yes because you have module-info file it doesn't require that now.  The bug maven had was with automatic module naming.  I can close this then and reopen to the other branch with just the expected content here without that part.

hazendaz:
3.x line already has the change so I guess nothing else to do here

Issue #3458, opened 2024-05-05
`isUnmodifiable` assertion for `Iterator` instances


scordio:
#### Feature summary
Similarly to #2102, we could have an `isUnmodifiable()` assertion for `Iterator` types.

#### Example

```java
Iterator<Integer> iterator = Collections.unmodifiableList(Arrays.asList(1, 2, 3)).iterator();

assertThat(iterator).isUnmodifiable();
```

dehasi:
@scordio if nobody works on this task, I'd like to pick it up. And it seems a bit more challenging than for collections :)

scordio:
Sure and thanks, @dehasi!

What challenges have you encountered?

dehasi:
To be short, I can't rely on that `iterator.remove()` will throw `UnsupportedOperationException`.

I.e. `Collections.EmptyIterator.remove()` throws `IllegalStateException`. Do we treat `EmptyIterator` as unmodifiable?

But can't rely on `IllegalStateException` as unmodifiability marker. `ArrayList<>().itertator()`  can also throw  `IllegalStateException` if we haven't called `next`. 

Initially, I wanted to check unmodifiability by just calling `iterator.remove()`. It's not enough (see to ArrayList example). 
Now, I'm thinking of making an additional check like `iterator.next(); iterator.remove()`. However, it will look a bit weird.

PR #3477, opened 2024-05-15
#3458 isUnmodifiable assertion for Iterator instances

dehasi:
#### Check List:
* Fixes: #3458 
* Unit tests: YES
* Javadoc with a code example (on API only): YES

#### What was done
I can only check if an iterator is unmodifiable by calling `remove()` and checking if the iterator throws `UnsupportedOperationException`. However not all unmodifiable iterators behave like that, see [my comment](https://github.com/assertj/assertj/issues/3458#issuecomment-2111116003) in the ticket. 
That's why I check explicitly for `Collections$EmptyIterator`  and `Collections$EmptyListIterator`.

cc @scordio

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractIteratorAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractIteratorAssert.java
@@ -14,8 +14,11 @@ package org.assertj.core.api;
 
 import java.util.Iterator;
 
+import org.assertj.core.annotations.Beta;
 import org.assertj.core.internal.Iterators;
 import org.assertj.core.util.VisibleForTesting;
+
+import static org.assertj.core.error.ShouldBeUnmodifiable.shouldBeUnmodifiable;
 
 /**
  * <p>Base class for all implementations of assertions for {@link Iterator}s.</p>
@@ -93,4 +96,50 @@ public abstract class AbstractIteratorAssert<SELF extends AbstractIteratorAssert
     return new IterableAssert<>(IterableAssert.toIterable(actual));
   }
 
+  /**
+   * Verifies that the actual iterator is unmodifiable, i.e., throws an {@link UnsupportedOperationException} with
+   * any attempt to remove from the iterator.
+   * <p>
+   * Example:
+   * <pre><code class='java'> // assertions will pass
+   * assertThat(List.of().iterator()).isUnmodifiable();
+   * assertThat(Set.of().iterator()).isUnmodifiable();
+   *
+   * // assertions will fail
+   * assertThat(new ArrayList&lt;&gt;().iterator()).isUnmodifiable();
+   * assertThat(new HashSet&lt;&gt;().iterator()).isUnmodifiable();</code></pre>
+   *
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual iterator is modifiable.
+   * @since 3.27.0
+   */
+  @Beta
+  public SELF isUnmodifiable() {
+    isNotNull();
+    assertIsUnmodifiable();
+    return myself;
+  }
+
+  private void assertIsUnmodifiable() {
+    switch (actual.getClass().getName()) {
+    case "java.util.Collections$EmptyIterator":
+    case "java.util.Collections$EmptyListIterator":
+      // immutable by contract, although not all methods throw UnsupportedOperationException
+      return;
+    }
+
+    expectUnsupportedOperationException(() -> actual.remove(), "Iterator.remove()");
+  }
+
+  // Same as AbstractCollectionAssert#expectUnsupportedOperationException
+  private void expectUnsupportedOperationException(Runnable runnable, String method) {
+    try {
+      runnable.run();
+      throwAssertionError(shouldBeUnmodifiable(method));
+    } catch (UnsupportedOperationException e) {
+      // happy path
+    } catch (RuntimeException e) {
+      throwAssertionError(shouldBeUnmodifiable(method, e));
+    }
+  }
 }
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractIteratorAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractIteratorAssert.java
@@ -12,13 +12,13 @@
  */
 package org.assertj.core.api;
 
+import static org.assertj.core.error.ShouldBeUnmodifiable.shouldBeUnmodifiable;
+
 import java.util.Iterator;
 
 import org.assertj.core.annotations.Beta;
 import org.assertj.core.internal.Iterators;
 import org.assertj.core.util.VisibleForTesting;
-
-import static org.assertj.core.error.ShouldBeUnmodifiable.shouldBeUnmodifiable;
 
 /**
  * <p>Base class for all implementations of assertions for {@link Iterator}s.</p>
@@ -111,7 +111,7 @@ public abstract class AbstractIteratorAssert<SELF extends AbstractIteratorAssert
    *
    * @return {@code this} assertion object.
    * @throws AssertionError if the actual iterator is modifiable.
-   * @since 3.27.0
+   * @since 3.26.0
    */
   @Beta
   public SELF isUnmodifiable() {
@@ -128,7 +128,7 @@ public abstract class AbstractIteratorAssert<SELF extends AbstractIteratorAssert
       return;
     }
 
-    expectUnsupportedOperationException(() -> actual.remove(), "Iterator.remove()");
+    expectUnsupportedOperationException(actual::remove, "Iterator.remove()");
   }
 
   // Same as AbstractCollectionAssert#expectUnsupportedOperationException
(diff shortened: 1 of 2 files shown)


assertj-core/src/main/java/org/assertj/core/api/AbstractIteratorAssert.java
```diff
@@ -93,4 +96,49 @@ public IterableAssert<ELEMENT> toIterable() {
...
+    return myself;
+  }
+
+  private void assertIsUnmodifiable() {
+    switch (actual.getClass().getName()) {
+      case "java.util.Collections$EmptyIterator":
+      case "java.util.Collections$EmptyListIterator":
+        // immutable by contract, although not all methods throw UnsupportedOperationException
+        return;
+    }
```
scordio, line 128: > To be short, I can't rely on that `iterator.remove()` will throw `UnsupportedOperationException`.
> 
> I.e. `Collections.EmptyIterator.remove()` throws `IllegalStateException`. Do we treat `EmptyIterator` as unmodifiable?

As you've already figured out, I don't think we have a better choice for these JDK types, similar to `emptyList`, `emptySet`, etc.

assertj-core/src/main/java/org/assertj/core/api/AbstractIteratorAssert.java
```diff
@@ -93,4 +96,49 @@ public IterableAssert<ELEMENT> toIterable() {
...
+
+    expectUnsupportedOperationException(() -> actual.remove(), "Iterator.remove()");
```
scordio, line 130: > But can't rely on `IllegalStateException` as unmodifiability marker. `ArrayList<>().itertator()` can also throw `IllegalStateException` if we haven't called `next`.
> 
> Initially, I wanted to check unmodifiability by just calling `iterator.remove()`. It's not enough (see to ArrayList example). Now, I'm thinking of making an additional check like `iterator.next(); iterator.remove()`. However, it will look a bit weird.

According to the [`remove()`](https://docs.oracle.com/javase/8/docs/api/java/util/Iterator.html#remove--) contract, a preliminary call to `next()` is required (probably to be guarded by `hasNext()`). We might go this way and see how the modifiable implementations behave. WDYT?

assertj-core/src/test/java/org/assertj/core/api/iterator/IteratorAssert_isUnmodifiable_Test.java
```diff
@@ -0,0 +1,125 @@
...
+    Iterator<?> actual = null;
+    // WHEN
+    AssertionError assertionError = expectAssertionError(() -> assertThat(actual).isUnmodifiable());
+    // THEN
+    then(assertionError).hasMessage(shouldNotBeNull().create());
+  }
+
+  @ParameterizedTest
+  @MethodSource({"modifiableIterators", "startedModifiableIterators"})
+  void should_fail_if_actual_can_be_modified(/* GIVEN */ Iterator<?> actual, ErrorMessageFactory errorMessageFactory) {
```
scordio, line 58:
```suggestion
  void should_fail_if_actual_can_be_modified(Iterator<?> actual, ErrorMessageFactory errorMessageFactory) {
```

assertj-core/src/test/java/org/assertj/core/api/iterator/IteratorAssert_isUnmodifiable_Test.java
```diff
@@ -0,0 +1,125 @@
...
+    startedIterator.next();
+    Iterator<?> endedIterator = new ArrayList<>(List.of(1)).iterator();
+    endedIterator.next();
+    return Stream.of(arguments(startedIterator, shouldBeUnmodifiable("Iterator.remove()")),
+      arguments(endedIterator, shouldBeUnmodifiable("Iterator.remove()")));
+  }
+
+  @ParameterizedTest
+  @MethodSource("unmodifiableIterators")
+  void should_pass(/* GIVEN */ Iterator<?> actual) {
```
scordio, line 89:
```suggestion
  void should_pass(Iterator<?> actual) {
```

assertj-core/src/test/java/org/assertj/core/api/iterator/IteratorAssert_isUnmodifiable_Test.java
```diff
@@ -0,0 +1,125 @@
...
+      arguments(new LinkedList<>().iterator(), shouldBeUnmodifiable("Iterator.remove()",  new IllegalStateException())),
+      arguments(new HashSet<>().iterator(), shouldBeUnmodifiable("Iterator.remove()",  new IllegalStateException())),
+      arguments(newArrayList(new Object()).iterator(), shouldBeUnmodifiable("Iterator.remove()",  new IllegalStateException())),
+      arguments(newLinkedHashSet(new Object()).iterator(), shouldBeUnmodifiable("Iterator.remove()",  new IllegalStateException())),
+      arguments(newTreeSet("element").iterator(), shouldBeUnmodifiable("Iterator.remove()", new IllegalStateException())));
+  }
+
+  // If Iterator.remove() is called Iterator.next(), it doesn't throw exception
+  private static Stream<Arguments> startedModifiableIterators() {
+    Iterator<?> startedIterator = new ArrayList<>(List.of(1,2,3)).iterator();
```
scordio, line 79: The codebase is still targeting Java 8 so `List.of` cannot be used. You could use `org.assertj.core.util.Lists.list()` (statically imported):

```suggestion
    Iterator<?> startedIterator = new ArrayList<>(list(1,2,3)).iterator();
```

assertj-core/src/test/java/org/assertj/core/api/iterator/IteratorAssert_isUnmodifiable_Test.java
```diff
@@ -0,0 +1,125 @@
...
+    startedIterator.next();
+    Iterator<?> endedIterator = new ArrayList<>(List.of(1)).iterator();
```
scordio, line 81:
```suggestion
    Iterator<?> endedIterator = new ArrayList<>(list(1)).iterator();
```

scordio (commented):
Thanks @dehasi, here are some comments. The build is failing due to the usage of Java 9+ API (`List.of`).

assertj-core/src/main/java/org/assertj/core/api/AbstractIteratorAssert.java
```diff
@@ -93,4 +96,49 @@ public IterableAssert<ELEMENT> toIterable() {
...
+   * <pre><code class='java'> // assertions will pass
+   * assertThat(List.of().iterator()).isUnmodifiable();
+   * assertThat(Set.of().iterator()).isUnmodifiable();
+   *
+   * // assertions will fail
+   * assertThat(new ArrayList&lt;&gt;().iterator()).isUnmodifiable();
+   * assertThat(new HashSet&lt;&gt;().iterator()).isUnmodifiable();</code></pre>
+   *
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual iterator is modifiable.
```
scordio, line 113:
```suggestion
   * @throws AssertionError if the actual iterator is modifiable.
   * @since 3.27.0.
```

assertj-core/src/test/java/org/assertj/core/api/iterator/IteratorAssert_isUnmodifiable_Test.java
dehasi: Just wanted to show (off) a place where we can put "GIVEN" and every test requires GIVEN/WHEN/THEN comments. Good practice btw ;)

assertj-core/src/main/java/org/assertj/core/api/AbstractIteratorAssert.java
dehasi: Maybe I miss smth, but I can't find other iterator classes, only these two. Other collections create an interator "ad hoc".

assertj-core/src/main/java/org/assertj/core/api/AbstractIteratorAssert.java
dehasi: With the same thoughts, I wrote my comment in the ticket :) 
I thought of making two checks and expecting either `remove()` or `next(); remove()` will throw `UnsupportedOperationException`. I couldn't implement it nicely, which made me think that I'm overenginerring. 
Then I thought that if an iterator is unmodifiable, whyever on the earth somebody will implement any other logic in `remove()` except just throwing an exception? o_O That's why I left only this check.

However if you think it makes sense to call `next()` before `remove()` I can add it, ofc with guarded
`hasNext()`.  So I have three options for you:
1. Accept my argument to common sense and leave it as it is;
2. Call `next(); remove()` in the current check;
3. Add two checks, one with just `remove()` and another with `next(); remove()`;
4. Ofc I am open to any other suggestions ;) 

so, WDYT?

scordio: Let's be pragmatic and go with option 1!

dehasi:
@scordio I've squashed my changes. I think I can only struggle if the target branch is changed, and I might need assistance. Just afraid to mess with Git 🫤

scordio:
This is now merged. Thanks for your first contribution, @dehasi!

PR #3483, opened 2024-05-20
Replace simple date format by date time formatter

### LINKED ISSUE #3482: Evaluate `DateTimeFormatter` as replacement of `SimpleDateFormat` for internal usage
We use `SimpleDateFormat` in many date-related assertions and this brings some drawbacks, for example:
* https://github.com/assertj/assertj/issues/3382#issuecomment-2119269407
* https://github.com/assertj/assertj/pull/3467#discussion_r1596511220

We might instead consider using "DateTimeFormatter", which supports nanosecond granularity.

joel-costigliola:
Fix #3482

The only bit that was not addressed is the lenient parsing, this is why the tests are failing.

At least we are now supporting nanos, ex:
```java
Date date = Date.from(Instant.parse("1970-01-01T00:00:00.000000001Z"));
assertThat(date).isEqualTo("1970-01-01T00:00:00.000000001Z");
```

Not supporting lenient date parsing is breaking change, probably one we can remove in AssertJ 4 since it's not a great idea to accept date like `2001-01-34` anyway.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractDateAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractDateAssert.java
@@ -12,26 +12,32 @@
  */
 package org.assertj.core.api;
 
+import static java.lang.String.format;
+import static java.time.ZoneId.systemDefault;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.TimeUnit.HOURS;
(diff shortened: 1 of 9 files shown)


joel-costigliola:
integrated in ca5f9bc6cd68738a06bf19fbb7f70f20b9d8c067

Issue #3537, opened 2024-07-14
AllOf should report which condition failed 


joel-costigliola:
#### Feature summary

See https://stackoverflow.com/questions/78745304/assertj-how-to-know-which-conditions-failed-when-using-allof/78747459#78747459

#### Example

```java
Condition<String> condition1 = new Condition<String>(text -> text != null, "Input not null");
Condition<String> condition2 = new Condition<String>(text -> text.equalsIgnoreCase("abc"), "Input not matching");
assertThat(someString).satisfies(allOf(condition1, condition2));
```

JunHyungJang:
It looks interesting ! Can i get this issue?

joel-costigliola:
Sure thanks @JunHyungJang

injae-kim:
![image](https://github.com/user-attachments/assets/d3fcbb7b-bca0-498f-9a59-bdfadb9ba0c4)

Now we can check which condition failed inside allOf 👍

PR #3554, opened 2024-08-01
Fix : satisfies(AllOf.allOf(condition.....)) to indicate the false condition

JunHyungJang:
- https://stackoverflow.com/questions/78745304/assertj-how-to-know-which-conditions-failed-when-using-allof/78747459#78747459
AllOf should report which condition failed  #3537
I fix the above issue by fixing satisfies.allOf(conditions...) to indicate that which condition is failed and passed

Please let me know, if I should do something more or something to fix
Thank you

Check List:
Fixes AllOf should report which condition failed  #3537
Unit tests : YES
Javadoc with a code example (on API only) : YES
PR meets the [contributing guidelines](https://github.com/assertj/assertj/blob/main/CONTRIBUTING.md)


assertj-core/src/main/java/org/assertj/core/error/AbstractAssert_satisfiesAllOf_Test.java
```diff
@@ -0,0 +1,33 @@
+package org.assertj.core.error;
+
+import org.assertj.core.api.Condition;
+import org.assertj.core.condition.AllOf;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.assertj.core.api.BDDAssertions.then;
+
+public class AbstractAssert_satisfiesAllOf_Test {
```
joel-costigliola, line 10: Could you refactor that test so that it follows our contribution guidelines https://github.com/assertj/assertj/blob/3.x/CONTRIBUTING.md, especially the use of GIVEN WHEN THEN and `AssertionUtil.expectAssertionError`

assertj-core/src/main/java/org/assertj/core/error/AbstractAssert_satisfiesAllOf_Test.java
```diff
@@ -0,0 +1,33 @@
...
+  void should_pass_when_all_of_the_condition_is_met() {
+    Condition<String> conditionA = new Condition<>(text -> text != null, "Input not null");
+    Condition<String> conditionB = new Condition<>(text -> text.equalsIgnoreCase("ABC"), "Input not matching");
+    assertThat("ABC").satisfies(AllOf.allOf(conditionA, conditionB));
+  }
+
+  @Test
+  void should_fail_if_one_of_the_condition_is_not_met() {
+    Condition<String> condition1 = new Condition<>(text -> text != null, "Input not null");
+    Condition<String> condition2 = new Condition<>(text -> text.equalsIgnoreCase("abc"), "Input not matching");
```
joel-costigliola, line 22: add another failing condition to show that all failing conditions are reported (and not the first failing)

joel-costigliola (changes requested):
thanks for the PR, first round of comments :)

JunHyungJang:
Thank you for your first round of comments !
1. I add the testcase which fails every conditions
2. I fix the testcase following the contributing guidlines

joel-costigliola:
Integrated thanks @JunHyungJang !

injae-kim:
Nice work @JunHyungJang !

PR #3606, opened 2024-09-13
Guava TableAssert: assertions for row count added

mk868:
Hello,

This is another PR that enriches the assertions for Guava `Table`.
In this PR, I add the following methods to `TableAssert`:

- `hasRowCountGreaterThan(int boundary)`
- `hasRowCountGreaterThanOrEqualTo(int boundary)`
- `hasRowCountLessThan(int boundary)`
- `hasRowCountLessThanOrEqualTo(int boundary)`

The new methods are a follow-up to the existing `hasRowCount(int expectedSize)` method.

#### Check List:
* Unit tests : YES


mk868:
If changes are fine and get merged then I will create an equivalent PR for the table columns

scordio:
To avoid bloating the API, what if we added a `.rowCount()` navigation method, which would return integer assertions?

mk868:
I agree, I like this idea, I will prepare a PR with `.rowCount()` and `.columnCount()` methods

PR #3607, opened 2024-09-13
feat: support HashSet#contains logic for changed hashCodes

### LINKED ISSUE #1104: contains for hashset finds Object which is changed
#### Summary
The contains method of HashSet doesn´t find an object when a value changed which is used in the hashcode method of the object. Normally 'assertThat(anyHashSet).contains(changedObject)' should not find it. I wrote a Test and added it here. The Bug appeard in Version 3.8.0. 

#### Example

```java
    @Test
    public void hashSetCanFailToFindTheSameElementPreviouslyAdded() {
        Set<Date> hashSet = new HashSet<>();
        Date date = new Date();
        hashSet.add(date);
        date.setTime(123L);
        Assertions.assertThat(hashSet).contains(date);
        Assertions.assertThat(hashSet.contains(date)).isFalse();
    }
```

#### Java 8 specific ?

mat3e:
#### Check List:
* Fixes #1104
* Unit tests : YES
* PR meets the [contributing guidelines](https://github.com/assertj/assertj/blob/main/CONTRIBUTING.md) : YES

### DIFF
--- a/assertj-core/src/test/java/org/assertj/core/api/hashset/HashSetAssert_contains_Test.java
+++ b/assertj-core/src/test/java/org/assertj/core/api/hashset/HashSetAssert_contains_Test.java
@@ -1,0 +1,65 @@
+package org.assertj.core.api.hashset;
+
+import static java.time.Instant.EPOCH;
+import static java.util.Arrays.asList;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatNoException;
+import static org.assertj.core.api.BDDAssertions.then;
+import static org.assertj.core.error.ShouldContain.shouldContain;
+import static org.assertj.core.util.AssertionsUtil.expectAssertionError;
+import static org.mockito.Mockito.verify;
+
+import java.util.Date;
+import java.util.HashSet;
+
+import org.assertj.core.api.HashSetAssert;
+import org.assertj.core.api.HashSetAssertBaseTest;
+
+class HashSetAssert_contains_Test extends HashSetAssertBaseTest {
+  @Override
+  protected HashSetAssert<Object> invoke_api_method() {
+    return assertions.contains(someValues);
+  }
+
+  @Override
+  protected void verify_internal_effects() {
+    verify(iterables).assertContains(getInfo(assertions), getActual(assertions), someValues);
+  }
+
+  @HashSetFactory.Test
+  void should_pass(HashSetFactory hashSetFactory) {
+    // GIVEN
+    HashSet<String> hasSet = hashSetFactory.createWith("Yoda", "Luke", "Han");
+    // WHEN/THEN
+    assertThatNoException().isThrownBy(() -> assertThat(hasSet).contains("Han", "Luke"));
+  }
+
+  @HashSetFactory.Test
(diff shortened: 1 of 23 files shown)


joel-costigliola:
@mat3e there are build failure because the code needs to be java 8 compatible.

assertj-core/src/test/java/org/assertj/core/api/hashset/HashSetAssert_contains_Test.java
```diff
@@ -0,0 +1,44 @@
...
+class HashSetAssert_contains_Test extends HashSetAssertBaseTest {
+  @Test
+  void should_pass() {
+    // WHEN/THEN
+    assertThatNoException()
+      .isThrownBy(() -> assertThat(newHashSet(Set.of("Yoda", "Luke", "Han"))).contains("Han", "Luke"));
+  }
+
+  @Test
+  void should_fail_if_missing_but_also_if_hashCode_changed() {
```
joel-costigliola, line 27: Having separate tests for missing element and different hash code would be better

assertj-core/src/test/java/org/assertj/core/api/hashset/HashSetAssert_contains_Test.java
```diff
@@ -0,0 +1,44 @@
...
+    // GIVEN
+    Date first = Date.from(EPOCH.plusSeconds(1));
+    Date second = Date.from(EPOCH.plusSeconds(2));
+    HashSet<Date> dates = newHashSet(Set.of(first, second));
```
joel-costigliola, line 31: can't use `Set.of` since the code must be java 8 compatible

assertj-core/src/test/java/org/assertj/core/api/hashset/HashSetAssert_contains_Test.java
mat3e: Right, I'll ultimately split this, but wanted to emphasize the current behavior - first fail doesn't discover the changed `hashCode` and occurs just because of `missing` element. 

Not sure if this is the way to go - one might get a failure, fix something in code/tests and then get another failure (due to `hashCode` change). Ideally, fail should discover everything I think, but for that I might need to further improve the `ComparisonStrategy` perhaps to include both checks at once or do tricks like
```java
assertAll(() -> super.containsForProxy(values), () -> hashSets.assertContains(info, actual, values));
```
in the overriden `containsForProxy` 🤔

joel-costigliola: why not use `doesNotContain` to emphasize the missing value is not in the hash set ?

mat3e:
I allowed myself to resolve the conversations. Took me longer than I thought, but should be good now.

It's probably a breaking change, but #1104 is now assigned to `4.0.0-M1`, so shouldn't be a problem 👍🏻

mat3e:
I haven't yet updated `InstanceOfAssertFactories`, `BDDAssertions`, etc. Let me know if the rest looks conceptually OK.

joel-costigliola:
Integrated thanks @mat3e !

mat3e:
> Integrated thanks @mat3e !

Wow, thanks! 

I wanted to fix remaining classes during next longer time off from ordinary work, but glad you did it 😇

PR #3617, opened 2024-09-23
Add standard representation for `CharSequence`

Marcono1234:
Currently `CharSequence` has no special handling in `StandardRepresentation`, so when you accidentally compare a `String` with for example a `StringBuilder` you get something like:
> expected: test
> but was: "test"

This assumes that for custom `CharSequence` types users have overridden `toString`, as required by the `CharSequence` Javadoc. Otherwise it might be a bit confusing that the result is now enclosed in `"`.

#### Check List:
* Unit tests : YES

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
+++ b/assertj-core/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
@@ -235,6 +235,7 @@ public class StandardRepresentation implements Representation {
     if (object instanceof File) return toStringOf((File) object);
     if (object instanceof Path) return fallbackToStringOf(object);
     if (object instanceof String) return toStringOf((String) object);
+    if (object instanceof CharSequence) return toStringOf((CharSequence) object);
     if (object instanceof Character) return toStringOf((Character) object);
     if (object instanceof Comparator) return toStringOf((Comparator<?>) object);
     if (object instanceof SimpleDateFormat) return toStringOf((SimpleDateFormat) object);
@@ -401,6 +402,10 @@ public class StandardRepresentation implements Representation {
   }
 
   protected String toStringOf(String s) {
+    return toStringOf((CharSequence) s);
+  }
+
+  protected String toStringOf(CharSequence s) {
     return concat("\"", s, "\"");
   }
 
--- a/assertj-core/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
+++ b/assertj-core/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
@@ -401,12 +401,16 @@ public class StandardRepresentation implements Representation {
     return String.format("local class %s", c.getSimpleName());
   }
 
+  private static String toQuotedString(CharSequence s) {
+    return concat("\"", s, "\"");
+  }
+
   protected String toStringOf(String s) {
-    return toStringOf((CharSequence) s);
+    return toQuotedString(s);
   }
 
   protected String toStringOf(CharSequence s) {
-    return concat("\"", s, "\"");
+    return toQuotedString(s);
   }
 
   protected String toStringOf(Character c) {
--- a/assertj-core/src/main/java/org/assertj/core/presentation/UnicodeRepresentation.java
+++ b/assertj-core/src/main/java/org/assertj/core/presentation/UnicodeRepresentation.java
@@ -43,14 +43,15 @@ public class UnicodeRepresentation extends StandardRepresentation {
   }
 
   @Override
-  protected String toStringOf(String string) {
+  protected String toStringOf(CharSequence string) {
     return escapeUnicode(string);
   }
 
-  private static String escapeUnicode(String input) {
+  private static String escapeUnicode(CharSequence input) {
     StringBuilder b = new StringBuilder(input.length());
     Formatter formatter = new Formatter(b);
-    for (char c : input.toCharArray()) {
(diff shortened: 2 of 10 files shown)


assertj-core/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
```diff
@@ -401,6 +402,10 @@ protected String toStringOf(Class<?> c) {
   }
 
   protected String toStringOf(String s) {
+    return toStringOf((CharSequence) s);
+  }
+
+  protected String toStringOf(CharSequence s) {
     return concat("\"", s, "\"");
   }
```
Marcono1234, line 410: I wasn't sure whether the existing `toStringOf(String)` should be replaced (since `String` implements `CharSequence`), or whether that would be a breaking API change.

assertj-core/src/test/java/org/assertj/core/error/ShouldBeEqual_Test.java
```diff
@@ -66,6 +66,18 @@ void should_use_actual_and_expected_representation_in_AssertionFailedError_actua
                                   " but was: [1, 2, 3]"));
   }
 
+  @Test
+  void should_display_class_for_ambiguous_CharSequence() {
```
Marcono1234, line 70: Not completely sure where the 'right' place for this test is. All these test classes seem to perform variants of ambiguous `toString` checks:
- `ShouldBeEqual_Test` (this class here)
- `ShouldBeEqual_newAssertionError_differentiating_expected_and_actual_Test`
- `StandardRepresentation_unambiguousToStringOf_Test`

assertj-core/src/test/java/org/assertj/core/error/ShouldBeEqual_Test.java
```diff
@@ -154,7 +166,7 @@ void should_display_multiline_values_nicely_with_comparison_strategy() {
   }
 
   @Test
-  void should_display_multiline_values_nicely_for_ambiguous_representation_for_ambiguous_representation() {
+  void should_display_multiline_values_nicely_with_comparison_strategy_for_ambiguous_representation() {
```
Marcono1234, line 169: Renamed this because it contained `for_ambiguous_representation` twice, and the test in the lines in front also used `..._with_comparison_strategy` as name.

assertj-core/src/test/java/org/assertj/core/error/ShouldBeEqual_Test.java
joel-costigliola: That would be in `StandardRepresentation_unambiguousToStringOf_Test` where we should have exhaustive tests for representation, the other tests are just to check that `ShouldBeEqual` honors that too but we only one test proving that.

assertj-core/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
joel-costigliola: It's not a binary breaking change but if some users rely on that representation then it would be, so I'm inclined we don't change it. Reusing the String representation for char sequence is ok as if the contents are the same we would add the type to differentiate the values.

Marcono1234: Ah sorry my comment was a bit misleading. What I meant is:
- Should there be `toStringOf(String)` and `toStringOf(CharSequence)` (as done here currently)
- Or, should there only be `toStringOf(CharSequence)`, and `toStringOf(String)` is removed\
The code would then simply use the `CharSequence` variant for `String` as well. But removing the `toStringOf(String)` overload would technically be an incompatible API change since it is `protected` and users might have overridden it.

joel-costigliola: I think it would be better to extract `concat("\"", s, "\"");` to a private method and call it from both `toStringOf`, users can then override `toStringOf(CharSequence s)` without impacting `toStringOf(String s)`.

assertj-core/src/main/java/org/assertj/core/presentation/UnicodeRepresentation.java
```diff
@@ -33,6 +33,7 @@ public class UnicodeRepresentation extends StandardRepresentation {
   public String toStringOf(Object object) {
     if (hasCustomFormatterFor(object)) return customFormat(object);
     if (object instanceof String) return toStringOf((String) object);
+    if (object instanceof CharSequence) return toStringOf((CharSequence) object);
     if (object instanceof Character) return toStringOf((Character) object);
```
Marcono1234, line 37: I think technically all these `instanceof` checks are redundant because they are equivalent to what the parent class `StandardRepresentation` would do, but for consistency I added the check for `CharSequence` here as well.

assertj-core/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
Marcono1234: I hope the latest changes in 6952df829991457aed6cd10b667e8e27b68e56af are what you had in mind.

joel-costigliola: thanks, I'll review that next week, I'm off for a few days

joel-costigliola:
Integrated thanks @Marcono1234 !

PR #3618, opened 2024-09-23
Fix Unicode escapes in `inUnicode()` Javadoc

Marcono1234:
They are already processed by the Java compiler, so previously instead of the Unicode escape the corresponding character was shown.

See for example [`AbstractCharacterAssert#inUnicode()` Javadoc](https://javadoc.io/doc/org.assertj/assertj-core/latest/org/assertj/core/api/AbstractCharacterAssert.html#inUnicode()):
Javadoc screenshot
Notice how it does not show the Unicode escape.

I chose the HTML character reference `&bsol;` as replacement for `\`, but other approaches might work as well.

#### Check List:
* Unit tests : NA


joel-costigliola:
Awesome thanks for the fix @Marcono1234 !

PR #3645, opened 2024-11-01
hasFieldOrPropertyWithValue with getters throwing exception #3563

### LINKED ISSUE #3563: hasFieldOrPropertyWithValue() eats exceptions thrown by getters, and reports non-existent property instead
**Describe the bug**
If a getter throws an exception, it is not reported in the `AssertionError` from `hasFieldOrPropertyWithValue()`

* assertj core version: 3.26.3
* java version: 11
* test framework version: junit 5.10.2

```java

class ThrowingGetterTest {
  @Test
  void test() {
    assertThat(new MyThrowingPojo())
        .hasFieldOrPropertyWithValue("property", "value");
  }

  static class MyThrowingPojo {
    public String getProperty() {
      throw new RuntimeException();
    }
  }
}
```
Output:
```
java.lang.AssertionError: 
Expecting
  AssertJTest$MyThrowingPojo@be64738
to have a property or a field named "property"
(static and synthetic fields are ignored)

	at AssertJTest.test(AssertJTest.java:9)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
```
(no trace of the `RuntimeException` – note that the test succeeds when `getProperty()` returns `"value"`)

The problem appears to be that the nice `IntrospectionError` built by `PropertyOrFieldSupport#getSimpleValue()` is discarded in `org.assertj.core.internal.Objects#assertHasFieldOrProperty()`.

In my specific case, the object under test was implemented by proxy which was throwing the exception, which makes it really confusing.

etrandafir93:
#### Overview

Current behavior for _hasFieldOrPropertyWithValue_ / _hasFieldOrProperty_ 
- we try to invoke the getter/accessor
- if not present **or if it throws an exception** we verify the field directly 
- even if the field is missing, the original exception is swallowed 

As discussed in #3563, we want to keep using the field verification as a fallback in case the getter invocation fails.
However, if the field is missing, we want to rethrow the original exception raised by the getter.

#### About the PR

- In the first commit, I've added some tests for the current behavior of the lib
- The 2nd commit updates 2 of the tests to reflect the new desired behavior (failing test)
- In 3rd commit we have the actual code changes

- In `Objects::assertHasFieldOrProperty` we can decide if we want to throw the original getter exception as a runtime (This would be similar to what happens when something like _.extracting(...)_ throws an exception). Alternatively, we can wrap it within an `AssertionError` with a custom message.

#### Check List:
* Fixes #3563 
* Unit tests : YES 

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/internal/Objects.java
+++ b/assertj-core/src/main/java/org/assertj/core/internal/Objects.java
@@ -12,15 +12,26 @@
  */
 package org.assertj.core.internal;
 
+import org.assertj.core.api.AssertionInfo;
+import org.assertj.core.error.GroupTypeDescription;
+import org.assertj.core.internal.DeepDifference.Difference;
+import org.assertj.core.util.VisibleForTesting;
+import org.assertj.core.util.introspection.FieldSupport;
+import org.assertj.core.util.introspection.IntrospectionError;
+import org.assertj.core.util.introspection.PropertyOrFieldSupport;
+import org.assertj.core.util.introspection.PropertySupport;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.util.*;
+
 import static java.lang.String.format;
 import static java.lang.reflect.Modifier.isStatic;
 import static java.util.Arrays.asList;
 import static java.util.Arrays.stream;
 import static java.util.Objects.deepEquals;
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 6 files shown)


joel-costigliola:
Integrated thanks @etrandafir93!

PR #3652, opened 2024-11-09
Fix Issues/3650

### LINKED ISSUE #3650: Enhance AbstractFileAsert#hasSameBinaryContentAs
I just found that the [AbstractFileAssert#hasSameBinaryContentAs(File)](https://github.com/assertj/assertj/blob/835c1dc3aee2348eb234f3209488fdd078702c2f/assertj-core/src/main/java/org/assertj/core/api/AbstractFileAssert.java#L407) method does the following.

* Loads **whole** bytes of the *expected* file into memory(an array).
* And compares byte by byte from two input streams,
  * one from the `actual` file
  * and the other is the in-memory cached byte array (from the *expected* file)

We can(should) do like this.

* Pre-verify those files' sizes for a fast-fail manner, since comparing bytes from two different-sized files is meaningless.
* Compare two binary sequences (as is), **without reading all bytes (from the *expected* file)**

```java
        assertThat(decrypted)
                .hasSize(plain.length())        // not required when it's covered by the hasSameBinaryContentsAs
                .hasSameBinaryContentAs(plain); // should be more memory-friendly!
```

onacit:
#### Check List:
* Fixes #3650 
* Unit tests : YES

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractFileAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractFileAssert.java
@@ -46,6 +46,7 @@ import org.assertj.core.util.VisibleForTesting;
  * @author Valeriy Vyrva
  * @author Nikolaos Georgiou
  * @author Rostyslav Ivankiv
+ * @author Jin Kwon
  */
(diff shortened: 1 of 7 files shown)


joel-costigliola:
@onacit thanks for the PR, a few quick comments:

- can you run  `mvn spotless:apply` to fix format violations ?
- I did not see a test that shows we are checking binary content sizes, did I miss it ?

PR #3686, opened 2024-12-03
Add throwable stacktrace to ShouldNotContainCharSequence

shaikhu:
#### Check List:
* Fixes #1355 )
* Unit tests : YES
* Javadoc with a code example (on API only) :  NA
* PR meets the [contributing guidelines](https://github.com/assertj/assertj/blob/main/CONTRIBUTING.md)

PR #3691, opened 2024-12-07
Add throwable stacktrace to ShouldNotContainCharSequence - Correct message

shaikhu:
Apologies, double checking, I don't think the previous [PR ](https://github.com/assertj/assertj/pull/3686) was implemented correctly. 😓 Using [ShouldContainThrowable_create_Test](https://github.com/shaikhu/assertj/blob/3.x/assertj-core/src/test/java/org/assertj/core/error/ShouldContainThrowable_create_Test.java) as a basis, this PR has the correct implementation.

#### Check List:
* Fixes #1355
* Unit tests : YES


joel-costigliola:
Good catch!

PR #3694, opened 2024-12-08
Simplity is less than or eq to

cookieMr:
#### Check List:
* Fixes n/a
* Unit tests : NA

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/internal/AbstractComparisonStrategy.java
+++ b/assertj-core/src/main/java/org/assertj/core/internal/AbstractComparisonStrategy.java
@@ -66,7 +66,7 @@ public abstract class AbstractComparisonStrategy implements ComparisonStrategy {
 
   @Override
   public boolean isLessThanOrEqualTo(Object actual, Object other) {
-    return areEqual(actual, other) || isLessThan(actual, other);
+    return !isGreaterThan(actual, other);
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


assertj-core/src/main/java/org/assertj/core/internal/AbstractComparisonStrategy.java
```diff
@@ -61,12 +61,12 @@ public boolean arrayContains(Object array, Object value) {
 
   @Override
   public boolean isLessThan(Object actual, Object other) {
-    return !areEqual(actual, other) && !isGreaterThan(actual, other);
+    return !isGreaterThanOrEqualTo(actual, other);
```
cookieMr, line 64: simple De Morgan's law -> https://en.wikipedia.org/wiki/De_Morgan%27s_laws

assertj-core/src/main/java/org/assertj/core/internal/AbstractComparisonStrategy.java
```diff
@@ -61,12 +61,12 @@ public boolean arrayContains(Object array, Object value) {
...
   }
 
   @Override
   public boolean isLessThanOrEqualTo(Object actual, Object other) {
-    return areEqual(actual, other) || isLessThan(actual, other);
+    return !isGreaterThan(actual, other);
```
cookieMr, line 69: no need to call twice the `areEqual` (2nd call was negated)

joel-costigliola:
Thanks @cookieMr.

PR #3696, opened 2024-12-09
Custom equality comparator gh 3678

### LINKED ISSUE #3678: Support specialized equality methods
#### Feature summary
Often times, classes define specialized equality methods that have different semantics than `Object#equals`. It would be nice to be able to reference these directly via a `BiPredicate<T, T>` and have it work with `isEqualTo` and other relevant methods.

#### Example

```java
assertThat(actual)
    .usingEqualityMethod((x, y) -> x.equalsStrict(y))
    .isEqualTo(expected);

// or

assertThat(actual)
    .usingEqualityMethod(MyObject::equalsStrict)
    .isEqualTo(expected);
```

I can currently hack around this by (ab)using `usingComparator`, but it's not ideal.

```java
assertThat(actual)
    .usingComparator((x, y) -> x.equalsStrict(y) ? 0 : -1)
    .isEqualTo(expected)
```

cookieMr:
#### Check List:
* Fixes #3678
* Unit tests : YES

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -31,6 +31,7 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.BiPredicate;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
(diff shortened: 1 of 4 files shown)


cookieMr:
@scordio plz, let me know if there is some more work in this PR to do to make it ready for merge

i did just 2 classes with tests as they'll be the same for each implementation of the parent `AbstractAssert` class but if you think more is needed, i can do that

also i don't know if i put tests in the right place

joel-costigliola:
Integrated with a few updates, thanks @cookieMr

PR #3717, opened 2024-12-27
Enforce Java 17+ as build environment

Bananeweizen:
The build otherwise fails with low level bytecode class version related exceptions. Make this far more explicit, as the exception doesn't really tell what's wrong.

Also simplify the configuration of the enforcer plugin. It's not a good pattern to have separate executions in each module, rather just inherit a default configuration and add more specific rules. Note the "combine-children" attribute.

### DIFF
--- a/assertj-core/pom.xml
+++ b/assertj-core/pom.xml
@@ -291,29 +291,21 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-enforcer-plugin</artifactId>
-        <executions>
-          <execution>
-            <id>enforce-dependencies</id>
-            <goals>
-              <goal>enforce</goal>
-            </goals>
-            <configuration>
-              <rules>
-                <bannedDependencies>
-                  <includes>
-                    <include>net.bytebuddy:byte-buddy</include>
-                  </includes>
-                  <excludes>
-                    <exclude>org.assertj:assertj-core</exclude>
-                    <exclude>org.hamcrest:hamcrest-core</exclude>
-                    <exclude>*:*:*:jar:compile</exclude>
-                  </excludes>
-                </bannedDependencies>
-                <dependencyConvergence />
-              </rules>
-            </configuration>
-          </execution>
-        </executions>
+        <configuration>
+          <rules combine-children="append">
+            <bannedDependencies>
+              <includes>
+                <include>net.bytebuddy:byte-buddy</include>
+              </includes>
+              <excludes>
+                <exclude>org.assertj:assertj-core</exclude>
+                <exclude>org.hamcrest:hamcrest-core</exclude>
+                <exclude>*:*:*:jar:compile</exclude>
+              </excludes>
+            </bannedDependencies>
+            <dependencyConvergence />
+          </rules>
+        </configuration>
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -191,6 +191,22 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-enforcer-plugin</artifactId>
           <version>${maven-enforcer-plugin.version}</version>
+          <executions>
+            <execution>
+              <goals>
+                <goal>enforce</goal>
+              </goals>
+            </execution>
+          </executions>
+          <configuration>
+            <rules>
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -191,6 +191,22 @@
...
+            </execution>
+          </executions>
+          <configuration>
+            <rules>
+              <requireJavaVersion>
+                <version>[17,)</version>
+                <message>The build requires Java 17 or better.</message>
+              </requireJavaVersion>
+              <requirePluginVersions />
+            </rules>
```
scordio, line 208: We had this in the past, but we removed it in 0af36c24 as it was putting the burden of running with Java 17 on all other AssertJ modules that depend on the `assertj-parent`. However, this is not always possible, for example due to the `javax` vs. `jakarta` namespace.

We could still enforce the runtime version on `assertj-tests`. Plus, we should enforce Java 21 at least, since that's the language level we use in the tests:

https://github.com/assertj/assertj/blob/5c58e3623a331fd19806c92643757fed3dbc1702/assertj-tests/pom.xml#L23

Could you update the PR accordingly?

assertj-core/pom.xml
```diff
@@ -291,29 +291,21 @@
...
-                    <exclude>*:*:*:jar:compile</exclude>
-                  </excludes>
-                </bannedDependencies>
-                <dependencyConvergence />
-              </rules>
-            </configuration>
-          </execution>
-        </executions>
+        <configuration>
+          <rules combine-children="append">
```
scordio, line 295: Nice! 👍

pom.xml
Bananeweizen: @scordio I can surely update the PR, but I'm not sure if we look at the same things. The enforcer rule checks that the _JRE running Maven_ has the mentioned version (i.e. it makes sure that the used Maven plugins can run correctly, and at least the bnd-maven-plugin requires Java 17 AFAICS). It's completely unrelated to the compiler, source, target, release, ... settings when compiling the java code. Those are determined by the maven.compiler.xxx settings.
So lifting this restriction would only allow to build some of the modulese separately from a shell which uses Java 8 or 11, under the additional condition of removing the bdn-maven-plugin configuration for that specific module. Right now, all modules use that plugin, as it's configured in the parent POM. Therefore none of the modules can be built on a shell with Java 8/11 currently.

Do you still want changes?

scordio: What I meant is, if we add the rule to the root POM, `assertj-parent` will inherit it, and any AssertJ module using that parent will be forced to run on Java 17 during build time, which is less desirable for modules not in this repository.

Or did I miss something?

Bananeweizen: Now I got it, I looked only at modules in this repo. Even though I didn't find any outside maven module referencing this parent, we can surely move the java 17 rule to the core module, and have an additional 21 rule on the test module. That would then not cover building assertj-guava (or other modules) separately, but would be a compromise for catching many typical error patterns, but not spreading to modules outside the repo.

Does that sound good?

scordio: Just rethinking it again: do we really want to apply restrictions on version 3, knowing that we'll move on after 3.27.1 is released? Such a change wouldn't have any concrete impact on the upcoming release.

What if we merge the rules as you suggested and add the Java 21 rule to the test module only? We will reuse such a setup in version 4.

Bananeweizen: This change is not targeted to any release, it has zero compilation or runtime consequences. This PR tries to improve the life of contributors (myself included), by failing with a useful error message, if the local development environment runs a wrong Java version. Therefore it's useful on _every_ past or future branch. Assuming that version 3 might be maintained some more time and contributors run into the same kind of trouble on the 3.x maintenance branch, it seems rather logical to me to merge this as soon as possible and not to defer this to an arbitrary point of time in the future when someone decides that version 4 development starts. It's not related to version 3 or 4.

scordio:
**Team decision**:
* We don't want to add further changes to the `3.x` branch as it's likely to be updated by the AssertJ team only
* Given that the Java baseline of `main` is now 17, at the moment we don't see a concrete need to configure the enforcer plugin to fail fast, as the compiler plugin executions will anyway fail with an incompatible JRE
* We still want to cherry-pick the enforcer configuration improvements from this PR

scordio:
Superseded by 6cce2bfe6b0ad81da114a9365fc65e2a35968d1e.

PR #3718, opened 2024-12-27
Declare license using SPDX identifier

Bananeweizen:
That's the recommended way to work with licenses in the Maven POM, see https://maven.apache.org/pom.html#Licenses


scordio:
Good catch. Thanks, @Bananeweizen!

PR #3740, opened 2025-01-10
Improve failure message for instanceOf if actual is null (#3428)

### LINKED ISSUE #3428: isInstanceOf() should issue a better failure message when the tested object is null
Background: I often use `assertThat (catchException (...)).isInstanceOf (...)` instead of `assertThatThrowBy ...` or something similar. I guess it's a matter of taste what to prefer. I have noticed that `isInstanceOf()` issues a pretty confusing failure message when nothing is actually thrown, i.e. when `catchException()` returns `null`.

To illustrate, when I run this code:

    assertThat (catchException (() -> { })).isInstanceOf (RuntimeException.class);

I get this failure message:

In comparison, when the tested object is not `null`, but of a wrong class, failure message is much more clear. For example:

    assertThat (catchException (() -> { throw new Exception (); })).isInstanceOf (RuntimeException.class);

gives me this:

    Expecting actual throwable to be an instance of:
      java.lang.RuntimeException
    but was:
      java.lang.Exception

#### Suggestion

Rephrase the `null`-case message somewhat like this:

    Expecting actual throwable to be an instance of:
      java.lang.RuntimeException
    but was:
      null (no throwable)

taulpi:
#### Check List:
* Fixes #3428
* Unit tests : YES

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/internal/Objects.java
+++ b/assertj-core/src/main/java/org/assertj/core/internal/Objects.java
@@ -28,6 +28,7 @@ import static org.assertj.core.error.ShouldBeEqualToIgnoringFields.shouldBeEqual
 import static org.assertj.core.error.ShouldBeExactlyInstanceOf.shouldBeExactlyInstance;
 import static org.assertj.core.error.ShouldBeIn.shouldBeIn;
 import static org.assertj.core.error.ShouldBeInstance.shouldBeInstance;
+import static org.assertj.core.error.ShouldBeInstance.shouldBeInstanceButWasNull;
 import static org.assertj.core.error.ShouldBeInstanceOfAny.shouldBeInstanceOfAny;
 import static org.assertj.core.error.ShouldBeOfClassIn.shouldBeOfClassIn;
 import static org.assertj.core.error.ShouldBeSame.shouldBeSame;
@@ -126,6 +127,7 @@ public class Objects {
   }
 
   public void assertIsInstanceOf(AssertionInfo info, Object actual, Class<?> type) {
+    if (actual == null) throw failures.failure(info, shouldBeInstanceButWasNull(type));
     if (!isInstanceOfClass(actual, type, info)) throw failures.failure(info, shouldBeInstance(actual, type));
   }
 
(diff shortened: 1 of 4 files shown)


assertj-core/src/main/java/org/assertj/core/internal/Objects.java
```diff
@@ -126,6 +127,7 @@ public Failures getFailures() {
   }
 
   public void assertIsInstanceOf(AssertionInfo info, Object actual, Class<?> type) {
+    if (actual == null) throw failures.failure(info, shouldBeInstanceButWasNull(type));
```
scordio, line 130: I'm unsure if we should proceed with such a change as it's coupling the null check with the assertion-specific check. Furthermore, an `assertNotNull` call is already performed inside `isInstanceOfClass`, which should be removed if we decide to return an assertion-tailored error message when `actual` is null.

I see it as a more general topic as in many other places we leverage `isNotNull()` before executing the corresponding assertion, e.g.:

https://github.com/assertj/assertj/blob/e9b7839e838501bc998d595e683929279ad0da20/assertj-core/src/main/java/org/assertj/core/api/AbstractOptionalAssert.java#L102-L106

The general question is: should the null error message be generalized to give more context about the invoked assertion? If yes, how do we do that in a maintainable way without carrying hardcoded strings everywhere?

scordio (commented):
Hi @taulpi, thanks for raising the PR! I have a fundamental comment that we probably need to discuss within the team.

assertj-core/src/main/java/org/assertj/core/internal/Objects.java
taulpi: Hi @scordio,

thank you for the quick reply! Just a quick hint regarding removing the `assertNotNull` from `isInstanceOfClass`: That will probably break `isNotInstanceOf(null)`, as the current behaviour would throw a failure, and the change would make it pass. Thinking about it, this could be remedied by also executing a null check in `assertIsNotInstanceOf`. Then, the `assertNotNull` could be removed and `assertIsNotInstanceOf` becomes "symmetrical" to `assertIsInstanceOf`.

Your general point totally makes sense. Always giving more context in the failure message for `actual == null` would be a much fundamental change. Let me know, when there are any news or decisions regarding this issue.

PR #3750, opened 2025-01-22
Add warnings about Stream closing to assertThat JavaDocs

rpokorny:
#### Check List:
* NA

I noticed that since https://github.com/assertj/assertj/pull/1218 , `assertThat` automatically closes streams that are passed to it.  This is worth documenting for cases where running assertions on the the items in the stream is dependent on the stream remaining open.  For instance, I have just had a case where I wanted to test a stream that's backed by a `ZipFile`, where the items in the stream are only usable so long as the stream (and backing `ZipFile`) remain open.  This isn't possible with `assertThat(BaseStream)`.  This was easy enough to work around (by collecting the stream to an array myself), but it would be very helpful towards understanding the problem if the documentation specified that the stream gets closed by `assertThat`.

This PR adds a brief mention of this closing behavior to the relevant javadocs that I could find.


joel-costigliola:
@rpokorny Thanks for pointing that out and taking the time to improve the doc, appreciated.

joel-costigliola:
Integrated thanks @rpokorny !

PR #3759, opened 2025-01-30
Run OpenRewrite Java 17 recipe

scordio:
See https://docs.openrewrite.org/recipes/java/migrate/upgradetojava17.


assertj-core/src/main/java/org/assertj/core/internal/DeepDifference.java
```diff
@@ -254,7 +254,7 @@ private static List<Difference> determineDifferences(Object a, Object b, List<St
 
       // Special handle SortedSets because they are fast to compare
       // because their elements must be in the same order to be equivalent Sets.
-      if (key1 instanceof Collection<?> set) {
+      if (key1 instanceof SortedSet<?> set) {
```
scordio: Hi @timtebeek! May I ask your opinion on this one?

I executed the Java 17 recipe on the AssertJ codebase, and as far as I can tell, the source changes in this commit (the red ones) resulted from an incorrect migration.

Initially, they were:

https://github.com/assertj/assertj/blob/11a36324ea3edb3cb910ee1eb608dc7064e01eb9/assertj-core/src/main/java/org/assertj/core/internal/DeepDifference.java#L258-L259

https://github.com/assertj/assertj/blob/11a36324ea3edb3cb910ee1eb608dc7064e01eb9/assertj-core/src/main/java/org/assertj/core/internal/DeepDifference.java#L267-L268

Do you think it's something to report as an OpenRewrite issue?

assertj-core/src/main/java/org/assertj/core/internal/DeepDifference.java
```diff
@@ -263,7 +263,7 @@ private static List<Difference> determineDifferences(Object a, Object b, List<St
       }
 
       // Check List, as element order matters this comparison is faster than using unordered comparison.
-      if (key1 instanceof Collection<?> list) {
+      if (key1 instanceof List<?> list) {
```
scordio: @timtebeek this is the second one, which originally was:

https://github.com/assertj/assertj/blob/11a36324ea3edb3cb910ee1eb608dc7064e01eb9/assertj-core/src/main/java/org/assertj/core/internal/DeepDifference.java#L267-L268

assertj-core/src/main/java/org/assertj/core/internal/DeepDifference.java
timtebeek: Apologies for the late reply! I was out 🏂🏻. Indeed looks like a bug; I think it's best not to make any change if there's a mismatch between the righthand side of the `instanceof` and the cast. Would you mind logging an issue? I'm catching up here still.

Great to see you're running these recipes; I hope they've helped you so far already.

scordio: No worries!

Yes, they definitely helped! Still going through the review of the 400+ files 🙃 

I'll log an issue in the upcoming days 👍

PR #3770, opened 2025-02-14
Docs migration to Antora

VanRoy:
### Overview

Migration docs from central `assertj-docs` repository to aggregation by `Antora`


scordio:
Hi @VanRoy, do I understand correctly that your intention is to bring the docs into the 3.x branch?

Will we take up the 4.x (current) docs in a separate set of changes? I could imagine they'll just be a copy of the 3.x content, at least until we start adding new 4.x specific features.

VanRoy:
@scordio Indeed my first intention is to migrate the documentation of the 3.27.3 and also prepare all is needed to publish documentation at the of the release of 4.0. 

For the 4.0 , it's pretty easy , we just have to copy / cherry-pick the docs folder of this branch to main branch.
After the creation of the tag `assertj-build-4.0.0`, the Antora project will automatically detect the version ( with this settings : https://github.com/assertj/assertj.github.io/pull/1/files#diff-0e8a7c2f91f241f59cc684e048c14593122447d39c4b8bfa6953cb5e91151aa2R16 ) 

For the 3.27.3 it's a little bit tricky because the tag already exist. So I think we have these options : 

- Publish a new release `assertj-build-3.27.4` with the documentation ( maybe a little bit weird a release only for docs refactoring )
- Create a tag like `assertj-build-3.27.3-docs` , specifically to allow Antora project detect this version and generate the doc
- Instead a tag we can create a branch like `docs-3.27.3` ( and keep it ) to host  3.27.3 docs
- Don't document the 3.27.3 an publish only the docs for the 4.0.0
- Maybe another idea ?

PR #3815, opened 2025-03-27
Fix: hasNoNullFields() now detects null collections and arrays

### LINKED ISSUE #3810: hasNoNullFields does not detect null collections
**Describe the bug**
The hasNoNullFields assertion does not detect null values for collections while it works fine for other types
 
* assertj core version: 3.27.3
* java version: 21
* test framework version: junit-jupiter 5.10.3

The following test should fail for both assertions however the first passes while the second one correctly fails

```java
class NullCheckTest {

    public record Data(List<String> l, String s){};

    @Test
    void testNullString() {
        Data data1 = new Data(null, "");
        Data data2 = new Data(List.of(), null);

        assertThat(data1).usingRecursiveAssertion().hasNoNullFields();
        assertThat(data2).usingRecursiveAssertion().hasNoNullFields();
    }
}
```

reduan96:
Fixes #3810

### Description

This PR fixes a bug in `hasNoNullFields()` where fields of type `Collection`, `Map`, `Optional`, or `Array` that were `null` were not evaluated by the recursive assertion engine. As a result, the assertion passed even when some fields were actually `null`.

### GIVEN / WHEN / THEN

- GIVEN: an object with a `null` List, Map, Optional, or Array field
- WHEN: asserting using `assertThat(obj).usingRecursiveAssertion().hasNoNullFields()`
- THEN: the assertion should fail if any field is `null`, regardless of type

### Fix

We modified the condition in `RecursiveAssertionDriver#assertRecursively` to ensure that fields which are `null` are evaluated by the predicate, regardless of their type:

if (!isRootObject(fieldLocation) && (node == null || shouldEvaluateAssertion(nodeType))) {
    evaluateAssertion(predicate, node, fieldLocation);
}

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/recursive/assertion/RecursiveAssertionDriver.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/recursive/assertion/RecursiveAssertionDriver.java
@@ -66,7 +66,7 @@ public class RecursiveAssertionDriver {
     boolean nodeAlreadyVisited = markNodeAsVisited(node);
     if (nodeAlreadyVisited) return;
 
-    if (!isRootObject(fieldLocation) && shouldEvaluateAssertion(nodeType)) {
+    if (!isRootObject(fieldLocation) && (node == null || shouldEvaluateAssertion(nodeType))) {
       evaluateAssertion(predicate, node, fieldLocation);
     }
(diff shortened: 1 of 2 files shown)


joel-costigliola:
@reduan96 sorry for the long delay, I'm closing this PR because the actual fix was a documentation one.

thanks anyway for this PR

PR #3817, opened 2025-03-28
Do not show values for missing map keys

lmartelli:
I find it misleading/distracting to show the entry value when the reason the assertion fails is the key not being found.


joel-costigliola:
Integrated thanks @lmartelli !

PR #3836, opened 2025-05-17
Add option to ignore compared fields existence check

### LINKED ISSUE #3806: Allow to ignore check on compared fields in the recursive comparison (initial issue description was: Problems on method 'usingRecursiveFieldByFieldElementComparatorOnFields')
**Describe the bug**
From version 3.25.0, it seems that method `usingRecursiveFieldByFieldElementComparatorOnFields` on `AbstractIterableAssert` returns an error if one of the fields don't exist. The problem is that I have polymorphic objects.

* assertj core version: 3.25.0
* java version: 21
* test framework version: junit 5.11.4
* os (if relevant): 

```java

@Data
@SuperBuilder
public abstract sealed class AppliedExemptionResponse permits AppliedTotalExemptionResponse, AppliedPartialExemptionResponse {
	String exemptionCode;
	String description;
	String chargeItemCode;
	String description;
}

@Value
@SuperBuilder
public class AppliedTotalExemptionResponse extends AppliedExemptionResponse {
}

@Value
@SuperBuilder
public class AppliedPartialExemptionResponse extends AppliedExemptionResponse {
	BigDecimal value;
	String type;
	boolean replacement;
}

public class MyTest {

	@Test
	void test() {

		final List<AppliedExemptionResponse> appliedExemptions = List.of(
				AppliedTotalExemptionResponse.builder().exemptionCode( "E1" ).build(),
				AppliedPartialExemptionResponse.builder().value( BigDecimal.ZERO ).build()
		);
		final List<AppliedExemptionResponse> expectedAppliedExemptions = List.of(
				AppliedTotalExemptionResponse.builder().exemptionCode( "E1" ).build(),
				AppliedPartialExemptionResponse.builder().value( BigDecimal.ZERO ).build()
		);

		assertThat( appliedExemptions )
				.usingRecursiveFieldByFieldElementComparatorOnFields( "exemptionCode", "chargeItemCode", "description", "value" )
				.containsExactlyInAnyOrderElementsOf( expectedAppliedExemptions );
	}
}
```

The test fails with the following error:

`java.lang.IllegalArgumentException: The following fields don't exist: {value}`

How can I fix the problem?

kssumin:
This adds a new option (ignoringNonExistentFields) to recursive comparison that allows polymorphic objects with different field structures to be compared effectively. When enabled, the comparison will skip field existence validation, which is useful when comparing objects of different subtypes where some fields might only exist in certain implementations.

## Overview
This PR adds a new feature to the recursive comparison functionality that allows ignoring non-existent field checks. This is particularly useful when comparing polymorphic objects where some fields might not exist in all subtypes.

## Implementation Details
- Added `ignoreNonExistentFields` boolean flag to `RecursiveComparisonConfiguration`
- Added `ignoringNonExistentFields()` method to `RecursiveComparisonAssert`
- Added appropriate builder methods and setters
- Added early return in `checkComparedFieldsExist` when this option is enabled
- Added corresponding toString/description methods to show when this option is active
- Added unit tests to verify the behavior

## Example Use Case
When comparing objects like:
```java
class BaseClass {
    String commonField = "common";
}

class SubClass1 extends BaseClass {
    String subField1 = "sub1";
}

class SubClass2 extends BaseClass {
    // 'subField1' doesn't exist here
    String subField2 = "sub2";
}
```

Regular recursive comparison would fail due to field existence checks. With this new option:
```java
SubClass1 actual = new SubClass1();
SubClass2 expected = new SubClass2();

// This will now pass
assertThat(actual)
    .usingRecursiveComparison()
    .ignoringNonExistentFields()
    .isEqualTo(expected);
```

#### Check List:
* Fixes #3806 
* Unit tests : YES
* PR meets the [contributing guidelines](https://github.com/assertj/assertj/blob/main/CONTRIBUTING.md) : YES

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/RecursiveComparisonAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/RecursiveComparisonAssert.java
@@ -650,6 +650,46 @@ public class RecursiveComparisonAssert<SELF extends RecursiveComparisonAssert<SE
   }
 
   /**
+   * Allows to ignore non-existent fields in the recursive comparison.
+   * <p>
+   * This is useful when comparing polymorphic objects where some fields might not exist in all subtypes.
+   * <p>
+   * Example:
+   * <pre><code class='java'> class BaseClass {
+   *   String commonField = "common";
+   * }
+   *
+   * class SubClass1 extends BaseClass {
+   *   String subField1 = "sub1";
+   * }
+   *
+   * class SubClass2 extends BaseClass {
+   *   // 'subField1' doesn't exist here
+   *   String subField2 = "sub2";
+   * }
+   *
+   * SubClass1 actual = new SubClass1();
+   * SubClass2 expected = new SubClass2();
+   *
+   * // Regular recursive comparison - fails due to field existence check
+   * // assertThat(actual).usingRecursiveComparison().isEqualTo(expected); // fails!
+   *
+   * // Using ignoringNonExistentComparedFields() makes it pass
+   * assertThat(actual)
+   *   .usingRecursiveComparison()
+   *   .ignoringNonExistentComparedFields()
+   *   .isEqualTo(expected);
+   * </code></pre>
+   *
+   * @return this {@link RecursiveComparisonAssert} to allow fluent chaining.
+   */
+  @CheckReturnValue
+  public SELF ignoringNonExistentComparedFields() {
+    recursiveComparisonConfiguration.setIgnoreNonExistentComparedFields(true);
+    return myself;
+  }
+
+  /**
    * Makes the recursive comparison to ignore all <b>expected null fields</b>.
    * <p>
    * Example:
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
@@ -2515,6 +2515,52 @@ public abstract class AbstractIterableAssert<SELF extends AbstractIterableAssert
   }
 
   /**
+   * Allows to ignore non-existent fields in the recursive comparison when using
+   * {@link #usingRecursiveFieldByFieldElementComparatorOnFields(String...)} or similar methods.
+   * <p>
+   * This is useful when comparing collections of polymorphic objects where some fields might not exist in all subtypes.
+   * <p>
+   * Example:
+   * <pre><code class='java'>
+   * // GIVEN
+   * class BaseClass {
+   *   String common = "same";
+   * }
+   *
+   * class SubType1 extends BaseClass {
+   *   String onlyInSubType1 = "type1";
+   * }
+   *
+   * class SubType2 extends BaseClass {
+   *   // No 'onlyInSubType1' field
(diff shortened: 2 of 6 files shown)


joel-costigliola:
Thanks @kssumin, I will review your PR shortly !

assertj-core/src/test/java/org/assertj/core/internal/ConfigurableRecursiveFieldByFieldComparator_Test.java
```diff
@@ -87,4 +87,48 @@ void should_throw_an_NPE_if_given_RecursiveComparisonConfiguration_is_null() {
                    .hasMessage("RecursiveComparisonConfiguration must not be null");
   }
 
+  @Test
+  void should_success_ignore_non_existent_fields_when_configured() {
```
joel-costigliola, line 91: I'm not completely sure what we are testing here, we check that the config has the expected `ignoreNonExistentFields` value but I don't get the point of checking `int compare = configurableRecursiveFieldByFieldComparator.compare(actual, other);`.

Instead of these two tests, could we add a test in `RecursiveComparisonAssert_isEqualTo_comparingOnlyFields_Test` and reproduce the test example in the initial issue https://github.com/assertj/assertj/issues/3806#issuecomment-2745230319 ? if the implementation is correct, the test should pass

joel-costigliola: We could in addition also add a test with the code example from the javadoc (with BaseClass, Subclass1 and 2)

Thanks !

assertj-core/src/main/java/org/assertj/core/api/RecursiveComparisonAssert.java
```diff
@@ -649,6 +649,46 @@ public SELF ignoringActualEmptyOptionalFields() {
...
+   * assertThat(actual)
+   *   .usingRecursiveComparison()
+   *   .ignoringNonExistentFields()
+   *   .isEqualTo(expected);
+   * </code></pre>
+   *
+   * @return this {@link RecursiveComparisonAssert} to allow fluent chaining.
+   */
+  @CheckReturnValue
+  public SELF ignoringNonExistentFields() {
```
joel-costigliola, line 687: I still need to think if we keep that name or if we should rename to `ignoringNonExistentComparedFields` which is more accurate.

joel-costigliola:
Thanks @kssumin, first round of code review done.

kssumin:
@joel-costigliola Thank you for your review and suggestions! I've made the following changes based on your feedback:

1. Added test cases in RecursiveComparisonAssert_isEqualTo_comparingOnlyFields_Test that reproduce the example from issue #3806
2. Added another test with the BaseClass/SubClass example from the JavaDoc to verify the functionality
3. Updated the method name as suggested to be more descriptive (ignoringNonExistentComparedFields)

I've maintained the test structure with GIVEN/WHEN/THEN pattern and made sure all tests verify both success and failure cases. Let me know if you'd like me to make any additional changes!

joel-costigliola:
Integrated thanks @kssumin !

PR #3847, opened 2025-06-04
improve isThrown() method error msg

### LINKED ISSUE #3845: Mention expected `Throwable` type in failure message for `ThrowableTypeAssert.isThrownBy()`
#### Status Quo

In JUnit, we have an assertion similar to the following.

```java
assertThatExceptionOfType(JUnitException.class)
		.isThrownBy(() -> predicates.looksLikeIntendedTestClass(TestCase.class))
		.withMessage("...");
```

When that fails to throw a `JUnitException`, we get the following failure message in the `AssertionError`.

```
Expecting code to raise a throwable.
```

#### Expected Behavior

For the previous example, I would expect to get the following failure message in the `AssertionError` (or similar).

```
Expected code to throw org.junit.platform.commons.JUnitException.
```

At the very least, I would expect that the "expected type of exception" (`org.junit.platform.commons.JUnitException` in this case) is included in the failure message.

simhani1:
#### Description

This PR improves the clarity of error messages when no exception is thrown by a lambda passed to isThrownBy.

Previously, the failure message simply stated that “code should raise a throwable,” which could be vague in some contexts.
Now, the message explicitly mentions the expected exception type, e.g.:

```
Expecting code to throw java.lang.IllegalArgumentException, but no exception was thrown.
```

This change enhances the developer experience by providing more precise feedback during assertion failures, especially when working with soft assertions or parameterized exception checks.

I also updated related tests to align with the new message format.

#### Check List:
* Fixes #3845
* Unit tests : YES / NO / NA

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java
@@ -59,7 +59,9 @@ public abstract class AbstractThrowableAssert<SELF extends AbstractThrowableAsse
   protected void hasThrownExceptionOf(Class<? extends Throwable> expectedThrowableType) {
     if (actual == null) {
       String typeName = expectedThrowableType.getName();
-      throw Failures.instance().failure(info, new BasicErrorMessageFactory("%nExpecting code to throw %s, but no exception was thrown.".formatted(typeName)));
+      throw Failures.instance()
+                    .failure(info,
+                             new BasicErrorMessageFactory("%nExpecting code to throw %s, but no exception was thrown.".formatted(typeName)));
     }
     myself.isInstanceOf(expectedThrowableType);
   }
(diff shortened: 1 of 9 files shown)


simhani1:
While working on this change, I encountered a design concern regarding the following logic:

```java
public ThrowableAssertAlternative<T> isThrownBy(final ThrowingCallable throwingCallable) {
  Throwable throwable = Throwables.catchThrowable(throwingCallable);
  checkThrowableType(throwable);
  T castThrowable = (T) throwable;
  return buildThrowableTypeAssert(castThrowable).as(description);
}

protected void checkThrowableType(Throwable throwable) {
  assertThat(throwable).as(description).hasBeenThrown().isInstanceOf(expectedThrowableType);
}
```

Originally, hasBeenThrown() is used to check whether an exception was thrown, and it’s implemented in AbstractThrowableAssert. 

However, hasBeenThrown() only checks whether actual is null, which means it cannot include the expected exception type in the failure message — because actual itself is the thrown exception instance that would need to be checked.

```java
protected SELF hasBeenThrown() {
  if (actual == null) {
    throw Failures.instance().failure(info, new BasicErrorMessageFactory("%nExpecting code to raise a throwable."));
  }
  return myself;
}
```

So I make a new method `hasThrownExceptionOf(Class<? extends Throwable>)`, which accepts the expected exception type and produces a more informative failure message when no exception is thrown, such as:

```
Expecting code to throw java.lang.IllegalArgumentException, but no exception was thrown.
```

I’m open to suggestions if there’s a better way to include the expected exception type in the failure message.

joel-costigliola:
Integrated thanks @simhani1 !

PR #3854, opened 2025-06-14
Remove `extracting(Function)` from `Iterable` assertions

### LINKED ISSUE #1499: Ambiguous signature for `extracting()` when using Kotlin
#### Summary

When trying to use .extracting() with lambdas in kotlin, the call is ambigous and will not compile out-of-the-box. I think this is due to

```extracting(ThrowingExtractor<? super ELEMENT, V, EXCEPTION> extractor)```
and
```extracting(Function<? super ELEMENT, V> extractor)```

have the same effective signature because in kotlin, there are no checked exceptions (thank god).

#### Example

```kotlin
assertThat(highlights).hasSize(6).extracting(HighlightInfo::getText).doesNotContain("assertThat")
// nor does this work
assertThat(highlights).hasSize(6).extracting { it.text }.doesNotContain("assertThat")
```
I've got no idea on how to fix that in a backwards compatible manner other than to have a different method name for kotlin that won't clash or trying to add an extension that will cope with it.

PatrickWalter214:
In Kotlin, the call to extracting with a lambda for an iterable assert is ambiguous. This is due to the `extracting` method existing for `ThrowingExtractor` and `Function`, which have the same effective signature, as checked exceptions do not exist in Kotlin. We opt to continue using the extracting method for `ThrowingExtractor`, since it is the more flexible functional interface, as it supports that the lambda provided can throw an exception.

#### Check List:
* Fixes #1499 
* Unit tests : YES


assertj-core/src/main/java/org/assertj/core/extractor/ThrowingExtractors.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Copyright 2012-2025 the original author or authors.
+ */
+package org.assertj.core.extractor;
+
+import org.assertj.core.api.iterable.ThrowingExtractor;
+
+import java.util.function.Function;
+
+public class ThrowingExtractors {
```
scordio: I plan to refactor this area further, I want to keep new public APIs to the minimum, for the time being. For this reason, I'll convert the methods in this class to factory methods directly in `ThrowingExtractor`.

assertj-tests/assertj-integration-tests/assertj-core-kotlin/src/test/kotlin/org/assertj/tests/core/kotlin/api/Assertions_assertThat_Test.kt
scordio: I recently restructured the Kotlin tests a bit, so I'll move this into a separate class.

scordio (commented):
Many thanks @PatrickWalter214 for your efforts, and sorry for such a big delay!

There are a couple of things I'd like to do differently, but I'm taking care of them directly.

assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
```diff
@@ -1288,42 +1288,6 @@ public AbstractListAssert<?, List<? extends Tuple>, Tuple, ObjectAssert<Tuple>>
...
-   *
-   * Note that the order of extracted property/field values is consistent with the iteration order of the Iterable under
-   * test, for example if it's a {@link HashSet}, you won't be able to make any assumptions on the extracted values order.
-   *
-   * @param <V> the type of elements extracted.
-   * @param extractor the object transforming input object to desired one
-   * @return a new assertion object whose object under test is the list of values extracted
-   */
-  @CheckReturnValue
-  public <V> AbstractListAssert<?, List<? extends V>, V, ObjectAssert<V>> extracting(Function<? super ELEMENT, V> extractor) {
```
scordio: As discussed today with @filiphr, removing the overload accepting `Function` will break compatibility with all the existing `Function` implementations that cannot be easily transformed into a `ThrowingExtractor`, such as those coming from third parties.

A better strategy is likely to keep `extracting(Function)` and add to `ThrowingExtractor` factory methods that ease wrapping code that throws exceptions, similar to [how the Spring Framework does](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/function/ThrowingFunction.html#of(org.springframework.util.function.ThrowingFunction)).

I'll analyze the topic again and come up with a better proposal.

Issue #3354, opened 2024-02-03
Checking compared fields existence: handle containers 


joel-costigliola:
#### Summary

Checking `person.children.name` is not yet done (we ignore containers for the time being), in this example we need to get the first child and check if its class has a `name` field

We need to do the same for these containers type:

- [ ] array
- [ ] iterable
- [ ] optional 
- [ ] atomic reference.

chadongmin:
@joel-costigliola 

I've created a pull request to address this issue: https://github.com/assertj/assertj/pull/3857

The proposed solution refactors the validation logic to correctly unwrap various container types and ensures that an `IllegalArgumentException` is thrown as expected. I've also added a comprehensive set of tests to cover these cases.

I did my best to follow the contribution guidelines, and I am ready to make any necessary changes. Looking forward to your feedback.

Thank you for your great work on this project!

joel-costigliola:
Thanks @chadongmin, I'll review the PR later on this week (time permitting)

PR #3857, opened 2025-06-16
Handle non-existent field check in recursive comparison containers

chadongmin:
#### Check List:
* Fixes [#3354](https://github.com/assertj/assertj/issues/3354)
* Unit tests : YES
* PR meets the [contributing guidelines](https://github.com/assertj/assertj/blob/main/CONTRIBUTING.md): YES

---

## 📝 Description

This pull request resolves an issue where the recursive comparison did not throw an exception when checking for a non-existent field within various container or wrapper types.

### The Problem

When using `usingRecursiveComparison().comparingOnlyFields()`, specifying a path to a field that does not exist inside a container (e.g., `List`, `Set`, `Array`, `Optional`, `AtomicReference`) would silently pass instead of throwing an `IllegalArgumentException`.

For example, given a `Player` class with only a `name` field, the following assertion would incorrectly pass:
```java
// This test would pass, which is wrong.
assertThat(teamWithPlayers)
  .usingRecursiveComparison()
  .comparingOnlyFields("players.salary") // 'salary' does not exist on Player
  .isEqualTo(anotherTeam);
```
This was caused by the `checkComparedFieldExists` method in `RecursiveComparisonConfiguration`, which stopped its validation path as soon as it encountered a container, without inspecting the elements within it.

### The Solution

This PR refactors the `checkComparedFieldExists` method to correctly validate field paths that traverse through containers.

1.  **Refactored `checkComparedFieldExists` in `RecursiveComparisonConfiguration.java`**:
    - The `while` loop logic was restructured using an `if-else` block to clearly distinguish between container and non-container nodes.
    - When a container node is detected via `isContainer()`, it no longer halts the check.

2.  **Added `unwrapContainer` Helper Method**:
    - A new private helper method, `unwrapContainer`, has been introduced to handle the logic of extracting a sample element from various container types.
    - It currently supports `Optional`, `AtomicReference`, `Iterable`, and `Array`.
    - This separation of concerns improves the readability and maintainability of the main validation method.

The new logic ensures that the validation continues by "unwrapping" the container, taking the first element as a sample, and then proceeding to check for the existence of the subsequent field in the path on that element. This results in the expected `IllegalArgumentException` being thrown upfront.

## ✅ Testing

A new test class, `RecursiveComparison_NonExistentField_Test.java`, has been added to verify the fix.

-   The test class includes nested tests for `Iterable` (List/Set), `Array`, `Optional`, and `AtomicReference` to ensure broad coverage.
-   Each test confirms that an `IllegalArgumentException` is now correctly thrown when a non-existent field within these containers is specified for comparison.
-   The tests adhere to the project's contribution guidelines, including `package-private` visibility and underscore-based naming conventions.

With these changes, all new tests pass, confirming the bug is resolved.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonConfiguration.java
@@ -25,8 +25,10 @@ import static org.assertj.core.presentation.StandardRepresentation.STANDARD_REPR
 import static org.assertj.core.util.Lists.list;
 import static org.assertj.core.util.Sets.newLinkedHashSet;
 
+import java.lang.reflect.Array;
 import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
@@ -34,6 +36,7 @@ import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiPredicate;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


chadongmin:
Hi @scordio 

I've submitted this PR to fix issue #3354. Could you please take a look when you have a moment?

I'm happy to make any changes based on your feedback.

Thanks!

joel-costigliola:
Integrated thanks @chadongmin !

PR #3871, opened 2025-07-25
add new assertAny method to Assertions to verify that at least one of…

### LINKED ISSUE #3804: SoftAssertions -> Pass Test if any of the assertion is passed
#### Feature summary
I have a scenario in which I would like to pass the test if any of the assertion within a set of SoftAssertions passes. I could not find a way in which AssertJ has this functionality using SoftAssertions.

For example, below code will check all the assertions, and fail in the end if any of the assertion is failed.

```java
int age = 19;
int birthYear = 2025;
SoftAssertions.assertSoftly(sa -> {
    sa.assertThat(age).as("Age").isGreaterThanOrEqualTo(18);
    sa.assertThat(birthYear).as("Birth Year").isGreaterThanOrEqualTo(2005);
});
```

#### Example

What I want is exact opposite of above that if any of the assertion is passed, then the test should pass. Something. like below which checks all the assertions in end, and then passes the test if any of the assertion is passed. I understand this is possible via combining conditions, but this would make the tests a bit complicated, so a possibility to achieve in a similiar way would be helpful like adding a new `assertAny()`
An optional descriptive message could be outputted to warn the user if any assertion is failed, or more optimized way would be to just pass the test as soon as any of the assetion within this set has passed without continuing through rest.

```java
int age = 19;
int birthYear = 2025;
SoftAssertions.assertAny(sa -> {
    sa.assertThat(age).as("Age").isGreaterThanOrEqualTo(18);
    sa.assertThat(birthYear).as("Birth Year").isGreaterThanOrEqualTo(2005);
});
```

alihmzyv:
… the executables does not fail. closes #3804.

#### Check List:
* Fixes #3804
* Unit tests : YES

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/Assertions.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/Assertions.java
@@ -12,9 +12,11 @@
  */
 package org.assertj.core.api;
 
+import static java.util.Arrays.stream;
 import static java.util.Objects.requireNonNull;
 import static org.assertj.core.configuration.ConfigurationProvider.CONFIGURATION_PROVIDER;
 import static org.assertj.core.data.Percentage.withPercentage;
+import static org.assertj.core.util.Preconditions.checkArgument;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


alihmzyv:
@scordio kind reminder to review. Thanks

joel-costigliola:
Integrated thanks @alihmzyv !

PR #3876, opened 2025-07-29
Issue 3867

### LINKED ISSUE #3867: Incorrect example in documentation
https://github.com/assertj/assertj/blob/bc67ddb83ed9798731a3a1dfd6c572d83954e01e/assertj-core/src/main/java/org/assertj/core/api/filter/Filters.java#L235
`with(String fieldName)` method returns this:
```
  public Filters<E> with(String propertyOrFieldName) {
    validatePropertyOrFieldName(propertyOrFieldName);
    propertyOrFieldNameToFilterOn = propertyOrFieldName;
    return this;
  }
```

But there is no `greaterThan` in Filters class, so you cannot use an api from this example simply because it just doesn't exist at all.

https://github.com/assertj/assertj/blob/bc67ddb83ed9798731a3a1dfd6c572d83954e01e/assertj-core/src/main/java/org/assertj/core/api/filter/Filters.java#L35

leukim:
I have rewritten the Filters javadoc to reflect how the class should be actually used. As stated in the issue, the example was wrong (and not only in the actual API usage, but the explanation of the class was rather confusing).

I have taken the opportunity to add a matches(Condition) method that allows to filter the selected field (using'with(String)') on any arbitrary condition, and not just equals/notEquals. Since matches is a generic case to those, I have also refactored them to use matches(Condition) internally to avoid some duplication.

I have also added a new test class to test this new behaviour (and make sure that the code in the Javadoc is now correct!).

#### Check List:
* Fixes #3867 
* Unit tests : YES

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/filter/Filters.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/filter/Filters.java
@@ -29,26 +29,49 @@ import org.assertj.core.util.introspection.PropertyOrFieldSupport;
 /**
  * Filters the elements of a given <code>{@link Iterable}</code> or array according to the specified filter criteria.
  * <p>
- * Filter criteria can be expressed either by a {@link Condition} or a pseudo filter language on elements properties.
+ *  Filter criteria can be expressed using {@link Condition} in two different ways:
+ *  <ul>
+ *    <li>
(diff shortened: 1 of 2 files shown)


joel-costigliola:
Integrated thanks @leukim ! 

I simply renamed `matches` to `matching` for better readability, fixed the javadoc (can't use <> in code examples, it has to be &lt; and &gt;) and a few minor improvements.

PR #3951, opened 2025-09-22
Add missing export for `org.assertj.core.annotation`

### LINKED ISSUE #3950: Package org.assertj.core.annotation is not exported in module-info
### Context
`org.assertj.core.util.CanIgnoreReturnValue` was deprecated in favor of `org.assertj.core.annotation.CanIgnoreReturnValue`

* assertj core version: 3.27.4
* java version: 21.0.3
* test framework version: 5.13.4

### Problem
The new annotation cannot be found when working with JPMS because the `org.assertj.core.annotation` is not exported.

### Solution
Export package `org.assertj.core.annotation` in module-info if it is intended to be public.

duponter:
#### Check List:
* Fixes #3950
* Unit tests : TBD

### DIFF
--- a/assertj-core/src/main/java9/module-info.java
+++ b/assertj-core/src/main/java9/module-info.java
@@ -12,6 +12,7 @@
  */
 module org.assertj.core {
   // AssertJ Core's package API
+  exports org.assertj.core.annotation;
   exports org.assertj.core.annotations;
   exports org.assertj.core.api;
   exports org.assertj.core.api.exception;


scordio:
I sketched the following:

```java
class ModuleDescriptor_Test {

  @Test
  void test() {
    // GIVEN
    Module underTest = ModuleLayer.boot().findModule("org.assertj.core").orElseThrow();
    // WHEN
    Set<String> packages = underTest.getPackages().stream().filter(p -> !p.contains("internal")).collect(toSet());
    // THEN
    then(packages).allMatch(underTest::isExported);
  }

}
```

It can be executed on `main` as module descriptors are handled appropriately, but we don't have a chance on `3.x` as everything runs on the classpath and `findModule("org.assertj.core")` returns an empty optional.

I'm going ahead with the merge and will add the test above on `main` only, so that we can spot such issues when forward-porting changes.

scordio:
Thanks for your first contribution, @duponter!

duponter:
You're welcome. But I should thank you for maintaining this fine library and the swift response to my bug ticket.

PR #3957, opened 2025-10-01
feat: Add `assertj-protobuf` module

### LINKED ISSUE #3736: Add assertj-protobuf to assertj ecosystem.
#### Feature summary

Add assertj-protobuf to assertj ecosystem. This would benefit Java developers having to work with Protobuf and develop GRPC services.

Assertj is largely similar to [Google Truth's assertion library](https://truth.dev/). One of the extensions Google Truth has that assertj does not have is for protobuf. See [here](https://truth.dev/protobufs.html).

Curiostack has forked Google Truth's protobuf library and adapted it for AssertJ. See [here](https://github.com/curioswitch/curiostack/tree/master/common/testing/assertj-protobuf). Their library is under an MIT License.
Unfortunately, that library is now abandonware, as the owners of the library are now a Go shop rather than a Java shop.

Is it possible for AssertJ to take over this fork? If so, should it be a separate module like assertj-vavr? Or should it live in the `assertj / assertj` repo like assertj-guava?

#### Example

The following would be useufl

```java
MyProto expected = MyProto.newBuilder().addBar(3).addBar(5).setFoo("qux").build();
MyProto actual =   MyProto.newBuilder().addBar(5).addBar(3).setFoo("quizzux").build();
assertThat(actual).ignoringRepeatedFieldOrder().isEqualTo(expected);
```

belljun3395:
# AssertJ Protobuf Module Implementation

* Closes #3736

## Implemented Features

A new Maven module providing:

#### Main Classes

**`org.assertj.protobuf.api.Assertions`**
- Entry point for AssertJ Protobuf
- Provides `assertThat(Message)` static method

**`org.assertj.protobuf.api.MessageAssert<T extends Message>`**
- Main assertion class for Protobuf messages
- Extends `AbstractAssert` following AssertJ standards

**`org.assertj.protobuf.internal.MessageDifferencer`**
- Internal message comparison logic
- Supports repeated fields, nested messages, field exclusion, etc.

#### Provided Methods

```java
// Basic comparison
assertThat(actual).isEqualTo(expected);

// Ignore repeated field order
assertThat(actual).ignoringRepeatedFieldOrder().isEqualTo(expected);

// Exclude specific fields
assertThat(actual).ignoringFields("timestamp").isEqualTo(expected);

// Partial comparison (only expected fields)
assertThat(actual).comparingExpectedFieldsOnly().isEqualTo(expected);

// Field presence checks
assertThat(actual).hasField("name");
assertThat(actual).doesNotHaveField("deletedAt");
```

### Dependency Management
```xml
<!-- Compile -->
<dependency>
  <groupId>org.assertj</groupId>
  <artifactId>assertj-core</artifactId>
  <version>${project.version}</version>
</dependency>

<!-- Provided - users choose version -->
<dependency>
  <groupId>com.google.protobuf</groupId>
  <artifactId>protobuf-java</artifactId>
  <version>4.29.3</version>
  <scope>provided</scope>
</dependency>
```

#### Repeated Field Handling
Protobuf's repeated fields don't support `hasField()`, requiring special handling:

```java
if (field.isRepeated()) {
  // Use getField() instead of hasField()
  actualValue = actual.getField(field);
  expectedValue = expected.getField(field);
  
  // Treat empty lists as null
  List<Object> actualList = (List<Object>) actualValue;
  if (actualList.isEmpty()) actualValue = null;
}
```

#### AS_SET Mode (Order-Insensitive)
Algorithm for comparing repeated fields like sets:

1. For each item in expected list
2. Find matching item in actual list
3. Remove matched item from actual list (prevent duplicates)
4. Success if all items matched

#### Recursive Message Comparison
Nested messages are compared recursively, tracking field paths:
- `"field_name"` → `"nested.field_name"` → `"nested.child.field_name"`

### Basic Usage
```java
import static org.assertj.protobuf.api.Assertions.assertThat;

@Test
void should_compare_protobuf_messages() {
  // GIVEN
  UserProto expected = UserProto.newBuilder()
      .setName("John")
      .setAge(30)
      .build();
  
  UserProto actual = userService.getUser(userId);
  
  // WHEN/THEN
  assertThat(actual).isEqualTo(expected);
}
```

### Advanced Patterns
```java
@Test
void should_ignore_timestamp_and_repeated_order() {
  // GIVEN
  OrderProto expected = OrderProto.newBuilder()
      .setOrderId("ORD-123")
      .addItems("ITEM-1")
      .addItems("ITEM-2")
      .build();
  
  OrderProto actual = orderService.getOrder("ORD-123");
  
  // WHEN/THEN
  assertThat(actual)
      .ignoringFields("createdAt", "updatedAt")
      .ignoringRepeatedFieldOrder()
      .isEqualTo(expected);
}
```

### Build
```bash
# Build protobuf module only
./mvnw clean install -pl assertj-protobuf

# Include tests
./mvnw clean install -pl assertj-protobuf,assertj-tests/assertj-integration-tests/assertj-protobuf-tests
```

### Run Tests
```bash
./mvnw test -pl assertj-tests/assertj-integration-tests/assertj-protobuf-tests
```


genuss:
Hi everyone! I'm very much interested in the feature and was glad to see a PR. I tested it on a real project of mine and would like to share some results if you are interested.

First of all, it works as expected, I didn't notice any major issues. I only found two interesting points.

1. `MessageDifferencer` doesn't distinguish between protobuf maps and plain repeated fields. This causes assertions to fail when order of keys doesn't match. This could be fixed by extending the condition in `compareRepeatedField` to `repeatedFieldComparison == RepeatedFieldComparison.AS_SET || field.isMapField()`. There may be a specific separate path for maps, but this small addition is IMHO a good start.
2. When using `comparingExpectedFieldsOnly` primitive fields set to default values aren't checked at all. So the user choice becomes tricky: either you compare all expected+actual fields or you can't assert default values. One the one hand it's the whole point of using `comparingExpectedFieldsOnly`, but on the other it's a bit tedious to exclude fields manually. Particularly this problem arises with booleans: you can't check that the field is `false` at all.
Personally I'm not sure what is the best way to solve it:
* Create a scope `MessageDifferencer.Scope.PARTIAL_WITH_PRIMITIVES`?
* Handle bool type specifically?

belljun3395:
@genuss Thanks for the feedback. The first point is fixed, the second one is an API decision I would rather align on with the maintainers first.

## 1. Map fields

You were right, I was able to reproduce the issue: map fields are reflected as repeated `MapEntry` messages and were going through the ordinary index-based repeated field comparison, so two maps with identical entries compared unequal whenever their entries happened to be iterated in a different order.

I took your suggestion a step further. Routing maps through the existing `AS_SET` path does fix the ordering, but it still matches whole entries, so differences are reported in terms of the underlying `MapEntry` (`expected to contain <key: "env" value: "prod"> but was not found`). Maps now get a dedicated path that rebuilds both reflected entry lists into maps and compares them by key:

- map entry order is always ignored, independently of `ignoringRepeatedFieldOrder()` — this matches ProtoTruth, where map fields "are always compared by key";
- value mismatches are reported by key: `Map field <labels[env]>: expected <prod> but was <dev>`;
- missing and unexpected keys get their own diagnostics;
- message valued entries are compared recursively.

Covered by `MessageAssert_isEqualTo_with_map_fields_Test`.

## 2. Default valued fields in `comparingExpectedFieldsOnly()`

All the examples below use this message, and a test that only cares about one thing: the order was not paid.

```protobuf
message Order {
  string id = 1;
  bool paid = 2;             // scalar
  int32 retry_count = 3;     // scalar
  Address shipping = 4;      // message
  repeated Item items = 5;   // repeated
}
```

`paid` is a proto3 singular scalar declared without `optional`, so it uses implicit presence: `setPaid(false)` and never calling the setter produce indistinguishable messages. As a result `getAllFields()` — which is what `Scope.PARTIAL` derives the fields to compare from — does not report `paid` at all, and the bug goes through:

```java
Order expected = Order.newBuilder().setId("A-1").setPaid(false).build();
Order actual = orderService.get("A-1");  // paid is wrongly true, and retry_count, shipping, items are set

assertThat(actual).comparingExpectedFieldsOnly().isEqualTo(expected);  // passes, paid is never compared
```

Fields with explicit presence, such as proto3 `optional` fields, already work today. The unresolved case is singular non-message fields for which `FieldDescriptor.hasPresence()` is false. I see three ways to go.

**Option 0, accept the limitation and document it.** This is what ProtoTruth does, with the very same method name:

> Fields not set in the expected proto(s) are ignored. In particular, proto3 fields which have their default values are ignored, as these are indistinguishable from unset fields. If you want to assert that a proto3 message has certain fields with default values, you cannot use this method.
> — [`ProtoFluentAssertion#comparingExpectedFieldsOnly()`](https://truth.dev/api/latest/com/google/common/truth/extensions/proto/ProtoFluentAssertion.html)

It is a legitimate answer, and the cheapest one: no new API, and we stay aligned with the reference implementation. The user falls back to a full comparison, and lists everything the test does not care about:

```java
assertThat(actual).ignoringFields("retry_count", "shipping", "items")
                  .isEqualTo(expected);
```

**Option A, the `Scope.PARTIAL_WITH_PRIMITIVES` you suggested**, compares every scalar by value, including the ones the test never mentioned. Message and repeated fields stay partial, so `shipping` and `items` no longer have to be listed. But `retry_count` does: the test says nothing about it, so expected holds its default `0` — not as an assertion, simply as a consequence of saying nothing — and an actual order that legitimately retried 3 times fails with `expected <0> but was <3>`.

```java
assertThat(actual).comparingExpectedFieldsAndPrimitives()
                  .ignoringFields("retry_count")   // shipping and items no longer need listing
                  .isEqualTo(expected);
```

So the comparison silently stops being partial for scalars: every scalar the test did not set becomes an implicit assertion that the actual message holds the default value, and the more scalars the message has, the closer this gets to option 0.

**Option B opts in field by field, mirroring `ignoringFields()`:**

```java
assertThat(actual).comparingExpectedFieldsOnly()
                  .includingFields("paid")
                  .isEqualTo(expected);
```

`paid` is compared despite holding its default value, while `retry_count`, `shipping` and `items` stay out — nothing has to be listed. The partial comparison stays partial, and the test states exactly what it asserts. The cost is spelling the field name out.

These are the options I could come up with, and I am happy to implement whichever direction you prefer.

@scordio what direction would you like me to take?

genuss:
@belljun3395 , just wanted to say hello and report that you fix works like a charm!

As for the second question, I'm actually OK with all variants, but find Option B most promising.

PR #4004, opened 2025-11-04
Added withFormatter() for Duration assert

### LINKED ISSUE #3976: Support *easy* custom Representation for Duration objects in assertion messages
#### Feature summary
When using assertions on `Duration`s, the actual and expected values are not always formatted in a user-friendly way. For example:
```
Expecting actual:
  0.002512594S
to be less than or equal to:
  0.001S 
```

Thinking it would be nice to be able to provide my own format pattern for those values, I started digging and found `withRepresentation()`. I was encouraged that I could easily reach my goal 🙌 , until I saw that `StandardRepresentation.toStringof(Duration)` is `private`.  ☹️  It's pretty much the only overload of `toStringOf()` that's not overridable, is there a particular reason?
Can it be made protected so it can be overridden like all the other overloads?

erizzo:
#### Check List:
* Fixes #3976 
* Unit tests : YES

This PR adds a method to AbstractDurationAssert to make it easy to override the formatting of `Duration` objects in test failure messages.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractDurationAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractDurationAssert.java
@@ -22,13 +22,16 @@ import static org.assertj.core.error.ShouldHaveDuration.shouldHaveSeconds;
 import static org.assertj.core.util.Preconditions.checkArgument;
 
 import java.time.Duration;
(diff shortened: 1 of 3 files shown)


joel-costigliola:
Integrated thanks @erizzo !

PR #4048, opened 2025-12-03
feat: honor maxStackTraceElementsDisplayed in soft assertions

### LINKED ISSUE #4040: Include more stack frames in soft assertion failures
#### Feature summary

Currently, a soft assertion failure (i.e. the message of an `AssertJMultipleFailuresError`) only contains the topmost stack frame of each individual failure. While this is often sufficient, there are cases where it isn't - e.g. when using a custom assertion via the `softly.check(() -> ...)` idiom, only the stack frame within the custom assertion method is visible, not the one from where it was called.

Therefore, please increase the number of stack frames shown, or make it configurable (e.g. through an instance method in `AbstractSoftAssertions`, or a static method in some suitable class).

chanwonlee:
#### Check List:
* Fixes #4040
* Unit tests : YES

## Description
Changed the hardcoded `.limit(5)` in `Throwables.ERROR_DESCRIPTION_EXTRACTOR` to honor the `StandardRepresentation.getMaxStackTraceElementsDisplayed()` setting.

Now users can control the number of stack trace elements displayed in soft assertion failures by calling `Assertions.setMaxStackTraceElementsDisplayed(n)`.

### Question
@joel-costigliola When maxStackTraceElementsDisplayed is set to 0, should I display a special message instead of showing an empty stack trace section?
Right now it prints "cause first 0 stack trace elements:", which may look a bit odd.

### DIFF
--- a/assertj-core/src/test/java/org/assertj/core/util/Throwables_describeErrors_Test.java
+++ b/assertj-core/src/test/java/org/assertj/core/util/Throwables_describeErrors_Test.java
@@ -50,7 +50,6 @@ class Throwables_describeErrors_Test {
     then(descriptions).hasSize(1);
     String description = descriptions.get(0);
     then(description).contains("cause first 10 stack trace elements:");
-    System.out.println(description);
     long stackTraceElementCount = description.lines()
                                              .filter(line -> line.trim().startsWith("at "))
                                              .count();


chanwonlee:
@joel-costigliola Could you please take a look when you have a chance? Thank you!

joel-costigliola:
@chanwonlee I'm actually on it and extending the changes more (you'll get the full credits :slightly_smiling_face: )

joel-costigliola:
Integrated thanks @chanwonlee

Issue #3453, opened 2024-05-02
Deprecate `assertThat(Iterable, AssertFactory)` and `assertThat(Iterable, Class)`


scordio:
Both [`assertThat(Iterable, AssertFactory)`](https://www.javadoc.io/doc/org.assertj/assertj-core/latest/org/assertj/core/api/Assertions.html#assertThat(java.lang.Iterable,org.assertj.core.api.AssertFactory)) and [`assertThat(Iterable, Class)`](https://www.javadoc.io/doc/org.assertj/assertj-core/latest/org/assertj/core/api/Assertions.html#assertThat(ACTUAL,java.lang.Class)) have been added in version 3.5.0 to aid the creation of type-specific assertions for the elements of an `Iterable` instance.

Today, it's possible to write the following:

```java
Iterable<String> hobbits = Set.of("frodo", "sam", "Pippin");

// with AssertFactory
assertThat(hobbits, StringAssert::new).first()
                                      .startsWith("fro")
                                      .endsWith("do");

// with Class
assertThat(hobbits, StringAssert.class).first()
                                       .startsWith("fro")
                                       .endsWith("do");
```

However, with the introduction of `InstanceOfAssertFactory` in version 3.13.0 and the corresponding `first(InstanceOfAssertFactory)` variant in version 3.14.0, the above can be rewritten as:

```java
assertThat(hobbits).first(STRING) // static import of InstanceOfAssertFactories.STRING
                   .startsWith("fro")
                   .endsWith("do");
```

The main advantage of the latter is easier discoverability therefore we should deprecate the two `assertThat` entry points.

PR #4076, opened 2025-12-22
Add `withElementAssert` to `Iterable` assertions

### LINKED ISSUE #3490: Unable to construct iterable assertions across a specific type easily
Within the new AssertJ 3.26.0 release, I can see that the factory based navigable assertion for iterables has been deprecated internally, along with other mechanisms to do the same thing.

This raises an issue for me, as I extend the AssertJ API and provide additional custom assertions that wrap the assertj ones.

In this case, I have a requirement to be able to return an assertion object across a list of strings for the user to be able to perform further assertions on. The current implementation looks like this: https://github.com/ascopes/java-compiler-testing/blob/1a92ef59df49cf2f45ae83a109569be006b08df2/java-compiler-testing/src/main/java/io/github/ascopes/jct/assertions/JctCompilationAssert.java#L70.

Unfortunately with these APIs being deprecated, I can not find a way to return an iterable assert across strings that automatically expand to string assertions. I can see new API methods such as .first(AssertFactory) and last(AssertFactory), but no way to instruct AssertJ how to expose assertions on internal objects anymore such that the user does not need to explicitly cast these assertions with an explicit InstanceOfAssertFactory object.

Is there any advice for how to migrate off of this, or is this something that will need to be a breaking API change on my side?

scordio:
This PR introduces a new `withElementAssert` configuration method to the `Iterable` assertion and related subclasses.

This is a replacement for the factory-based navigable assertions deprecated in version [`3.26.0`](https://github.com/assertj/assertj/releases/tag/assertj-build-3.26.0), as custom AssertJ extensions lost the ability to get type-specific assertions when navigating to specific elements via methods like `first()`, `last()`, `element(int)`, and `singleElement()`. Although it's still possible to use the overloads with the `InstanceOfAssertFactory` parameter, end users must call the appropriate methods, and extension authors can no longer hide this detail in their implementations.

The new `withElementAssert` method allows extension authors to configure the assertion type to be returned when navigating to specific elements before returning the assertion object to the end user.

This change also completes the work started in #3453, by deprecating `FactoryBasedNavigableIterableAssert` and `FactoryBasedNavigableListAssert`.

* Closes #3490

### Open points

- [x] https://github.com/assertj/assertj/issues/3490#issuecomment-2180743115
- [x] Update deprecations introduced in ad950f03e4cae51e8b01db069d69cdaed8ef2fff to mention it as an alternative
- [x] Check compatibility with `InstanceOfAssertFactory` (https://github.com/assertj/assertj/pull/4076#discussion_r2678591517)

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/Assertions.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/Assertions.java
@@ -685,7 +685,7 @@ public class Assertions implements InstanceOfAssertFactories {
    * <p>
    * Otherwise, the element assertion factory can be configured on the assertion object via
    * {@link AbstractIterableAssert#withElementAssert(AssertFactory) withElementAssert}:
-   * <pre><code class='java'>assertThat(hobbits).withElementAssert(StringAssert::new)
+   * <pre><code class='java'>assertThat(hobbits).withElementAssert(Assertions::assertThat)
    *                   .first()
    *                   .startsWith("fro")
    *                   .endsWith("do");</code></pre>
@@ -746,7 +746,7 @@ public class Assertions implements InstanceOfAssertFactories {
    * <p>
    * Otherwise, the element assertion factory can be configured on the assertion object via
    * {@link AbstractIterableAssert#withElementAssert(AssertFactory) withElementAssert}:
-   * <pre><code class='java'>assertThat(hobbits).withElementAssert(StringAssert::new)
+   * <pre><code class='java'>assertThat(hobbits).withElementAssert(Assertions::assertThat)
    *                   .first()
    *                   .startsWith("fro")
    *                   .endsWith("do");</code></pre>
@@ -815,7 +815,7 @@ public class Assertions implements InstanceOfAssertFactories {
    * <p>
    * Otherwise, the element assertion factory can be configured on the assertion object via
    * {@link AbstractListAssert#withElementAssert(AssertFactory) withElementAssert}:
-   * <pre><code class='java'>assertThat(hobbits).withElementAssert(StringAssert::new)
+   * <pre><code class='java'>assertThat(hobbits).withElementAssert(Assertions::assertThat)
    *                   .first()
    *                   .startsWith("fro")
    *                   .endsWith("do");</code></pre>
@@ -876,7 +876,7 @@ public class Assertions implements InstanceOfAssertFactories {
    * <p>
    * Otherwise, the element assertion factory can be configured on the assertion object via
    * {@link AbstractListAssert#withElementAssert(AssertFactory) withElementAssert}:
-   * <pre><code class='java'>assertThat(hobbits).withElementAssert(StringAssert::new)
+   * <pre><code class='java'>assertThat(hobbits).withElementAssert(Assertions::assertThat)
    *                   .first()
    *                   .startsWith("fro")
    *                   .endsWith("do");</code></pre>
--- a/assertj-core/src/main/java/org/assertj/core/api/AssertionsForInterfaceTypes.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AssertionsForInterfaceTypes.java
@@ -386,7 +386,7 @@ public class AssertionsForInterfaceTypes extends AssertionsForClassTypes {
    * <p>
    * Otherwise, the element assertion factory can be configured on the assertion object via
    * {@link AbstractIterableAssert#withElementAssert(AssertFactory) withElementAssert}:
-   * <pre><code class='java'>assertThat(hobbits).withElementAssert(StringAssert::new)
+   * <pre><code class='java'>assertThat(hobbits).withElementAssert(Assertions::assertThat)
    *                   .first()
    *                   .startsWith("fro")
    *                   .endsWith("do");</code></pre>
@@ -419,7 +419,7 @@ public class AssertionsForInterfaceTypes extends AssertionsForClassTypes {
    * <p>
    * Otherwise, the element assertion factory can be configured on the assertion object via
    * {@link AbstractIterableAssert#withElementAssert(AssertFactory) withElementAssert}:
-   * <pre><code class='java'>assertThat(hobbits).withElementAssert(StringAssert::new)
+   * <pre><code class='java'>assertThat(hobbits).withElementAssert(Assertions::assertThat)
    *                   .first()
    *                   .startsWith("fro")
    *                   .endsWith("do");</code></pre>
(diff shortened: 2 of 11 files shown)


scordio:
> * [Unable to construct iterable assertions across a specific type easily #3490 (comment)](https://github.com/assertj/assertj/issues/3490#issuecomment-2180743115)

The issue can be reproduced with the following test:

```java
  @Test
  void should_allow_chaining_element_specific_assertions_with_factory_returning_assertion_superclass() {
    // GIVEN
    List<String> actual = List.of("Homer", "Marge");
    AssertFactory<String, AbstractCharSequenceAssert<?, String>> assertFactory = StringAssert::new;
    // WHEN/THEN
    assertThat(actual).withElementAssert(assertFactory) // does not compile
                      .hasSize(2)
                      .first()
                      .startsWith("Hom");
  }
```

assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
```diff
@@ -106,7 +106,7 @@
 public abstract class AbstractIterableAssert<SELF extends AbstractIterableAssert<SELF, ACTUAL, ELEMENT, ELEMENT_ASSERT>,
                                              ACTUAL extends Iterable<? extends ELEMENT>,
                                              ELEMENT,
-                                             ELEMENT_ASSERT extends AbstractAssert<ELEMENT_ASSERT, ELEMENT>>
+                                             ELEMENT_ASSERT extends AbstractAssert<? extends ELEMENT_ASSERT, ELEMENT>>
```
scordio, line 112: This is currently not compatible with `InstanceOfAssertFactory` instances.

Generally, I think the `AbstractAssert` upper bound on `ELEMENT_ASSERT` is too restrictive, though I haven't had time to prove that the two are related.

I checked 4d21a2dd, which initially introduced the upper bound, and couldn't find any strong reason to prefer `AbstractAssert` over `Assert`. @joel-costigliola do you happen to remember anything about it?

However, the change would be binary-incompatible, so we should ship it only in version 4.

assertj-core/src/main/java/org/assertj/core/api/AbstractCollectionAssert.java
```diff
@@ -131,4 +137,32 @@ private <E extends ELEMENT> Collection<E> emptyCollection() {
...
+    protected ELEMENT_ASSERT toAssert(ELEMENT value, String description) {
+      return assertFactory.createAssert(value).as(description);
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    protected SELF newAbstractIterableAssert(Iterable<? extends ELEMENT> iterable) {
+      return (SELF) new FactoryBasedCollectionAssert<>(newArrayList(iterable), assertFactory);
+    }
+
```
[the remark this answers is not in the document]
scordio: @copilot `FactoryBasedCollectionAssert` inherits from `AbstractCollectionAssert`, the latter implementing `withElementAssert` and returning a new `FactoryBasedCollectionAssert` instance.

The following test case works correctly:

```java
  @Test
  void test() {
    // GIVEN
    Collection<String> actual = List.of("Homer", "Marge");
    AssertFactory<String, StringAssert> assertFactory = StringAssert::new;
    // WHEN/THEN
    assertThat(actual).withElementAssert(assertFactory)
                      .withElementAssert(assertFactory)
                      .hasSize(2)
                      .first()
                      .startsWith("Hom");
  }
```

If the issue really exists, provide a test case that demonstrates it.

assertj-core/src/main/java/org/assertj/core/api/CollectionAssert.java
```diff
@@ -26,25 +26,39 @@
  *
  * @since 3.21.0
  */
-public class CollectionAssert<ELEMENT> extends
-    AbstractCollectionAssert<CollectionAssert<ELEMENT>, Collection<? extends ELEMENT>, ELEMENT, ObjectAssert<ELEMENT>> {
+public class CollectionAssert<SELF extends CollectionAssert<SELF, ACTUAL, ELEMENT, ELEMENT_ASSERT>,
+                              ACTUAL extends Collection<? extends ELEMENT>,
+                              ELEMENT,
+                              ELEMENT_ASSERT extends AbstractAssert<? extends ELEMENT_ASSERT, ELEMENT>>
+  extends AbstractCollectionAssert<SELF, ACTUAL, ELEMENT, ELEMENT_ASSERT> {
```
scordio: As expected, the changes at 61c4e258 are binary-incompatible, though they would be a cleaner implementation. Therefore, we'll address them in version 4.

I'll create a follow-up task for it.

assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
joel-costigliola: No recollection, but as I'm not a generic expert, I might have defined a too restrictive upper bound.
It is fine to relax it if you need that.

assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
```diff
@@ -126,12 +126,42 @@ protected AbstractIterableAssert(ACTUAL actual, Class<?> selfType) {
...
       SortedSet<ELEMENT> sortedSet = (SortedSet<ELEMENT>) actual;
       Comparator<? super ELEMENT> comparator = sortedSet.comparator();
       if (comparator != null) usingElementComparator(sortedSet.comparator());
     }
   }
 
+  /**
+   * Configures the {@link AssertFactory} used to create assertions whenever a navigation method is invoked.
+   * <p>
+   * This is generally more convenient for domain-specific assertions, so that users of the domain assertions
```
scordio:
```suggestion
   * This is generally more convenient for custom assertions, so that users of such assertions
```

assertj-core/src/main/java/org/assertj/core/api/CollectionAssert.java
scordio: Raised #4098.

assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
scordio: The incompatibility is actually correct, since any `InstanceOfAssertFactory` type is an `AssertFactory<Object, ASSERT>`, free from any constraints by design, while `withElementAssert` expects a factory that is aligned with the `ELEMENT` type of the `Iterable`.

PR #4089, opened 2026-01-06
Add keys() method to MapAssert for fluent key assertions

### LINKED ISSUE #4078: Provide containsAnyKeyOf() similar to containsAnyOf() for Map
#### Feature summary
It would be nice if `containsAnyKeyOf()` similar to `containsAnyOf()` for `Map` was provided.

#### Example

To achieve it now, `keySet()` could be used as follows:

```java
Map<SomeType, List<String>> map = Map.of(SomeType.TYPE_A, List.of("a", "aa"));
assertThat(map.keySet()).containsAnyOf(SomeType.TYPE_A, SomeType.TYPE_B);
```

It would be nice if it could be done as follows:

```java
Map<SomeType, List<String>> map = Map.of(SomeType.TYPE_A, List.of("a", "aa"));
assertThat(map).containsAnyKeyOf(SomeType.TYPE_A, SomeType.TYPE_B);
```

Ogu1208:
#### Check List:
* Fixes #4078
* Unit tests : YES

### DIFF
--- a/assertj-core/src/test/java/org/assertj/core/api/map/MapAssert_keys_Test.java
+++ b/assertj-core/src/test/java/org/assertj/core/api/map/MapAssert_keys_Test.java
@@ -15,9 +15,9 @@
  */
 package org.assertj.core.api.map;
 
-import static org.assertj.core.api.Assertions.from;
 import static org.assertj.core.api.BDDAssertions.then;
 import static org.assertj.core.api.BDDAssertions.thenNullPointerException;
+import static org.assertj.core.api.InstanceOfAssertFactories.STRING;
 
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


assertj-core/src/test/java/org/assertj/core/api/map/MapAssert_keys_Test.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  @Test
+  void should_return_collection_assertion_with_right_generic_type() {
+    // GIVEN
+    Map<String, String> map = Map.of("first", "one");
+    // WHEN/THEN
+    then(map).keys()
+             .singleElement()
+             .returns('f', from(s -> s.charAt(0)));
```
joel-costigliola: since `returns` is available for object assertions, this does not really check that we honor the generic type (String here), so let's use a String assertion like startsWihh

Ogu1208: @joel-costigliola Thanks for the feedback!
Updated to use `singleElement(STRING).startsWith("f")` to properly verify the generic type.

joel-costigliola:
Integrated thanks @Ogu1208 !

PR #4127, opened 2026-01-28
Fix ClassCastException in SoftAssertions with custom list assertions

### LINKED ISSUE #3797: Calling final methods through a custom assertion, which is inherited from AbstractListAssert, results in a java.lang.ClassCastException
**Describe the bug**
Calling final methods through a custom assertion, which is inherited from AbstractListAssert, results in a java.lang.ClassCastException. In fact, any call to final methods inherited from the underlying Abstract Assert class leads to a java.lang.ClassCastException.

Could you advise on any workaround or point out where I might be handling the custom assertion implementation incorrectly?

* assertj core version: 3.27.3
* java version: 1.17
* groovy version: 4.0.18
* test framework version: spock 2.3-groovy-4.0
* os (if relevant): MacOS/Linux

I've implemented an example project that describes the issue: 
https://github.com/AlexeyAkentyev/assertj-custom-assertions-example/blob/d13162c1537ec1f7ead34df42294a5d9640a7cce/src/test/groovy/org/example/custom/assertions/tests/DomainObjectsAssertionSpec.groovy#L136
 
Project contains example of domain related assertions.
Test fails with the error:
`Caused by: java.lang.ClassCastException: class [Ljava.lang.Object; cannot be cast to class [Lorg.example.custom.assertions.domain.DomainObject; ([Ljava.lang.Object; is in module java.base of loader 'bootstrap'; [Lorg.example.custom.assertions.domain.DomainObject; is in unnamed module of loader 'app')`

HuitaePark:
Modified `containsExactlyInAnyOrderForProxy` to accept a `Collection` instead of arrays. This prevents `ClassCastException` caused by strict array type checking in proxies.

- **Signature Change**: `ForProxy` method now takes `Collection`.
- **New Utility**: Added `IterableUtil.toArrayList` for safe conversion.
- **Test**: Added `SoftAssertions_proxied_AbstractListAssert_Test` to verify the fix.

I have applied this fix only to `containsExactlyInAnyOrder` as a proof of concept.
Please confirm if this direction is correct. Once approved, I will apply this pattern to all other affected methods (e.g.,`ForProxy`).

#### Check List:
* Fixes #3797 
* Unit tests : YES

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
@@ -32,6 +32,7 @@ import static org.assertj.core.internal.Iterables.byPassingAssertions;
 import static org.assertj.core.internal.TypeComparators.defaultTypeComparators;
 import static org.assertj.core.util.Arrays.isArray;
 import static org.assertj.core.util.IterableUtil.toArray;
+import static org.assertj.core.util.IterableUtil.toArrayList;
 import static org.assertj.core.util.Lists.newArrayList;
 import static org.assertj.core.util.Preconditions.checkArgument;
 import static org.assertj.core.util.Preconditions.checkNotNull;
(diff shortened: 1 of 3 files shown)


joel-costigliola:
thanks @HuitaePark, we'll wait for https://github.com/assertj/assertj/pull/4199 to be merged before integrating this one.

HuitaePark:
@joel-costigliola Understood. Please let me know if there's anything else I need to do!

joel-costigliola:
will do !

joel-costigliola:
Integrated thanks @HuitaePark !
I actually only kept the test as we removed the byte buddy proxying that was the issue root cause.

I checked that the test failed in commits previous to removing byte buddy and succeeds afterward.

HuitaePark:
Thanks for sharing the progress. I'm glad it worked out well!

Issue #3167, opened 2023-09-01
Better indentation of strings in assertion messages containing multi-line expected/actual


Bananeweizen:
#### Feature summary
Some of the String assertion messages indent both the actual and expected value for readability with 2 spaces, for example
https://github.com/assertj/assertj/blob/d02a738d7522dc58c54ca1b782fa50db9fcb67fc/assertj-core/src/main/java/org/assertj/core/error/ShouldBeEqualIgnoringWhitespace.java#L34

If actual and expected consist of multiple lines of text, that indentation can lead to more confusion than being helpful, because only the first line is indented, the others are not. Example output:
```
// isEqualToIgnoringWhitespace
Expecting actual:
  "foo
bar
baz"
to be equal to:
  "moreFoo
bar
baz"
when ignoring whitespace differences
```

I think it would be beneficial to
* either indent or not indent _every_ line of actual and expected
* eventually unify the indentation usage in all String/Charsequence related assertions, since that indentation is currently hard-coded in each sub class of the base error message class. Also it seems the base error message class has some capabilities around multi-line arguments, which are not used in the sub classes.

joel-costigliola:
I completely agree, that's something I had in mind but could not find to address, we already have done it for arrays and iterables btw

CodingMasterLSW:
Hi @joel-costigliola, @scordio 

I've submitted a PR (#4108) this issue. Could you please take a look when you have a moment? I'd appreciate your feedback. Thanks!

CodingMasterLSW:
Hi @joel-costigliola,

I’ve submitted a PR (#4114) for this issue.

As I worked on it, I realized that the number of classes to be updated is quite large. To make the review process more manageable, I thought it might be better to split the task into multiple PRs. Does this approach work for you? If you prefer having everything in a single PR, I will consolidate the subsequent changes into the current one.

In the meantime, I have categorized the potential refactoring targets below. One thing I’m uncertain about is the `Pattern/Match group`. I wonder if they strictly require this refactoring, or if they should be left as they are. I’d appreciate your thoughts on this.
### Contain CharSequence 
- ShouldContainCharSequence
- ShouldContainCharSequenceOnlyOnce
- ShouldBeSubstring
- ShouldNotContainCharSequence
- ShouldContainSequenceOfCharSequence
- ShouldContainSubsequenceOfCharSequence
- ShouldNotContainSequenceOfCharSequence
- ShouldNotContainSubsequenceOfCharSequence

[...]

PR #4129, opened 2026-01-31
Fix: Implement Blank/Whitespace in IndentWrapper

CodingMasterLSW:
Related Issue: https://github.com/assertj/assertj/issues/3167

## Summary

  - apply IndentWrapper to blank/whitespace in error message
  
## Updated Classes
- ShouldContainOnlyWhitespaces 
- ShouldNotContainOnlyWhitespaces
- ShouldContainOneOrMoreWhitespaces
- ShouldNotContainAnyWhitespaces
- ShouldNotStartWithWhitespaces
- ShouldNotEndWithWhitespaces

## Question
| Class / Scenario | Error Message Screenshot |
| :--- | :--- |
| **ShouldBeBlank**<br>(Current: No indentation) |  |
| **ShouldNotBeBlank**<br>(Current: No indentation) |  |
| **Other Classes**<br>(Target: Indented correctly) |  |

I noticed that the error messages in ShouldBeBlank and ShouldNotBeBlank are not indented. Is this behavior intentional
I was wondering if we should apply indentation to these classes as well for the sake of consistency and multi-line support.
 
I haven't modified them yet as I wanted to confirm the original design intent first. If these should be updated, please let me know, and I will include the changes in this PR.


joel-costigliola:
Good point @CodingMasterLSW, I think we should indent `ShouldBeBlank` and `ShouldNotBeBlank` value for consistency

CodingMasterLSW:
@joel-costigliola,
I've pushed the commits with the requested changes.
please take a look when you have a moment.

assertj-core/src/test/java/org/assertj/core/api/bytearray/ByteArrayAssert_asHexString_Test.java
```diff
@@ -84,7 +84,8 @@ void should_fail_with_soft_assertions_capturing_all_errors() {
                         .hasMessageContainingAll("-- error 1 --",
                                                  shouldBeEqualMessage("\"FF0001\"", "\"010203\""),
                                                  "-- error 2 --",
-                                                 "Expecting blank but was: \"FF0001\"");
+                                                 "Expecting blank but was:",
+                                                 "\"FF0001\"");
```
joel-costigliola: Was that necessary ? If not, could we revert this change since the original version is more accurate (it checks `Expecting blank but was: \"FF0001\"` is present whereas the current version checks that `Expecting blank but was:` and `\"FF0001\"` are present which can work with `Expecting blank but was: foobarbaz \"FF0001\"`.

this comment applies to other tests too.

assertj-core/src/test/java/org/assertj/core/error/ShouldBeBlank_create_Test.java
```diff
@@ -0,0 +1,54 @@
...
+
+import static java.lang.String.format;
+import static org.assertj.core.api.BDDAssertions.then;
+import static org.assertj.core.error.ShouldBeBlank.shouldBeBlank;
+
+import org.assertj.core.description.TextDescription;
+import org.assertj.core.presentation.StandardRepresentation;
+import org.junit.jupiter.api.Test;
+
+class ShouldBeBlank_create_Test {
```
joel-costigliola, line 26: thanks for adding this missing test !

assertj-core/src/test/java/org/assertj/core/api/bytearray/ByteArrayAssert_asHexString_Test.java
CodingMasterLSW: Oh, This is my mistake.

However, simply reverting the changes causes the tests to fail due to line break issues. To address this, I updated the tests to use String.format. This ensures we verify the exact complete message, resolving the accuracy issue you mentioned without causing failures.

commit: 355cbe4c9a24ee4f1d0dd7b1a24fcdc13795a687

If I were to revert this test code, I would need to revert all changes to `ShouldBeBlank` and `ShouldNotBeBlank`

If there are any concerns with this solution, please let me know!

CodingMasterLSW:
Hi @joel-costigliola,

I fully agree with your point regarding the loose assertions. However, simply reverting the changes caused the tests to fail due to line break discrepancies. Therefore, I updated the tests to use String.format to verify the exact complete message.

I also specifically used String.format() instead of .formatted() to maintain consistency with the existing conventions in the codebase.

While working on these updates, I came across one question regarding `ByteArrayAssert_asHexString_Test.java` (specifically inside `should_fail_with_soft_assertions_capturing_all_errors()`). (same for other tests)

| **Before** | **After** |
| :---: | :---: |
| Image 2 | Image 1 |

As shown in the screenshots, the formatting between error1 and error2 appears inconsistent. Although the changes to `ShouldBeBlank` and `ShouldNotBeBlank` now align them with other blank/whitespace classes, the resulting combined error message in this test looks a bit awkward.

I would appreciate your thoughts on how best to handle this.

joel-costigliola:
Integrated thanks @CodingMasterLSW !

PR #4145, opened 2026-02-14
Use `https://` URLs in documentation

Marcono1234:
Replaces `http://` URLs with `https://` URLs in the documentation. Exceptions:
- Apache License header URL
- Maven XML schema URLs
- example URLs for `URI` and `URL` asserts
- arizona.edu URLs in `MyersDiff.java`; those seem to be dead links
- sun.com URL in `Introspection.java`; seems to be a dead link and not sure if a good replacement exists


assertj-core/src/main/java/org/assertj/core/api/AutoCloseableBDDSoftAssertions.java
```diff
@@ -137,7 +137,7 @@
  *
  * @author Brian Laframboise
  *
- * @see <a href="http://beust.com/weblog/2012/07/29/reinventing-assertions/">Reinventing Assertions (inspired this feature)</a>
+ * @see <a href="https://www.beust.com/weblog/reinventing-assertions/">Reinventing Assertions (inspired this feature)</a>
```
Marcono1234, line 140: This is the redirect target of that URL, is it fine that I replaced it here (and for the other occurrences)?

joel-costigliola: yes :+1:

joel-costigliola:
Thanks @Marcono1234 !

PR #4149, opened 2026-02-14
Improve Javadoc regarding default `Charset` usage

Marcono1234:
- Fixed missing mention of default Charset usage (see review comment)
- Fixed wrong mention of default Charset usage
- Added `{@link ...}` to all default Charset references, for consistency
- Changed "default character set" to "default charset", for consistency

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractByteArrayAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractByteArrayAssert.java
@@ -1270,7 +1270,8 @@ public abstract class AbstractByteArrayAssert<SELF extends AbstractByteArrayAsse
    * Converts the actual byte[] under test to a String and returns assertions for the computed String
    * allowing String specific assertions from this call.
    * <p>
-   * The byte[] conversion to a String by decoding the specified bytes using the platform's default charset.
+   * The byte[] conversion to a String by decoding the specified bytes using the platform's
+   * {@link Charset#defaultCharset() default charset}.
    * <p>
    * Example :
    * <pre><code class='java'> byte[] bytes = new byte[] { -1, 0, 1 };
@@ -1300,8 +1301,6 @@ public abstract class AbstractByteArrayAssert<SELF extends AbstractByteArrayAsse
    * Converts the actual byte[] under test to a String by decoding the specified bytes using the given charset
    * and returns assertions for the computed String
    * allowing String specific assertions from this call.
-   * <p>
-   * The byte[] conversion to a String by decoding the specified bytes using the platform's default charset.
    * <p>
    * Example :
    * <pre><code class='java'> byte[] bytes = new byte[] { -1, 0, 1 };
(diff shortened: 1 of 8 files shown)


assertj-core/src/main/java/org/assertj/core/api/AbstractByteArrayAssert.java
```diff
@@ -1301,8 +1302,6 @@ public AbstractStringAssert<?> asString() {
    * and returns assertions for the computed String
    * allowing String specific assertions from this call.
    * <p>
-   * The byte[] conversion to a String by decoding the specified bytes using the platform's default charset.
-   * <p>
```
Marcono1234, line 1305: This looked wrong (maybe a copy & paste error?); this method takes an explicit `Charset`, as mentioned also by the previous Javadoc paragraph.

assertj-core/src/main/java/org/assertj/core/api/AbstractInputStreamAssert.java
```diff
@@ -228,6 +228,9 @@ private void assertIsNotEmpty() {
    * This will change in AssertJ 4.0 where newlines will be taken into account, in the meantime, to get this behavior
    * one can use {@link #asString(Charset)} and then chain with {@link AbstractStringAssert#isEqualTo(String)}.
    * <p>
+   * The {@link Charset#defaultCharset() default charset} is used for decoding the bytes of the stream to a String.
+   * To use a different charset for decoding, use {@link #asString(Charset)}.
+   * <p>
```
Marcono1234, line 233: Usage of the default Charset was previously not mentioned here.

The link to `asString` might be a bit redundant because the previous Javadoc paragraph mentions it as well, but I assume that other paragraph will be removed after version 4.0?

assertj-core/src/main/java/org/assertj/core/api/AbstractPathAssert.java
```diff
@@ -180,7 +180,7 @@ public SELF hasSameTextualContentAs(Path expected, Charset expectedCharset) {
    *
    * // The following assertion succeeds:
    * String expectedContent = "Gerçek Başka bir yerde mi" + org.assertj.core.util.Compatibility.System.lineSeparator();
-   * byte[] binaryContent = expectedContent.getBytes(turkishCharset.name());
+   * byte[] binaryContent = expectedContent.getBytes(turkishCharset);
```
Marcono1234, line 183: Not directly related to the default Charset, but I simplified this here nonetheless, to refer to `String#getBytes(Charset)`.

assertj-core/src/main/java/org/assertj/core/api/AbstractByteArrayAssert.java
joel-costigliola: yes, likely a copy/paste error

assertj-core/src/main/java/org/assertj/core/api/AbstractInputStreamAssert.java
joel-costigliola: I'm fine keeping the extra asString mention, that does not hurt

joel-costigliola:
thanks for the PR @Marcono1234, appreciated !

PR #4158, opened 2026-02-21
Update EqualsVerifier and fix JDK 26 final field mutation warnings

jqno:
#### Check List:
* Fixes https://github.com/jqno/equalsverifier/issues/1149
* Unit tests : YES

This PR follows up on @scordio's ticket in EqualsVerifier, about JDK 26's "final means final" warnings.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/FieldLocation.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/FieldLocation.java
@@ -61,17 +61,12 @@ public final class FieldLocation implements Comparable<FieldLocation> {
     if (this == obj) return true;
     if (!(obj instanceof FieldLocation)) return false;
     FieldLocation that = (FieldLocation) obj;
-    return Objects.equals(pathToUseInRules, that.pathToUseInRules)
-           && Objects.equals(decomposedPath, that.decomposedPath)
-           && Objects.equals(pathsHierarchyToUseInRules, that.pathsHierarchyToUseInRules);
+    return Objects.equals(decomposedPath, that.decomposedPath);
   }
 
   @Override
   public int hashCode() {
-    int result = Objects.hashCode(pathToUseInRules);
-    result = 31 * result + Objects.hashCode(decomposedPath);
-    result = 31 * result + Objects.hashCode(pathsHierarchyToUseInRules);
-    return result;
+    return Objects.hashCode(decomposedPath);
   }
 
   @Override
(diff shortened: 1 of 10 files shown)


assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/FieldLocation.java
```diff
@@ -61,17 +61,12 @@ public boolean equals(Object obj) {
     if (this == obj) return true;
     if (!(obj instanceof FieldLocation)) return false;
     FieldLocation that = (FieldLocation) obj;
-    return Objects.equals(pathToUseInRules, that.pathToUseInRules)
-           && Objects.equals(decomposedPath, that.decomposedPath)
-           && Objects.equals(pathsHierarchyToUseInRules, that.pathsHierarchyToUseInRules);
+    return Objects.equals(decomposedPath, that.decomposedPath);
```
jqno, line 64: I noticed, when updating the test for this class, that the fields I removed are directly derived from `decomposedPath`, and therefore didn't really contribute to equals.

If you disagree, I'm happy to restore these lines (and the ones in hashCode below)

scordio: @joel-costigliola WDYT about this one? The rest looks good to me 👍

scordio (approved):

scordio:
@jqno as we still maintain version 3 with Java 26 pipeline, would you be fine with using `3.x` as the base branch? You'd need to rebase this branch and force-push it, or I can also take care of that before merging (at least, I change the PR target to signal it).

jqno:
Sorry, I wasn't aware of that! Yes of course that's fine. I have to attend to life stuff now, but I can probably do it tonight. Or if you want to do it now, that's totally fine with me too

scordio:
> at least, I change the PR target to signal it

I restored `main`, otherwise thousands of files appear to be changed... let's see who's faster with the rebase 🙃

assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/FieldLocation.java
joel-costigliola: fine by me !

jqno:
I rebased and changed the PR's target branch. Hope I did it correctly :sweat_smile:

scordio:
Many thanks and congrats on your first contribution, @jqno!

jqno:
It's nice to do something back!

PR #4180, opened 2026-03-06
Define comparator `toString` in `AssertWithComparator.usingEquals`

lukas-krecan:
If `customEqualsDescription` is not set, the comparator `toString` is used in the exception message. As it's a lambda, the message look like this:

```
when comparing values using AssertWithComparator$$Lambda/0x000003800027c098"
```

Even if the predicate does not have meaningful `toString`, using it is more logical.

#### Check List:
* Fixes NA
* Unit tests : YES


scordio:
Thanks, @lukas-krecan!

PR #4195, opened 2026-03-21
Streamline implementation of assertions returning to navigation origin

scordio:
This change streamlines the implementation of assertion types that allow returning to the navigation origin.

Constructors in concrete and abstract classes that don't align with storing the origin instance in the abstract class have been deprecated and will be removed in version 4.

### Scope

* `AbstractBigDecimalScaleAssert`
* `AbstractFileSizeAssert`
* `AbstractIterableSizeAssert`
* `AbstractMapSizeAssert`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,6 +60,8 @@ We prefer integrating PR by squashing all the commits and rebasing it to `main`;
 
 ## Naming Conventions
 
+### Assertions
+
 Here are some of the `ThrowableAssert` assertions: `hasMessage`, `hasNoCause`, `hasMessageContaining`; for each of them we have a test class, note the naming convention:
 * `ThrowableAssert_hasMessage_Test`
 * `ThrowableAssert_hasNoCause_Test`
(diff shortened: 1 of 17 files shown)


joel-costigliola:
LGTM besides formatting violation

joel-costigliola:
Actually after giving it more thoughts, I find `returnToOrigin` to be less expressive than the existing code because it's generic, ex:
```java
    then(file).size()
              .isGreaterThan(4L)
              .returnToOrigin().hasContent("actual");
```
vs
```java
    then(file).size()
              .isGreaterThan(4L)
              .returnToFile().hasContent("actual");
```
On the plus side, it standardizes the back navigation.

@scordio it's not a strong opinion, I'm happy to go that way if you feel like it.

scordio:
> On the plus side, it standardizes the back navigation.

That is mostly my motivation.

As of today, the method names are specific to their domain:
* `returnToBigDecimal` for `BigDecimalScaleAssert`
* `returnToFile` for `FileSizeAssert`
* `returnToIterable` for `IterableSizeAssert`
* `returnToMap` for `MapSizeAssert`
* `returnToInitialThrowable` for `SuppressedExceptionsAssert` (4.x only)

However, there's a lack of cohesion and a maintenance burden in keeping them in sync (e.g., consistent `@CheckReturnValue`, stronger reference to navigation in the Javadoc, appropriate return types, etc.).

We might just keep the status quo, but I wanted to give us a chance to think about it once more.

joel-costigliola:
Another point to consider, if we chain multiple navigation, it might not be obvious what `returnToOrigin would do.
Ex: 

```java
    then(exception).message().size()
              .isGreaterThan(4L)
              .returnToOrigin(); // is it message or exception ?
```

scordio:
### Team Decision
* ~Keep `OriginAwareAssert` and its type parameter as a marker interface, but without any method to be implemented~
* Keep the existing assertion-specific names as they are clearer than just a generic name, but implement them in the related abstract class in a backward-compatible manner
* Future method names should follow the pattern `returnTo<OriginActualType>`, which means `returnToInitialThrowable` from 4.x will be introduced in 3.x as `returnToThrowable`
* Apply `@CheckReturnValue` consistently
* Ensure that the cases mentioned at https://github.com/assertj/assertj/pull/4195#issuecomment-4106006993 adhere to the previous points
* Document this design choice in `CONTRIBUTING.md`

Issue #2563, opened 2022-04-12
Improve the performance of methods `SoftAssertions#assertThat`


JohnnyBFG:
#### Summary
We recently upgraded from AssertJ 3.8.0 to 3.22.0 and we noticed significant slow down in tests using _SoftAssertions#assertThat_.
I look into it and found that AssertJ switched from cglib to ByteBuddy in 3.9.1.
I did some simple performance corporation between version and found that proxy generation is ~4x slower compare to AssertJ 3.8.0. with cglib. Speed difference is quite noticeable especially on first running test. TypeCache in SoftProxies is kinda helping, but only tests running after type was already cached.
I also compare times from AssertJ 3.22.0 to Mockito 3.2.4 (which is quite old, but it's too using ByteBuddy to generate sub-classes) and found that execution times for generating classes are almost identical.

This slowdown make SoftAssertions#assertThat unsuitable for some of our tests, as they must run under 2s + it makes harder to predict actual execution time of whole test. We found workaround in _SoftAssertions#check(() -> assertThat)_ but this syntax can be harder to read in longer tests so we would prefer to use _SoftAssertions#assertThat_ as it's more natural.

I want to ask you if you could please look into this problem and found if there is any space to improvement?

I understand that there isn't probably any space on ByteBuddy side for improvements, but maybe there could be done something on ours?
Maybe change some strategy or use ByteBuddy to generate class from interface (not actual class) and delegate calls to real object (but that would require create interfaces for each Assert class and I'm not sure if it would be possible due generic + if such change wouldn't result if worse runtime performance).
This problem can be also environment problem and sub-class generation is much faster on newer versions of JDK (but I have doubts about that, I experienced similar problems on up-to-date versions of JDK 11 too).

#### Testing environment
java version "11.0.12" 2021-07-20 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.12+8-LTS-237)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.12+8-LTS-237, mixed mode)

Tests were performed in Intellij Idea 2020.2.4

#### AssertJ performance comparison
Used tests:
```java
//used for AssertJ
public class MainTest {
    private final Exception exception = new Exception("lol");
    private final Map<Object, Object> map = Map.of();
    private final Object object = new Object();
    private final List<String> list = List.of();

    @Test
    public void text() { SoftAssertions.assertSoftly(softly -> softly.assertThat("text")); }
    @Test
[...]

PR #4199, opened 2026-03-26
Remove bytebuddy use for SoftAssertions and Assumptions

filiphr:
This is a stab at #2563. I had an initial idea to use code generation to do the proxies. However, after a bit more thinking I got another idea, basically make all assertions fully aware that they might be part of an assumption or soft assertion.

The pattern for this PR is the `AbstractAssert` to expose 2 methods:

* `protected SELF executeAssertion(Runnable body)`
* `protected <T> T executeAssertionNavigation(Supplier<T> body)`

In order to allow implementation to participate in soft assertions or assumptions.

The `AbstractAssert` then has the `AssertionErrorHandler`, which is set by the soft assertions or assumptions and handles the required logic.

This does mean that all implementations would need to be adjusted.

e.g.

Currently:

```java
public SELF isNotNull() {
  return myself;
}
```

Becomes

```java
public SELF isNotNull() {
  return executeAssertion(() -> objects.assertNotNull(info, actual));
}
```

----

Important to mention that a big portion of this was done with Claude Code, but I did a review of the core parts and the API. We need to go through the changes in the different assertion implementations throughly if we like this approach.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
@@ -722,8 +722,10 @@ public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, 
   @Override
   @CheckReturnValue
   public AbstractStringAssert<?> asString() {
-    objects.assertNotNull(info, actual);
-    return Assertions.assertThat(actual.toString()).withAssertionState(myself);
+    return executeAssertionNavigation(() -> {
+      objects.assertNotNull(info, actual);
+      return Assertions.assertThat(actual.toString()).withAssertionState(myself);
+    });
   }
 
   /**
@@ -1141,14 +1143,22 @@ public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, 
 
   private SELF matches(Predicate<? super ACTUAL> predicate, PredicateDescription predicateDescription) {
     requireNonNull(predicate, "The predicate must not be null");
-    if (predicate.test(actual)) return myself;
-    throw Failures.instance().failure(info, shouldMatch(actual, predicate, predicateDescription));
+    return executeAssertion(() -> {
+      if (predicate.test(actual)) {
+        return;
+      }
+      throw Failures.instance().failure(info, shouldMatch(actual, predicate, predicateDescription));
+    });
   }
 
   private SELF doesNotMatch(Predicate<? super ACTUAL> predicate, PredicateDescription predicateDescription) {
     requireNonNull(predicate, "The predicate must not be null");
-    if (predicate.negate().test(actual)) return myself;
-    throw Failures.instance().failure(info, shouldNotMatch(actual, predicate, predicateDescription));
+    return executeAssertion(() -> {
+      if (predicate.negate().test(actual)) {
+        return;
+      }
+      throw Failures.instance().failure(info, shouldNotMatch(actual, predicate, predicateDescription));
+    });
   }
 
   public static void setCustomRepresentation(Representation customRepresentation) {
@@ -1191,9 +1201,7 @@ public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, 
    * @return a new {@link AbstractListAssert}.
    */
   protected <E> AbstractListAssert<?, List<? extends E>, E, ObjectAssert<E>> newListAssertInstance(List<? extends E> newActual) {
-    ListAssert<E> listAssert = new ListAssert<>(newActual);
-    listAssert.withAssertionState(myself);
-    return listAssert;
+    return new ListAssert<>(newActual);
   }
 
   SELF withAssertionState(@SuppressWarnings("rawtypes") AbstractAssert assertInstance) {
@@ -1205,16 +1213,8 @@ public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, 
 
   private void propagateAssertionInfoFrom(AbstractAssert<?, ?> assertInstance) {
     this.info.useRepresentation(assertInstance.info.representation());
-    if (assertInstance.info.overridingErrorMessage() != null) {
-      // When overridingErrorMessage is set, always propagate both it and the description
-      // (even if null) so the overridingErrorMessage is used as-is without any description prefix.
-      this.info.overridingErrorMessage(assertInstance.info.overridingErrorMessage());
-      this.info.description(assertInstance.info.description());
-    } else if (assertInstance.info.description() != null) {
-      // Only overwrite description if the source has one set.
-      // This preserves descriptions set by extracting/navigation methods (via .as()).
-      this.info.description(assertInstance.info.description());
-    }
+    this.info.description(assertInstance.info.description());
+    this.info.overridingErrorMessage(assertInstance.info.overridingErrorMessage());
   }
 
   // this method is meant to be overridden and made public in subclasses that want to expose it
@@ -1271,7 +1271,7 @@ public abstract class AbstractAssert<SELF extends AbstractAssert<SELF, ACTUAL>, 
     String extractedPropertyOrFieldDescription = extractedDescriptionOf(propertyOrField);
     String description = mostRelevantDescription(info.description(), extractedPropertyOrFieldDescription);
     @SuppressWarnings("unchecked")
-    ASSERT result = (ASSERT) assertFactory.createAssert(value).as(description).withAssertionState(myself);
+    ASSERT result = (ASSERT) assertFactory.createAssert(value).withAssertionState(myself).as(description);
     return result;
   }
 
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
@@ -891,7 +891,7 @@ public abstract class AbstractIterableAssert<SELF extends AbstractIterableAssert
     List<Object> values = FieldsOrPropertiesExtractor.extract(actual, byName(propertyOrField));
     String extractedDescription = extractedDescriptionOf(propertyOrField);
     String description = mostRelevantDescription(info.description(), extractedDescription);
-    return propagateState(newListAssertInstanceForMethodsChangingElementType(values).as(description));
+    return newListAssertInstanceForMethodsChangingElementType(values).as(description);
   }
 
   /**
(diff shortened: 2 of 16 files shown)


assertj-core/src/test/java/org/assertj/core/api/BDDSoftAssertionsTest.java
```diff
@@ -577,7 +577,9 @@ void should_collect_all_errors_when_using_extracting() {
           .isEmpty();
     // THEN
     assertThat(softly.errorsCollected()).extracting(Throwable::getMessage)
-                                        .containsExactly("error 1", "error 2", "error 3");
+                                        .containsExactly("[Extracted: first] error 1",
+                                                         "[Extracted: first] error 2",
+                                                         "[Extracted: last] error 3");
```
filiphr, line 582: I believe that this change is actually more correct. The reason is that when using regular assertions.

e.g.

```java
Assertions.assertThat(names)
          .extracting("first")
          .overridingErrorMessage("error 1")
          .contains("gandalf");
```

The error message is:

```
[Extracted: first] error 1
```

assertj-core/src/main/java/org/assertj/core/api/AbstractCompletableFutureAssert.java
```diff
@@ -263,11 +271,12 @@ public SELF isCompletedWithValue(RESULT expected) {
    * @throws AssertionError if the actual {@code CompletableFuture} does not succeed within the given timeout with the satisfying value.
    */
   public SELF isCompletedWithValueMatchingWithin(Predicate<RESULT> resultPredicate, Duration completionDuration) {
-    RESULT actualResult = futures.assertSucceededWithin(info, actual, completionDuration);
-    if (!resultPredicate.test(actualResult)) {
-      throw Failures.instance().failure(info, shouldMatch(actualResult, resultPredicate, PredicateDescription.GIVEN));
-    }
-    return myself;
+    return executeAssertion(() -> {
+      RESULT actualResult = futures.assertSucceededWithin(info, actual, completionDuration);
```
filiphr: One of the things that should be evaluated is this. Should this be within the execution or outside. I think it should be outside, but it should be discussed. There was one place where it failed if it was inside, but seems like this one didn't fail any test.

joel-costigliola: Why do you think it should be outside `executeAssertion` ?

assertj-core/src/test/java/org/assertj/core/api/BDDSoftAssertionsTest.java
joel-costigliola: Agreed !

assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
```diff
@@ -928,7 +891,7 @@ public AbstractListAssert<?, List<?>, Object, ObjectAssert<Object>> extracting(S
     List<Object> values = FieldsOrPropertiesExtractor.extract(actual, byName(propertyOrField));
     String extractedDescription = extractedDescriptionOf(propertyOrField);
     String description = mostRelevantDescription(info.description(), extractedDescription);
-    return newListAssertInstanceForMethodsChangingElementType(values).as(description);
+    return propagateState(newListAssertInstanceForMethodsChangingElementType(values).as(description));
```
joel-costigliola: I'm not convinced we need `propagateState` since it simply calls `withAssertionState`, moreover I find  `withAssertionState` clearer because it is obvious which assertion state is being replicated.

this line of code can be rewritten as 
```java
newListAssertInstanceForMethodsChangingElementType(values).withAssertionState(myself).as(description)
```

assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
```diff
@@ -1142,19 +1191,30 @@ public SELF doesNotHaveSameHashCodeAs(Object other) {
...
     return myself;
   }
 
   private void propagateAssertionInfoFrom(AbstractAssert<?, ?> assertInstance) {
     this.info.useRepresentation(assertInstance.info.representation());
-    this.info.description(assertInstance.info.description());
-    this.info.overridingErrorMessage(assertInstance.info.overridingErrorMessage());
+    if (assertInstance.info.overridingErrorMessage() != null) {
+      // When overridingErrorMessage is set, always propagate both it and the description
+      // (even if null) so the overridingErrorMessage is used as-is without any description prefix.
```
joel-costigliola: This is a bit convoluted, if some tests don't propagate the right info, I feel it should be fixed at the source. 

But TBH we haven't done a great job a propagating the assertion info everywhere, it's likely should be addressed in a separate issue.

joel-costigliola:
thanks @filiphr, I have some minor comments only which I'm happy to address myself after we discuss them.

I usually get PR locally and rework them slightly (but amending the commits so that the original author gets the full credit for the work), I find it a better way to review code and address minor issues like formatting and renaming stuff for clarity.

assertj-core/src/main/java/org/assertj/core/api/AbstractCompletableFutureAssert.java
filiphr: There are tests for `succeedsWithin` and `failsWithin` that expect an assertion error.

e.g.

https://github.com/assertj/assertj/blob/bf3bbd5953fa58ab74f9153c338727aad28b04f8/assertj-core/src/test/java/org/assertj/core/api/SoftAssertions_future_Test.java#L63-L73

https://github.com/assertj/assertj/blob/bf3bbd5953fa58ab74f9153c338727aad28b04f8/assertj-core/src/test/java/org/assertj/core/api/SoftAssertions_future_Test.java#L161-L170

I initially had

```java
private ObjectAssert<RESULT> internalSucceedsWithin(Duration timeout) {
  return executeAssertionNavigation( () -> {
    RESULT result = futures.assertSucceededWithin( info, actual, timeout );
    return newObjectAssert( result );
  } );
}
```

However, the tests above failed. That's why my gut feeling says that it should behave the same everywhere. I'll create another PR that adds tests for those scenarios and then we can align it.

assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
filiphr: The line would be 

```java
newListAssertInstanceForMethodsChangingElementType(values).as(description).withAssertionState(myself)
```

However, yes you are right. This is a tricky one since `newListAssertInstanceForMethodsChangingElementType` does call `withAssertionState(myself)` through `newListAssertInstance`. However, the `as(description)` messes up the errors a bit.

In any case I'll remove the `propagateState`

assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
filiphr: I agree about this one. It's mostly around the navigation assertions where this was causing some problems. I'll try to address this separately.

assertj-core/src/main/java/org/assertj/core/api/AbstractCompletableFutureAssert.java
filiphr: @joel-costigliola, I've created #4203 for this, at least for the (Completable)Future asserts.

filiphr:
>thanks @filiphr, I have some minor comments only which I'm happy to address myself after we discuss them.

@joel-costigliola I'm more than happy to address the feedback as well. I've already commented back on your comments.

I think the main things were `propagateState` and the `withAssertionState`. I've removed the `propagateState` and reverted the changes in `withAssertionState`. I also reviewed all the places where we call it and made sure that it gets called before any setting of the description. This lead to needing some more adjustments in the soft tests. I've checked the adjustments and actually it's more correct now, it's identical like with regular assertions.

>I find it a better way to review code and address minor issues like formatting and renaming stuff for clarity.

I'm the same. I try to do the same for MapStruct

assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
filiphr: I've removed `propagateState` and re-evaluated all the call sides of the `withAssertionState`. This lead to some more test adjustments which are in the same direction as the changes already done in `BDDSoftAssertionsTest`

assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
filiphr: I found a way better solution for this, and reverted all the changes here

assertj-core/src/main/java/org/assertj/core/api/AbstractCompletableFutureAssert.java
filiphr: I've integrated the changes from #4203 here.

joel-costigliola:
Thanks @filiphr, looks good to me

joel-costigliola:
let's wait for @scordio feedback before merging it

filiphr:
Thanks for the review @joel-costigliola. One thing to not forget, we should also document this, especially for custom assertions that they need to use the `executeAssertion` and `executeAssertionNavigation` if they want to participate in soft assertions.

joel-costigliola:
Thank you so much for this one @filiphr !

scordio:
Thanks @filiphr! I'm still going through the review, I might just do some small updates on `main` directly.

joel-costigliola:
@scordio sorry I thought your thumbs up meant you had done the review :facepalm:

filiphr:
Thanks @joel-costigliola and @scordio. Yes, feel free to do adjustments, I'm also mora than happy to do some adjustments in new PRs as well. Feel free to share any feedback

scordio:
I think we could fully clean up all Byte Buddy dependencies from the POMs, right?

scordio:
FYI, I'll trigger the Copilot review post-mortem to aid my own review. You don't have to react on any findings (assuming they are valid)

joel-costigliola:
I pushed two commits to 
- favor `SoftAssertionProvider.soft` over `SoftAssertionProvider.proxy`
- remove completely references to byte buddy

scordio:
I spotted `org.assertj.core.util.Throwables.isProxiedAssertionClass` having a ByteBuddy reference.

With a quick look, I think we can simply delete it and clean up the corresponding calling side. Do you see the same?

filiphr:
Thanks for that push @joel-costigliola. Maybe a question regarding `SoftAssertionProvider.proxy`. Should we perhaps deprecate it in favor of `SoftAssertionsProvider.soft`?

joel-costigliola:
Good idea, I don't see a use case for it but not 100% sure there isn't one so deprecating it is a good option

PR #4218, opened 2026-04-12
Skip soft assertion errors

joel-costigliola:


joel-costigliola:
fixed by [@filiphr ](https://github.com/assertj/assertj/pull/4219)

PR #4219, opened 2026-04-12
Prevent `extracting` from leaking RuntimeException in soft assertions

### LINKED ISSUE #4213: Soft assertions fail in confusing ways with null values
**Describe the bug**

See the test case.

* assertj core version: 3.27.7
* java version: 21.0.9
* test framework version: -
* os (if relevant): -

```java
        try (var softly = new AutoCloseableSoftAssertions ()) {
            Object  value = null;  // in real tests this would be a result of invoking some method, for example
            softly.assertThat (value).isNotNull ().extracting ("foo").isEqualTo ("bar");
        }
```

expecting this to produce exactly one failure like this (got this by removing `.extracting ("foo").isEqualTo ("bar")`):

    Exception in thread "main" org.assertj.core.api.SoftAssertionError: 
    The following assertion failed:
    1) 
    at Foo.main(Foo.java:15)
    
            at org.assertj.core.error.AssertionErrorCreator.multipleSoftAssertionsError(AssertionErrorCreator.java:102)
            at org.assertj.core.api.AbstractSoftAssertions.assertAll(AbstractSoftAssertions.java:41)
            at org.assertj.core.api.AbstractSoftAssertions.assertAll(AbstractSoftAssertions.java:46)
            at org.assertj.core.api.AutoCloseableSoftAssertionsProvider.close(AutoCloseableSoftAssertionsProvider.java:34)
            at Foo.main(Foo.java:16)

Instead it dies with an exception, meaning that it doesn't work "softly" anymore, and the output is also pretty confusing:

    Exception in thread "main" java.lang.IllegalArgumentException: The object to extract property/field from should not be null
            at org.assertj.core.util.Preconditions.checkArgument(Preconditions.java:134)
            at org.assertj.core.util.introspection.PropertyOrFieldSupport.getValueOf(PropertyOrFieldSupport.java:53)
            at org.assertj.core.extractor.ByNameSingleExtractor.apply(ByNameSingleExtractor.java:32)
            at org.assertj.core.api.AbstractAssert.extracting(AbstractAssert.java:1266)
            at org.assertj.core.api.AbstractObjectAssert.extracting(AbstractObjectAssert.java:806)
            at org.assertj.core.api.ObjectAssert$ByteBuddy$UbgnH5po.extracting$accessor$oMeG1udP(Unknown Source)
            at org.assertj.core.api.ObjectAssert$ByteBuddy$UbgnH5po$AssertJ$SoftProxies$rsSMPYj6.call(Unknown Source)
            at org.assertj.core.api.ProxifyMethodChangingTheObjectUnderTest.intercept(ProxifyMethodChangingTheObjectUnderTest.java:82)
            at org.assertj.core.api.ObjectAssert$ByteBuddy$UbgnH5po.extracting(Unknown Source)
            at Foo.main(Foo.java:15)
            Suppressed: org.assertj.core.api.SoftAssertionError: 
    The following 2 assertions failed:
    1) 
    at Foo.main(Foo.java:15)
    2) 
    at Foo.main(Foo.java:15)
    
                    at org.assertj.core.error.AssertionErrorCreator.multipleSoftAssertionsError(AssertionErrorCreator.java:102)
                    at org.assertj.core.api.AbstractSoftAssertions.assertAll(AbstractSoftAssertions.java:41)
                    at org.assertj.core.api.AbstractSoftAssertions.assertAll(AbstractSoftAssertions.java:46)
                    at org.assertj.core.api.AutoCloseableSoftAssertionsProvider.close(AutoCloseableSoftAssertionsProvider.java:34)
                    at Foo.main(Foo.java:13)

filiphr:
This is a spike on skipping errors in chained assertions. Similar to PR #4218 and should fix #4213.

---- 
When `actual` is null, `extracting` internally fails its null-guard with an AssertionError (collected by soft assertions), then continues to apply the extractor to null and leaks a RuntimeException that bypasses soft collection.

Introduce a `skipAssertions` dead-chain flag on AbstractAssert, propagated via `withAssertionState`. When a soft null-guard in `extracting` fails, return a dead-chain navigated assertion so downstream assertions no-op instead of throwing or piling up errors that cannot succeed.

#### Check List:
* Fixes #4213
* Unit tests : YES 


assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
```diff
@@ -1281,12 +1290,30 @@ protected RecursiveAssertionAssert usingRecursiveAssertion() {
...
+   * errors that cannot succeed.
+   */
+  protected <A extends AbstractAssert<?, ?>> A markAsDeadChain(A deadChain) {
+    deadChain.withAssertionState(myself);
+    deadChain.skipAssertions = true;
+    return deadChain;
+  }
+
+  private <T, ASSERT extends AbstractAssert<?, ?>> ASSERT deadChainNavigation(AssertFactory<T, ASSERT> assertFactory) {
+    return markAsDeadChain(assertFactory.createAssert((T) null));
```
joel-costigliola, line 1314: I think this could be a bit simpler, since we can't really extract anything, we could simply return this assertion after setting the `skipAssertions` to true, I mean at this point since we failed the navigation, it would make sense not to navigate at all.

I suggest 

- removing `markAsDeadChain`, 
- changing visibility of this method to `protected` so that users can override it, 
- implementing it like:  

```java
skipChainedAssertions = true;
// Noop assert since we just have skipChainedAssertions
return (ASSERT)myself;
```

For navigation that returns a different type of assertion, we can - as you did for the list assertion - return an assertion with actual set to null.

WDYT ?

filiphr: I am not sure that I am following.

In this particular case we can't do what you are suggestion since this method is for navigation, i.e. we can't just do `(ASSERT)myself) it'll lead to a class cast exception. 

This method is used in the 2 `extracting` from `AbstractAssert` that take an `AssertFactory` and return a different type of assertion. 

Also in the `AbstractObjectAssert`, I can't return `myself` as it expects a `AbstractListAssert` as a return type.

Perhaps we can only keep `markAsDeadChain` and do

```java
if (actual == null) return markAsDeadChain(assertFactory.createAssert((T) null));
```

If you like the approach I can go ahead and implement it for the rest of the classes to see how it would look like fully.

joel-costigliola: Since extracting a single value returns `AbstractAssert`, you can return any abstract assert including `myself` as long as you set the skip assertions flag to true, since it's a "dead" assert instance it does not matter to build a precise subtype of assert (I'm probably not much clearer than my previous comment but I don't know how to express my point in a clearer way).

I have tried with my suggested implementation on your branch branch locally, ie   this:
```java
  protected <ASSERT extends AbstractAssert<?, ?>> ASSERT extracting(String propertyOrField,
                                                                    AssertFactory<Object, ASSERT> assertFactory) {
    requireNonNull(propertyOrField, shouldNotBeNull("propertyOrField")::create);
    requireNonNull(assertFactory, shouldNotBeNull("assertFactory")::create);
    isNotNull();
    if (actual == null) {
      skipAssertions = true;
      return (ASSERT)myself;
    }
   ....  
``` 
and all the SoftAssertions_extracting_on_null_actual_Test are green.

I also tested with strongly typed extracting and it works too, ex:

```java
  @Test
  void should_not_throw_when_extracting_by_name_without_explicit_isNotNull() {
    // GIVEN / WHEN
    softly.assertThat((Object) null).extracting("foo", STRING).startsWith("bar"); // extract as String
    // THEN - extracting's internal isNotNull collects one error; downstream is noop
    List<Throwable> errors = softly.errorsCollected();
    then(errors).hasSize(1);
    then(errors.get(0)).hasMessageContaining("Expecting actual not to be null");
  }
```

@filiphr  I'm more than happy to be proven wrong, could you think of a test that would break my suggested implementation ?

we can't do that for extracting multiple values since it returns a list assert, in this case we obviously can't return `myself`.

filiphr: I've checked this a bit more. Yes you are right about the `AbstractAssert#extracting`. However, that pattern won't work for the `AbsractIterable#first(InstanceOfAssertFactory)`.

I'll try to find some time to apply a general pattern for everything

joel-costigliola: I agree it won't work with `first()` since `myself` is an iterable assert and first and object assert.
In that case, simply returning an `ObjectAssert` would work. 

BTW I find the `deadChain` term to be quite good and expressive
Keen to see where your exploration is going to go

filiphr: @joel-costigliola I did some more work on this.  Have a look at how it turned out. Using Claude I also found some places where the pattern was missing and I've adjusted all those places to use this new pattern. Let me know what you think.

I see that Java 26 is failing, if you agree on the pattern in the PR, I'll investigate it to see why it fails.

joel-costigliola: thanks, I'll look at it during the week end

assertj-core/src/test/java/org/assertj/core/api/SoftAssertions_navigations_on_null_actual_Test.java
```diff
@@ -0,0 +1,1052 @@
...
+
+  @Nested
+  class Extracting {
+
+    @Test
+    void should_not_throw_when_chaining_extracting_by_name_after_failed_isNotNull() {
+      // GIVEN / WHEN
+      softly.assertThat((Object) null).isNotNull().extracting("foo").isEqualTo("bar");
+      // THEN - no RuntimeException, only soft-collected AssertionErrors
+      List<Throwable> errors = softly.errorsCollected();
```
joel-costigliola: could use `List<AssertionError> errors = softly.assertionErrorsCollected();` here and skip the `then(errors).allMatch(e -> e instanceof AssertionError);` check

assertj-core/src/test/java/org/assertj/core/api/SoftAssertions_navigations_on_null_actual_Test.java
```diff
@@ -0,0 +1,1052 @@
...
+    softly = new SoftAssertions();
+  }
+
+  @Nested
+  class Extracting {
+
+    @Test
+    void should_not_throw_when_chaining_extracting_by_name_after_failed_isNotNull() {
+      // GIVEN / WHEN
+      softly.assertThat((Object) null).isNotNull().extracting("foo").isEqualTo("bar");
```
joel-costigliola: use `assertThatObject(null)` to avoid casting

assertj-core/src/test/java/org/assertj/core/api/SoftAssertions_navigations_on_null_actual_Test.java
```diff
@@ -0,0 +1,1052 @@
...
+      // GIVEN / WHEN
+      softly.assertThat((Object) null)
+            .extracting("foo")
+            .extracting("bar")
+            .isEqualTo("baz");
+      // THEN - only the first extracting's internal isNotNull collects an error;
+      // the second extracting is on a dead chain and must not add an error or throw.
+      List<Throwable> errors = softly.errorsCollected();
+      then(errors).hasSize(1);
+      then(errors.get(0)).hasMessageContaining("Expecting actual not to be null");
```
joel-costigliola: FTR we could use:
```java
      then(softly.errorsCollected()).singleElement(THROWABLE)
                                    .hasMessageContaining("Expecting actual not to be null");
```

assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
```diff
@@ -1281,12 +1294,53 @@ protected RecursiveAssertionAssert usingRecursiveAssertion() {
...
+   * {@code AbstractOptionalAssert}, which {@code myself} always is).
+   * <p>
+   * <strong>Not safe</strong> when the declared return type is a caller-supplied type variable
+   * (e.g. {@code <ASSERT extends AbstractAssert<?, ?>> ASSERT extracting(..., AssertFactory<?, ASSERT>)}).
+   * In that case the compiler inserts a synthetic {@code checkcast} at the call site to the inferred
+   * concrete type, which {@code myself} will generally not satisfy — prefer
+   * {@link #markAsDeadChain(AbstractAssert)} with a factory-built instance of the correct type.
+   */
+  @SuppressWarnings("unchecked")
+  protected <A extends AbstractAssert<?, ?>> A deadChainSameType() {
```
joel-costigliola: Maybe `deadChain()` is clear enough here ?

assertj-core/src/main/java/org/assertj/core/api/AbstractByteArrayAssert.java
```diff
@@ -1250,7 +1250,7 @@ public AbstractStringAssert<?> asString() {
       objects.assertNotNull(info, actual);
       String actualAsString = new String(actual);
       return assertThat(actualAsString).withAssertionState(myself);
-    });
+    }, () -> new StringAssert(null));
```
joel-costigliola: Introduce dead chain static method factory like `StringAssert.deadChainStringAssert()`  to avoid code duplication and move the responsibility to the target assert class.

Let's do the same for all the other target assert classes (the one we navigate to)

assertj-core/src/test/java/org/assertj/core/api/SoftAssertions_navigations_on_null_actual_Test.java
```diff
@@ -0,0 +1,1052 @@
...
+ * when the source {@code actual} is {@code null} under soft assertions (see assertj/assertj#4213).
+ * <p>
+ * The contract is:
+ * <ul>
+ *   <li>the navigation's internal null-guard collects one {@link AssertionError},</li>
+ *   <li>the navigated assertion is returned as a dead chain so subsequent assertions no-op,</li>
+ *   <li>no {@link RuntimeException} leaks out of the soft-assertion block.</li>
+ * </ul>
+ */
+class SoftAssertions_navigations_on_null_actual_Test {
```
joel-costigliola, line 41: Add a test where we navigate twice, a bit of a corner case, but still interesting to check it behaves as expected, i.e. we skip the second navigation

joel-costigliola (commented):
LGTM, main comment is the factory method, the other ones are  minor

assertj-core/src/main/java/org/assertj/core/api/AbstractByteArrayAssert.java
filiphr: We still want to do

```java
markAsDeadChain(deadChainFactory.get())
```

in the `executeAssertionNavigation`, right?

filiphr:
Thanks for the review @joel-costigliola. All the feedback makes sense. I'll apply it and let you know when I am done.

filiphr:
@joel-costigliola I've applied all your feedback

assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
```diff
@@ -162,6 +172,7 @@ protected <T> T executeAssertionNavigation(Supplier<T> body) {
         throw new RuntimeException(e);
       }
     }
+    if (skipAssertions) return markAsDeadChain(deadChainFactory.get());
```
joel-costigliola: It feels a bit strange to have to mark as dead chain a value that comes from a deadChainFactory, so either deadChainFactory returns a dead chain that don't need changing or deadChainFactory is actually a simple assertFactory that returns a null based assert instance.

I would vote for the latter because it means the dead chain is done in one place, if it needs evolving we wouldn't have to change all the factories and the specific assert don't need to be aware of the dead chain concept, they just need to provide a factory method for a null value assert.

If I'm not mistaken the code in this PR does set twice `skipAssertions = true`, once in `markAsDeadChain` and another in the factory, for example in `deadChainStringAssert()` with the `deadChain()` call.

WDYT ?

assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
```diff
@@ -1253,6 +1265,7 @@ protected RecursiveAssertionAssert usingRecursiveAssertion() {
     requireNonNull(propertyOrField, shouldNotBeNull("propertyOrField")::create);
     requireNonNull(assertFactory, shouldNotBeNull("assertFactory")::create);
     isNotNull();
+    if (actual == null) return markAsDeadChain(assertFactory.createAssert((Object) null));
```
joel-costigliola, line 1268: we could add a default method in  AssertFactory to build a null based instance and avoid repeating casting null

with a method like 
```java
default ASSERT createAssertForNullValue() { return createAssert((T) null);}
```
we can replace `assertFactory.createAssert((Object) null)`  or `assertFactory.createAssert((T) null)`  by `assertFactory.createAssertForNull()` which is a bit more elegant.

joel-costigliola (commented):
Another round of review, waiting to know what you think

assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
filiphr: I am also more for that second one. I can call it `emptyAssertProvider`. It does mean that I should chain the `deadChainStringAssert` to `nullStringAssert` or something like that.

assertj-core/src/main/java/org/assertj/core/api/AbstractAssert.java
filiphr: To be honest I don't think that it's worth exposing such a method for this purpose. Yes it would make it more elegant from within AssertJ, but for implementors of the `AssertFactory` it's a bit verbose, I really do not think that it's worth extending the interface for that.

filiphr:
@joel-costigliola I've applied some of your feedback. Let me know what you think.

joel-costigliola:
thanks @filiphr I have refactored a bit this PR and integrated it in this commit https://github.com/assertj/assertj/commit/d4ded1a1c4fd5e84e0b165ff65b160cf05e1de51 

I started doing minor refactoring, then realized not all navigation methods were calling `executeAssertionNavigation`, doing that led to cleaner code, ex `extracting` had code like:

```java
    isNotNull();
    if (actual == null) return markAsDeadChain(assertFactory.createAssert((Object) null));
```
which was a bit strange, we check for null with `isNotNull` and right after we have to check if actual is null ! 

when wrapped with `executeAssertionNavigation`, there is no need for checking if actual is null since when `isNotNull` fails, `executeAssertionNavigation` will return an assertion instance where assertions are skipped.

this meant no need for `deadChain` method, only `markAsDeadChain` that is now private.

responsilbilites are clearer:

- navigation methods don't need to know about dead chain assertion, they just provide a null/empty based assert instance 
- executeAssertionNavigation makes it skip all assertions

I didn't ask you to update the PR since I came to this in several refactoring steps, I did not see the final result right away, but you'll get the credit for the whole PR anyway (well deserved!)

joel-costigliola:
I also realized there are other navigation methods that are not implemented with `executeAssertionNavigation` but that's for another piece of work (ex: size() in map assert)

filiphr:
thanks for integrating this @joel-costigliola. I went through the changes that you did and they are mostly OK. However, I do not agree with everything.

You've moved the `requireNonNull` checks inside the assert navigation. In my opinion this is not correct. The `requireNonNull` are user errors, they are not assertions. e.g. take the following into consideration

```java
Person person = null;
softly.assertThat(person)
    .extracting(Person::getName)
    .isNotNull()
    .extracting(null)
    .isEmpty();
```

For this particular case the `requireNonNull` would be fully ignored since the second `.extracting(null)` will be skipped, so the `requireNonNull` there will never be executed.

If you agree I can look into it.

>I also realized there are other navigation methods that are not implemented with `executeAssertionNavigation` but that's for another piece of work (ex: size() in map assert)

If you are not looking into it, I can try to carve some time to check those out.

joel-costigliola:
I agree on the principle that we can move the preconditions outside the assert navigation.

I'm not sure that it will change anything though because runtime exceptions are rethrown in `executeAssertionNavigation`, but I might be missing your point.

I tried the test example, and it behaves the same with preconditions outside the assert navigation, could you craft a test showing that it makes a difference ?

FYI I tried this 
```java

      Person person = null;
      softly.assertThat(person)
            .extracting(Person::name)
            .isNotNull()
            .extracting((Function<? super String, ?>[]) null)
            .isEmpty();
```

filiphr:
>I'm not sure that it will change anything though because runtime exceptions are rethrown in `executeAssertionNavigation`, but I might be missing your point.

They are. However, `executeAssertionNavigation` has:

```java
if (skipAssertions) return markAsDeadChain(assertSupplier.get());
```

which means that the 

```
.isNotNull()
.extracting((Function<? super String, ?>[]) null)
.isEmpty()
```

would be skipped. There is only a single assertion in the soft assertions.

Basically, I would expect the following assertion to pass

```java
 assertThatThrownBy(() -> {
        Person person = null;
        softly.assertThat(person)
              .extracting(Person::name)
              .isNotNull()
              .extracting((Function<? super String, ?>[]) null)
              .isEmpty();
      })
        .isInstanceOf(NullPointerException.class)
        .hasMessage(shouldNotBeNull("extractors").create());
```

However it fails, since no exception is thrown.

joel-costigliola:
@filiphr I agree it is better to move the preconditions outside the `executeAssertionNavigation`/`executeAssertion`

PR #4239, opened 2026-05-09
Fix ignoringCollectionOrder false positives with duplicate elements

### LINKED ISSUE #4228: ignoringCollectionOrder false positive with duplicate expected elements
**Describe the bug**

Method usingRecursiveComparison().ignoringCollectionOrder() passes when it should fail: when _expected_ contains duplicate elements that are not all present in actual, the assertion wrongly succeeds.

**Actual behavior**: The assertion passes (no exception is thrown).
Expected behavior: The assertion should throw an AssertionError because expected contains two "a"s but actual only has one.

**Root cause**: in _doCompareUnorderedIterables_, the actual-elements map is rebuilt from scratch on every loop iteration. actualIterator.remove() only removes from the temporary per-iteration map, so the same actual element can match multiple expected elements — bijection is never enforced.

**Fix**: build the map once before the loop so removals persist. Additionally, when a candidate pair does not match, backfill its VisitedDualValues entry with the actual differences found — otherwise the cycle-detection shortcut can later return "equal" for pairs that differ only at a leaf level, causing wrong matches in subsequent iterations.

  - assertj core version: 3.27.3
  - java version: 17
  - test framework version: JUnit 5
  - os (if relevant): —

```java
@Test
void should_fail_when_expected_has_duplicate_elements_not_present_in_actual() {
  List<String> actual = List.of("a", "b");
  List<String> expected = List.of("a", "a"); // duplicate — "b" has no match in expected

  // should fail: expected requires two "a"s but actual only has one
  assertThat(actual).usingRecursiveComparison()
                    .ignoringCollectionOrder()
                    .isEqualTo(expected);
}
```

chanwonlee:
#### Check List:
* Fixes #4228
* Unit tests : YES

#### Summary
`usingRecursiveComparison().ignoringCollectionOrder().isEqualTo(...)` produces a false positive when `expected` contains duplicate elements that are not all present in `actual`.
There are two distinct bugs in `doCompareUnorderedIterables` / `searchExpectedElementIn`, detailed below.

#### Bug 1 — hash buckets rebuilt every iteration
In `doCompareUnorderedIterables`, the `actualElementsGroupedByHashCode(actual)` map was rebuilt inside the loop over `expected`. Although individual bucket entries are removed via `Iterator.remove()` when an `actual` element is matched, rebuilding the map at the start of each iteration restored every previously matched element, breaking the one-to-one matching guarantee.

**Fix**: the map is now built once before the loop, so removals persist across iterations.

#### Bug 2 — visited cache not storing differences for unmatched pairs
In `searchExpectedElementIn`, when a candidate `(actualElement, expectedElement)` pair is compared and differences are found, those differences are recorded against child dual values (typically a leaf field), not against the parent pair itself. The corresponding `VisitedDualValues` entry for the parent pair stays empty.

If the same pair is encountered again (e.g. when `actual` contains the same reference twice), the cycle-detection cache returns an empty differences list, which `searchExpectedElementIn` interprets as "match found" — producing a false positive.

**Fix**: when a candidate fails to match, the differences found during the sub-comparison are added to this pair's visited entry, so that subsequent cache lookups correctly resolve as "different".

#### Other changes
Updated the in-line comment in `doCompareUnorderedIterables` describing the hash-bucket optimisation.

P.S. @joel-costigliola Could you please reassign issue #4228 to me? Thank you!

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/recursive/comparison/RecursiveComparisonDifferenceCalculator.java
@@ -652,15 +652,14 @@ public class RecursiveComparisonDifferenceCalculator {
   private static void doCompareUnorderedIterables(DualValue dualValue, Iterable<?> actual, Iterable<?> expected,
                                                   ComparisonState comparisonState) {
     List<Object> expectedElementsNotFound = list();
+    // speed up comparison by selecting actual elements matching expected hash code, note that the hash code might not be
+    // relevant if fields used to compute it are ignored in the recursive comparison, it's a good heuristic though to check
+    // the first actual elements that could match the expected one, worst case we compare all actual elements.
+    // the map is built once before the loop so that actual elements removed via Iterator.remove() when matched stay
+    // removed across iterations, ensuring one-to-one matching between actual and expected elements.
+    Map<Integer, ? extends List<?>> actualElementsGroupedByHashCode = actualElementsGroupedByHashCode(actual);
     for (Object expectedElement : expected) {
       boolean expectedElementMatched = false;
-      // speed up comparison by selecting actual elements matching expected hash code, note that the hash code might not be
-      // relevant if fields used to compute it are ignored in the recursive comparison, it's a good heuristic though to check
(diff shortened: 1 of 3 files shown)


joel-costigliola:
Integrated thanks @chanwonlee !

PR #4026, opened 2025-11-22
Enable mutation testing

barbulescu:
* See #3843

PR #4275, opened 2026-06-05
add mutation testing

### LINKED ISSUE #3843: Move to https://www.arcmutate.com/
Looks like the best to configure pitest is through https://www.arcmutate.com/ which is free for opensource (see https://github.com/jqno/equalsverifier/tree/main)

it seems that https://github.com/assertj/assertj/blob/main/cdg-pitest-licence.txt is much like https://github.com/jqno/equalsverifier/blob/main/arcmutate-licence.txt but we should reach out to arcmutate team for guidance

barbulescu:
copy of https://github.com/assertj/assertj/pull/4026 targeting 3.x branch

* Closes #3843


scordio:
Superseded by #4276.

PR #4280, opened 2026-06-06
feat: support Iterable in usingRecursiveAssertion

### LINKED ISSUE #4269: usingRecursiveAssertion - Iterable
### Problem

Current implementation of `usingRecursiveAssertion()` only supports collection assertions on types of Collection and arrays. Therefore, types that implement Iterable, but not Collection, cannot take advantage of recursive assertions.

Examples of libraries that implement Iterable on collections, but not Java's Collection include, but may not be limited to https://github.com/vavr-io/vavr.

### Proposed Solution

Extend support of recursive assertion to include types implementing Iterable.

The current solution appears  to already supports Iterable, other than requiring the type Collection, because all it does is use an enhanced for loop. See the code snippet below

    private void recurseIntoCollection(Predicate<Object> predicate, Collection<?> collection, FieldLocation fieldLocation) {
      // TODO handle collection if needed by policy
      if (collection == null) {
        return; // no way to recursive into the collection, anyway the collection node has already been visited
      }
      int index = 0;
      for (Object element : collection) {
        assertRecursively(predicate, element, safeGetClass(element), fieldLocation.field(format(INDEX_FORMAT, index)));
        index++;
      }
   }

mvanhorn:
## Summary

`usingRecursiveAssertion()` now recurses into any `Iterable`, not just `Collection`. `RecursiveAssertionDriver` previously special-cased `Collection.class`, so fields typed as a non-Collection `Iterable` (Guava's `FluentIterable`, custom domain iterables, etc.) were treated as plain leaf nodes and their elements never visited. The container checks and recursion now key on `Iterable`, with the existing `COLLECTION_OBJECT_AND_ELEMENTS` / `ELEMENTS_ONLY` / `COLLECTION_OBJECT_ONLY` policies applying unchanged.

## Why this matters

#4269 reports that recursive assertions silently skip elements inside `Iterable` fields, which makes `allFieldsSatisfy` pass on objects whose nested elements would fail the predicate - a false positive in test code, where silent gaps matter most. Maps, arrays, and `Optional` keep their existing specific handling; the `Iterable` check runs after those so nothing that previously matched a more specific branch changes behavior.

## Testing

`RecursiveAssertionDriver_CollectionPolicyTest` gains cases asserting a non-Collection `Iterable` field is recursed under each collection policy, mirroring the existing `Collection` cases. Full recursive-assertion test pack: 78 run, 0 failures (`assertj-core-tests`, JDK 25).

#### Check List:
* Fixes #4269
* Unit tests : YES
* Javadoc with a code example (on API only) : NA (no public API signature change)

### DIFF
--- a/.github/workflows/pitest-publish.yml
+++ b/.github/workflows/pitest-publish.yml
@@ -1,0 +1,33 @@
+name: Mutation Testing Result
+
+on:
+  workflow_run: # Executed only when this workflow file exists on the default branch
+    workflows: [ Mutation Testing ]
+    types: [ completed ]
+
+permissions: {}
+
+jobs:
+
+  publish:
+
+    name: Publish
+    if: github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Set up Java
+        uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
+        with:
+          distribution: 'zulu'
+          java-version: '21'
+          cache: 'maven'
+      - name: Download Results
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: pitest
+          path: target/pit-reports-ci
+          run-id: ${{ github.event.workflow_run.workflow_id }}
+      - name: Publish Results
+        run: ./mvnw $MAVEN_ARGS pitest-github:github -DrepoToken=${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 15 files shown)


joel-costigliola:
@mvanhorn thanks for the PR.

The PR looks good, but I think we should rename `CollectionAssertionPolicy` to `IterableAssertionPolicy`,  `COLLECTION_OBJECT_ONLY` to `ITERABLE_OBJECT_ONLY` and `COLLECTION_OBJECT_AND_ELEMENTS` to `ITERABLE_OBJECT_AND_ELEMENTS`, it's a breaking change can do in `main`  since main is a new major version.

We would need to backport this PR in 3.x adding `IterableAssertionPolicy` and deprecating `CollectionAssertionPolicy`.

mvanhorn:
Done in 5dca083: `CollectionAssertionPolicy` is now `IterableAssertionPolicy` with `ITERABLE_OBJECT_ONLY` and `ITERABLE_OBJECT_AND_ELEMENTS` (test classes renamed to match). Verified locally: assertj-core compiles and the RecursiveAssertion test suite passes 78/78. Heads up that the Binary Compatibility check will flag the rename since it compares against the latest release - expected for the intentional break on main. Happy to open the 3.x backport as a follow-up PR adding `IterableAssertionPolicy` and deprecating `CollectionAssertionPolicy` if that's how you'd like to sequence it.

joel-costigliola:
Thanks @mvanhorn, waiting for the backport if you can make it :)

.github/workflows/pitest-publish.yml
```diff
@@ -0,0 +1,33 @@
+name: Mutation Testing Result
```
scordio, line 1: This shouldn't have been part of this PR. I'll clean it up separately.

scordio: Done in adadd327.

mvanhorn:
Thank you for shepherding this, @joel-costigliola. Iterable support rounds out usingRecursiveAssertion nicely.

joel-costigliola:
backported in 3.x with 65327a37e24637cf7e25f723c98b8792dfb567c7

PR #4285, opened 2026-06-09
Expose FileInfo as actual/expected for file/path diff assertions

filiphr:
When file-vs-file content assertions fail, pass both files as opentest4j FileInfo actual/expected values on the thrown AssertionFailedError so IDEs can render a side-by-side file diff.

A new internal FileContent carrier (path + bytes, no opentest4j dependency) flows through the existing 4-arg Failures.failure overload; AssertionErrorCreator reflectively converts it to org.opentest4j.FileInfo when opentest4j is present, keeping the opentest4j dependency optional.

----
This was inspired by some work I was doing for MapStruct in https://github.com/mapstruct/mapstruct/pull/4063. Keep in mind that in the existing tests if `expectAssertionError` is removed IntelliJ **will not** render what is there in the MapStruct PR, and that is due to the fact that once the test is done the temporary directories are removed. To really see it, a test is needed where the files won't be deleted if the test is done.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/error/AssertionErrorCreator.java
+++ b/assertj-core/src/main/java/org/assertj/core/error/AssertionErrorCreator.java
@@ -17,18 +17,21 @@ package org.assertj.core.error;
 
 import static org.assertj.core.util.Arrays.array;
 
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


scordio:
Superseded by #4286.

joel-costigliola:
merge commit holding the PR content 60d61370c66ac072faf030fca51cae15cc743b01

Issue #3977, opened 2025-10-17
Evaluate Javadoc migration to Markdown


scordio:
AssertJ extensively uses Javadoc to showcase API usage, which can sometimes become a bit complex to maintain.

Since [we now build with Java 25](https://github.com/assertj/assertj/issues/3944), we could evaluate migrating to Markdown and have a more readable syntax, especially for lists and code snippets.

PR #4295, opened 2026-06-20
Convert Javadoc to Markdown on assertj-tests modules

barbulescu:
This is work on https://github.com/assertj/assertj/issues/3977 on modules supporting Java 23+.

Most of the changes were done with Openrewrite with several bugs fixed on the way.

Initial review was done by Claude followed by several fixes and a final manual review.

---

Blocked by:
* https://github.com/assertj/assertj/pull/4298


scordio:
Hi @barbulescu, thanks a lot for raising this!

AFAIK, what should matter is only the Java version at build time, not the language level we target. This means we can do the same for all other modules, and we might also consider targeting `3.x` since we're building it with Java 25, and then forward-port the changes to `main`.

I'll double-check my assumption and apply your changes either in `3.x` or `main`.

scordio:
FYI, to make sure I'm not clashing with your work, I started experimenting with the Guava module in #4298.

PR #4310, opened 2026-07-01
Add tests for PathNaturalOrderComparator comparison and null-handling behaviour

vasiliy-mikhailov:
Adds tests covering `PathNaturalOrderComparator`'s `compare` method: equal paths return zero, natural ordering returns negative/positive, and the inherited null-safe contract returns -1 for a null first argument, 1 for a null second argument, and 0 when both are null. Previously only the `toString` description was tested.

| metric | before | after |
|---|---|---|
| mutation score | 0% | 100% |
| test methods added | n/a | 6 |

The additions are append-only (no existing test is modified) and pass against the current code.

---

**How this was produced**

This PR was generated with an AI-assisted pipeline built around mutation testing (PIT). The pipeline mutates the target class (flipping conditions and changing boundary/edge cases) and runs the existing tests against each mutant. Where a mutant survives (the existing tests do not catch that edge case), it writes a focused test for that case and reruns PIT to confirm the new test actually kills that specific mutant. So every added test is verified to catch a concrete edge case the suite missed before, rather than being speculative or redundant. The change is additive only (no production code modified), and the module builds green under its CI JDK.

### DIFF
--- a/assertj-core/src/test/java/org/assertj/core/util/PathNaturalOrderComparatorTest.java
+++ b/assertj-core/src/test/java/org/assertj/core/util/PathNaturalOrderComparatorTest.java
@@ -17,6 +17,9 @@ package org.assertj.core.util;
 
 import static org.assertj.core.api.BDDAssertions.then;
 
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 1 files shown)


scordio:
Thanks for raising it, @vasiliy-mikhailov.

The new tests don't fully follow the project's [style guide](https://github.com/assertj/assertj/blob/main/CONTRIBUTING.md) (missing `GIVEN`/`WHEN`/`THEN`). Also, I'm not sure this is the testing strategy I'd use, given the simplicity of `PathNaturalOrderComparator` and the fact that the actual logic is in `NaturalOrderComparator`.

Nevertheless, since gaps in mutation testing are a broader topic, it's more effective for us to address this internally as part of #4311.

vasiliy-mikhailov:
Thanks for the thoughtful review and for closing the loop, @scordio. Both points are fair: the tests should have followed the GIVEN/WHEN/THEN structure from the contributing guide, and you're right that `PathNaturalOrderComparator` is a thin delegator over `NaturalOrderComparator`, so exercising the underlying comparator directly is the better target. Addressing the broader coverage gaps internally through #4311 makes sense. Thanks again for taking the time to explain the reasoning.

PR #4330, opened 2026-07-20
feat: add allElementsAreEqual assertion for iterables

### LINKED ISSUE #4302: Chaining-friendly assertion that all elements are equal
#### Feature summary

Provide a chaining-friendly way to assert that all elements are the same:
```java
assertThat(list).allElementsTheSame();
```

The current best-practice requires using `allMatch()` and referencing the first element of the list:
`assertThat(list).allMatch(e -> e.equals(list.get(0));`. However that patter falls short in more complex use cases where the list is produced by `filteredOn()` and `extracting()` sequences and cannot be easily referenced.

#### Example

My current real-world test that verifies that the output table is aligned:
```java
List<String> dataLines = Arrays.stream(output.split("\n"))
	.filter(s -> s.contains("|"))
	.toList();

assertThat(dataLines)
		.hasSize(20)
		.allMatch(s -> s.matches(".* \\| .* --- \\w+"));

List<int[]> pipePositions = dataLines.stream()
	.map(s -> new int[] {s.indexOf('|'), s.indexOf("---")})
	.toList();

assertThat(pipePositions)
		.allMatch(p -> Arrays.equals(p, pipePositions.get(0)));
```
Note how I cannot join those 2 checks into one because i have to reference the first element.

How it could look like:
```java
assertThat(output.split("\n"))
        .filteredOn(s -> s.contains("|"))
        .hasSize(20)
        .allMatch(s -> s.matches(".* \\| .* --- \\w+"))
        .extracting(s -> new int[] {s.indexOf('|'), s.indexOf("---")})
        .allElementsTheSame()
```

#### Special Consideration

As shown in my example, sometimes `equals()` check on elements is not adequate and a special check is needed (for example `Arrays.equals()`).

seonwooj0810:
Closes #4302

## Why

There is currently no chaining-friendly way to assert that all elements of an iterable are equal to each other. The workaround `assertThat(list).allMatch(e -> e.equals(list.get(0)))` requires a direct reference to the first element, which is awkward when the iterable is produced by a chain of `filteredOn`/`extracting` calls and cannot be easily referenced. As discussed in the issue, @joel-costigliola agreed to add `allElementsAreEqual` (equality-based, not reference-based).

## What

Adds `allElementsAreEqual()` to `AbstractIterableAssert`, so it is available on all `Iterable`-based assertions:

```java
assertThat(dataLines).allElementsAreEqual();
```

- Passes when all elements are equal, and (vacuously) for empty or single-element iterables — consistent with `allMatch`/`allSatisfy`.
- Fails otherwise, reporting every element that is not equal to the first.
- Honors a custom element comparator set via `usingElementComparator(...)` (the assertion goes through the internal `comparisonStrategy`).

New error factory `ShouldHaveAllElementsEqual` produces a message such as:

```
Expecting all elements to be equal to:
  "a"
but the following element(s) were not:
  ["b"]
in:
  ["a", "b"]
```

## Tests

- `Iterables_assertAllElementsAreEqual_Test` — engine behaviour: all-equal, single-element and empty pass; null actual; failure reports all unequal elements; readable error message; and the custom (case-insensitive) comparison-strategy pass/fail paths.
- `IterableAssert_allElementsAreEqual_Test` — verifies the public API delegates to the internal engine (`IterableAssertBaseTest`).

Verification done: ran `./mvnw -pl assertj-core test` for both new test classes (10 tests, all green); `spotless:apply` and `license:format` produce no further changes.

I followed the discussion in #4302; if you would prefer this also exposed on `ObjectEnumerableAssert` (arrays) or want the pair/fold assertions tracked separately, happy to adjust.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java
@@ -548,6 +548,31 @@ public abstract class AbstractIterableAssert<SELF extends AbstractIterableAssert
     return executeAssertion(() -> iterables.assertDoesNotHaveDuplicates(info, actual));
   }
 
+  /**
+   * Verifies that all the elements of the actual iterable are equal to each other, using the {@link Object#equals(Object)}
+   * implementation of the elements (or the comparator set with {@link #usingElementComparator(java.util.Comparator)}).
+   * <p>
+   * This is convenient when the reference element cannot easily be extracted from the iterable, for example when the
+   * iterable is produced by a chain of {@code filteredOn}/{@code extracting} calls.
+   * <p>
+   * Example:
+   * <pre><code class='java'> // assertions succeed
+   * assertThat(list(&quot;a&quot;, &quot;a&quot;, &quot;a&quot;)).allElementsAreEqual();
+   * assertThat(list(&quot;a&quot;)).allElementsAreEqual();
+   * assertThat(emptyList()).allElementsAreEqual();
+   *
+   * // assertion fails
+   * assertThat(list(&quot;a&quot;, &quot;a&quot;, &quot;b&quot;)).allElementsAreEqual();</code></pre>
+   *
+   * @return {@code this} assertion object.
+   * @throws AssertionError if the actual iterable is {@code null}.
+   * @throws AssertionError if the actual iterable contains at least two elements that are not equal.
+   * @since 4.0.0
+   */
+  public SELF allElementsAreEqual() {
+    return executeAssertion(() -> iterables.assertAllElementsAreEqual(info, actual));
+  }
+
   @Override
   @SafeVarargs
   public final SELF startsWith(ELEMENT... sequence) {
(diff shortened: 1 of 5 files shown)


seonwooj0810:
Circling back on this assertion addition — happy to tweak the naming or API shape if you have preferences.

scordio:
Hi @seonwooj0810, thanks for your contribution!

I'd like to discuss an alternative name for this feature; see https://github.com/assertj/assertj/issues/4302#issuecomment-5646944724.

seonwooj0810:
Thanks @scordio! Happy to rename. One thing to flag: in #4302 @joel-costigliola steered away from "same" since it could read as reference identity, which is why I went with `allElementsAreEqual`. Keeping the `containsOnly*` family is a good point though — would something like `containsOnlyEqualElements` work for both concerns? I'll update the PR once you two settle on the name.

scordio:
Fair enough, comment updated for consistency. WDYT @joel-costigliola?

scordio:
@seonwooj0810 just FYI, there is no need to polish your PR description or comments through an LLM. We much prefer direct and authentic conversation, as keeping things direct makes the review process faster and more natural for everyone.

PR #4338, opened 2026-07-23
 Refactor global configuration reset handling

scordio:
Prior to this change, test-order pollution was caused by the global mutable AssertJ configuration not being reset consistently between tests. The `@MutatesGlobalConfiguration` JUnit extension introduced in #2670 was an attempt to address this problem, but it couldn't work properly due to the lack of runtime retention.

This PR addressed the original issue in two ways.

## Tests under `assertj-core-tests`

A new `ResetGlobalConfigurationExtension` is defined as a [global JUnit extension](https://docs.junit.org/6.1.2/extensions/registering-extensions.html#registration-automatic); note that this requires a new `provides` directive in the test-scoped `module-info.java`.

With this, the configuration is always reset after each test execution in the whole test suite.

## Tests under `assertj-core`

Today, we achieve module-path test execution by patching the module descriptor with the `argFile` content. If we were to define a global extension here as well, we would need a `java` command line option equivalent to the `provides` directive. However, such a command line option does not exist, as of today.

For this reason, configuration reset is done programmatically via an explicit `@AfterEach` method in the areas where the state is mutated (`EntryPointAssertionsBaseTest`, the assumptions test, and now shared `DateAssertBaseTest` for date assertions tests).

Migrating such tests to `assertj-core-tests` will allow deleting the `@AfterEach` methods.


scordio:
@joel-costigliola I'll go ahead with the merge so we can move forward with other PRs (including Dependabot ones), but please review the changes when you get a chance and let me know if you would do anything differently.

joel-costigliola:
LGTM

PR #4345, opened 2026-08-01
fix: show Timestamp nanos in StandardRepresentation

### LINKED ISSUE #4241: Unintuitive behaviour of afterOrEqualTo() for Instant
I need to check that a method-under-test is setting a `java.sql.Timestamp` correctly using `System.currentTimeMillis()`.

To avoid test flakiness because of timing issues, I typically record the timestamp before and after invoking the method-under-test, and then check that the timestamp set by the method-under-test falls in the range. This pattern works successfully but very fast methods, it is not sufficient to use `isAfter()` and `isBefore()` - I must use `isAfterOrEqualTo()`.

This pattern was working fine until I switched from passing a `java.sql.Timestamp` in to `isAfterOrEqualTo()`, and for convenience started using `java.time.Instant`. The assertion started to fail when the `Instant` was equal to the `Timestamp`.

I upgraded to 3.27.7 but still had the same problem.

I can work around this by changing my arguments back into a Timestamp, but thought I would raise it anyway.

arimu1:
#### Check List:
* Fixes #4241
* Unit tests : YES

### Description

When asserting a millis-only `java.sql.Timestamp` (e.g. built from `System.currentTimeMillis()`) with `isAfterOrEqualTo(Instant)`, AssertJ converts the `Instant` via `Timestamp.from(instant)`, which keeps nanoseconds. Comparison is therefore nano-aware and can fail when the values share the same millisecond but differ in sub-ms precision.

The failure message used `SimpleDateFormat` with only millisecond precision, so both sides looked identical and the assertion error appeared to contradict itself.

### Change

`StandardRepresentation` now includes sub-millisecond digits from `Timestamp.getNanos()` (trailing zeros stripped, Instant-style) so the precision mismatch is visible in the error message.

Assertion behavior is unchanged — only the representation is improved.

### Example

```java
Instant withMicros = Instant.parse("2011-01-01T00:00:00.123456Z");
Timestamp actual = new Timestamp(withMicros.toEpochMilli()); // millis only

assertThat(actual).isAfterOrEqualTo(withMicros);
// still fails (correct), but the message now shows ...123 vs ...123456
```

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
+++ b/assertj-core/src/main/java/org/assertj/core/presentation/StandardRepresentation.java
@@ -33,6 +33,7 @@ import java.io.StringWriter;
 import java.lang.reflect.Method;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Path;
+import java.sql.Timestamp;
 import java.text.SimpleDateFormat;
 import java.time.Duration;
 import java.time.LocalDate;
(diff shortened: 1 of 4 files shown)


joel-costigliola:
good catch, thanks @arimu1

joel-costigliola:
Integrated in 3.x with https://github.com/assertj/assertj/commit/6a4a584401741f57be78ed4946892a85ee51b452

PR #4350, opened 2026-08-04
Add hasEnclosingClass and hasNoEnclosingClass assertions

### LINKED ISSUE #4343: Assertions for enclosing classes
Following [`Class#getEnclosingClass()`](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getEnclosingClass--), we could add assertions to test whether a given class has an enclosing class, which can be handy to enforce top-level classes.

The cases we should cover are:

* Top-level classes
* Nested classes (static member classes)
* Inner classes (non-static member classes)
* Local classes (named classes declared within a method)
* Anonymous classes

Possible assertions:
- `hasEnclosingClass(Class)` --> `null` rejected
- `hasNoEnclosingClass()` --> passes for top-level classes, fails for all others

The implementation is likely very similar to `hasSuperclass(Class)`/`hasNoSuperclass()`.

chanwonlee:
Closes #4343

Added two new assertions:
- `hasEnclosingClass`
- `hasNoEnclosingClass`

Both are implemented similarly to `hasSuperclass(Class)`/`hasNoSuperclass()`

#### Check List:
* Unit tests : YES


chanwonlee:
@scordio would you please take a look when you have a chance? Thanks!

Issue #864, opened 2017-01-01
SoftAssertions#fail(String failureMessage, Throwable realCause) display information about realCause?


PascalSchumacher:
While taking a look at https://github.com/joel-costigliola/assertj-core/issues/539 I noticed that SoftAssertions#fail(String failureMessage, Throwable realCause) could be more useful, at the moment it does not show any information about realCause.

```java
@Test
public void test() {
    softly.fail("failure", new RuntimeException("abc"));
    softly.assertAll();
}
```

just displays:

>org.assertj.core.api.SoftAssertionError:
>The following assertion failed:
>1) failure
>
>    at org.assertj.core.api.SoftAssertions.assertAll(SoftAssertions.java:116)
>    at org.assertj.core.api.SoftAssertionsTest.test(SoftAssertionsTest.java:115)

The error message etc. of `realCause` is lost.

joel-costigliola:
I agree, we should at least display the real cause message and probably the stack trace (although it might get more difficult to read the other errors because of a stack trace can be quite long).
WDYT ?

PascalSchumacher:
The real cause message (without the stacktrace) would be the best compromise between being concise and providing helpful information. (Stacktrace can also be added later on if there is demand for this feature.)

joel-costigliola:
What about displaying the 5 first stack trace elements (and adding the fact that it is a truncated stack trace) ?

PascalSchumacher:
Sure, that should be enough information in almost all cases I guess.

sherfert:
I think this regressed. While it seems the behavior was fixed in 3.23.1 (https://github.com/assertj/assertj/commit/c487af4cdb37b01103e4ec0f9971c9a53c766db5), I can reproduce the original issue in 3.27.6.

sherfert:
After having digged a bit, I think this does not work if we succeed in creating a `org.opentest4j.MultipleFailuresError` in `AssertionErrorCreator`

```
public AssertionError multipleSoftAssertionsError(List<? extends Throwable> errors) {
    Optional<AssertionError> multipleFailuresError = tryBuildingMultipleFailuresError(errors);
    return multipleFailuresError.orElse(new SoftAssertionError(describeErrors(errors)));
  }
```

The or-else-returned `SoftAssertionError` still calls `describeErrors` introduced in the above linked commit.

scordio:
Thanks for reporting it, @sherfert! I'll add it to 3.28.0 for analysis.

[...]

PR #4375, opened 2026-08-26
Include failure causes in `AssertJMultipleFailuresError` message

### LINKED ISSUE #4125: Regression on `SoftAssertions#fail(String, Throwable)`
> I think this regressed. While it seems the behavior was fixed in 3.23.1 (https://github.com/assertj/assertj/commit/c487af4cdb37b01103e4ec0f9971c9a53c766db5), I can reproduce the original issue in 3.27.6. 

 _Originally posted by @sherfert in [#864](https://github.com/assertj/assertj/issues/864#issuecomment-3804628164)_

jjh75607:
Fixes #4125

Make `AssertJMultipleFailuresError#getMessage` render the cause of each failure that has one, so `SoftAssertions#fail(String, Throwable)` shows `cause message: ...` and the first stack trace elements of the cause again, as in 3.23.1. Failures without a cause are formatted exactly as before.

The cause was lost because `getMessage` only used the failure message; `Throwables.describeErrors` (the #864 fix) was reached only through the `SoftAssertionError` fallback, which never runs when opentest4j is on the classpath.

`SoftAssertionsErrorDescriptionTest` did not catch this since it asserted the opentest4j prefix but not the cause; it now asserts the cause too. `main` is not affected: 452cee73 already renders the cause in `MultipleAssertionsError`.

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/error/AssertJMultipleFailuresError.java
+++ b/assertj-core/src/main/java/org/assertj/core/error/AssertJMultipleFailuresError.java
@@ -17,6 +17,7 @@ package org.assertj.core.error;
 
 import static java.lang.String.format;
 import static org.assertj.core.util.Throwables.addLineNumberToErrorMessages;
+import static org.assertj.core.util.Throwables.describeErrors;
 
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


joel-costigliola:
Integrated thanks @jjh75607 !

PR #4387, opened 2026-09-03
Fix SoftAssertions#fail(String, Throwable) losing realCause info in aggregated error (#4125)

### LINKED ISSUE #4125: Regression on `SoftAssertions#fail(String, Throwable)`

wylovelyi:
## Summary

Fixes #4125 — a regression of #864 where `SoftAssertions#fail(String failureMessage, Throwable realCause)` no longer shows the `realCause` information in the aggregated error reported by `assertAll()` (reported as broken again in 3.27.6).

## Root cause

The original 2017 fix (commit `c487af4`) rendered the cause's message and first five stack-trace elements when building the `SoftAssertionError`. After soft assertions started rendering the aggregated error through `AssertionErrorCreator.multipleSoftFailuresError` → `AssertJMultipleFailuresError`, the cause is no longer shown, because `AssertJMultipleFailuresError.getMessage()` only uses each failure's `getMessage()` and does not render the cause.

## Fix

In `AbstractSoftAssertions.fail(String, Throwable)` we now embed the cause description into the failure message (message + `cause message:` + first five stack-trace elements), matching the original 2017 behaviour. `fail(Throwable)` keeps an empty message and only sets the cause, so existing behaviour for the no-message variant is unchanged.

This is a targeted change that restores the lost information without altering the global error-formatting classes.

## Verification

- Added `SoftAssertionsTest.should_display_real_cause_in_aggregated_error` asserting that the error thrown by `assertAll()` contains `cause message: abc`.
- Ran `SoftAssertionsTest` and `BDDSoftAssertionsTest` — all green.
- `spotless:check` passes.

This targets the `3.x` line (the reported regression is in 3.27.6).

### DIFF
--- a/assertj-core/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
+++ b/assertj-core/src/main/java/org/assertj/core/api/AbstractSoftAssertions.java
@@ -17,6 +17,7 @@ package org.assertj.core.api;
 
 import static java.lang.String.format;
 
+import java.util.Arrays;
 import java.util.List;
 
 import org.assertj.core.annotation.CanIgnoreReturnValue;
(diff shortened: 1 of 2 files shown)


scordio:
Thanks, but we already have a PR for this: #4375.

PR #4410, opened 2026-09-20
Add UTF-8 convenience methods (#4150)

### LINKED ISSUE #4150: Add convenience methods using UTF-8 charset
#### Feature summary
UTF-8 is a common (if not the most common) charset being used. Even the JDK uses it by default for some of its API, e.g. the methods in `java.nio.file.Files` such as `Files#readString(Path)`.

There are multiple parts of the AssertJ API which perform `byte[]` to `String` decoding, e.g. asserts for `Path` or `InputStream`. However, currently AssertJ either requires you to explicitly specify the charset, or uses the default charset.

It would be useful if AssertJ made it easier to use UTF-8 as charset, either by:
- adding additional convenience methods, such as `contentUtf8()` / `utf8Content()`
- OR, changing the methods which currently use the default charset to instead use UTF-8
(breaking change)

#### Example

```java
assertThat(file).contentUtf8().startsWith("hello");
```

#### Additional notes
- With JDK >= 18 the default charset is actually UTF-8 by default ([JEP 400](https://openjdk.org/jeps/400))
  - Contra: This reduces the need for dedicated AssertJ API for UTF-8.
  - Pro: Some AssertJ users might still prefer to explicitly specify UTF-8 as desired charset for the assertion instead of implicitly relying on it due to it being the default charset.
- Contra: Users could add a static import for `StandardCharsets.UTF_8`, so it just becomes `assertThat(file).content(UTF_8)`. That might not be too bad?
- Pro (for changing current AssertJ methods using default charset to UTF-8): Using the default charset can make tests error-prone and non-portable.

harrisleesh:
## Summary

Adds UTF-8 convenience methods so users don't have to pass `StandardCharsets.UTF_8` explicitly, as requested in #4150.

These are purely additive, non-breaking methods that delegate to the existing charset-taking navigation methods (no new decoding logic, no behavior change):

- `AbstractPathAssert#utf8Content()` → `content(StandardCharsets.UTF_8)`
- `AbstractFileAssert#utf8Content()` → `content(StandardCharsets.UTF_8)`
- `AbstractInputStreamAssert#asStringUtf8()` → `asString(StandardCharsets.UTF_8)`

```java
assertThat(path).utf8Content().startsWith("hello");
assertThat(file).utf8Content().startsWith("hello");
assertThat(inputStream).asStringUtf8().startsWith("hello");
```

Each method has full Javadoc with passing + failing examples and an `@since 4.0.0` tag.

## Notes on naming

I proposed this API shape on the issue first (https://github.com/assertj/assertj/issues/4150#issuecomment-5749298853) and I'm happy to adjust before merging.

For `Path`/`File` the existing entry point is `content(Charset)`, so `utf8Content()` reads naturally. For `InputStream` the existing entry point is `asString(Charset)` (there is no `content(Charset)`), so `asStringUtf8()` mirrors it. If you'd prefer a single unified naming scheme across all three (e.g. `utf8Content()` everywhere), just let me know and I'll update.

## Tests

New test classes mirroring the existing `content`/`asString` tests:

- `PathAssert_utf8Content_Test`
- `FileAssert_utf8Content_Test`
- `InputStreamAssert_asStringUtf8_Test`

All green locally on JDK 25, `spotless:apply` applied. No regressions in the Path/File assert test packages (257 tests) or the InputStream tests.

Closes #4150

🤖 Generated with [Claude Code](https://claude.com/claude-code)


scordio:
Hi @harrisleesh, thanks for the PR! I don't think we should proceed with it, see https://github.com/assertj/assertj/issues/4150#issuecomment-5749601948, so I'm closing it.

PR #4412, opened 2026-09-21
Null values now pass in "notInstance" checks

### LINKED ISSUE #4411: Avoid failures in isNotInstanceOf assertions for null objects
**Describe the bug**

I have a unit test for a method that can return multiple subtypes. One of these operation types is expected to be null (since HTTP DELETE operations generally do not return a JSON body). However, null values are not supported by `isNotInstanceOf()`, so I have to use the following.

```java
    // Only patch operation types should return a PatchRequest.
    assertThat(BulkOperation.post("/Users", node).getDataAsScimResource())
        .isNotInstanceOf(PatchRequest.class);

    assertThat(BulkOperation.delete("/Users/userID").getDataAsScimResource())
        .satisfies(data -> assertThat(data instanceof PatchRequest).isFalse());
```

Specifically, if the `satisfies()` call is replaced with `isNotInstanceOf(PatchRequest.class)`, an assertion error is thrown, which is unideal:
```
java.lang.AssertionError: 
```
&nbsp;

I would like to suggest changing this behavior. It seems reasonable that if Java considers `null instanceof Object` to always be false, then the following assertions could be considered passing:

```java
    assertThat(null).isNotInstanceOf(String.class);
    assertThat(null).isNotInstanceOfAny(String.class);
    assertThat(null).isNotExactlyInstanceOf(String.class);
```

I'll open a PR with proposed changes to update this behavior if you all agree on whether this seems appropriate. I've confirmed that this code change resolves the issue in my test.
&nbsp;

* assertj core version: 3.27.7
* java version: 17, 21, 25
* test framework version: TestNG 7.12.0
* os (if relevant): N/A

```java
  @Test
  public void test() {
    String nullString = null;
    assertThat(nullString)
        .satisfies(s -> assertThat(s instanceof String).isFalse())
        .isNotInstanceOf(String.class);
  }
```

kqarryzada:
Updated the implementation of methods such as Objects.isNotInstanceOf() to pass when the 'actual' value is null. This is intended to align with the way Java treats the 'instanceof' keyword.

#### Check List:
* Fixes #4411
* Unit tests : YES


assertj-core/src/test/java/org/assertj/core/internal/objects/Objects_assertIsNotExactlyInstanceOf_Test.java
```diff
@@ -42,16 +40,14 @@ void should_pass_if_actual_is_not_exactly_instance_of_type() {
...
   }
 
   @Test
-  void should_fail_if_actual_is_null() {
-    assertThatExceptionOfType(AssertionError.class).isThrownBy(() -> objects.assertIsNotExactlyInstanceOf(someInfo(), null,
-                                                                                                          String.class))
-                                                   .withMessage(actualIsNull());
+  void should_throw_error_if_type_is_null() {
+    assertThatNullPointerException().isThrownBy(() -> objects.assertIsNotExactlyInstanceOf(someInfo(), "Yoda", null))
+                                    .withMessage("The given type should not be null");
```
kqarryzada, line 50: The diff looks strange for these because I left "should_pass" tests above the "should_throw" ones, but I didn't modify any "should_throw" tests.

